Issue #4, opened 2016-12-29
check if file exists before trying to add it to the package?


caarlos0:

caarlos0:
maybe a default files section and those files could be checked?

caarlos0:
done in #9

PR #9, opened 2016-12-30
Alt licence

jspc:
From https://www.reddit.com/r/golang/comments/5l3i9b/deliver_go_binaries_as_fast_and_easy_as_possible/dbss8t5/

This PR introduces expansion of the default files to add code in `config.go`'s `fix()` function.

The original rationale behind this change is to allow the tool to recognise and use the alternative spelling of 'Licence' that Real English ™️ uses. The scope has crept a little in that the original implementation would bail when no licence was included.

I have also expanded the testing of this method; it had happy-path'd file selection a tad.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"errors"
 	"io/ioutil"
+	"log"
 	"os"
 	"path"
 	"path/filepath"
@@ -11,7 +12,10 @@ import (
 	yaml "gopkg.in/yaml.v1"
 )
 
-var emptyBrew = Homebrew{}
+var (
+	emptyBrew    = Homebrew{}
+	filePatterns = []string{"LICENCE.*", "LICENSE.*", "README.*"}
+)
 
 // Homebrew contains the brew section
 type Homebrew struct {
@@ -70,10 +74,13 @@ func fix(config ProjectConfig) ProjectConfig {
 	if len(config.Files) == 0 {
 		config.Files = []string{}
 
-		for _, f := range []string{"README.md", "LICENCE.md", "LICENSE.md"} {
-			if _, err := os.Stat(f); err == nil {
-				config.Files = append(config.Files, f)
+		for _, pattern := range filePatterns {
+			matches, err := globPath(pattern)
+			if err != nil {
+				log.Fatalf("Error searching for %q: %v", pattern, err)
 			}
+
+			config.Files = append(config.Files, matches...)
 		}
 	}
 	if config.Token == "" {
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -66,9 +66,12 @@ func Load(file string) (config ProjectConfig, err error) {
 
 func fix(config ProjectConfig) ProjectConfig {
 	if len(config.Files) == 0 {
-		config.Files = []string{
-			"README.md",
-			"LICENSE.md",
+		config.Files = []string{}
+
+		for _, f := range []string{"README.md", "LICENCE.md", "LICENSE.md"} {
```
caarlos0: I guess it would be nice to get `README`s and `LICENSE`s with other extensions too, so I was thinking about changing this to `README.*`, `LICENSE.*` and `LICENCE.*`...

What do you think?

caarlos0:
refs #4

config/config.go
jspc: Thats not a bad idea. I think the patterns should probably be abstracted out of that function, too; it'd be more maintainable in a var statement at the top.

I'll get something going tomorrow

caarlos0: nice, if you can't by any reason, let me know and I work on it =)

thanks for the contribution, by the way =D

jspc:
@caarlos0 ♻️ 

## A note

https://github.com/goreleaser/releaser/pull/9/files#diff-3baf47c64847a8fb8aaa8cc2e088513bR80

This is pure laziness on my behalf to avoid writing error catchers. I'm happy to do so, but if I do I think I'm going to change `fix()` to use the signature and make that entire block more idiomatic.

```golang
(*ProjectConfig)func fix() error
```

to avoid lots of shadowing/ declaration cruft.

nochso:
Not everybody uses Markdown (especially for LICENSE files), so how about also looking for files without an extension?

I suppose it comes down to this RE: `(LICEN[SC]E|README)(\.(md|txt))?`

It might also be a good idea to include a `CHANGELOG` file.

jspc:
@nochso latest commit has a naive implementation of that pattern (I didn't play much with the globbing in `path/filepath` so I don't know how good the pattern matching is).

nochso:
`filepath.Match` in combination with "LICENSE.*" will not match files without an extension.

jspc:
Good spot; latest commit is a little more permissive.

nochso:
Good enough for me. Otherwise `strings.TrimSuffix(path, filepath.Ext(path))` should work for getting the file name without any extension.

caarlos0:
Looks good, thanks!

PR #25, opened 2017-01-01
cleanup pipe

caarlos0:
removing the `dist` folder :)


caarlos0:
thinking better now, this is pretty much useless.

nervermind

jspc:
👍  - I'm using `dist/` for my docker builds, this PR would have made it complicated

PR #29, opened 2017-01-02
gometalinter on CI

caarlos0:
refs #28

PR #28, opened 2017-01-02
Draft: making contributing easier

lucapette:
Hi! First of all, thank you very much for this project! I wanted to write such a thing for a while but never managed to get to it. 

As I do not consider this PR ready (will explain with a short list what I think we miss),  it is not meant to be merged yet. It is more a proposal and I find polite to bring some code in when proposing changes to an open source project.

I like the project a lot and would like to add a few things I believe are good way  to increase the chances people will contribute to it.

I'm going to make an initial list of things I would love to help with:

- [ ] Makefile
   I already added a draft of the Makefile I would add in the PR so it is easier to discuss about it. I use a similar solution in all the golang projects I work on and it works nicely.
- [ ] Adding a CONTRIBUTING.md
  Some documentation about how to get the project up and running locally using the Makefile we have just discussed
- [x] Adding a code of conduct
   I start off  http://contributor-covenant.org/ an awesome project that makes it very easy for people to add a good code of conduct to their projects. I consider it good practise too, it shows care and increases the chances of getting people to contribute to your project
- [ ] Update the README
  It makes sense to link the contributing.md and the code of conduct

Right now, `make test` shows this output to me:

```sh
go test  -cover $(glide novendor) -run . -timeout=30s
ok      github.com/goreleaser/releaser/config   0.025s  coverage: 51.7% of statements
ok      github.com/goreleaser/releaser/config/git       0.068s  coverage: 100.0% of statements
?       github.com/goreleaser/releaser/pipeline [no test files]
ok      github.com/goreleaser/releaser/pipeline/brew    0.022s  coverage: 16.3% of statements
?       github.com/goreleaser/releaser/pipeline/build   [no test files]
ok      github.com/goreleaser/releaser/pipeline/compress        0.036s  coverage: 6.5% of statements
ok      github.com/goreleaser/releaser/pipeline/release 0.118s  coverage: 16.1% of statements
ok      github.com/goreleaser/releaser/split    0.027s  coverage: 100.0% of statements
ok      github.com/goreleaser/releaser/uname    0.021s  coverage: 100.0% of statements
?       github.com/goreleaser/releaser  [no test files]
```

I personally love I can get a sense of the coverage right away! 

The `make lint` command reports interesting findings too:

```sh
errcheck $(glide novendor)
config/config_test.go:35:10:    os.Chdir("./.test/1")
config/config_test.go:36:16:    defer os.Chdir(cwd)
config/config_test.go:49:10:    os.Chdir("./.test/2")
[...]

PR #30, opened 2017-01-02
added errcheck

caarlos0:
refs #28 and #29 

not sure this is the right way to fix the `defer x.Close()` issues... I'm basically ignoring the possible errors..

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -73,8 +73,14 @@ func assertFiles(t *testing.T, dir string, files []string) {
 	assert := assert.New(t)
 
 	cwd, _ := os.Getwd()
-	os.Chdir(dir)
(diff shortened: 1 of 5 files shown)


config/config_test.go
```diff
@@ -73,8 +73,14 @@ func assertFiles(t *testing.T, dir string, files []string) {
 	assert := assert.New(t)
 
 	cwd, _ := os.Getwd()
-	os.Chdir(dir)
-	defer os.Chdir(cwd)
+	if err := os.Chdir(dir); err != nil {
+		t.Error(err)
```
jspc: This is probably worth `panic`ing on to fail fast; it's not a problem with the code, it's a problem with the test or the underlying os

PR #33, opened 2017-01-02
added editorconfig

caarlos0:


jspc:
LGTM. Perhaps it'd be good to enforce these with some pre commit hooks running `go fmt` would be good, given these are practically that

Issue #42, opened 2017-01-03
replace oses and arches by goos and goarch?


caarlos0:
Seems more brainless... oses and arches may be confusing...

medhoover:
Why not using `os, arch` ?
Even better to support a combined syntax `-osarch linux/amd64`.
Taken from gox syntax https://github.com/mitchellh/gox#usage

caarlos0:
@medhoover because `goos` and `goarch` are clearer..

`os` and `arch` doesn't explicitly tells in what form they should be.

`goos` and `goarch` do...


I like the `gox` syntax (`goos/goarch`). Will think more about it, maybe for version 2...

PR #51, opened 2017-01-06
renaming Oses to GoOS and Arches to GoArch

caarlos0:
YAML will still be `oses` and `arches` for now.

Addresses #42


caarlos0:
check #81

PR #69, opened 2017-01-14
Mention Git tags. Closes #68.

### LINKED ISSUE #68: Mention that releasing uses Git tags
... and that GoReleaser expects a tag to exist.

jorinvo:
Feel free to improve on what I'm trying to say here.
I just think it's worth mentioning that you need to create a Git tag first to use GoReleaser.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# GoReleaser 
+# GoReleaser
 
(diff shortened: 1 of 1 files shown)


caarlos0:
Looks good, thanks!

PR #70, opened 2017-01-14
improved create/update formulae logic

### LINKED ISSUE #66: Brew releasing requires formula file to exist
Brew releasing only works when the formula file already exist in the specified Git repository.
I added an empty file to make it work.
I think the problem is the use of `UpdateFile` here: https://github.com/goreleaser/releaser/blob/master/pipeline/brew/brew.go#L74
Maybe we can consider using `CreateFile` when the updating fails?

caarlos0:
closes #66

### DIFF
--- a/pipeline/brew/brew.go
+++ b/pipeline/brew/brew.go
@@ -3,8 +3,6 @@ package brew
 import (
 	"bytes"
 	"context"
-	"crypto/sha256"
-	"fmt"
 	"log"
 	"strings"
 	"text/template"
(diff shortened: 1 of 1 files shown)


jorinvo:
Looks good to me :)
 
Guess for this 3rd party dependent things it's more work writing tests than what you gain in confidence from it...

caarlos0:
@jorinvo I intend on mocking everything and writing tests, but just when it's in pre-1.0 state...

right now everything can change and it doesn't worth IMO...

jorinvo:
@caarlos0 completely agree :)

PR #61, opened 2017-01-12
release init

caarlos0:
closes #47

@jspc what do you think?

PR #78, opened 2017-01-14
Optional config

### LINKED ISSUE #74: Make more configuration optional
I also like the idea of a quick way to create a default config file (https://github.com/goreleaser/releaser/issues/47), but many projects probably have the same configuration and wouldn't even need any configuration file if the defaults are set right.

- I think in most cases the `repo` option could be parse from the `git remote`. 

- Also, many Go binaries follow the convention of naming the binary like the directory containing the `main.go` file. This could be used as `binary_name`.

Would it be useful to make those two settings optional?

jorinvo:
This fixes #74 and removes the need for #61.

- [x] optional `repo`
- [x] optional `binary_name`
- [x] optional config file


caarlos0:
btw: conflict, I merged #76 before seein your PR, sorry :/

jorinvo:
@caarlos0 don't worry I'll update everything and fix the PR :)

jorinvo:
After a change this big it was simpler to rewrite the changes on top of the new architecture than merging them.
I documented the optionals in the README.
The `valid` package didn't seem necessary to me anymore. If it's still needed for some other reasons we can sure add it back.

caarlos0:
LGTM, thanks!

PR #81, opened 2017-01-14
Improved config documentation and the config itself

### LINKED ISSUE #42: replace oses and arches by goos and goarch?

### LINKED ISSUE #80: Add a complete and commented goreleaser.yml.example
I was thinking, maybe instead of showing everything that can be done in the README, we can add a `goreleaser.yml.example` with every section and comment everything there...

Maybe wrap it a unit test to prevent it from being outdated...

caarlos0:
What I did:

- created a `goreleaser.example.yml` with all options filled and commented
- renamed `Github` usages to `GitHub`
- renamed `compress` to `archive`
- removed customization out of readme and linked the example files.

While documenting stuff, I realized that some configs might be in the wrong place before, so I changed them:
- `files` now belong to `archive`
- `binary_name` now belongs to `build`
- `oses` and `arches` were renamed to `goos` and `goarch` ( closes #42 #51  )

closes #80 

This is a breaking change, but I think it's the right move. 
@jspc @jorinvo what do you guys think?


goreleaser.example.yml
```diff
@@ -0,0 +1,81 @@
...
+  # - Arch
+  # The default is `{{.BinaryName}}_{{.Os}}_{{.Arch}}`
+  name_template: "{{.BinaryName}}_{{.Version}}_{{.Os}}_{{.Arch}}"
+
+  # Archive format. Valid options are `tar.gz` and `zip`.
+  # Default is `zip`
+  format: zip
+
+  # Replacements for GOOS and GOARCH on the archive name.
+  # The left values should be a valid GOOS or GOARCH name, and the right
```
jorinvo: Maybe: "The keys should be valid GOOS or GOARCH values followed by your custom replacements."

goreleaser.example.yml
```diff
@@ -0,0 +1,81 @@
...
+release:
+  # Repo in which the release will be created.
+  # Default is extracted from the origin remote URL.
+  repo: user/repo
+
+
+# The brew section specifies how the formula should be created
+# Check this link for details: https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md
+brew:
+  # Reporitory to push the tap
```
jorinvo: "... to push the tap to"

goreleaser.example.yml
```diff
@@ -0,0 +1,81 @@
...
+  repo: user/homebrew-tap
+
+  # Folder inside the repository to put the formula.
+  # Default is the root folder
```
jorinvo: "root folder."

config/config.go
```diff
@@ -6,36 +6,41 @@ import (
 	yaml "gopkg.in/yaml.v1"
 )
 
-// Homebrew contains the brew section
-type Homebrew struct {
+// HomebrewConfig contains the brew section
```
jorinvo: I think the API would be simpler when going with the [convention](https://golang.org/doc/effective_go.html#package-names) of stripping the `Config` suffix for all types.
Then you could use them from other packages like `config.Homebrew` or `config.Project`.

caarlos0: Makes sense

caarlos0:
@jorinvo yeah, I'm trying to find all repos using it and will open a PR fixing the configs :)

PR #88, opened 2017-01-17
Update CONTRIBUTING.md

caarlos0:
a more fail-proof way of cloning the project.


jorinvo:
Good point. While you are at it, would you mind also changing `releaser` in this file to `goreleaser`?

Issue #72, opened 2017-01-14
Clarify usage of Git tags


jorinvo:
The main use case for GoReleaser seems to be running it via CI for Git tags only.
However, nothing stops you from running it on any other branch.

By running it on another branch GoReleaser still uses the latest Git tag to name the release, 
but for building it actually uses the source of the current active branch (if I don't miss anything here?).
Also, it's not possible to create a release that doesn't depend on a Git tag.

I would suggest to either always use the source of the latest tag
or to allow releasing without a relation to Git tags.

The first solution would require to get the source from the tag for building or to only allow building when the current branch is the latest tag.

The second solution would need more thinking about other scenarios for releasing Go projects.

caarlos0:
Yeah, right now it only releases tags...

I don't know what is best here... maybe this is a v2 thing...

jorinvo:
@caarlos0 I think releasing tags is definitely the most useful thing and I don't think we even need to support other use cases.

However, currently you can only correctly release a tag when you switch to the tagged branch before running GoReleaser.
If I'm on my `master` branch, for example, and I have one commit more there than my latest tag, The binary will be built including this one new commit. 
It should use the source of the tag instead of the current branch.

caarlos0:
Hmmm, that makes a lot of sense... If not that, at least exiting 1 to let the user know...

`git checkout` the tag may not be so simple, since there can be unchanged files, etc...

caarlos0:
I'll move it to v1

jorinvo:
I will look into this.

PR #98, opened 2017-01-19
Always use latest tag as source

jorinvo:
This should help with #72.
As you can see from the commits I also fixed another small bug and included it here.

To clean up `git stash` and `git checkout` I added a new interface `Cleaner`. 
Please let me know if you have suggestions for a better solution.

### DIFF
--- a/main.go
+++ b/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/goreleaser/goreleaser/pipeline/git"
 	"github.com/goreleaser/goreleaser/pipeline/release"
 	"github.com/goreleaser/goreleaser/pipeline/repos"
+	"github.com/goreleaser/goreleaser/pipeline/source"
 	"github.com/urfave/cli"
 )
 
@@ -26,6 +27,8 @@ var pipes = []pipeline.Pipe{
 	env.Pipe{},
 	git.Pipe{},
 	repos.Pipe{},
+
+	&source.Pipe{},
 
 	// real work
 	build.Pipe{},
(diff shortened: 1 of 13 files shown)


caarlos0:
It looks good, but I think it's more safe to just fail if the current ref isn't the tag ref.

Stashing and un-stashing things looks like a good idea, but the idea is that goreleaser would be integrated within the CI, so there isn't anything to stash.

If running locally, IMHO is better to alert the user that the code should not be changed and HEAD should be the tag while running goreleaser. The user might change files in another shell or editor while goreleaser is running and it will then be "wrong" anyways... 

What do you think?

jorinvo:
Personally, I can still see the use case for running GoReleaser locally and manually. 
In this case it's really convenient to not having to switch branches and stash changes. 

I also understand the issue of the additional complexity. I guess we would need to see in practice if any problems come up.

caarlos0:
@jorinvo yeah, ok

caarlos0:
yeah, travis won't let we stash changes...

```
Using source from latest tag
 -> Stashing changes
failed stashing changes: 
*** Please tell me who you are.
Run
  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"
to set your account's default identity.
Omit --global to set the identity only in this repository.
fatal: empty ident name (for <travis@testing-docker-25b1f0fa-18ec-44fb-a019-8cb8b231b923.(none)>) not allowed
Cannot save the current index state
exit status 1
```

https://travis-ci.org/goreleaser/goreleaser

Issue #94, opened 2017-01-18
get/latest could be better


caarlos0:
We can put `https://github.com/goreleaser/get/blob/master/latest` as a asset on our jekyll site, so the url become cleaner, e.g.: `http://goreleaser.github.io/get` (instead of `https://raw.githubusercontent.com/goreleaser/get/master/latest`)

jorinvo:
Also see https://github.com/goreleaser/goreleaser/issues/93#issuecomment-273438717, we should not only offer the `latest` version. Should give users the choice to keep using one stable version.

caarlos0:
@jorinvo that makes sense, maybe we should update docs then.

Using `go get` indeed makes more sense...

caarlos0:
About getting an specific version, I guess `gopkg.in` kind of solves that, no?

jorinvo:
@caarlos0 Nice! I didn't know about `gopkg.in` before :)
Should we reference it as the goto way of using GoReleaser? 
Or should we stick with the `latest` option as default? As soon as we are stable on 1.0 I think we should commit to a stable API in any case...

caarlos0:
@jorinvo I think that for now we can use `go get ` or leave as is.

As soon as v1 is released, we can change docs to use `http://gopkg.in/goreleaser/goreleaser.v1`, that way people using `go get` will still get the latest version, but can change it to v1 if needed.

medhoover:
About the version link, a solution for the version issue is currently being prepared.
[The Saga of Go Dependency Management](https://blog.gopheracademy.com/advent-2016/saga-go-dependency-management/)

caarlos0:
@medhoover yeah, but that could take a while...

caarlos0:
I guess we can close this one...

Issue #50, opened 2017-01-06
Homebrew formulae improvements


caarlos0:
The homebrew tap is a little hacky to support linuxbrew (getting os from uname).

I was thinking:

- is linuxbrew support really required?
- is anyone still using i386 macs?

Maybe we could simplify it by just using the `Darwin_x86_64` build directly.

caarlos0:
also check https://github.com/Linuxbrew/brew/issues/220

jorinvo:
Also would be nice to add the `sha` value to the formula.
Right now I get a warning when installing:
```
Warning: Cannot verify integrity of name.tar.gz
A checksum was not provided for this resource
For your reference the SHA256 is: ...
```

PR #99, opened 2017-01-21
Replaced README with docs/README

caarlos0:
@medhoover did an awesome job on #50, but it seems like most of the README was inside docs... so I just moved the docs/README over the README and did a few adjustments.

What do you guys think?


caarlos0:
this also closes #94

jorinvo:
Looks good to me! I also prefer it accessible right in the Readme. 
Could you also fix the link to the brew formula while you are at it? 
I'm without computer at the moment. 
Thanks :)

caarlos0:
@jorinvo sure!

PR #103, opened 2017-01-23
Keep all environment variables for the build command execution. 

stffabi:
- cmd.Env overwrites all environment variables of the goreleaser process.
- Go build may search for gcc in PATH and then fails with
  ```
  exec: "gcc": executable file not found in %PATH%
  ```

### DIFF
--- a/pipeline/build/build.go
+++ b/pipeline/build/build.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"strings"
 
 	"github.com/goreleaser/goreleaser/context"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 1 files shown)


caarlos0:
Thanks!

It is a valid suggestion, but the implementation can be simpler than that.

I opened the PR #103, take a look :)

stffabi:
@caarlos0 Thanks, I'm new to golang so there are still things I could learn :)

Just a quick question just for my understanding.
```
cmd.Env = append(cmd.Env, "GOOS="+goos, "GOARCH="+goarch)
cmd.Env = append(cmd.Env, os.Environ()...)
```

What happens if for example GOOS and GOARCH are already set from the outside? Won't this end up with having multiple GOOS/GOARCH entries? Or maybe the goos/goarch values form the build pipe get overwritten by the os.Environ() entries, depending on the implementation of cmd.Exec?

caarlos0:
@stffabi usually you won't have any GOOS or GOARCH set, but if you did, I guess that they will be overriden by the second append.

I'll change that, just in case. :)

stffabi:
I won't expect that these are set, but one never knows 😄.

caarlos0:
indeed, fixed on 415aadb346cf654e3310dd2ab5e6d513bca4f050 :)

stffabi:
Thanks.

PR #107, opened 2017-01-24
Mention single run for CI

### LINKED ISSUE #92: Update docs on how to avoid deploying multiple times
When running Travis CI I like to test multiple OSes and maybe even different Go version.
However, only one of them should deploy the binaries.
My current solution is a test like this:
https://github.com/qvl/sleepto/blob/master/.travis.yml#L9

I don't know if we should mention this in the docs.

Also, what I don't like about this solution is, that it deploys as soon as this particular CI run was successful. As far as I know there is no way of running code only after all OS-version combination have been completed?

jorinvo:
We cannot include code here since there are too many different scenarios. A note has to be enough. Closes #92.

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,8 @@ deployment:
       - goreleaser
 ```
 
+(Note that if you test multiple versions or multiple OSes you probably want to make sure GoReleaser is just run once)
+
 ---
 
 Would you like to fix something in the documentation? Feel free to open an [issue](https://github.com/goreleaser/goreleaser/issues).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -307,6 +307,8 @@ deployment:
       - goreleaser
 ```
 
+(Note that if you test multiple versions or multiple OSes you probably want to make sure GoReleaser is just run once)
```
caarlos0: maybe remove the parenthesis and prefix the line with `>` (so it become a quote?)

jorinvo: Not sure if it's a good fit since it's not an actual quote.
What do you think about making it italic?

PR #109, opened 2017-01-30
added fpm support

### LINKED ISSUE #10: create packages with fpm
See: https://www.reddit.com/r/golang/comments/5l3i9b/deliver_go_binaries_as_fast_and_easy_as_possible/dbsrx16/

caarlos0:
- we were using the tag as version, neither brew nor fpm really liked that. Now goreleaser removes the `v` prefix from the tag name and uses that as version for both cases
- added support for calling fpm (only dependencies optional supported for now)

What do you guys think?

Agree with the API? Once we agree, I'll update the docs!

closes #10

### DIFF
--- a/pipeline/fpm/fpm.go
+++ b/pipeline/fpm/fpm.go
@@ -1,0 +1,100 @@
+package fpm
+
+import (
+	"bytes"
+	"errors"
+	"log"
+	"os/exec"
+	"path/filepath"
+
+	"golang.org/x/sync/errgroup"
+
+	"github.com/goreleaser/goreleaser/context"
+)
+
+var linuxArchives = []struct {
+	Key  string
+	Arch string
+}{
+	{
+		Key:  "linuxamd64",
+		Arch: "x86_64",
+	},
+	{
+		Key:  "linux386",
+		Arch: "i386",
+	},
+}
+
+// ErrNoFPM is shown when fpm cannot be found in $PATH
+var ErrNoFPM = errors.New("fpm not present in $PATH")
+
+// Pipe for fpm packaging
+type Pipe struct{}
+
+// Description of the pipe
+func (Pipe) Description() string {
+	return "Creating Linux packages with fpm"
+}
+
+// Run the pipe
+func (Pipe) Run(ctx *context.Context) error {
+	cmd := exec.Command("which", "fpm")
+	if err := cmd.Run(); err != nil {
+		return ErrNoFPM
+	}
+	if len(ctx.Config.FPM.Formats) == 0 {
+		log.Println("No output formats configured")
+		return nil
+	}
+	var g errgroup.Group
+	for _, format := range ctx.Config.FPM.Formats {
+		for _, archive := range linuxArchives {
+			if ctx.Archives[archive.Key] == "" {
+				continue
+			}
+			archive := archive
+			g.Go(func() error {
+				return create(
+					ctx,
+					format.Name,
+					ctx.Archives[archive.Key],
+					archive.Arch,
+					format.Dependencies,
+				)
+			})
+		}
+	}
+	return g.Wait()
+}
+
+func create(ctx *context.Context, format, archive, arch string, deps []string) error {
+	var path = filepath.Join("dist", archive)
+	var file = path + ".deb"
+	var name = ctx.Config.Build.BinaryName
+	log.Println("Creating", file)
+
+	var options = []string{
+		"-s", "dir",
+		"-t", format,
+		"-n", name,
+		"-v", ctx.Version,
+		"-a", arch,
+		"-C", path,
+		"-p", file,
+		"--force",
+	}
+	for _, dep := range deps {
+		options = append(options, "-d", dep)
+	}
+	options = append(options, name+"="+filepath.Join("/usr/local/bin", name))
+	cmd := exec.Command("fpm", options...)
+	log.Println(cmd)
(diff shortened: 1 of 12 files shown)


caarlos0:
cc/ @jorinvo @jspc

pipeline/fpm/fpm.go
```diff
@@ -0,0 +1,90 @@
...
+	return "Creating Linux packages with fpm"
+}
+
+// Run the pipe
+func (Pipe) Run(ctx *context.Context) error {
+	if len(ctx.Config.FPM.Formats) == 0 {
+		log.Println("No output formats configured, skipping")
+		return nil
+	}
+	cmd := exec.Command("which", "fpm")
```
jorinvo: Maybe simplerto use *exec.LookPath* (https://golang.org/pkg/os/exec/#LookPath) instead?

pipeline/fpm/fpm.go
```diff
@@ -0,0 +1,90 @@
...
+		"-v", ctx.Version,
+		"-a", arch,
+		"-C", path,
+		"-p", file,
+		"--force",
+	}
+	for _, dep := range ctx.Config.FPM.Dependencies {
+		options = append(options, "-d", dep)
+	}
+	options = append(options, name+"="+filepath.Join("/usr/local/bin", name))
```
jorinvo: It's not really clear to me what this line does. Could you explain it? Even better, adding a code comment instead :)

pipeline/git/git.go
```diff
@@ -30,5 +34,7 @@ func (Pipe) Run(ctx *context.Context) (err error) {
 		PreviousTag: previous,
 		Diff:        log,
 	}
+	// removes usual `v` prefix
+	ctx.Version = strings.TrimPrefix(tag, "v")
```
jorinvo: I was checking and it seems that at least all popular Go projects follow the "vX.XX.XX" convention.
But I still think we shouldn't rely on it. If someone sets a tag name starting with "v" but not being of the same format, it might be a big surprise.
How about a quick check before trimming? A regex like "^v[0-9.]+" would do.

pipeline/release/release.go
```diff
@@ -67,9 +73,13 @@ func description(diff string) string {
...
 }
 
-func upload(client *github.Client, releaseID int, archive string, ctx *context.Context) error {
-	archive = archive + "." + ctx.Config.Archive.Format
-	file, err := os.Open("dist/" + archive)
+func upload(ctx *context.Context, client *github.Client, releaseID int, archive, format string) error {
+	archive = archive + "." + format
+	var path = filepath.Join("dist", archive)
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		return nil
```
jorinvo: These errors are just ignored? Shouldn't we fail if the archive doesn't exist?
Also, if it's an error *os.IsNotExist* doesn't catch it's currently not handled.

caarlos0: Yeah, I added this part because the current code will try to add the mac build `.deb` too..

In this case I'll just ignore it... If the archive wasn't generated and should, it should have failed in the archive or fpm pipe I think..

Maybe it would be better to add a list of uploads in the context...

pipeline/git/git.go
caarlos0: I actually made it error if the final version doesn't start with a number :D

pipeline/release/release.go
jorinvo: Makes sense. 
But since the intention here is not clear from just looking at the code it would be nice to add what you just described as a code comment.

pipeline/git/git.go
jorinvo: That's a little too opinionated for my taste. 
I think we should also allow users to use a custom version scheme. Maybe we can change it to not error and instead ignore other types of formats?

caarlos0: FPM won't work if the version doesn't start with a number apparently :/

caarlos0: Not sure if FPM or if the Debian format won't accept it...

jorinvo: If it's a requirement for Fpm we don't have a choice. 
But if a user only deploys to Github or Homebrew,  I think we should still allow other version formats. 
Maybe we can move the check to the fpm pipe?

caarlos0:
@jorinvo just to confirm, is a debian-ish thing:

```console
❯ docker run -v $PWD/dist:/tmp -it ubuntu /bin/bash
root@1b14f1638848:/# dpkg -i /tmp/test.deb
dpkg: error processing archive /tmp/test.deb (--install):
 parsing file '/var/lib/dpkg/tmp.ci/control' near line 2 package 'test':
 error in 'Version' field string 'v2.0.0': version number does not start with digit
Errors were encountered while processing:
 /tmp/test.deb
```

pipeline/git/git.go
caarlos0: Yeah, I thought about that too... 

Seemed kind of OK for me to make it standard for everyone (ldflag, cask, fpm, yet-to-be-implemented stuff).

jorinvo: Ok. Can leave it this way. Can still change it later if we get any troubles.

caarlos0: Seems like homebrew also ideally expects semantic versioning too (`^\d+\.\d+\.\d+`).

I'll just point that out in the README.

Issue #110, opened 2017-01-30
Release to GitLab


jorinvo:
As @bernardoVale mentioned  #on Slack it would be nice to support releasing to GitLab.
There is a Go package for the API https://github.com/xanzy/go-gitlab.
Since we also plan on adding Bintray support in #15 we might be able to combine some of the releasing logic.

jorinvo:
Currently we parse the default repository to push to from the *git remote*.
We explicitly check for "github" (https://github.com/goreleaser/goreleaser/blob/master/pipeline/defaults/remote.go#L21). 
But we could also make this more general and map from the detected remote to a service.

This would allow using GitLab without any configuration.

But for Bintray we would need additional configuration.

And we should also support use cases where people like to upload to multiple services.

I also think in some scenario a user might not want to upload the release at all and instead has a custom script or some manual process to run after GoReleaser is done (Maybe I should open another issue to make this a separate feature request).

Anyways, I think we should keep it simple so that a user doesn't need any configuration for the most basic use case and can use GiHub and GitLab out of the box.

For customisations we should probably split the `release` section by service.
For GitHub we currently only have one setting, but for other services there might be more options to set later on. Therefore I would create a config object per service:
```yaml
release:
  github: 
    repo: user/repo
  gitlab: 
    repo: user/repo
  bintray:
    ...
```

This would break the current configuration format. We could think about creating an alias from `release.repo` to `release.github.repo`.

We could also think about removing the `release` section completely and adding each service on the top level to the configuration.
`brew` is kind of the same as those other services to release to. Would be better to handle them all the same way I think.

And all service specific authentication I would handle via env vars like we already do for GitHub.

What do you think @caarlos0 @jspc?

caarlos0:
@jspc yeah, I have a WIP stash on my machine that does basically that... 

The only problem I see (and was also present on my branch) is that this allows both github and gitlab repo to be setup and also doesn't support custom hosts (github enterprise, gitlab hosted, etc).

Maybe we should do something like this:

```yaml
release:
  repo:
    provider: github
    name: user/repo
    host: my.server.com
```

jorinvo:
@caarlos0 does this configuration format allow using multiple providers at the same time?
And what do you think about seeing Homebrew releasing on the same level as GitHub, GitLab, ...?

zet4:
If GitLab support was to be added, would Gogs/Gitea be supported as well?

jorinvo:
[...]

PR #118, opened 2017-02-16
Changed confix syntax to accept multiple repository providers

caarlos0:
refs #110 

This will break compatibility with previous versions config.


caarlos0:
@jorinvo yeah, I had the same feeling while doing this... maybe we should just fallback to add a `provider` property instead of changing all this

jorinvo:
@caarlos0 good idea! I wasn't that happy with the name "repo" anyways since some possible future release targets / providers could also be of different shape than a repository.

codeskyblue:
if goreleaser.yml has a version, break changes can be easily commit

jorinvo:
The most flexible solution I can think of would be to have one configuration option `release` which is a list of different providers:
``` yaml
release:
  - github:
      repo: user/repo
      host: my.server.com
  - gitlab:
      repo: user/repo
      host: my.server.com
  - brew:
      repo: user/homebrew-tap
      folder: Formula
      dependencies:
        - git
        - zsh
```

This format would allow as many different release as anyone desires.
Each item would also be easy parsable by keeping settings for each type of provider in it's own property.
And yes, I would also treat brew as a release target. 

I don't think this would complicate things and it wouldn't stop use from keeping the default behavior of having no configuration at all.

caarlos0:
@jorinvo yeah, we shall break compatibility to do that though...

Maybe we should add the new syntax and start to warn about the new syntax and let it be for a while...

jorinvo:
@caarlos0 we can keep supporting the old syntax by keeping it as fallback.
First we try to parse the new format. If it didn't work we parse the old format.

titpetric:
Just throwing it out there: what if you'd like to publish to two different github repositories? Would it be enough just to add:

```
- github1:
    repo: repo/name
    host: github.com
- github2:
    repo: repo/name2
   host: github.com
```

What exactly are the asumptions behind this? Does it need a HTTP/S API to push releases or is it managed with a git somehow (branch, remote?)? I actually wouldn't be that terribly upset if I deployed releases to a special git branch and pushed them to whatever origins I have (...). Maybe that's a different PR altogether :)

jorinvo:
@titpetric 

- My idea was to make it a simple list as seen above. That way you can add as many release targets of the same type as you like.

Having two GitHub repos to release to would look like:
 
``` yaml
release:
  - github:
      repo: user/repo
  - github:
      repo: user2/repoX
      # This one being a GitHub enterprise instance
      host: my.server.com
```

As you can see `github` is just a property set for one item in the `release` list. 
For each item we would need to restrict it to setting only one provider property though.

- The `github` release target is only intended to create regular GitHub releases. It should work with github.com and also GitHub enterprise systems.
For other release mechanisms that use Git in some other way we would need to create a new provider.

caarlos0:
@jorinvo I agree with your solution. I'll start a new branch for it.

About breaking compatibility, I guess this is the time to do it. We are not 1.0 stable yet, so, better now than later.

caarlos0:
I re-implemented this in #145

caarlos0:
@jorinvo I end up with an even simple syntax, check #145 when you can :)

PR #119, opened 2017-02-21
First implementation of plist field for brew formulae

lucapette:
I need a plist field in one of my go projects and I did not want to give up on goreleaser :)  So I drafted a PR to discuss how to better add the feature. 

It's already working as it is but I'm not sure this is the right direction. A couple of notes:

- Wherever I added fields to structs I felt the need to reorder the fields alphabetically and put an empty line between groups (I would group the fields by type). But that's probably too personal and I'm not sure you'd like that.
- I changed the tests very little because I'm not sure you'd like to have more there. The rest of the brew tests follow the same approach.

I've tried the branch [here](https://github.com/lucapette/homebrew-tap/blob/master/tracker.rb) and it works fine.

This project gets better every time I use it. Thank you so much for your work!

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,6 +274,11 @@ brew:
   conflicts:
     - svn
     - bash
+
+  # Packages that run as a service
+  plist:
+    <?xml version="1.0" encoding="UTF-8"?>
+    ...
 ```
 
 By defining the `brew` section, GoReleaser will take care of publishing the Homebrew tap.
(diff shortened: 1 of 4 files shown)


caarlos0:
I'm OK with this.

Our tests are indeed not sufficient in many cases because the API may not be very stable yet. It should improve by the v1.0.0 release :)

BTW your tracker project looks great :D

lucapette:
@caarlos0 somehow I missed the notification! Thank you very much for merging it!

PR #122, opened 2017-02-22
support for custom build flags

### LINKED ISSUE #120: Support add build flag -tags
For example

```
go build -tags dev
```

caarlos0:
closes #120

### DIFF
--- a/pipeline/build/build.go
+++ b/pipeline/build/build.go
@@ -51,7 +51,7 @@ func build(name, goos, goarch string, ctx *context.Context) error {
 	}
 	cmd := []string{"go", "build"}
 	if ctx.Config.Build.Flags != "" {
-		cmd = append(cmd, strings.Split(ctx.Config.Build.Flags, " ")...)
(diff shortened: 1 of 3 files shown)


pipeline/build/build.go
```diff
@@ -49,7 +49,11 @@ func build(name, goos, goarch string, ctx *context.Context) error {
 			return err
 		}
 	}
-	cmd := []string{"go", "build", "-ldflags=" + ldflags, "-o", output, ctx.Config.Build.Main}
+	cmd := []string{"go", "build"}
+	if ctx.Config.Build.Flags != "" {
+		cmd = append(cmd, strings.Split(ctx.Config.Build.Flags, " ")...)
```
codeskyblue: Is  `strings.Fields(ctx.Config.Build.Flags)` better?

jorinvo: Good call! Even though in this case it probably doesn't affect the behaviour much, it makes the intent more obvious :)

caarlos0: Indeed, I didn't know about that function, always good to learn, Thanks 😄

Issue #132, opened 2017-03-01
Skip upload to GitHub


titpetric:
I would like to build the binaries only, because either:

1. I'm not publishing them to GitHub but some other service (or may not even be using Git),
2. I'm separating build/release stages to run some tests/whatever before doing a release

Currently there are two requirements from goreleaser, in regards to this:

- we MUST set `release/repo` in yaml (or have a git repository available where we are running),
- we MUST set GITHUB_TOKEN

I'd like to skip both however. Ideally I would like to make these steps configurable with something like:

~~~
goreleaser build
goreleaser deploy
~~~

The first one would just build the required binaries/run pre/post hooks, the second one would push whatever is available in the dist/ folder (and could be replaced with something else, like `rsync`, or other service).

jorinvo:
With the new syntax (https://github.com/goreleaser/goreleaser/issues/110#issuecomment-279025617) we should allow something like:

```yaml
release:
  disabled: true
```

I also think it would be helpful to add `build` and `deploy` sub commands as you suggested.
Maybe we can move this part to a separate issue?

titpetric:
@jorinvo done, also recreated the PR just to verify Travis is building it. 👍

titpetric:
@jorinvo any chance of merging this in? I'm wrinting a [12FA book](https://leanpub.com/12fa-docker-golang) and would love to include goreleaser in it.

jorinvo:
@titpetric have to discuss this in #134 :)

caarlos0:
check #136

PR #133, opened 2017-03-01
Skip upload to GitHub

titpetric:
This implements #132 

* Repo can now be set to "none" for github or brew, allowing to skip the release stage (brew modified only for compatibility),
* Defaults are unchanged (if repo is empty, it will still try to detect it with `remoteRepo` (only with explicit "none" the release step is skipped)
* GITHUB_TOKEN isn't validated in `env` anymore, but when creating the GitHub client. If the release step is skipped, then GITHUB_TOKEN is not needed

### DIFF
--- a/pipeline/brew/brew.go
+++ b/pipeline/brew/brew.go
@@ -85,7 +85,7 @@ func (Pipe) Description() string {
 
 // Run the pipe
 func (Pipe) Run(ctx *context.Context) error {
-	if ctx.Config.Brew.Repo == "" {
+	if ctx.Config.Brew.Repo == "" || ctx.Config.Brew.Repo == "none" {
 		return nil
 	}
 	client := clients.GitHub(ctx)
--- a/pipeline/brew/brew.go
+++ b/pipeline/brew/brew.go
@@ -88,7 +88,10 @@ func (Pipe) Run(ctx *context.Context) error {
 	if ctx.Config.Brew.Repo == "" || ctx.Config.Brew.Repo == "none" {
 		return nil
 	}
-	client := clients.GitHub(ctx)
+	client, err := clients.GitHub(ctx)
+	if err != nil {
+		return err
+	}
 	path := filepath.Join(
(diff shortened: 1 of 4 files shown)


pipeline/release/release.go
```diff
@@ -22,7 +22,13 @@ func (Pipe) Description() string {
 
 // Run the pipe
 func (Pipe) Run(ctx *context.Context) error {
-	client := clients.GitHub(ctx)
+	if ctx.Config.Brew.Repo == "" || ctx.Config.Brew.Repo == "none" {
```
jorinvo: Why skip the Github release when Brew is skipped?

pipeline/brew/brew.go
```diff
@@ -85,7 +85,7 @@ func (Pipe) Description() string {
 
 // Run the pipe
 func (Pipe) Run(ctx *context.Context) error {
-	if ctx.Config.Brew.Repo == "" {
+	if ctx.Config.Brew.Repo == "" || ctx.Config.Brew.Repo == "none" {
```
jorinvo: Does it even need the `"none"` behavior or can you directly set it to `""` in the YAML config?
I haven't tried it though.

pipeline/release/release.go
titpetric: My bad, fixing

pipeline/brew/brew.go
titpetric: For brew you can, but GitHub is a requirement for brew, and I wanted to keep pairity (ie, if github = none, brew = none). Can't say there's a better way

titpetric:
@jorinvo thanks for the comment & review. Travis seems to be having some issues so I'll just close and reopen this PR.

jorinvo:
@titpetric yeah, travis.org already has issues since yesterday. But it's all over the place. Seems to be everything S3 https://news.ycombinator.com/item?id=13755673 ...

PR #134, opened 2017-03-01
Skip upload to GH

titpetric:
This implements #132, recreated PR to confirm travis build

### DIFF
--- a/clients/github.go
+++ b/clients/github.go
@@ -6,10 +6,16 @@ import (
 	"golang.org/x/oauth2"
 )
 
+// ErrMissingToken indicates an error when GITHUB_TOKEN is missing in the environment
+var ErrMissingToken = errors.New("Missing GITHUB_TOKEN")
+
 // GitHub client for the given token
-func GitHub(ctx *context.Context) *github.Client {
+func GitHub(ctx *context.Context) *github.Client, error {
+	if ctx.Token == "" {
+		return nil, ErrMissingToken
+	}
 	ts := oauth2.StaticTokenSource(
 		&oauth2.Token{AccessToken: ctx.Token},
 	)
-	return github.NewClient(oauth2.NewClient(ctx, ts))
+	return github.NewClient(oauth2.NewClient(ctx, ts)), nil
 }
--- a/clients/github.go
+++ b/clients/github.go
@@ -1,6 +1,8 @@
 package clients
 
 import (
+	"errors"
+
 	"github.com/google/go-github/github"
 	"github.com/goreleaser/goreleaser/context"
 	"golang.org/x/oauth2"
@@ -10,7 +12,7 @@ import (
 var ErrMissingToken = errors.New("Missing GITHUB_TOKEN")
 
 // GitHub client for the given token
-func GitHub(ctx *context.Context) *github.Client, error {
+func GitHub(ctx *context.Context) (*github.Client, error) {
 	if ctx.Token == "" {
 		return nil, ErrMissingToken
 	}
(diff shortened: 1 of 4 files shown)


caarlos0:
I would prefer to add a `skip bool` to the config instead of setting the repo to `none`.

What do you think?

titpetric:
This change is backwards-compatible in the sense that I'm pretty sure that nobody uses "none" as their github account/repository name (*cough*), and if you skip setting it, it keeps the autodetect feature for GH.

If you're in favor of `skip`, it should be applied to `brew` as well, and brew as I suspect would have a different default (skip=true), which begs the question if setting `repo=...` is enough for brew, or would you also have to set `skip=false`?

Right now I'm for merging this in (self-serving, I'm aware :)).

In terms of 1.0, I would refer to #135 with specific extension:

* `goreleaser push -u github` - push a release to GitHub and set to "track" changes
* `goreleaser push s3` - push a release to S3...
* `goreleaser push` - push to all tracked destinations

And you could even have `goreleaser init` that would ask you for some defaults. I think [npm](https://docs.npmjs.com/cli/init) and [glide](https://github.com/Masterminds/glide)** are good examples of what kind of API I'd want/need from goreleaser. Whatever is written in the yml should be in the background. :)

caarlos0:
I still don't like the using of `repo: none`. It would require some special explanation on the `repo` sections on README, while skip is good enough.

Default for bools is false, so skip would always be false by default. In the case of `brew`, I guess we should check for `!(skip || repo == "")`.

caarlos0:
About the commands, I'm not sure.

The idea of goreleaser is to wrap it up with the CI and let it work automatically. It is not an interactive tool like others around (for node, for example). 

Other than that, I agree that the idea of being able to skip things is valid, so, maybe it makes more sense to just add flags like `skip-build`, `skip-release` etc.  I'm not yet sure what is the best way of doing all this though

titpetric:
Your call. I think either one needs documentation, and I don't like that
skip has "special" cases from the start. The opinion about 1.0 I hope is
valid :)

titpetric:
I get your point about "one command for all" for CI. I just wish there was a sane way to specify deployment and order. Currently "release" is "hidden away" (even by naming it "release" and not "github").

One could of course specify this all in yml with something like this:

```
deploy:
  - github:
    token: [...]
    repo: ...
  - brew
    token [...]
    repo: ...
```

and then completely remove any of these sections. Again, this is just thinking out loud for 1.0 (or some point down the line). I would go somewhere in the way like buildkite does for their build pipelines with (free signup and quota for 30 days i think). Currently, there's too much of fixed structures that are getting in the way a bit :)

titpetric:
I'm aware currently this is hardcoded in pipe.go I think. Anyway, would be nice to have a separate build/deploy step, if one needs it. It's what I'm trying to do :)

jorinvo:
While keeping the main focus in running GoReleaser automated via CI,
I still like the idea of helping out users with other scenarios aswell.
Add additional commands without changing the default behavior and adding more flexible options for  using multiple release targets (or none) would be something I think is in the scope of this projects.
But @caarlos0 should decide where things are going :)

titpetric:
Ok, let me know if I should work on skip: true, or perhaps something more along the lines of this:

~~~
skip:
  - github
  - brew
~~~

caarlos0:
Let me think about this for a while :)

jorinvo:
Since the title of this PR is "Skip upload to GH" implementing #135 would also be sufficient solution for this problem, right?

titpetric:
All three ways work for me :) mainly I'm interested in demonstrating build
by itself, and then release by itself - as release part on my side is more
than just github/brew (creating a docker container & releasing that). If
you have any preference, please let me know and I can give a hand.

caarlos0:
check #136

PR #136, opened 2017-03-04
allow skipping release processes

### LINKED ISSUE #135: Add helper commands
Hello,

it would be helpful to add helper commands, that execute only part of the build/release processes. For example:

```
gorelease build
gorelease deploy/push github
gorelease deploy/push brew
```

I'm thinking about it mainly from the perspective that sometimes I want only to make a build (but not a deploy), and that there should be some familiarity to git for example (git status, git push,...).

caarlos0:
closes #135 #134 #132 

Basically, the `--skip-release` flag allows to only build and archive, skipping validations and the upload to github and others.

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,23 +22,6 @@ import (
 
 var version = "master"
 
-var pipes = []pipeline.Pipe{
-	// load data, set defaults, etc...
-	defaults.Pipe{},
-	env.Pipe{},
-	git.Pipe{},
-	repos.Pipe{},
-
-	source.Pipe{},
-
-	// real work
-	build.Pipe{},
-	archive.Pipe{},
-	fpm.Pipe{},
-	release.Pipe{},
-	brew.Pipe{},
-}
-
 func main() {
 	var app = cli.NewApp()
 	app.Name = "goreleaser"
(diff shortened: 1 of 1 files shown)


titpetric:
Lovely. I'll check it out in the next few days :)

titpetric:
I hope I'm not running an old version, but it seems that release: repo: is still required even with --skip-release. It still tries to get Git info. I'm trying to build a project without any git repo.

~~~
Getting Git info
exit status 128: fatal: Not a git repository (or any parent up to mount point /go/src/app)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
~~~

If I at least do a 'git init' and a dummy commit, but it's already getting hack-y.

caarlos0:
Yeah, without a git repo it would still not work...

titpetric:
Alas, with my `none` hack it skipped it completely :) Ah well. I'll see what I can do, will have to move on if I can't have separate build / release stages, but maybe I'll come back to revisit goreleaser :(

caarlos0:
check #137

PR #137, opened 2017-03-06
skip the git log/diff as well

caarlos0:
refs https://github.com/goreleaser/goreleaser/pull/136#issuecomment-284532299


titpetric:
Looks fine to me, I can take a spin when it's merged.

PR #139, opened 2017-03-09
homebrew: add custom install command

morpheu:
Add custom install script with multiple commands support for homebrew formulas

### DIFF
--- a/pipeline/brew/brew.go
+++ b/pipeline/brew/brew.go
@@ -37,9 +37,21 @@ const formula = `class {{ .Name }} < Formula
   {{- end }}
   {{- end }}
 
+  {{- if .Install }}
+
+  def install
+    {{- range $index, $element := .Install }}
+    {{ . -}}
+		{{- end }}
+  end
+
+	{{- else }}
+
   def install
     bin.install "{{ .BinaryName }}"
   end
+
+	{{- end }}
 
   {{- if .Caveats }}
 
@@ -71,6 +83,7 @@ type templateData struct {
 	Format       string
 	SHA256       string
 	Plist        string
+	Install      []string
 	Dependencies []string
 	Conflicts    []string
 }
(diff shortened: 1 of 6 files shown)


pipeline/brew/brew.go
```diff
@@ -37,10 +37,22 @@ const formula = `class {{ .Name }} < Formula
   {{- end }}
   {{- end }}
 
+  {{- if .Install }}
+
+  def install
+    {{- range $index, $element := .Install }}
+    {{ . -}}
+		{{- end }}
```
caarlos0: this `{{- end }}` seem unaligned...

pipeline/brew/brew.go
```diff
@@ -37,10 +37,22 @@ const formula = `class {{ .Name }} < Formula
...
+  end
+
+	{{- else }}
```
caarlos0: wouldn't it be simpler to set the default `bin.install "binary"` when the config doesn't set anything?

If so, you could just remove the entire `if` block, because the `range` would deal with everything, no?

pipeline/brew/brew.go
```diff
@@ -186,6 +177,12 @@ func dataFor(
 	} else {
 		description = *rep.Description
 	}
+	if ctx.Config.Brew.Install != "" {
+		installCmds = ctx.Config.Brew.Install
+	} else {
+		installCmds = "bin.install \"" + ctx.Config.Build.BinaryName + "\""
```
caarlos0: there is a [`defaults` pipe](https://github.com/goreleaser/goreleaser/blob/master/pipeline/defaults/defaults.go) that is used to set defaults, can you add it there?

morpheu: Sure, I didn't see it :) Sorry

caarlos0: np :)

caarlos0:
Thanks!

PR #140, opened 2017-03-16
homebrew: allow choose which versions creates brew formulas

morpheu:
Allow homebrew task only run for some version match. Useful for only build it for stable releases, for example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,6 +269,10 @@ brew:
   # Default is empty.
   caveats: "How to use this binary"
 
+  # Choose which versions brew should be build. Eg: for stable releases only without -rc suffix
+  # Default to any version
+  version_match: '^[0-9\.]+$'
+
   # Dependencies of your package
   dependencies:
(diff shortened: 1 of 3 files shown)


jorinvo:
You could also achieve this by customising your build script (for example in your `.travis.yml` file) to use different GoReleaser config files for different versions (via `goreleaser --config file.yml`).

caarlos0:
@morpheu I agree with @jorinvo .  I think it's easy enough to hack it via travis.yml itself.

I'll close this, if you have a different opinion, feel free to comment it anyways.

Cheers!

morpheu:
I think it's kind ugly to have different yamls to accomplish almost same goal including some travis trick instead having only one single field in yaml file. Homebrew formulas only make sense is most of time when you are working with stable releases - sure, there is a devel option, but for now goreleaser doesn't have this option for create homebrew formulas

PR #145, opened 2017-03-23
changing repo config

### LINKED ISSUE #93: Change `binary_name` to `binary`
Wouldn't this still be descriptive enough, especially since it is in the `build` sub-section of the configuration file now?

caarlos0:
Basically changing all configs that need to be changed for v1!

- `repo` is now a struct and for now is only available as github, e.g. `release.github` and `brew.github`. We may add more later to implement #110 for example
- changed from `binary_name` to `binary` only. This should fix #93

Side notes:

- This shall be released in `0.8.[4-6]` (this week), and I plan to remove the previous syntaxes by `0.9.\d` (2 weeks from now maybe?)
- Some parts of the code look ugly right now, I know, but since I'll remove them very soon, I'll leaving them as-is (including the lack of unit tests)

Next steps:

- We may now be able to start implementing GitLab support and etc (assuming a cleaned up branch without the ugly `if` statements)


caarlos0:
cc/ @jorinvo @jspc

jorinvo:
Nice work @caarlos0!
Thanks for adding all the deprecation warnings! 
I think we should give people more than two weeks to update their configuration files though 😉 

Your decision to go for a hash for the different providers seems go to me. It keeps things cleaner and simpler than my initial list idea. 
The trade-offs are that the user has no control over the order of releases and the user cannot release to the same type of provider twice within one configuration file. 
I think you made the right decision to keep things simple instead.

And just as a note fore myself, there are three changes here:
- `build.binary_name` is renamed to `build.binary` 
- `release` syntax is changed to a struct of providers with `release.github` implemented
- `brew` section is also changed to the same new syntax as `release`

caarlos0:
@jorinvo fair enough, shall we make it ~1 month then? Let's say, remove it by Apr 21?

jorinvo:
@caarlos0 sounds better to me :) I mean, it doesn't really stop us from moving on while keeping the deprecated code.

PR #148, opened 2017-03-23
pre/post-build hooks should only run once

### LINKED ISSUE #146: Support for upx
I've been [using `upx --brute` to reduce the size of my binaries](https://blog.filippo.io/shrink-your-go-binaries-with-this-one-weird-trick/) by 40%. 

I'd like to use goreleaser in my releasing pipeline, but I found the post hook does not run at the right time to add `upx`.

I was wondering, would you be interested in specifically adding a flag for automatically running `upx` after building in goreleaser? If so, I can work on a PR.

caarlos0:
They are running for each build, which doesn't even make sense since we don't pass the binary just built as a param to the underneath script.

This patch makes it run only one time, the `pre-hook` before all builds, and the `post-build` after all builds are done.

A valid use case of this is described in #146. With this, we can now do:

> goreleaser.yml:
```yml
  hooks:
    post: ./upx.sh
```

> upx.sh:
```sh
#!/bin/bash
set -ex
upx dist/goreleaser*/goreleaser*
```

closes #146

### DIFF
--- a/pipeline/build/build.go
+++ b/pipeline/build/build.go
@@ -47,7 +47,9 @@ func (Pipe) Run(ctx *context.Context) error {
 			})
 		}
 	}
-	err := g.Wait()
+	if err := g.Wait(); err != nil {
+		return err
+	}
 	if ctx.Config.Build.Hooks.Post != "" {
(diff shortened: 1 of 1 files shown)


schollz:
This works great for me.

Thanks a lot for the awesome tool!

Issue #162, opened 2017-04-11
Upload checksums for binaries produced


esiqveland:
It would be nice if `gorelease` would also publish signatures or checksums, of the binaries produced.

caarlos0:
Nice suggestion! Would be glad to see a PR with that =D

esiqveland:
I think this should be two features then. One for checksums (easy) and one for signatures (harder, requires config).

After having a look around it should be possible to make checksums as a pipe, and then either upload checksum files as artifacts or include the checksums in the release description text. Or ideally both.

I think checksums should be done first as it is easier to implement and gives much of the value.

What do you think about that?

esiqveland:
I have two suggestions for this. One is checksumming the bundled archives, and one is checksumming each file produces, and including this checksum inside the archive. I think I like the first approach better as it lets you check the archive before unpacking it, which might not even work if the archive is corrupt.

This means I am scrapping this approach.
Getting the checksums to be uploaded during the release phase is a bit trickier and doesn't really work with how the release pipe is coded now, as it tries to guess the final path of each file to upload using a fixed format.

On another note, I found this TODO: https://github.com/goreleaser/goreleaser/blob/master/pipeline/release/release.go#L58

Such a list of files to upload would be very helpful for this feature.

caarlos0:
@esiqveland yeah, the first approach seems to make more sense to me.

Just need to be careful to not breadk th ability to run goreleaser multiple times, as dicussed in #49.
It should be simple, just append to file and the second time goreleaser is ran it should update the checksums file, which will then contain the checksums for all the binaries produced...

caarlos0:
I just fixed that TODO. Moving to the Checksumming now

esiqveland:
Alright, you can see my suggestion here:
https://github.com/goreleaser/goreleaser/pull/169

PR #169, opened 2017-04-14
Artifacts: create checksums of archive

esiqveland:
Adds a pipeline for creating checksums of archive and adds them to artifact list for releasing.
Currently supports sha256.

Adresses #162


caarlos0:
@esiqveland sorry, only now I saw your PR.

Anyways, I believe my impl is more complete, check #170 (with md5sum et al)

PR #170, opened 2017-04-14
Uploading archive checksums

### LINKED ISSUE #162: Upload checksums for binaries produced

caarlos0:
- For each artifact, creates a `artifact.checksums` file, containing both MD5 and SHA256 sums
- Upload those checksums into the release as well

Example release created with this branch: https://github.com/caarlos0/test/releases/tag/v1.1.2

closes #162


caarlos0:
Question: should this be configurable?

esiqveland:
I see, I like this approach better. 

I don't think it should be configurable, but I think a sha256 would suffice. Keeping one checksum per file also makes it easier to automate using unix tools like `shasum` and `mdsum`.

pipeline/checksums/checksums.go
```diff
@@ -0,0 +1,64 @@
...
+
+// Description of the pipe
+func (Pipe) Description() string {
+	return "Calculating checksums"
+}
+
+// Run the pipe
+func (Pipe) Run(ctx *context.Context) (err error) {
+	var g errgroup.Group
+	for _, artifact := range ctx.Artifacts {
```
esiqveland: Is it safe to loop over `ctx.Artifacts` while doing operations that add to it?
See `ctx.AddArtifact` on line 62.

caarlos0: Yep. AFAIK when you `range list` it is the state of the list at that given time, later operation on it will not affect the loop.

e.g: https://play.golang.org/p/Xcr6YRkmGo

caarlos0:
> Keeping one checksum per file also makes it easier to automate using unix tools like shasum and mdsum.

I think automation is still quite easy:

```
~/Code/Go/src/github.com/caarlos0/test/dist master
❯ grep $(gmd5sum test_linux_64bits.deb | cut -f1 -d' ') test_linux_64bits.deb.checksums
md5sum 3ab92ebbd33ca1fa954cf84699183b7d

~/Code/Go/src/github.com/caarlos0/test/dist master
❯ grep $(gsha256sum test_linux_64bits.deb | cut -f1 -d' ') test_linux_64bits.deb.checksums
sha256sum 8ac9b8b5ed1c095b8cf7d3d6a8460ea5c2505e7ac90670dd7e82b805bf5b9149
```

pipeline/checksums/checksums.go
esiqveland: I'm only asking because I had an issue with this in my first version.

It seem to not be safe at least ref (https://golang.org/ref/spec#For_statements):

> If map entries are created during iteration, that entry may be produced during the iteration or may be skipped

esiqveland:
👍

caarlos0:
We could maybe put the summings like this:

```
3ab92ebbd33ca1fa954cf84699183b7d  test_linux_64bits.deb
8ac9b8b5ed1c095b8cf7d3d6a8460ea5c2505e7ac90670dd7e82b805bf5b9149  test_linux_64bits.deb
```

Then we could check with:

```
~/Code/Go/src/github.com/caarlos0/test/dist master
❯ gsha256sum -c test_linux_64bits.deb.checksums
test_linux_64bits.deb: OK
gsha256sum: WARNING: 1 line is improperly formatted

~/Code/Go/src/github.com/caarlos0/test/dist master
❯ gmd5sum -c test_linux_64bits.deb.checksums
test_linux_64bits.deb: OK
gmd5sum: WARNING: 1 line is improperly formatted
```

What do you think?

esiqveland:
I think that is even better. It also opens up for adding more checksums for any potetial extra archives in a more standardized way.

caarlos0:
Either way, I agree on your point about MD5. I'll remove it. Simpler is better =)

pipeline/checksums/checksums.go
caarlos0: Yeah, but it is not a map, it's a slice :)

esiqveland: Ah, I thought it was a map ;)

Issue #176, opened 2017-04-17
Better checksums


caarlos0:
Isn't kind of strange to have several `.checksums` files?

Maybe a `{{ .BinaryName }}_checksums.txt` with all checksums would be better...

What do you guys think?

caarlos0:
cc/ @esiqveland @jorinvo @jspc

caarlos0:
Something like this:

```diff
diff --git a/pipeline/checksums/checksums.go b/pipeline/checksums/checksums.go
index a8323a0..6e11042 100644
--- a/pipeline/checksums/checksums.go
+++ b/pipeline/checksums/checksums.go
@@ -4,8 +4,8 @@ package checksums
 
 import (
 	"fmt"
-	"io/ioutil"
 	"log"
+	"os"
 	"path/filepath"
 
 	"github.com/goreleaser/goreleaser/checksum"
@@ -23,31 +23,38 @@ func (Pipe) Description() string {
 
 // Run the pipe
 func (Pipe) Run(ctx *context.Context) (err error) {
+	file, err := os.OpenFile(
+		filepath.Join(
+			ctx.Config.Dist,
+			fmt.Sprintf("%v.%v", ctx.Config.Build.Binary, "-checksums.txt"),
+		),
+		os.O_APPEND|os.O_WRONLY|os.O_CREATE|os.O_EXCL,
+		0644,
+	)
+	if err != nil {
+		return err
+	}
+	ctx.AddArtifact(file.Name())
+	defer func() {
+		_ = file.Close()
+	}()
 	var g errgroup.Group
 	for _, artifact := range ctx.Artifacts {
 		artifact := artifact
 		g.Go(func() error {
-			return checksums(ctx, artifact)
+			return checksums(ctx, file, artifact)
 		})
 	}
 	return g.Wait()
 }
 
-func checksums(ctx *context.Context, name string) error {
+func checksums(ctx *context.Context, file *os.File, name string) error {
 	log.Println("Checksumming", name)
 	var artifact = filepath.Join(ctx.Config.Dist, name)
 	sha, err := checksum.SHA256(artifact)
 	if err != nil {
 		return err
 	}
-	var file = filepath.Join(
-		ctx.Config.Dist,
-		fmt.Sprintf("%v.%v", name, "checksums"),
-	)
-	var content = fmt.Sprintf("%v\t%v\n", sha, name)
-	if err := ioutil.WriteFile(file, []byte(content), 0644); err != nil {
-		return err
[...]

PR #177, opened 2017-04-17
better checksums

caarlos0:
Proposal for #176 

This way, a user can download all archives and a single checksum file and do:

```sh
$ sha256sum -c test_checksums.txt
test_linux_64bits.tar.gz: OK
test_linux_64bits.deb: OK
test_macOS_64bits.tar.gz: OK
```

Example release: https://github.com/caarlos0/test/releases/tag/v2.0.1

### DIFF
--- a/pipeline/checksums/checksums.go
+++ b/pipeline/checksums/checksums.go
@@ -4,8 +4,8 @@ package checksums
 
 import (
 	"fmt"
-	"io/ioutil"
 	"log"
+	"os"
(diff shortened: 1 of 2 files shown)


esiqveland:
Usually you'll only want to download one archive (the one for your platform). How does this interact with that usecase?

caarlos0:
Will show OK for the file you have and not found for the others, but still exit 0 :)

Issue #183, opened 2017-04-19
Add hook to insert custom release notes


bep:
First, great job on this!

I'm in the process of automating the semi-manual release process of Hugo with Goreleaser. None of the points mentioned below are blockers in that regard, and I think these are connected enough to warrant keeping them in this single issue.

* Release notes are very important for Hugo. A release may contain 200 commits, and while the commit log is nice and all, it doesn't make sense to put it in the changelog as-is. So we create our own release notes, semi-manually, and that is fine. **But there is currently no hook in Goreleaser to insert this.**
* The release notes are part of the release. I.e. they should be in a file in the released source/tag. **The release notes in Goreleaser are created after the tag is created and not part of the distribution.**
* Goreleaser creates a release on GitHub and publishes it. I wish there were an option to create a draft, so it could be possible to polish the release notes by hand etc. before releasing it to the world.

In the bullets above, I would hope to get from Goreleaser, in prioritized order:

1) Hook to insert custom release notes.
2) A flag to publish release as GitHub draft.

I understand the you have to make some priorities here, but I suggest that you for 1.0 at least think about any future changes that may come in this department. I think you underestimate the value of the release notes.

But again, the above are no blockers for me. This is thousands times better than what I had (I'm not even sure if the latest Hugo is available via Brew yet, and our build process is incredibly spread out) -- and once I get it up and running, I may have a look doing some contribution for the Chocolatey integration (Windows) and maybe also Snapcraft.

caarlos0:
@bep Hi, I agree, the release notes are not optimal. 

Just so I can fully understand, what exactly would the hook do?

Is the hook an executable (like https://github.com/skywinder/github-changelog-generator)? Or is it a text file with what you want as changelog already written down?

Thanks

bep:
>Just so I can fully understand, what exactly would the hook do?

I have not thought too hard about it, but in its simplest form, I guess a prepared text file would be nice enough (aka `goreleaser --release-notes=/mypath/release-notes-v1.01.txt`. My main motivation is to not have to re-invent the "update GitHub release"  part. I can understand you want to keep the "created by @goreleaser" part, and that is fine -- just append that at the end.

caarlos0:
@bep nice, so, quite easy to implement. Will see if I can do that throughout the day :)



> I can understand you want to keep the "created by @goreleaser" part, and that is fine

😄

caarlos0:
@bep check #184 :)

caarlos0:
[...]

PR #185, opened 2017-04-19
adding draft support

caarlos0:
ref #183 

Question remaining: what should happen in the brew pipe when the release is a draft? Should we skip it?


bep:
That is a good question. How would you "finish" that brew publish when it's paused?

I would say that, for now, just make it crystal clear that this is for GitHub and GH only. And then think about a more thought through solution for the others (there will be others); maybe some kind of holding persistent "holding queue" ...

caarlos0:
@bep let me know what you think:

"Normal" release process:

- create release
- upload artifacts
- open branch on tap repo
- edit the formula in that branch
- open PR
- merge the PR 
- delete the branch

"Draft" release process:

- create draft release
- upload artifacts
- open branch on tap repo
- edit the formula in that branch
- open PR

So, to approve, the user need to publish the release and merge the PR.

What do you think?

bep:
I'm not totally sure that I follow you.

Currently you

1) Manually create a tag on any branch you want
2) Run goreleaser which creates and publishes a release on GitHub

There is not automatic pull request merging in the above, to my knowledge, which is fine.

So, in the refined logic, I would assume was limited to the "not auto publish the GitHub release?" but there may be logic re. Homebrew etc. that I have not touched yet.

caarlos0:
@bep Yeah, maybe I just went full crazy in this one (I'm on painkillers, sorry about that).

The idea was that, if the release is a draft, instead of creating/updating the tap, it would open a PR and let it open, so in order to really publish the release, the user would have to Publish the release AND merge that PR.

But maybe this is too much... I'm actually not sure at all how this should be handled.

bep:
I will sleep on this, but I suspect you in parts reinvent the wheel that GitHub does pretty great -- and  if you don't produce any artifacts (other than the binaries and the changelog that does not (as I currently see it) have any place in the released source; I'm not sure what you're "pulling in" as a pull request. 

I have built a kind of harness/workflow in Hugo to handle version updates, release notes etc., and, while I may contribute some of this back to Goreleaser when I figure out what may be general enough, I think many projects have to specific needs to tailor in a good way in this department.

What I would like in a draft feature would be:

1) First of all the obvious: Be able to mark the release as a draft in GitHub
2) On a more advanced level, a way to "hold" the other publication channels (Brew, ...) until I somehow approve it (aka `release --continue``).

caarlos0:
@bep yeah, I guess your idea is better.

So, I'll just leave this PR here and start a new one, which will just mark the release as draft and skip the brew pipe entirely. 

Next: serialize the current context to a `ctx.json` file or something, and be able to load it back and continue with `--continue` flag.

bep:
Probably overkill for this, but BoltDB is great for lightweight embedded db stuff.

PR #188, opened 2017-04-21
Add fpm options

bep:
This is needed to create a proper debian control file, but will also be used for the .rpm packages.

I did not find any easy way to add a unit test for this, but I have tested it myself and verified that the debian control file looks fine.


caarlos0:
Looks good, I'll add the tests later.

PR #189, opened 2017-04-21
Reduce the number of concurrent builds

bep:
With many GOOS OARCH combos, the CPU gets glowing.


caarlos0:
Thanks!

arran4:
Anyway to change this number?

jorinvo:
@arran4 currently no. Can I ask for the reason why you like to change it?

arran4:
Build exceeds memory on my machine when n is greater than 1. (16gb)

Issue #205, opened 2017-04-22
Provide a make install


bep:
With the new folder layout it's not possible to do a simple "go install" on the top level anymore, so it would be nice if the Makefile had a `go install` variant.

jorinvo:
Personally, I try to use a `cmd` sub-directory only for projects that consist of more than one binary and for projects where the main artefact is a _package_ instead of a binary.
The way I understand GoReleaser the main artefact is the `goreleaser` binary. 
We can still provide the same functionality as a _package_ by moving the `goreleaser.go` file to `goreleaser/goreleaser.go`.
Is there any reason for moving the `main.go` file to `cmd/goreleaser` that I'm missing right now?

caarlos0:
Wouldn't a `go get github.com/goreleaser/goreleaser/cmd/...` do?

caarlos0:
@jorinvo not really, I was just trying to better organize the code...  AFAIK there isn't a pattern for that, I personally like to have the cmd package because I restrain myself to do cli stuff there... maybe it is dumb to do that, idk.

bep:
>Wouldn't a go get github.com/goreleaser/goreleaser/cmd/.../go install ./cmd/... 

Maybe. 

But it would also give you the _Usability Fail Award of the Year 2017 Award_ :-)

caarlos0:
yeah, I added the task anyway :D

bep:
But I suspect most projects of this kind have a main func on the top level, making it "go gettable".

jorinvo:
@caarlos0 we could keep the cli logic in a sub-directory and still re-add a top level `main.go` simply referencing it. Would that help keeping things organised?

caarlos0:
closed by #206

PR #206, opened 2017-04-23
make goreleaser installable via go get #205

jorinvo:
Move `main.go` file to top level and move Go library to `goreleaserlib` package.

This organisation would allow GoReleaser to be installable via `go get` and at the same time make the functionality usable from Go via the `goreleaserlib`.

Calling the package `goreleaserlib` instead of just `goreleaser` allow the default behaviour of `go build` to still create a binary called `goreleaser`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ install:
   - gem install fpm
 script:
   - make test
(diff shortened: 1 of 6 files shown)


caarlos0:
thanks!

caarlos0:
refs https://github.com/goreleaser/goreleaser/issues/205

PR #214, opened 2017-04-25
added a way to ignore builds

### LINKED ISSUE #191: Allow more flexible GOOS/GOARCH combinations
I'm in the process of moving the Hugo builds from goxc/manual to Goreleaser. And we build for many OSes (because we can!) ... But the restrictions put in place by `goxc` isn't currently possible to express in Goreleaser config, I think. The following does neither build nor make sense:

```
  goos:
    - darwin
    - linux
    - freebsd
    - netbsd
    - openbsd
    - dragonfly
  goarch:
    - amd64
    - 386
    - arm
    - arm64
```

Looking at the [goxc tests](https://github.com/laher/goxc/blob/f8fbb3b8c4df6c161348ae8a0047d562fcde8855/platforms/buildconstraints_test.go):

```
	"":                   "[{darwin 386} {darwin amd64} {linux 386} {linux amd64} {linux arm} {freebsd 386} {freebsd amd64} {openbsd 386} {openbsd amd64} {windows 386} {windows amd64}]",
```

The above may be too restrictive (It may make sense to build arm for more than Linux), but we need some way of adding some restrictions. I would be happy to submit a PR if we could agree on a way to express this.

caarlos0:
closes #191

cc/ @bep 

FWIW: I'm not liking this impl right now. Will think more about it and look again tomorrow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,15 @@ build:
     - 6
     - 7
 
+  # List of combinations of GOOS + GOARCH + GOARM to ignore.
+  # Default is empty.
+  ignore:
+    - goos: darwin
+      goarch: 386
+    - goos: linux
+      goarch: arm
+      goarm: 7
+
   # Hooks can be used to customize the final binary, for example, to run
   # generator or whatever you want.
   # Default is both hooks empty.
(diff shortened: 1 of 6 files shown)


caarlos0:
closed in #219

PR #216, opened 2017-04-25
Fix ldflags key name in README

bep:
Which can lead to some hard-to-debug issues ...

See https://github.com/spf13/hugo/issues/3382


caarlos0:
Thanks!

Issue #215, opened 2017-04-25
Proposal: goreleaser init


lucapette:
I'm setting up goreleaser for a small project today and I thought it would be nice to have a small command that speeds up the setup. Something along the line of:

```sh
$ goreleaser init # or --init. I have no particular preferences
goreleaser.yml created. Please edit accordingly to your needs.
```

It could be as simple as creating a stub or we could try to guess some sensible defaults (not sure how easy is that). If you like the idea, I can give it a try in the coming days.

Thank you one more time for this awesome project, I find myself using it for _all_ my Go projects nowadays.

jorinvo:
We had this idea before. See #47. Back then we thought having sensible defaults can replace the need for a `goreleaser.yml` file.
There have been many new features added since then and now it seems like no project will use GoReleaser relying solely on the defaults.

I would welcome a simple *init* script that creates a configuration file with common options and some helpful comments.
What are your thoughts on this @caarlos0?

lucapette:
Thank you for your answer @jorinvo and sorry! For some reason I couldn't find the existing discussion despite searching. 

I'd be more than happy with a skeleton with good comments. I think it would still be a better user experience than copying the config from the README.

caarlos0:
I think it's a good idea now!

jorinvo:
@lucapette If you like, you can give it a try then! 
I think simply adding a `goreleaser.yml` file would be enough. No need for any interactive options at this point :)

bep:
It would also be good if that same sample config would be sent into an integration test of sorts, so we could have some guarantee that it is correct.

caarlos0:
Yes, I thought about creating a sample file and a fake repo, and let it run every time the CI runs... but I didn't find a good way of validating it (without writing a lot of code at least)...

lucapette:
I will have some time tomorrow and I'll give it a try. Any preference on `goreleaser --init` `goreleaser init`? Maybe the former is a bit easier to implement.

As for the test, I think it may be simple enough to assert the output is a valid `goreleaser.yml` file with some default properties. I would love to do it at acceptance level with a fake repo but, as @caarlos0 says, it may be hard without having some infrastructure for that kind of testing already in place. What I'm thinking now is that we could have a separate issue to discuss an acceptance test suite and aim to get it in for the next major version. What do you think?

bep:
>goreleaser init

Definetively. It should be its own subcommand, it only vaguely related to the main command -- and the implementation would look funky.

[...]

PR #221, opened 2017-04-27
Draft of the init command

lucapette:
This PR is a WIP for #215. As it's a work in progress and I'm not sure what's going to be the final result, I left docs changes for the README out. 

Here are my doubts:

- I introduced a default configuration as discussed but I'm not sure what's a good set of sensitive defaults. I'm sure you all know better how people are using the config and help me choose a good set of default. Right now, I literally copied the first example from the readme.
- The test I added is pretty simple but, as long as we don't put any logic into the template, I can't personally justify testing more than "it's a vaild yml for the project struct"
- The code checks if there's already a `goreleaser.yml`. The code is a bit cumbersome and maybe you think the check itself is overkill (I don't like the idea of skipping the check but it may be a personal think). Not sure there's a better way of handling that.

Any other comment about it more than welcome! I'll have more time later today or tomorrow morning so we can hopefully finish it this week :)


caarlos0:
One weird idea: we could run the defaults Pipe (will probably require moving some code but whatever) and use the its output config in init.

It's pretty simple actually:

- Create an empty context;
- Run the defaults pipe;
- Get the config from the context;
- Marshal into a yaml;
- Save.

This way the "init" goreleaser.yml will be the defaults for that project... things like repo, binary name and etc will already be right (its supposed to be at least).

What do you think?

lucapette:
@caarlos0 that sounds awesome! My first try was dumping an empty `config.Project` (so same thinking but not a good approach) and I did not like the output (everything was empty). I **love** your approach and will give it a go as soon as I can!

caarlos0:
```patch
diff --git a/config/config.go b/config/config.go
index ad4736e..acda4c8 100644
--- a/config/config.go
+++ b/config/config.go
@@ -104,16 +104,3 @@ func Load(file string) (config Project, err error) {
 	err = yaml.Unmarshal(data, &config)
 	return
 }
-
-// DefaultConfig returns a default configuration for the init command
-func DefaultConfig() string {
-	return `
-build:
-  binary: drum-roll
-  goos:
-    - windows
-    - darwin
-    - linux
-  goarch:
-    - amd64`
-}
diff --git a/config/config_test.go b/config/config_test.go
index 092319b..d912156 100644
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -1,16 +1 @@
 package config
-
-import (
-	"testing"
-
-	yaml "gopkg.in/yaml.v1"
-)
-
-func TestDefaultConfig(t *testing.T) {
-	cfg := DefaultConfig()
-	config := Project{}
-
-	if err := yaml.Unmarshal([]byte(cfg), &config); err != nil {
-		t.Errorf("Not a valid config. %s", err.Error())
-	}
-}
diff --git a/goreleaserlib/goreleaser.go b/goreleaserlib/goreleaser.go
index daee1b0..9ea56a0 100644
--- a/goreleaserlib/goreleaser.go
+++ b/goreleaserlib/goreleaser.go
@@ -1,11 +1,12 @@
 package goreleaserlib
 
 import (
+	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
 
-	"fmt"
+	yaml "gopkg.in/yaml.v1"
 
 	"github.com/goreleaser/goreleaser/config"
 	"github.com/goreleaser/goreleaser/context"
@@ -82,7 +83,7 @@ func Release(flags Flags) error {
 
 // InitProject creates an example goreleaser.yml in the current directory
 func InitProject() error {
-	filename := "goreleaser.yml"
+	var filename = "goreleaser.yml"
 	if _, err := os.Stat(filename); !os.IsNotExist(err) {
 		if err != nil {
 			return err
@@ -90,6 +91,14 @@ func InitProject() error {
 
 		return fmt.Errorf("%s already exists", filename)
 	}
-
-	return ioutil.WriteFile(filename, []byte(config.DefaultConfig()), 0644)
+	var ctx = context.New(config.Project{})
+	var pipe = defaults.Pipe{}
+	if err := pipe.Run(ctx); err != nil {
+		return err
+	}
+	out, err := yaml.Marshal(ctx.Config)
+	if err != nil {
+		return err
+	}
+	return ioutil.WriteFile(filename, out, 0644)
 }
diff --git a/main.go b/main.go
index 95889d7..cd6813b 100644
--- a/main.go
+++ b/main.go
@@ -52,9 +52,9 @@ func main() {
 			Aliases: []string{"i"},
 			Usage:   "generate goreleaser.yml",
 			Action: func(c *cli.Context) error {
-				err := goreleaserlib.InitProject()
+				var err = goreleaserlib.InitProject()
 				if err != nil {
-					return err
+					return cli.NewExitError(err.Error(), 1)
 				}
 
 				log.Printf("goreleaser.yml created. Please edit accordingly to your needs.")

```

this works, but it gens a ugly file, maybe we should add ommitempty in the struct fields:

```yaml
release:
  github:
    owner: goreleaser
    name: goreleaser
  draft: false
brew:
  github:
    owner: ""
    name: ""
  folder: ""
  caveats: ""
  plist: ""
  install: bin.install "goreleaser"
  dependencies: []
  conflicts: []
  description: ""
  homepage: ""
build:
  goos:
  - linux
  - darwin
  goarch:
  - amd64
  - "386"
  goarm:
  - "6"
  main: .
  ldflags: -s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}}
  flags: ""
  binary: goreleaser
  hooks:
    pre: ""
    post: ""
archive:
  format: tar.gz
  format_overrides: []
  name_template: '{{ .Binary }}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end
    }}'
  replacements:
    "386": i386
    amd64: x86_64
    darwin: Darwin
    freebsd: FreeBSD
    linux: Linux
    netbsd: NetBSD
    openbsd: OpenBSD
    windows: Windows
  files:
  - LICENSE.md
  - README.md
fpm:
  formats: []
  dependencies: []
  conflicts: []
  vendor: ""
  homepage: ""
  maintainer: ""
  description: ""
  license: ""

```

lucapette:
@caarlos0 I changed the code based on your suggestion. I like the defaults much more now but I'm not sure they're good enough for you so feedback is more than welcome.

If we like the current solution, the only thing that's definitely still needed is docs change in the README which I'll do as soon as we're all comfortable with the direction of this.

main.go
```diff
@@ -46,6 +46,23 @@ func main() {
...
+	app.Commands = []cli.Command{
+		{
+			Name:    "init",
+			Aliases: []string{"i"},
+			Usage:   "generate goreleaser.yml",
+			Action: func(c *cli.Context) error {
+				var filename = "goreleaser.yml"
+				err := goreleaserlib.InitProject(filename)
+				if err != nil {
+					return err
```
caarlos0: should return `cli.NewExitError(err.Error(), 1)`

caarlos0: you can also:

```go
if err := goreleaserlib.InitProject(filename); err != nil {
   return cli.NewExitError(err.Error(), 1)
}
```

goreleaserlib/goreleaser_test.go
```diff
@@ -152,3 +155,38 @@ release:
...
+	}
+
+	out, err := ioutil.ReadAll(file)
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+
+	config := config.Project{}
+	if err := yaml.Unmarshal(out, &config); err != nil {
+		t.Errorf("Not a valid config. %s", err.Error())
```
caarlos0: can you change those `t.Errorf` and `t.Fatal` statements to use the `assertify` package as the other tests do?

lucapette:
@caarlos0 I changed the code according to your feedback (thank you for that!). I added a test too but I'm not sure it's enough. I actually can't run `make test` without getting some failures so I always scope them with `TEST_PATTERNS`. I suspect the test suite now makes some assumptions I don't know about. I'll write more in the issue you opened about end to end testing as soon as I find a minute (I intend to write a lot :D)

goreleaserlib/goreleaser.go
```diff
@@ -77,3 +81,26 @@ func Release(flags Flags) error {
...
+func InitProject(filename string) error {
+	if _, err := os.Stat(filename); !os.IsNotExist(err) {
+		if err != nil {
+			return err
+		}
+
+		return fmt.Errorf("%s already exists", filename)
+	}
+
+	var ctx = context.New(config.Project{})
```
lucapette: I've noticed you prefer this style of declaration to `foo := bar` so I did the same here. But I'm curious to understand if it's just a taste preference or there's a technical reason for that.

caarlos0: just preference, looks prettier I think :P

lucapette: 👍

caarlos0:
@lucapette I'll merge this and try to cover the missing parts.

lucapette:
@caarlos0 thank you for taking over and finishing it up!

caarlos0:
@lucapette no problem, wasn't able to cover it all, but got a little up...

Thanks for all the work :D

Issue #225, opened 2017-05-10
Invalid binary for alpine amd64 (x86_64)


kamilsk:
Host machine:

```
$ uname -a
Darwin MacBook-Pro.beeline 16.5.0 Darwin Kernel Version 16.5.0: Fri Mar  3 16:52:33 PST 2017; root:xnu-3789.51.2~3/RELEASE_X86_64 x86_64
```

golang:alpine

```
$ docker run --rm golang:alpine uname -a
Linux 0e0a30d12b5b 4.9.13-moby #1 SMP Sat Mar 25 02:48:44 UTC 2017 x86_64 Linux

$ docker run --rm golang:alpine go version
go version go1.8.1 linux/amd64
```

golang:latest

```
$ docker run --rm golang:latest uname -a
Linux f912eb34c93b 4.9.13-moby #1 SMP Sat Mar 25 02:48:44 UTC 2017 x86_64 GNU/Linux

$ docker run --rm golang:latest go version
go version go1.8.1 linux/amd64
```

The problem:

```
$ docker run --rm golang:alpine \
  /bin/sh -c 'apk update --no-cache \
           && apk add --no-cache ca-certificates git wget \
           && update-ca-certificates &>/dev/null \
           && wget -q -O /tmp/goreleaser.tar.gz \
              https://github.com/goreleaser/goreleaser/releases/download/v0.17.1/goreleaser_Linux_x86_64.tar.gz \
           && mkdir /tmp/goreleaser && tar xf /tmp/goreleaser.tar.gz -C /tmp/goreleaser \
           && /tmp/goreleaser/goreleaser --version'
...
/bin/sh: /tmp/goreleaser/goreleaser: not found
```

```
$ docker run --rm golang:latest \
  /bin/sh -c 'wget -q -O /tmp/goreleaser.tar.gz \
              https://github.com/goreleaser/goreleaser/releases/download/v0.17.1/goreleaser_Linux_x86_64.tar.gz \
           && mkdir /tmp/goreleaser && tar xf /tmp/goreleaser.tar.gz -C /tmp/goreleaser \
           && /tmp/goreleaser/goreleaser --version'
...
goreleaser version 0.17.1, commit 5b06bef0666fd105b65a41d53d45e2d9db6e4862, built at 2017-05-02T22:24:02Z

$ docker run --rm golang:latest \
  /bin/sh -c 'wget -q -O /tmp/goreleaser.tar.gz \
[...]

PR #226, opened 2017-05-10
Use external, static linking.

jorinvo:
Possible fix for #225 

I saw [Hugo](https://github.com/spf13/hugo/blob/master/goreleaser.yml#L4) uses this option and the Hugo release works on Alpine Linux.
But I don't think Hugo is release from Travis CI.
The only way to test this, is by making it part of a tag and let it run on Travis. 
Could also create a new repository to try this out.
But before trying this, @caarlos0, maybe you have any idea?


caarlos0:
I think this is it.

We will probably also need to force netgo, but let's try this first.

PR #230, opened 2017-05-11
Add LoadReader function to prevent clients from having to deal with YAML

client9:
Even added test cases.  No changes to existing functionality.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -3,6 +3,7 @@
 package config
 
 import (
+	"io"
 	"io/ioutil"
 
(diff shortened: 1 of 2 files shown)


client9:
not fair.   there were 0% tests in this directory and I even added some bonus tests to boost coverage of existing code).   The % is low since there are like 8 LOC in entire file.  lulz

caarlos0:
Yeah, package had no tests, so it was being ignored by codecov. Don't worry about it :)

caarlos0:
Thanks for the PR BTW 🍻

Issue #229, opened 2017-05-11
Support globbing for "archive: > files:"


dogancelik:
`archive: > files:` setting should support globbing.
It only supports including files in root folder.

* If you do `public/`, it only creates the folder, **no files inside.**
* If you do `public/*`, it should only add immediate (only root) files of `public/` → **errors.**
* If you do `public/**`, it should add all files inside *public* folder → **errors.**

```
$ goreleaser --skip-validate --skip-publish
Running goreleaser 0.17.3
Setting defaults
Getting and validating git state
 -> Skipped validations because --skip-validate is set
Loading environment variables
 -> GITHUB_TOKEN not validated because publishing has been disabled
Building binaries
 -> Building dist\seps_windows_amd64\seps.exe
 -> Building dist\seps_linux_amd64\seps
 -> Building dist\seps_mac_amd64\seps
Creating archives
 -> Creating dist\seps_mac_amd64.tar.gz
 -> Creating dist\seps_linux_amd64.tar.gz
 -> Creating dist\seps_windows_amd64.zip
open public/**: The filename, directory name, or volume label syntax is incorrect.
```

Goreleaser version: `0.17.3, commit 2af8d94f94b9e67134787e63c042943084f12b0d, built at 2017-05-10T13:51:16Z`

caarlos0:
can you check #231?

caarlos0:
Done on v0.18.0

PR #231, opened 2017-05-11
support globs on archive.files

caarlos0:
refs #229 

- now, the `defaults` pipe won't search for files anymore, but would set the default globs into  `config.archive.files`
- the archive pipe will then iterate over `config.archive.files` and deal with all of them as globs
- removed an unneeded read dir from the archive pipe, which required moving `extFor` function to another pkg

It should work for:

- `*.ext`
- `blah*`
- `folder/*`
- `folder/**/*`

It doesn't work, however, for `folder` only, because the archive impls require files instead of folders. We may change this in the future.

This seems to be more robust than the previous solution, not sure if ideal though. Feedback appreciated!

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1,4 +1,4 @@
-memo = "85c15b94eaf073e2c6e87a85ac2088f1d61889c8ddcc8542c09145f0f816c6af"
+memo = "3cbc12d80513bcbb0ac166d404600ffbc63b1c301b5c6d3722dad30db6f6a549"
 
 [[projects]]
   name = "github.com/davecgh/go-spew"
@@ -23,6 +23,12 @@ memo = "85c15b94eaf073e2c6e87a85ac2088f1d61889c8ddcc8542c09145f0f816c6af"
   name = "github.com/google/go-querystring"
   packages = ["query"]
   revision = "53e6ce116135b80d037921a7fdd5138cf32d7a8a"
+
+[[projects]]
+  branch = "master"
+  name = "github.com/mattn/go-zglob"
+  packages = [".","fastwalk"]
+  revision = "95345c4e1c0ebc9d16a3284177f09360f4d20fab"
 
 [[projects]]
   name = "github.com/pmezard/go-difflib"
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ setup: ## Install all the build and lint dependencies
 	gometalinter --install
 
 test: ## Run all the tests
-	gotestcover $(TEST_OPTIONS) -covermode=atomic -coverprofile=coverage.txt $(SOURCE_FILES) -run $(TEST_PATTERN) -timeout=30s
+	gotestcover $(TEST_OPTIONS) -covermode=atomic -coverprofile=coverage.txt $(SOURCE_FILES) -run $(TEST_PATTERN) -timeout=2m
 
 cover: test ## Run all the tests and opens the coverage report
 	go tool cover -html=coverage.txt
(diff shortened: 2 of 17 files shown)


dogancelik:
Still getting the error  🤔

```
$ goreleaser
Running goreleaser dev
Setting defaults
Getting and validating git state
 -> Skipped validations because --skip-validate is set
Loading environment variables
 -> GITHUB_TOKEN not validated because publishing has been disabled
Building binaries
 -> Building dist\seps_windows_amd64\seps.exe
 -> Building dist\seps_mac_amd64\seps
 -> Building dist\seps_linux_amd64\seps
Creating archives
 -> Creating dist\seps_windows_amd64.zip
 -> Creating dist\seps_mac_amd64.tar.gz
 -> Creating dist\seps_linux_amd64.tar.gz
open public/**: The filename, directory name, or volume label syntax is incorrect.
```

jorinvo:
@caarlos0 nice, definitely a simpler and more flexible solution than previously!

@dogancelik unfortunately the *glob* function in the Go standard library doesn't support the double star `**` syntax. I'm afraid you have to specify the files further. @caarlos0 maybe we should look into the selection of folders to serve this use case..

jorinvo:
@caarlos0 how about supporting folders by using [`filepath.Walk`](https://golang.org/pkg/path/filepath/#Walk)? Could check if we have a directory in [`findFIles`](https://github.com/goreleaser/goreleaser/blob/0ae005b42fe83704bc93dc992cc1f7ac36278086/pipeline/archive/archive.go#L75) and add all of its contents.

client9:
@jorinvo correct on golang.  https://github.com/golang/go/issues/11862   Looks like https://github.com/mattn/go-zglob is a quick win.  I've used https://github.com/gobwas/glob in the past but it's much more complicated.

caarlos0:
@mattn zglob seems awesome, just pushed a version using it.
@dogancelik can you check if it attends your needs?

dogancelik:
Am I doing something wrong?
Here's what I do to build the source:

```sh
wget https://github.com/goreleaser/goreleaser/archive/globs.zip # globs branch
cd goreleaser-globs
go get ./...
go build # no error
```

* `public\**` and `public/**` fails:
  * `open public\**: The filename, directory name, or volume label syntax is incorrect.`
  * Tried it on Command Prompt, Git Bash.

This is the latest commit because it has `%GOPATH%\src\github.com\mattn\go-zglob`
I built it in a new *GOPATH* folder.

caarlos0:
@dogancelik should probably be `public/**/*`

caarlos0:
Actually, no, it should work:

image

```diff
diff --git a/goreleaser.yml b/goreleaser.yml
index 0b62b0e..e74636e 100644
--- a/goreleaser.yml
+++ b/goreleaser.yml
@@ -4,15 +4,12 @@ build:
   env:
    - CGO_ENABLED=0
   goos:
-    - linux
     - darwin
-    - windows
   goarch:
-    - 386
     - amd64
-    - arm
-    - arm64
 archive:
+  files:
+    - pipeline/**
   format_overrides:
     - goos: windows
       format: zip
```

caarlos0:
I see you're running on windows, so, I've built it for you:

[goreleaser_Windows_i386.zip](https://github.com/goreleaser/goreleaser/files/994302/goreleaser_Windows_i386.zip)
[goreleaser_Windows_x86_64.zip](https://github.com/goreleaser/goreleaser/files/994301/goreleaser_Windows_x86_64.zip)


let me know if it works.

dogancelik:
@caarlos0 Weird, your build worked 😄

caarlos0:
@dogancelik nice, I'll release this!

PR #238, opened 2017-05-18
breaking: do not default to uname values

### LINKED ISSUE #236: behavior of archive naming with and without archive.replacements sections
hello!

in goreleaser releases we have `Darwin` and `Linux` (capital letters) in the file names

```
goreleaser_Darwin_x86_64.tar.gz
goreleaser_Linux_arm64.deb
```

ok, but why are they not `darwin` and `linux` given the `goreleaser.yaml file`?   Why are they not using the GOOS names (all lower case?).  I don't see any archive/overriders in the file:

https://github.com/goreleaser/goreleaser/blob/master/goreleaser.yml

```yaml
build:
  env:
   - CGO_ENABLED=0
  goos:
    - linux
    - darwin
    - windows
  goarch:
    - 386
    - amd64
    - arm
    - arm64
archive:
  format_overrides:
    - goos: windows
      format: zip
```

I saw this previously in my other projects when goreleaser was not reading the yaml file, it would use the upper-case names.   When I fixed the YAML name, they started reporting correctly.  See #235

caarlos0:
check discussion on #236

closes #236

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,8 +276,6 @@ archive:
   # Replacements for GOOS and GOARCH on the archive name.
   # The keys should be valid GOOS or GOARCH values followed by your custom
   # replacements.
-  # By default, `replacements` replace GOOS and GOARCH values with valid outputs
-  # of `uname -s` and `uname -m` respectively.
   replacements:
     amd64: 64-bit
     386: 32-bit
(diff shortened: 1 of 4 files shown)


caarlos0:
v0.19.0 is out!

Issue #243, opened 2017-05-26
Naked releases to GitHub (raw binary, no tar, no zip)


client9:
I've seen a few places do this 

https://github.com/mvdan/sh/releases

They just upload raw binaries, no tar, no zip.

Not urgent, low priority.  But nice to support, so that people could switch to goreleaser and keep their existing formats.

jorinvo:
```yaml
archive:
  skip: true
```

Would that work?

client9:
let me look here

tj:
👍 this would be great, I usually just shove bins up as well, great tool though nice work!

caarlos0:
@tj this is WIP on #257, will try to release it in 2 days :)

caarlos0:
@tj it is out, not yet working homebrew and fpm though. Working on that part :)

tj:
awesome, works! the brew stuff sounds awesome, I've never bothered with brew, definitely will now haha

caarlos0:
I created new issues to deal with the things that are not working with binary releases right now.

Will close this since the initial thing is done :D

PR #257, opened 2017-06-05
Add option to skip archive.

jorinvo:
See #243.
Upload binaries directly instead of creating archive.
This option, however currently doesn't work together with brew and fpm
since they both require archived artifacts.

I'm not that happy that my current solution doesn't work together with brew and fpm, 
@caarlos0 do you have an idea how to generate archived + naked versions side by side so we can use the archived ones for brew/fpm and the other ones as artifacts?


pipeline/archive/archive.go
```diff
@@ -73,6 +76,13 @@ func create(ctx *context.Context, platform, name string) error {
 	return nil
 }
 
+func skip(ctx *context.Context, platform, name string) error {
+	log.Println("Skip archiving")
```
caarlos0: maybe add the platform and/or name to the log as well?

caarlos0:
I don't see why codecov is complaining about coverage... seems to me that all new code is covered...

jorinvo:
I also don't get it.

caarlos0:
ok, let's just wait for @client9, so we can have a PR prepared on godownloader as well :)

jorinvo:
@caarlos0 any thoughts regarding brew/fpm support?

caarlos0:
@jorinvo I think fpm may be kind of easy, no idea about how to do it with homebrew though...

caarlos0:
@jorinvo most formulae in the brew repos build/compile the software they install, e.g.: https://github.com/Homebrew/homebrew-core/blob/master/Formula/shfmt.rb

jorinvo:
@caarlos0 I like how our brew formulas don't depend on the `go` tool though..

Currently brew also only works if a `darwinamd64` build is included right?
We could also remove that dependency and always build one version specifically for brew.
Then there would be no conflict with the `skip` archive option either.

caarlos0:
@jorinvo I thinks it's strange to build for darwin if the user didn't added it to the list of targets.

as far as I remember there is a warning or something when brew is enabled and there is no darwin builds, we may make it a failure while validating the config instead so it would be clearer...

jorinvo:
@caarlos0 yes, you're right. 
Maybe we also don't need to support brew/fpm together with `archive: skip: true` for now until someone actually uses it.

caarlos0:
@jorinvo  Agreed!

client9:
just saw this... will look tomorrow am.

client9:
Ah added to ticket, not PR.

Don't know much about brew.

I think `skip: true` is confusing should be `format: binary` if I understand the issue correct.

jorinvo:
@client9 good idea using the existing `format` option. I will look into it tomorrow.

jorinvo:
@client9 I updated it to use `format` instead of `skip`.

caarlos0:
LGTM, @client9 what do you think?

caarlos0:
@jorinvo I changed the base branch to `multiple-builds` and I'm doing the changes so both features can work together :)

caarlos0:
For the future:

- create packages with fpm on naked releases
- create brew formulas for naked releases

PR #262, opened 2017-06-19
Add missing loop variable capture

### LINKED ISSUE #261: FPM builds only one package format
The FPM builder is missing a loop variable capture and builds only on archive format.

dominikschulz:
Fixes #261


caarlos0:
good catch, thanks!

Issue #268, opened 2017-06-29
Mistakenly things that the repo is in dirty state


fsouza:
I use git short status (https://github.com/fsouza/dotfiles/blob/master/.gitconfig#L17) and it confuses goreleaser:

```
% goreleaser
Running goreleaser dev
Setting defaults
Getting and validating git state
git is currently in a dirty state:
## master...origin/master
```

I currently work around this issue with ``--skip-validation``, but it would be good to see this fixed :D :D 

Thank you for the awesome tool!

jorinvo:
GoReleaser also uses the short version: https://github.com/goreleaser/goreleaser/blob/master/pipeline/git/git.go#L135
I think the test fails because you have the `branch` option set. 
I changed it to use the `--porcelain` flag which should be more stable.

caarlos0:
fixed in v0.21.3

caarlos0:
thanks for the report @fsouza, thanks @jorinvo for the quick fix!

fsouza:
![you-got-it](https://user-images.githubusercontent.com/108725/27714838-63c6823e-5d01-11e7-9e34-6b6c4abcd60e.gif)

PR #269, opened 2017-06-29
Use git status --porcelain

jorinvo:
to ensure output is always readable no matter what configuration a user has set (cc #268)


caarlos0:
good one!

PR #272, opened 2017-07-02
using a valid email to publish formulas

caarlos0:
I created a github account called @goreleaserbot and thought it would be nice to use it to publish formulas et al.

What do you guys think?

Ex: https://github.com/caarlos0/homebrew-tap/blob/master/Formula/test.rb


jorinvo:
Nice 👍

PR #291, opened 2017-07-08
Config validation

### LINKED ISSUE #290: Binary diff support
I'm not familiar with the existing formats or pkgs, but @c4milo was suggesting binary diffs for https://github.com/tj/go-update which is would be a really cool addition to tarballs. Would you be into supporting diffs along side the existing options?

caarlos0:
added a `XXX` map field in each config struct matching all non previously matched fields.

this `XXX` map should always be empty.

Kind of ugly but works. Inspired by prometheus/alertmanager 🍺 

closes #290

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,7 +1,6 @@
-homepage: &homepage http://goreleaser.github.io
-description: &description Deliver Go binaries as fast and easily as possible
+nnnnnnnn: a
 builds:
-  -
+  - y: 1
     env:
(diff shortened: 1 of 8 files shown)


jorinvo:
Nice! I didn't know that you can define a catch-all for parsing 🎉

caarlos0:
@jorinvo me neither, I discover that while poking into prometheus' code :D

jorinvo:
@caarlos0 guess the same also works for other formats like json :)

Issue #280, opened 2017-07-05
make fpm packages work with binary releases


caarlos0:
refs #257 and #243

PR #301, opened 2017-07-13
WIP: fixing fpm on multiple naked binaries releases

### LINKED ISSUE #298: Refactory: we need to able to get the "pretty name" of a binary
For both #280 and #281 we need to be able to "go back" to the pretty name of a binary.

Maybe a simple map on the context will do:

```go
map[string]string{
  "goreleaser_Darwin_amd64_v0.28.0": "goreleaser",
  "goreleaser_Windows_amd64_v0.28.0.exe": "goreleaser.exe",
  // etc...
}
```

Right now it is not possible to do that.

caarlos0:
closes #298 , #280

### DIFF
--- a/context/context.go
+++ b/context/context.go
@@ -21,13 +21,18 @@ type GitInfo struct {
 	Commit     string
 }
 
+// Binary with pretty name and path
+type Binary struct {
+	Name, Path string
+}
+
 // Context carries along some data through the pipes
 type Context struct {
 	ctx.Context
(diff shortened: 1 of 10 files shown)


caarlos0:
this looks ugly, ideas?

jorinvo:
I think it's good for now as long as it works.
Can clean it up later, but I also can't think of a cleaner way to do this at the moment.

PR #305, opened 2017-07-15
Fixes spelling mistake.

marianogappa:


jorinvo:
Thanks @marianogappa!
What do you think about changing it to something like `config created, please adjust it to your needs`?

jorinvo:
can keep it like this for now, thanks @marianogappa 👍

PR #306, opened 2017-07-15
allow configuring parallelism via flag

### LINKED ISSUE #304: Add config for number of parallel builds
As @arran4 mentioned in #189, there are valid cases where it is required to be able to limit the number of parallel builds.

caarlos0:
closes #304


jorinvo:
@caarlos0 personally I think `parallel` would be distinct enough, but `parallelism` is also ok.

PR #310, opened 2017-07-17
Fix config comments in README

tisba:
The comments for `plist` and `test` got mixed up, so I fixed it.


jorinvo:
thanks @tisba!

caarlos0:
My bad, thanks!

PR #317, opened 2017-07-27
add the snapcraft pipeline

elopio:
Snapcraft is the new universal and secure package being developed by Ubuntu. It allows upstream developers to push their software to all Linux distros on their own schedule, without intermediaries, and with automatic updates.

https://snapcraft.io/

### DIFF
--- a/pipeline/snapcraft/snapcraft.go
+++ b/pipeline/snapcraft/snapcraft.go
@@ -135,10 +135,16 @@ func create(ctx *context.Context, folder, arch string, binaries []context.Binary
 		return err
 	}
 
-	snap := metadata.Name + "_" + metadata.Version + "_" + arch + ".snap"
-	cmd := exec.Command("snapcraft", "snap", "--output", snap)
+	cmd := exec.Command("snapcraft", "clean")
 	cmd.Dir = path
 	if out, err := cmd.CombinedOutput(); err != nil {
+		return errors.New(string(out))
+	}
+
+	snap := metadata.Name + "_" + metadata.Version + "_" + arch + ".snap"
+	cmd = exec.Command("snapcraft", "snap", "--output", snap)
+	cmd.Dir = path
+	if out, err = cmd.CombinedOutput(); err != nil {
 		return errors.New(string(out))
 	}
 	ctx.AddArtifact(filepath.Join(path, snap))
--- a/pipeline/snapcraft/snapcraft.go
+++ b/pipeline/snapcraft/snapcraft.go
@@ -4,6 +4,7 @@ package snapcraft
 import (
 	"errors"
 	"io/ioutil"
+	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
@@ -27,7 +28,6 @@ type SnapcraftMetadata struct {
 	Confinement   string `yaml:",omitempty"`
 	Architectures []string
 	Apps          map[string]AppsMetadata
-	Parts         map[string]PartsMetadata
 }
 
 // AppsMetadata for the binaries that will be in the snap package
@@ -35,13 +35,6 @@ type AppsMetadata struct {
 	Command string
 	//	Plugs []string
 	//	Daemon string
-}
-
-// PartsMetadata for the binaries that will be in the snap package
-type PartsMetadata struct {
-	Source string
-	Plugin string
-	Prime  []string
 }
 
 // Pipe for snapcraft packaging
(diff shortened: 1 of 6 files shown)


elopio:
I based this in the fpm integration, please take a look and let me know if you have any corrections.

I have a few remaining questions. The snaps are secure because each binary (`app` in snapcraft terminology) has to declare the permissions it requires. I couldn't find a way to specify extra metadata for the binaries generated. For now this is useful for all snaps that don't require extra permissions, but we need to solve this before it's useful for all of them. Also, some binaries can be declared as daemons, and they will be autostarted.

Should I extend the `builds` structs to accept permissions for the binaries? Something like:

```
builds:
  - binary: binary_x
    plugs: [network]
    daemon: simple.
```

Or should I do it in the `snapcraft` section? Like:

```
snapcraft:
  ...
  apps:
    binary_x:
      plugs: [network]
      daemon: simple
```

The problem with the first one is that those values don't apply to fpm, or tar, so they would just be ignored.
The problem with the second approach would be to validate that the app name, `binary_x` in this example, corresponds to a binary generated by goreleaser. It seems easy to do, just some extra code.

Please, let me know your thoughts.

A few days ago we made a video explaining some details about go snaps, in case you are curious: https://www.youtube.com/watch?v=b2x6mumSrpw

pura vida.

caarlos0:
Hi @elopio!

Awesome contribution, thank you very much!


I definitely prefer:

```yaml
snapcraft:
  ...
  apps:
    binary_x:
      plugs: [network]
      daemon: simple
```

In this scenario, `apps` will be something like `map[string]struct{Plugs []string, Daemon string}`. This should work as expected I think.

I'll take a better look at the code now!

pipeline/snapcraft/snapcraft.go
```diff
@@ -0,0 +1,152 @@
...
+	Grade         string `yaml:",omitempty"`
+	Confinement   string `yaml:",omitempty"`
+	Architectures []string
+	Apps          map[string]AppsMetadata
+	Parts         map[string]PartsMetadata
+}
+
+// AppsMetadata for the binaries that will be in the snap package
+type AppsMetadata struct {
+	Command string
```
caarlos0: this accepts multiple commands, right?

since goreleaser can build multiple binaries, maybe we can get this to work from the start :)

caarlos0:
Quick question: can I build snaps from OSX or only from another linux box?

pipeline/snapcraft/snapcraft.go
elopio: There is one command per app, but there can be multiple apps per snap.

So I added a loop in there that turns each binary into an app.

elopio:
Thanks for your review @caarlos0! If you don't mind, I would like to leave the plugs and daemon values for a following PR. I agree on doing it as part of the `snapcraft` section, but I would prefer to keep PRs small and focused. We don't need that to generate the goreleaser snap, for example.

About OSX, we've spent this morning discussing about it. We are close to release snapcraft with homebrew. It won't be able to do everything that it can do on Ubuntu, like installing debs inside the snap, but it should be enough for cases like this, which have the binaries already compiled. I will rewrite a few things to rely less on snapcraft.

I will ping you again when this PR is ready for another review.

pura vida.

caarlos0:
@elopio agreed on the multiple PRs :)

Nice, great to see this kind of tool becoming popular =D

elopio:
@caarlos0 please take another look.
With the change I have just pushed, snapcraft is required only to create the squashfs compressed `.snap` from the directory that contains the snap metadata and the binaries. For now, it skips if the OS is not Linux, but once we have snapcraft ready to run in Windows and Mac I'll come back and remove that skip.

Thanks again! Any suggestions are welcome.

pipeline/snapcraft/snapcraft.go
```diff
@@ -0,0 +1,142 @@
...
+
+// Description of the pipe
+func (Pipe) Description() string {
+	return "Creating Linux packages with snapcraft"
+}
+
+// Run the pipe
+func (Pipe) Run(ctx *context.Context) error {
+	if ctx.Config.Snapcraft.Summary == "" {
+		log.Info("no snapcraft summary defined, skipping")
```
caarlos0: maybe this should be a warn or an error?

pipeline/snapcraft/snapcraft.go
```diff
@@ -0,0 +1,142 @@
...
+		return nil
+	}
+	if ctx.Config.Snapcraft.Summary == "" {
+		log.Info("no snapcraft description defined, skipping")
```
caarlos0: maybe this should be a warn or an error?

pipeline/snapcraft/snapcraft.go
```diff
@@ -0,0 +1,142 @@
...
+
+// SnapcraftMetadata to generate the snap package
+type SnapcraftMetadata struct {
+	Name          string
+	Version       string
+	Summary       string
+	Description   string
+	Grade         string `yaml:",omitempty"`
+	Confinement   string `yaml:",omitempty"`
+	Architectures []string
```
caarlos0: seems like this is unused...

caarlos0:
@elopio I reviewed your code, looks great! just a few minors 👍

pipeline/snapcraft/snapcraft.go
elopio: Done.

pipeline/snapcraft/snapcraft.go
elopio: You are right. This should be error.

pipeline/snapcraft/snapcraft.go
elopio: It is used here: https://github.com/goreleaser/goreleaser/pull/317/files#diff-0c1a079410d0d8f77011355f28769eb6R112

caarlos0:
@elopio released v0.28.0 🎉

PR #321, opened 2017-08-02
small improvements

caarlos0:
I noticed that the final artifact was in a subfolder inside dist, different from the other pipes, so I change that. (see https://travis-ci.org/goreleaser/goreleaser/builds/260157591#L891)

I also dealt with a possible error while linking files.

Finally, I passed gometalinter again on the code (disabled on build because it always time out on travis for some reason) and fixed the things it complained about 👍 

cc/ @elopio

### DIFF
--- a/pipeline/snapcraft/snapcraft.go
+++ b/pipeline/snapcraft/snapcraft.go
@@ -52,7 +52,7 @@ func (Pipe) Run(ctx *context.Context) error {
 		log.Error("no snapcraft summary defined, skipping")
 		return nil
 	}
-	if ctx.Config.Snapcraft.Summary == "" {
+	if ctx.Config.Snapcraft.Description == "" {
 		log.Error("no snapcraft description defined, skipping")
 		return nil
 	}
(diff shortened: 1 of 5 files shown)


pipeline/snapcraft/snapcraft.go
```diff
@@ -131,11 +134,14 @@ func create(ctx *context.Context, folder, arch string, binaries []context.Binary
 		return err
 	}
 
-	snap := metadata.Name + "_" + metadata.Version + "_" + arch + ".snap"
+	snap := filepath.Join(
+		ctx.Config.Dist,
+		metadata.Name+"_"+metadata.Version+"_"+arch+".snap",
+	)
```
elopio: This will fail because we are setting cmd.Dir before running the command, so it will try to put the .snap in an unexisting dist dir, in the wrong path.

It will work with this diff:
http://paste.ubuntu.com/25233953/

caarlos0: thanks @elopio, just pushed the fix

PR #328, opened 2017-08-07
allow to overwrite the snap name

elopio:


elopio:
From your comment in bug #328, I realized that we need to allow to overwrite the `name` in the snap.yaml in case the project name doesn't match with the valid values of the Ubuntu store.

caarlos0:
@elopio the name must be in the format `name_version_arch.snap` or it can anything?

elopio:
@caarlos0 The .snap name can be anything. The `name` in the .yaml needs to be `^[a-z0-9][a-z0-9+-]*$`.

caarlos0:
@elopio got it. We can add the templating in another time :)

PR #329, opened 2017-08-08
wip snapcraft filename template

### LINKED ISSUE #322: Name template for snap packages
Our "final" pipes all have the option of templating the final name, maybe we should add this feature to snaps as well.

caarlos0:
closes #322

### DIFF
--- a/pipeline/defaults/defaults.go
+++ b/pipeline/defaults/defaults.go
@@ -12,7 +12,7 @@ import (
 )
 
 // NameTemplate default name_template for the archive.
-const NameTemplate = "{{ .Binary }}_{{.Version}}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}"
+const NameTemplate = "{{ .ProjectName }}_{{.Version}}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}"
 
 // SnapshotNameTemplate represents the default format for snapshot release names.
 const SnapshotNameTemplate = "SNAPSHOT-{{ .Commit }}"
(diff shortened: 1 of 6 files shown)


caarlos0:
cc/ @elopio

pipeline/defaults/defaults.go
```diff
@@ -37,6 +37,12 @@ func (Pipe) Run(ctx *context.Context) error {
 	if ctx.Config.ProjectName == "" {
 		ctx.Config.ProjectName = ctx.Config.Release.GitHub.Name
 	}
+	if ctx.Config.Snapcraft.Name == "" {
+		ctx.Config.Snapcraft.Name = ctx.Config.ProjectName
```
elopio: This feels like the wrong place, don't you think?
The snapcraft pipe should take care of all the things snapcraft. This default file seems like a catch all that goes against the abstraction and modularity of releaser.

But well, this was just a comment that nobody asked about possible changes in the design, for the future. If it works now, it is good for me :D

pipeline/snapcraft/snapcraft.go
```diff
@@ -143,10 +141,17 @@ func create(ctx *context.Context, folder, arch string, binaries []context.Binary
...
 	}
 
-	snap := filepath.Join(
-		ctx.Config.Dist,
-		ctx.Config.ProjectName+"_"+metadata.Version+"_"+arch+".snap",
-	)
+	var target = buildtarget.Target{OS: "linux", Arch: arch}
+	if strings.Contains(arch, "arm") {
+		target.Arch = "arm"
+		target.Arm = strings.Replace(arch, "arm", "", 1)
```
elopio: could this transformation be hidden inside ForSnap, instead of a prerequisite step?

caarlos0:
this was the wrong way of fixing it anyways

PR #330, opened 2017-08-09
Add fpm path mappings

lox:
This adds a way to add arbitrary extra files or dirs to fpm packages:

```
fpm:
  files:
    "scripts/etc/init.d/": "/etc/init.d"
```


caarlos0:
awesome, thanks!

lox:
I'll follow up with adding all the various hooks.

PR #335, opened 2017-08-17
fix RPM cross-compiling (OSX => amazonlinux/CentOS)

matthewmueller:
This change fixes an error you'll receive if you try building an RPM package on OSX and installing it on linux:

```
rpm is intended for a different operating system
```

From `fpm -h`, you'll see:

```
--rpm-os OS                   (rpm only) The operating system to target this rpm for. You want to set this to 'linux' if you are using fpm on OS X, for example
```

There's probably a more generic way of doing this, but for this use case the patch will work well.

### DIFF
--- a/pipeline/fpm/fpm.go
+++ b/pipeline/fpm/fpm.go
@@ -46,7 +46,7 @@ func (Pipe) Run(ctx *context.Context) error {
 			arch := archFor(platform)
 			for folder, binaries := range groups {
 				g.Go(func() error {
-					return create(ctx, format, folder, arch, binaries)
+					return create(ctx, format, folder, arch, platform, binaries)
 				})
 			}
 		}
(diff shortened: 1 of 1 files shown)


pipeline/fpm/fpm.go
```diff
@@ -99,6 +99,11 @@ func create(ctx *context.Context, format, folder, arch string, binaries []contex
 		options = append(options, "--conflicts", conflict)
 	}
 
+	// FPM requires --rpm-os=linux if your rpm target is linux
+	if format == "rpm" && strings.Contains(platform, "linux") {
```
caarlos0: we can remove the `strings.Contains(platform, "linux")`, as right now this we only call this func when the platform is linux. Check:

https://github.com/goreleaser/goreleaser/blob/658d46397f6f6913d767987ab21c2de4d8ed9681/pipeline/fpm/fpm.go#L41

We can then also remove the additional `platform` parameter.

caarlos0:
thanks for the PR!

I have just one small comment :)

caarlos0:
I'll fix it

matthewmueller:
ooo good catch and thanks for the merge!

PR #336, opened 2017-08-17
improved logs

caarlos0:
- following the same template of the other logs: `skipping because X`
- improved validation of snapcraft pipe

### DIFF
--- a/pipeline/snapcraft/snapcraft.go
+++ b/pipeline/snapcraft/snapcraft.go
@@ -19,6 +19,12 @@ import (
 // ErrNoSnapcraft is shown when snapcraft cannot be found in $PATH
 var ErrNoSnapcraft = errors.New("snapcraft not present in $PATH")
 
+// ErrNoDescription is shown when no description provided
+var ErrNoDescription = errors.New("no description provided for snapcraft")
+
+// ErrNoSummary is shown when no summary provided
+var ErrNoSummary = errors.New("no summary provided for snapcraft")
+
 // SnapcraftMetadata to generate the snap package
 type SnapcraftMetadata struct {
 	Name          string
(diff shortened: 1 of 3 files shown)


pipeline/snapcraft/snapcraft_test.go
```diff
@@ -17,24 +18,23 @@ func TestDescription(t *testing.T) {
 }
 
 func TestRunPipeMissingInfo(t *testing.T) {
-	for name, snap := range map[string]config.Snapcraft{
-		"missing summary": {
+	for eerr, snap := range map[error]config.Snapcraft{
```
elopio: why the extra `e` in `eerr`? Was that a typo, or some golang styling?

pipeline/snapcraft/snapcraft.go
```diff
@@ -48,13 +54,15 @@ func (Pipe) Description() string {
 
 // Run the pipe
 func (Pipe) Run(ctx *context.Context) error {
-	if ctx.Config.Snapcraft.Summary == "" {
-		log.Error("no snapcraft summary defined, skipping")
+	if ctx.Config.Snapcraft.Summary == "" && ctx.Config.Snapcraft.Description == "" {
+		log.Warn("skipping because no summary nor description were provided")
```
elopio: Now that you are returning errors, shouldn't this one be an error too?

caarlos0: yeah, I'll make that in another patch. I'm thinking about creating a global `SkipError` or something like that...

pipeline/snapcraft/snapcraft_test.go
caarlos0: `err` is already being used. `eerr` stand for `expected error`

elopio: There's only one thing I hate about go, and is the short variable names. It's so hard for me to parse them...
I'm +1 on this PR. Thanks!

caarlos0: @elopio yeah, I don't like it so much either, but it's ok for small methods IMHO...

thaks

PR #337, opened 2017-08-18
building rpm packages as well

caarlos0:
it fails on my macOS box, let's try on a linux :)


pipeline/fpm/fpm_test.go
```diff
@@ -80,7 +65,7 @@ func TestNoFPMInPath(t *testing.T) {
 	var ctx = &context.Context{
```
bpicode: Hint: one should supply a Version, otherwise fpm->rpmbuild will not work.

caarlos0: thanks, apparently that's why rpmbuild was failing!

PR #341, opened 2017-08-20
skip pipe as an error type and improved output

caarlos0:
- skipping a pipe is now an error type
- improved some logs
- improved snapcraft test coverage


elopio:
nice refactor! I like it thanks.

Late again, I'm sorry, I have been jumping between conferences.

caarlos0:
no problem, thanks for reviewing @elopio :)

PR #347, opened 2017-08-27
padronizing snapcraft packages name

### LINKED ISSUE #322: Name template for snap packages

caarlos0:
it bugs me that the names are different:

image

I was trying to solve it the wrong way on #329, this one is the right way :)

closes #322


caarlos0:
closes #322

PR #349, opened 2017-08-27
checksums name template support

### LINKED ISSUE #343: Checksum file for each release?
Hi :)

Would it make sense to offer the option to generate one checksum file per archive within the dist folder? My use-case is that I'd like to simply rsync my local dist folder up to out assets server which might already contain other releases that are not available locally. In this case I'd pollute the shared checksums.txt file and potentially remove entries for it by accident.

caarlos0:
closes #343

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -11,8 +11,10 @@ builds:
       - amd64
       - arm
       - arm64
+checksum:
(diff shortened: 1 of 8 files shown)


zerok:
Awesome! Thank you :D

PR #352, opened 2017-09-09
Print a more verbose error message if archiving fails

zerok:
## Types of changes

I recently ran into an issue where one of the glob patterns in my `.goreleaser.yml` file didn't result an any existing files and therefore caused the archive step to fail. Sadly, the error message only said that a file didn't exist but not what file and in what step.

With this PR I've added a couple of more verbose error messages that should help debug these issues in the future :)

### DIFF
--- a/pipeline/archive/archive.go
+++ b/pipeline/archive/archive.go
@@ -4,6 +4,7 @@
 package archive
 
 import (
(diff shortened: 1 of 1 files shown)


caarlos0:
Thanks!

Issue #320, opened 2017-07-31
Add Docker-Image-Building / Push to Dockerhub (like Homebrew)


andygrunwald:
goreleaser has support for Homebrew (to push new releases, after a release is done).
Docker is also a wide used tool. Maybe this would also make sense to implement a docker build + tag + push to the configured registry?

One disadvantage would be to couple goreleaser to Docker.
Maybe a "Hook"-System would be good. With this people can develop use cases to push releases to different sources. Another use case for this would be a push to [Artifactory](https://www.jfrog.com/open-source/) or AWS S3

caarlos0:
The docker thing should be reasonable easy, basically:

- create a dockerfile from scratch (or assume one is created)
- docker build
- docker push

I think i'll look into that today

andygrunwald:
Yes, exactly.
The thing is more that this is kind of extendable. Because i think a lot of people have different use cases. Another use case could be to push to Artifactory or AWS S3.

I would be happy to review or test such a feature (more general or specific to docker).

caarlos0:
hey @andygrunwald , can you take a look, plz? https://github.com/goreleaser/goreleaser/pull/355

caarlos0:
TODO:

- [ ] templating in the image name (e.g.: user/image:{{.Tag }}_blah)
- [ ] deal with image names when its a snapshot
- [ ] a way to also tag and push latest (eg: latest: true in the config)
- [ ] docs (after the feature is complete)
- [ ] set the docker repository url

caarlos0:
I think this is done!

andygrunwald:
Thanks a lot @caarlos0!

PR #355, opened 2017-09-12
draft: Docker integration

caarlos0:
## Types of changes

------

So, this is not complete yet because I want to make sure the API is good enough first.

What I thought:

```yaml
dockers:
  - goos: linux
    goarch: amd64
    goarm: ''
    dockerfile: Dockerfile
    image: goreleaser/goreleaser
```

## This allows the user to:

- have several docker images (for different platforms or not)
- write their own dockerfiles

## How it works:

User can define multiple dockerfiles anywhere in the project, the dockerfile for the specific binary and platform is copied into the respective dist folder and `docker build` is run from that folder.

Yeap, that simple.

## What I would add in other PRs:

- templating in the image name (e.g.: `user/image:{{.Tag }}_blah`)
- deal with image names when its a snapshot
- a way to also tag and push latest (eg: `latest: true` in the config)
- docs (after the feature is complete)
- release goreleaser with docker as well
- set the docker repository url

---

What do you guys think?

refs #320

### DIFF
--- a/pipeline/docker/docker.go
+++ b/pipeline/docker/docker.go
@@ -9,10 +9,9 @@ import (
 
 	"github.com/apex/log"
 	"github.com/goreleaser/goreleaser/config"
-	"github.com/pkg/errors"
-
 	"github.com/goreleaser/goreleaser/context"
 	"github.com/goreleaser/goreleaser/pipeline"
+	"github.com/pkg/errors"
 )
 
 // Pipe for checksums
--- a/pipeline/docker/docker.go
+++ b/pipeline/docker/docker.go
@@ -1,0 +1,107 @@
+// Package docker provides a Pipe that creates and pushes a Docker image
+package docker
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os/exec"
+	"path/filepath"
+
+	"github.com/apex/log"
+	"github.com/goreleaser/goreleaser/config"
+	"github.com/pkg/errors"
+
+	"github.com/goreleaser/goreleaser/context"
+	"github.com/goreleaser/goreleaser/pipeline"
+)
+
+// Pipe for checksums
+type Pipe struct{}
+
+// Description of the pipe
+func (Pipe) Description() string {
+	return "Creating Docker images"
+}
+
+// Run the pipe
+func (Pipe) Run(ctx *context.Context) (err error) {
+	if ctx.Config.Docker[0].Image == "" {
+		return pipeline.Skip("docker section is not configured")
+	}
+	if ctx.Config.Release.Draft {
+		return pipeline.Skip("release is marked as draft")
+	}
+	for _, docker := range ctx.Config.Docker {
+		var imagePlatform = docker.Goos + docker.Goarch + docker.Goarm
+		for platform, groups := range ctx.Binaries {
+			if platform != imagePlatform {
+				continue
+			}
+			for folder, binaries := range groups {
+				for _, binary := range binaries {
+					if binary.Name != docker.Binary {
+						continue
+					}
+					var err = doRun(ctx, folder, docker, binary)
+					if err != nil && !pipeline.IsSkip(err) {
+						return err
+					}
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func processDocker(ctx *context.Context, docker config.Docker) error {
+
+}
+
+func doRun(ctx *context.Context, folder string, docker config.Docker, binary context.Binary) error {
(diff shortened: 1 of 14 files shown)


pipeline/docker/docker.go
```diff
@@ -0,0 +1,110 @@
...
+	var dockerfile = filepath.Join(root, "Dockerfile")
+	var image = fmt.Sprintf("%s:%s", docker.Image, ctx.Git.CurrentTag)
+
+	bts, err := ioutil.ReadFile(docker.Dockerfile)
+	if err != nil {
+		return errors.Wrap(err, "failed to read dockerfile")
+	}
+	if err := ioutil.WriteFile(dockerfile, bts, 0755); err != nil {
+		return err
+	}
```
andygrunwald: Why do we read the dockerfile here and write it after it directly with the same content? Or is the purpose to set chmod to 0755 ?

pipeline/docker/docker.go
```diff
@@ -0,0 +1,110 @@
...
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func doRun(ctx *context.Context, folder string, docker config.Docker, binary context.Binary) error {
+	var root = filepath.Join(ctx.Config.Dist, folder)
+	var dockerfile = filepath.Join(root, "Dockerfile")
```
andygrunwald: Would it make sense to make the name of the file (in this case `Dockerfile`) also configurable, but the default? The reason is: I saw already a few projects that owns multiple Dockerfiles. One purpose of this can be a `Dockerfile` (for prod) and a `Dockerfile.dev` (for specific dev envs with profilers, etc.)
In a `docker build -t ...` you can add the `-f Dockerfile.dev` parameter to tell Docker which Dockerfile to build

pipeline/docker/docker.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	if err := dockerPush(image); err != nil {
+		return err
+	}
+	return nil
+}
+
+func dockerBuild(root, image string) error {
+	log.WithField("image", image).Info("building docker image")
+	var cmd = exec.Command("docker", "build", "-t", image, root)
```
andygrunwald: If the name of the `Dockerfile` will be configurable in the goreleaser.yml, then we need to apply the `-f` parameter here

andygrunwald:
@caarlos0 Wow. First review looks good. I made two small comments. I didn't tested it yet.
Need to setup a test repo and test case for a live test. Will try this during the day, ok?

pipeline/docker/docker.go
caarlos0: I'm was copying the file... there is no way in the stdlib for doing so :/ 

https://github.com/golang/go/issues/8868

anyway, I realized I could just link the files instead of copying... using ioutil to do that wasn't a good idea anyway

pipeline/docker/docker.go
caarlos0: I don't see why anyone would use a Dockerfile.dev within goreleaser... 

anyway, we can add in the future if people ask for it :D

andygrunwald: According to the PR description and the sample configuration, i have implied you had think of this feature:

Copy from the PR text:
```
dockers:
  - goos: linux
    goarch: amd64
    goarm: ''
    dockerfile: Dockerfile
    image: goreleaser/goreleaser
```

Or what is the use case for `dockerfile: Dockerfile`?

One project with a `Dockerfile.dev` is https://github.com/cloudflare/complainer.
And here you have a bunch of projects that have a `Dockerfile.dev`: https://github.com/search?utf8=%E2%9C%93&q=filename%3ADockerfile.dev&type=Code
To be fair, this is not only golang projects., but this shows that there is a use case for this.

caarlos0: ohh, I think I get what you meant... lemme fix that real quick

pipeline/docker/docker.go
andygrunwald: Thanks for explaination

skisulli:
We use an internal Artifactory server, would it be out of scope to allow setting the docker remote URL as part of the config?

caarlos0:
@skisulli I think it is, will do in a future PR so this one doesn't get too big

andygrunwald:
Thanks a lot @caarlos0!

@skisulli Afaik this is already possible. Docker compose supports something like `my-artifactory.jfrog.io/imageName:tagName`as an image name.

So the host name can be part of the image name here.

PR #361, opened 2017-09-15
support github enterprise

caarlos0:


pipeline/brew/template.go
```diff
@@ -22,7 +22,13 @@ type templateData struct {
 const formulaTemplate = `class {{ .Name }} < Formula
   desc "{{ .Desc }}"
   homepage "{{ .Homepage }}"
-  url "https://github.com/{{ .Repo.Owner }}/{{ .Repo.Name }}/releases/download/{{ .Tag }}/{{ .File }}"
+
+  {{ if .Repo.DownloadsURL }}
+    url "{{ .Repo.DownloadsURL }}{{ .Repo.Owner }}/{{ .Repo.Name }}/releases/download/{{ .Tag }}/{{ .File }}"
```
caarlos0: this is actually wrong, should be the URL of the release repo...

sklevenz: Right. Would it make sense to pass whole config object as parameter? Or alternatively add the downloads url of the release repo as single parameter.

caarlos0: yeah, I'm thinking about that... 

Can we assume that github enterprise is used only for "enterprises"?

If yes, we can setup the URLs in a single time in the config, and use them everywhere, as if it is being used by a company, it's very likely that they will release only to their github, right?

If uploading artifacts to a github enteprise and the formula to a public github repo is not a use case, this can be implemented in a cleaner way...

sklevenz: @caarlos0 Yes. This sound good. I do not see a need for a mixed mode. It's either all github.com or all enterprise.

caarlos0: yeah, I think that makes sense.

Will fix this later today :)

caarlos0: another question: can the upload and download URLs be guessed to defaults like `uploads.mygithub.com` and `download.mygithub.com`?

if yes, we can have only one URL in the config :)

sklevenz: Unfortunately not. Here is what we have:

api url: https://github.foo.home/api/v3/
download url: https://github.foo.home/
uploads url: https://github.foo.home/api/uploads/

caarlos0: hmm, OK, good to know :)

caarlos0: @sklevenz I pushed the changes (with some delay, sorry), mind to take a look?

caarlos0:
@sklevenz I pushed the changes (with some delay, sorry), mind to take a look?

shairozan:
Still can't find any actual information in the docs as to what keys to actual set for this to work. Can someone point to documentation regarding this key / keys in the drone yml file?

PR #373, opened 2017-09-30
Add config field to overwrite brew commit author

### LINKED ISSUE #292: Override the commit author for brew
The commit author is currently hard coded to [goreleaserbot](https://github.com/goreleaser/goreleaser/blob/e8bdba5752e3e3b6a2a9850c64d333e1e4f81be7/internal/client/github.go#L34-L35). Is it possible to override that via `.goreleaser.yml` or maybe just use `ctx.Config.Brew.GitHub.Owner`?

jorinvo:
Default author name and email can be set in config file.
Close #292.

I think it is most flexible if we allow the user to set the commit author in the config file instead 
of guessing a correct value.
In most cases users won't care so I left the old default values.
In other cases I think the logged in user is not the best guess since organizations mostly rather specify a separate email to indicate that this commit was made by a bot, not a human.

I hope I didn't forget to update any relevant code.
And I am not completely sure I put it in the best place in the config file - open to suggestions here.

## Types of changes

- New feature (non-breaking change which adds functionality)
- I have updated the documentation accordingly.
- I have updated the tests to cover my changes.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -36,19 +36,26 @@ func (r Repo) String() string {
 
 // Homebrew contains the brew section
 type Homebrew struct {
-	GitHub       Repo     `yaml:",omitempty"`
-	Folder       string   `yaml:",omitempty"`
-	Caveats      string   `yaml:",omitempty"`
-	Plist        string   `yaml:",omitempty"`
-	Install      string   `yaml:",omitempty"`
-	Dependencies []string `yaml:",omitempty"`
-	Test         string   `yaml:",omitempty"`
-	Conflicts    []string `yaml:",omitempty"`
-	Description  string   `yaml:",omitempty"`
-	Homepage     string   `yaml:",omitempty"`
+	GitHub       Repo         `yaml:",omitempty"`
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -22,13 +22,20 @@ type GitHubURLs struct {
 
 // Repo represents any kind of repo (github, gitlab, etc)
 type Repo struct {
-	Owner string `yaml:",omitempty"`
-	Name  string `yaml:",omitempty"`
+	Owner  string       `yaml:",omitempty"`
+	Name   string       `yaml:",omitempty"`
+	Author CommitAuthor `yaml:",omitempty"`
```
caarlos0: I would put this in the brew section, since repo can be used in the release as well and it does not apply there..

jorinvo: Agree. Was thinking that this is easier to understand but guess we can fix with a comment in the docs.

jorinvo:
@caarlos0 I changed it to be `brew.commit_author`. Thant should be more clear.

PR #382, opened 2017-10-02
feat: Add archive config 'wrap_in_directory'

### LINKED ISSUE #251: sub directory inside the archive
It's most common in open source that `project-1.2.3.tar.gz` would extract into `project-1.2.3/` rather than just into `.`.

Extracting into `.` is surprising and dangerous you might well extract over peoples files by accident.

Can you please add a option to do the same? If off-by-default then a flag in the yaml file to activate this behaviour would be great.

This is quite a long standing standard and it's quite surprising now to see anything still not doing it.  See for example point 6.1 in http://ftp.icm.edu.pl/packages/linux-tsx-11/docs/HOWTO/Software-Release-Practice-HOWTO and indeed the downloadable tarballs github makes

thanks!

jorinvo:
The field is optional. When set to true, files in archive are wrapped
in a directory, which has the same name as the archive itself.
Tests have also been extended to cover this.

Closes #251

If you think another name for the option is more clear or we need further customization, please let me know.

## Types of changes

- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -106,6 +106,7 @@ type Archive struct {
 	Format          string            `yaml:",omitempty"`
 	FormatOverrides []FormatOverride  `yaml:"format_overrides,omitempty"`
 	NameTemplate    string            `yaml:"name_template,omitempty"`
+	WrapInDirectory bool              `yaml:"wrap_in_directory,omitempty"`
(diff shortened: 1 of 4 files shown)


caarlos0:
Awesome, thanks @jorinvo !

PR #401, opened 2017-10-15
feat: Changelog customization

caarlos0:
## Types of changes

- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.


jorinvo:
> I turned myself into a pipe morty!
> 
> PipeRick!!!
 
😄 
`changelog.filters.excludes` as a regex looks like a good and simple solution to me. It should solve the majority of cases and otherwise people can still create their own file from scratch.

caarlos0:
> changelog.filters.excludes as a regex looks like a good and simple solution to me. It should solve the majority of cases and otherwise people can still create their own file from scratch.

@jorinvo done

jorinvo:
Nice! Should merge this. Then we can also use it for the changelog of GoReleaser itself :)

tj:
piperick!

Issue #284, opened 2017-07-05
Changelog transform command


tj:
Hey hey!

I was thinking a clean way to implement changelog massaging without getting into templates would be to simply pass it through stdio of an optional command (shell).

Then something like `changelog: sort | grep -v docs` could group them together and filter some of the less important stuff. 

<img width="707" alt="screen shot 2017-07-04 at 6 29 48 pm" src="https://user-images.githubusercontent.com/25254/27845850-988bac80-60e7-11e7-912e-4601995ea279.png">

caarlos0:
hey @tj 🎉 

hmm, right you can do it with ~~some~~ a lot of hacking:

```
goreleaser --release-notes <(git log --pretty=oneline --abbrev-commit $(git describe --tags --abbrev=0 $(git describe --tags --abbrev=0)^)..$(git describe --tags --abbrev=0) | sort | grep -v docs)
```

I agree that this is not ideal though. The entire changelog is kind of crappy and very basic right now... I'm thinking about improving it by adding PRs merged, issues closed, contributors and that kind of fluffy stuff... For sure personalization would be in mind for all that =D

I was thinking about doing something like this:

```
changelog: # new pipe yay
  sort:
    by: title / commit / none
    direction: asc / desc
  filter:
    - word1
    - word2
```

This would cover your current request, and we may continue adding new features later.

What do you think?

tj:
hmm yea I wonder where to draw the line, having some things baked in would be convenient. I'd be happy with even just being able to specify --release-notes in the config, then we can chuck that one-liner in there or maybe I'll write a little Go program to organize it a bit

tj:
Actually --release-notes is probably fine! Feel free to close if no one else wants it

caarlos0:
Nice!

FWIW I think hugo has some custom scripts as well, you can check the source or maybe ping <at>bep :P 

I'll leave the issue here for now... will think more about it later :)

jorinvo:
I like the current approach via `--release-notes`. 
By dealing with plain files you are flexible to do unixy things as @caarlos0 showed above, which would be more difficult with a setting in the configuration file.
If we would add changelog options to GoReleaser, it would be hard to decide what should be added and what should be skipped.
I guess in many cases users also like to have a completely custom changelog.

Maybe we can add some examples for different formats to the documentation in the README.
We can start by adding the example from above and later on we can add more examples commands to generate the changelog.

[...]

PR #408, opened 2017-10-20
feat: validate changelog config for invalid fields

caarlos0:
If the user configures the changelog section with an invalid key, show
an error.


caarlos0:
refss #284

PR #430, opened 2017-12-02
refactor: defaulter interface

caarlos0:
Right now the code looks weird because the defaults
of a pipe are far away of the implementation of the pipe.

the intend of this PR is to bring them closer by having a
Defaulter interface.

I also renamed the Pipe interface to Piper, and removed
the Description method in favor for fmt.Stringer.

The next step (in another PR) would probably be some sort 
of refactor in the config itself, by bringing each pipe's config
closer to the pipe itself.

What do you guys think?

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)
- [ ] I have read the **CONTRIBUTING** document.


pipeline/default.go
```diff
@@ -0,0 +1,15 @@
+package pipeline
+
+import (
+	"fmt"
+
+	"github.com/goreleaser/goreleaser/context"
+)
+
+// Defaulter interface
```
jorinvo: Maybe can be a bit more descriptive here, something like:
`Defaulter can be implemented by a Piper to set default values for its configuration.`

goreleaserlib/goreleaser.go
```diff
@@ -26,7 +26,7 @@ import (
 	yaml "gopkg.in/yaml.v2"
 )
 
-var pipes = []pipeline.Pipe{
+var pipes = []pipeline.Piper{
 	defaults.Pipe{},  // load default configs
```
jorinvo: Maybe it would be nicer if `defaults` is not a `Piper` and instead of explicitly listing

caarlos0: the problem is the order of things, the default's order is different from the pipe's order...

jorinvo: Ah missed that part. Never mind then :)

PR #437, opened 2017-12-05
feat: env vars for Docker tag_template

magiconair:
* fix `loadEnv()` function to split env vars only into two pars
* move env vars into `context.Context`
* add env var support for Docker `tag_template`

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] I have read the **CONTRIBUTING** document.
- [ ] `make ci` passes on my machine. (fpm not installed)
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.

Issue #446, opened 2017-12-06
use commitlint probot instead of gitcop


caarlos0:
pros:

- gitcop has its config via web interface while commitlint probot is configurable via config files
- commitlint seems more versatile

cons:

- ?

@jorinvo what do you think?

caarlos0:
actually seems like right now the commitlint bot don't read the config files https://github.com/ahmed-taj/commitlint-bot/issues/1

ahmed-taj:
Hey @caarlos0, glad you like commitlint bot.

Yes, it doesn't support custom confrontation yet, I'm working on it, with help of @marionebl the author of commitlint  package

Disclaimer: I didn't use Gitcop but I can see it's a business so I will assume there is an effort to keep the service running 24/7. But for my commitlint bot, I can't afford hosting costs, so it will only be running on a free plan of Heruko (or similar services), I can't promise service availability.

The good news is that the project is open source under MIT, and you can always host your own instance :)

caarlos0:
@jorinvo hello 👋 can you give me permissions on gitcop? I can't edit anything..

I would like to remove the need of a commit body... it is extremely annoying while doing big refactories.

Also the bot is commenting 2 times for every commit...

caarlos0:
> Also the bot is commenting 2 times for every commit...

oh: it was because the webhook was duplicated... just removed one... should be ok now...

jorinvo:
@caarlos0 I disabled the required body. 

The Gitcop interface seems super basic. Can't see any way to add multiple people. If you like, you can setup a new account and I remove mine. But adding two people doesn't seem to work unfortunately :/

jorinvo:
Can also setup commitlint of course. Can see if it does everything we want. Then we can remove Gitcop all together. For me it doesn't really matter since they seem to just do their job, both of them :D

caarlos0:
thanks @jorinvo, commitlint is not configurable yet :/

when they do that, I'll be glad to make the migration, since we can then have the config as code =)

caarlos0:
not stale

caarlos0:
it is not a problem right now, let's ignore this for now

Issue #377, opened 2017-10-01
Add Git Commit Message Guidelines to CONTRIBUTING.md


jorinvo:
I really like this section Hugo has which describes the format to use for commit messages:
https://github.com/gohugoio/hugo/blob/master/CONTRIBUTING.md#git-commit-message-guidelines

This would allow for generating a more useful changelog for the releases.
Would you be interested in adding this? Any specific format perferences?

caarlos0:
@jorinvo Yes!

I was thinking about it some days ago.

I sometimes commit with really bad messages, a way of enforcing this would also be neat...

jorinvo:
@caarlos0 I'm afraid we can't really enforce it. All we can do is having one place and point to it whenever we see a not so nice commit message.

Should we just go with the same conventions as Hugo or do you have any other preferences?

caarlos0:
Maybe using gitcop? https://gitcop.com

jorinvo:
@caarlos0 nice! That looks way better! Can give it a try. I can do the setup if you like.

caarlos0:
Yeah, feel free to do it 🗡 

I saw it here: https://github.com/pagarme/pilot/pull/187#issuecomment-324526348

It's in portuguese, but basically, the action should either be add, remove, refactor, move, bump, update or release, which I think is fine, if agree we can copy it =D

jorinvo:
@caarlos0 I cannot add the webhook. 
Could you add https://hooks.gitcop.com/github
here: https://github.com/goreleaser/goreleaser/settings/hooks/new 
Also need a secret. I'll send it to you in Slack ok?

caarlos0:
@jorinvo ok 🍻

jorinvo:
It is set up. I merged their suggested list with the one you listed above @caarlos0.
Now looks like this:
![screen shot 2017-10-01 at 19 54 26](https://user-images.githubusercontent.com/738978/31057411-8331a9a2-a6e2-11e7-8efe-cae1a6622f25.png)

jorinvo:
I will also add this to CONTRIBUTING.md

caarlos0:
Awesome 🎉

PR #449, opened 2017-12-06
Inject binary into dockerfile

magiconair:
* allow env vars for `archive.name_template`
* inject binary filename into `Dockerfile`

Let me provide some background for this PR and if you agree with the approach then I'll add tests and update the documentation.

I want to use goreleaser to release https://github.com/fabiolb/fabio. I have some special requirements for naming and packaging most of which were already supported by goreleaser. At this point I don't want to change the naming of the artifacts since this would break the debian/ubuntu packages that people are building and which they have automated. 

For fabio I need to publish releases with the binary artifacts for multiple platforms with the following names:

```
fabio-VERSION-go1.x.x-GOOS_GOARCH
fabio-1.5.3-go1.9.2-linux_amd64
```

I also need to publish Docker containers which contain the go version in the tag since I sometimes publish the current version built with a newer go version, e.g.

```
fabiolb/fabio:1.5.3-go1.8.5
fabiolb/fabio:1.5.3-go1.9.2
```

#437 added support for the env vars in the Docker tags which allows me to inject the go version into the Docker tags.

This change now adds support for env vars in the `name_template` so that I can build artifacts with the go version. 

```yaml
archive:
  name_template: '{{ .ProjectName }}-{{ .Version }}-{{ .Env.GOVERSION }}-{{ .Os }}_{{ .Arch }}'
  format: binary
```

However, that breaks the generation of the Docker container since the name of the binary is no longer known upfront in a static Dockerfile. `COPY fabio /` no longer works since the binary is now `fabio-1.5.3-go1.9.2-linux_amd64`. 

The Dockerfile has an `ENV` directive which allows for some templating but I couldn't figure out how to overwrite the ENV variable from the command line with `docker build`. If that is possible then the change becomes much simpler and works outside of goreleaser as well and I would actually prefer that. (https://docs.docker.com/engine/reference/builder/#environment-replacement)

The approach in this PR is to copy the Dockerfile and expand the `${BINARY}` variable in the `Dockerfile` to the true filename of the artifact. Then you can write something like this:

```
FROM scratch
ENV BINARY fabio
COPY ${BINARY} /fabio
ENTRYPOINT ["/fabio"]
```

## Types of changes


GitCop:

 * Commit: 78da472a4594389fdda74792c9b5b9e47409d7a0
  * Your commit message does not contain a body

 * Commit: 3f0e06b1fdd04bfaf26317e9ae2a0b333622a655
  * Your commit message does not contain a body

---

GitCop:

 * Commit: 78da472a4594389fdda74792c9b5b9e47409d7a0
  * Your commit message does not contain a body

 * Commit: 3f0e06b1fdd04bfaf26317e9ae2a0b333622a655
  * Your commit message does not contain a body

---

caarlos0:
both changes seem ok to me :)

can you add tests and update the documentation as well?

thanks

magiconair:
Thx and I’ll do that later tonight.

magiconair:
@caarlos0 ready for review.

pipeline/docker/testdata/Dockerfile
```diff
@@ -1,2 +1,2 @@
 FROM scratch
-ADD mybin /
+ADD ${BINARY} /
```
caarlos0: shouldn't this one still be `mybin`?

GitCop:

 * Commit: a1213f98b462df1414726c3f1d426ef8328cb2c3
  * Your commit message does not contain a body

---

GitCop:

 * Commit: a1213f98b462df1414726c3f1d426ef8328cb2c3
  * Your commit message does not contain a body

---

magiconair:
yes. changed the wrong file. Pushed a fix.

BTW, the requirement to add a body for every commit runs a bit counter to having small commits. Usually the subject is enough. What's your take on that?

caarlos0:
> BTW, the requirement to add a body for every commit runs a bit counter to having small commits. Usually the subject is enough. What's your take on that?

yeah, I get mad at it sometimes too...
check #446 and #377 for ref

caarlos0:
@magiconair after investigating a bit further, I think that the problem is actually the `archive.format: binary`, not the name template itself...

the binary format + docker integration is broken :(

the name template with environment variables seem good though, I cherry-picked it to #451

caarlos0:
will look into this tomorrow. it's late here already

/remind me to fix this bug in 18 hourss

PR #453, opened 2017-12-08
WIP: fix: docker images with binary archive format

caarlos0:
closes #449 

@magiconair can you test if this branch works for your case? If yes, I'll finish and polish it up 🍻 

FWIW: our `context` package looks bad AF and probably needs a redesign.

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)
- [ ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/context/context.go
+++ b/context/context.go
@@ -25,7 +25,7 @@ type GitInfo struct {
 
 // Binary with pretty name and path
 type Binary struct {
-	Name, Path string
+	Name, Path, BuildName string
 }
 
 // Context carries along some data through the pipes
@@ -36,7 +36,7 @@ type Context struct {
 	Token        string
 	Git          GitInfo
 	Binaries     map[string]map[string][]Binary
-	Artifacts    []string
+	Artifacts    []Artifact
 	Dockers      []string
 	Brews        []string
 	ReleaseNotes string
@@ -56,13 +56,22 @@ var (
 	brewsLock     sync.Mutex
 )
 
+type Artifact struct {
+	Name, Path string
+}
+
 // AddArtifact adds a file to upload list
-func (ctx *Context) AddArtifact(file string) {
+func (ctx *Context) AddArtifact(name, file string) {
 	artifactsLock.Lock()
 	defer artifactsLock.Unlock()
 	file = strings.TrimPrefix(file, ctx.Config.Dist+string(filepath.Separator))
-	ctx.Artifacts = append(ctx.Artifacts, file)
-	log.WithField("artifact", file).Info("new release artifact")
+	ctx.Artifacts = append(ctx.Artifacts, Artifact{Name: name, Path: file})
+	log.WithFields(log.Fields{"file": file, "name": name}).Info("new release artifact")
+}
+
+func (ctx *Context) AddArtifactFromFile(file string) {
+	_, name := filepath.Split(file)
+	ctx.AddArtifact(name, file)
 }
 
 // AddBrew adds a brew tap to the brews list
@@ -82,7 +91,7 @@ func (ctx *Context) AddDocker(image string) {
 }
 
 // AddBinary adds a built binary to the current context
-func (ctx *Context) AddBinary(platform, folder, name, path string) {
+func (ctx *Context) AddBinary(platform, folder, name, buildName, path string) {
 	binariesLock.Lock()
 	defer binariesLock.Unlock()
 	if ctx.Binaries == nil {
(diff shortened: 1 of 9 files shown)


caarlos0:
FWIW, I just released [this release](https://github.com/caarlos0/test/releases/tag/v6.1.1) with those changes, and it worksss! 🎉

caarlos0:
tests will surely break though 🗡

magiconair:
I'll have a look.

context/context.go
```diff
@@ -56,13 +56,22 @@ var (
 	brewsLock     sync.Mutex
 )
 
+type Artifact struct {
+	Name, Path string
+}
+
 // AddArtifact adds a file to upload list
-func (ctx *Context) AddArtifact(file string) {
+func (ctx *Context) AddArtifact(name, file string) {
```
magiconair: I'd probably accept an `Artifact` here.

```go
func (ctx *Context) AddArtifact(a Artifact) { ... }
```

pipeline/archive/archive.go
```diff
@@ -97,7 +97,7 @@ func create(ctx *context.Context, platform string, groups map[string][]context.B
 				return fmt.Errorf("failed to add %s -> %s to the archive: %s", binary.Path, binary.Name, err.Error())
 			}
 		}
-		ctx.AddArtifact(archivePath)
+		ctx.AddArtifactFromFile(archivePath)
```
magiconair: If `AddArtifactFromFile` doesn't need additional context you could have an `NewArtifactFromFile` constructor for an `Artifact`.

magiconair:
Ah, missed that this wasn't merged yet. I'll test it and make a more thorough review.

magiconair:
How does that solve the problem of injecting the name of the binary into the `Dockerfile`?

My problem is that I need to have the `go1.x` version in the binary name which I solve by using the env var approach for the binary artifacts. With that it isn't clear what the name of the binary is that needs to be in the `Dockerfile`.

e.g. 

```
# linux binary artifact
# archive:
#    name_template: '{{ .ProjectName }}-{{ .Version }}-{{ .Env.GOVERSION }}-{{ .Os }}_{{ .Arch }}'
#    format: binary
fabio-1.5.3-go1.9.2-linux_amd64

# Dockerfile
FROM scratch
COPY ??? /fabio <-- which name do I need to use here?
ENTRYPOINT ["/fabio"]
```

caarlos0:
> My problem is that I need to have the go1.x version in the binary name which I solve by using the env var approach for the binary artifacts. With that it isn't clear what the name of the binary is that needs to be in the Dockerfile.

that wouldn't be a problem anymore. The binary itself inside the `dist` folder will have a clean name, so you can use `COPY fabio /fabio`. The problem was that when archive is `binary`, it builds the binary in `folder_with_name_template/binary_with_name_template`, now it will be `folder_with_name_template/binary`, so everything should work as expected. 

The environment variables were actually irrelevant to reproduce this issue, it was always happening when `archive.format = binary`..

caarlos0:
just to make sure:

- https://github.com/caarlos0/test/blob/master/.goreleaser.yml#L63
- https://github.com/caarlos0/test/releases/tag/v6.1.2

magiconair:
Actually, I need `binary_with_template_name` as the binary artifact since I don't ship a `.tar.gz` file. The files I publish to Github are `fabio-1.5.3-go1.9.2-linux_amd64`. The *same* binary then gets injected into the Docker container. 

The only way I can think of you're solving this is that during the Github publish step you publish the binary artifact under a different name. However, then the challenge is that all the names in the checksums won't match.

See https://github.com/fabiolb/fabio/releases/tag/v1.5.3 for an example.

magiconair:
OK, you're publishing the artifact under a different name. But this messed up the checksums.

```
$ cat dist/release-test-0.2.14.sha256
5039c136a8c971e16c69013e3ba10543a53a55d48f96a5cc2a26ac7cdff0d7df  release-test-0.2.14-go1.9.2-linux_amd64/release-test
6972ca85bd2c8ba7125acadf31b78c5436d7b4a840ed0683c9f89e18f7c244a0  release-test-0.2.14-go1.9.2-darwin_amd64/release-test
```
https://github.com/magiconair/release-test/releases/tag/v0.2.14

magiconair:
I think building one file and publishing it under a different name will cause a number of issues downstream since you have to code and remember that translation everywhere. `goreleaser` was doing the right thing before. The only thing that was missing was to tell Docker which file to use. I still think this is the right approach.

caarlos0:
> I think building one file and publishing it under a different name will cause a number of issues downstream since you have to code and remember that translation everywhere. goreleaser was doing the right thing before. The only thing that was missing was to tell Docker which file to use. I still think this is the right approach.

I still not sure about this. Let me think about it for a while, I'm tired right now and don't want to rush this in since it would be real hard to undo it in case it's the wrong move...

magiconair:
ofc. Take your time. Late here (NL) as well.

caarlos0:
ok, so I though about it and still think this is the valid approach.

I think that injecting the binary in the dockerfile may also be needed, but for different reasons...

imho this is only a complicated thing to do because of the design of the ctx package, which looks really bad and wasn't made thinking on all those things...

I'll try to redesign it having in mind all these use cases, let me know what you think.

caarlos0:
I'll do this in another PR.

PR #452, opened 2017-12-08
WIP: gpg signing (followup of #420)

magiconair:
* gpg_signing -> sign
* optionally sign checksum file only
* make signature file extension confiugrable
* support encrypted signing keys

This is based on PR #420 with some additional features. 

Tests and docs are still missing.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] I have read the **CONTRIBUTING** document.
- [ ] `make ci` passes on my machine.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.

PR #420, opened 2017-11-23
WIP: Adds GPG signing pipeline

simonswine:
This is WiP, (state works for me)

Just like to get ideas that need to be supported in config other than the key id. I can look after tests and and documentation then.

Fixes #166 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] I have read the **CONTRIBUTING** document.
- [x] `make ci` passes on my machine.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.

PR #460, opened 2017-12-13
feat: add artifact signing pipeline

### LINKED ISSUE #166: Upload signatures for binaries produced
Create and upload signatures for the binaries (or all files) produced.

This will probably involve some kind of config for signing key or method, like PGP for creating the signatures.

magiconair:
This patch adds a generic artifact signing pipeline.

This is a replacement for both #420 and #452 and this implementation relies on the tools the user has already installed.

Need to add a test.

Closes #420 
Closes #452 
Fixes #166 

## Types of changes


magiconair:
To enable signing artifacts with gpg the only thing users have to do is to add 

```yaml
sign:
  artifacts: checksum
```

Assuming that `gpg` is in the path and that a default signing key is available everything will just work including the decryption of keys. See the docs for full customization.

magiconair:
I'll write the test after dinner.

pipeline/sign/sign.go
```diff
@@ -0,0 +1,102 @@
...
+
+type Pipe struct{}
+
+func (Pipe) String() string {
+	return "signing artifacts"
+}
+
+
+func (Pipe) Run(ctx *context.Context) error {
+	// set defaults
```
caarlos0: defaults can be set by implementing the `pipeline.Defaulter` interface and adding it to the `pipeline/defaults` pipe. :)

caarlos0:
> Assuming that gpg is in the path and that a default signing key is available everything will just work including the decryption of keys. See the docs for full customization.

seems simple enough to me 👍

magiconair:
@caarlos0 I've added a test. This is ready for review.

magiconair:
dang. overlooked one thing. one sec.

magiconair:
now it should be ready to review

magiconair:
looking

magiconair:
Is there a reason travis needs to upgrade the kernel every time?

magiconair:
Maybe the gpg version is too old.

caarlos0:
@magiconair apparently it is a dep for snapcraft:

```
sudo apt-get install --yes snapd rpm

Reading package lists... Done

Building dependency tree       

Reading state information... Done

The following additional packages will be installed:

  debugedit liblua5.2-0 librpm3 librpmbuild3 librpmio3 librpmsign1

  linux-generic-lts-xenial linux-headers-4.4.0-103

  linux-headers-4.4.0-103-generic linux-headers-generic-lts-xenial

  linux-image-4.4.0-103-generic linux-image-extra-4.4.0-103-generic

  linux-image-generic-lts-xenial rpm-common rpm2cpio squashfs-tools systemd

Suggested packages:

  rpm-i18n fdutils linux-lts-xenial-tools alien elfutils systemd-ui

Recommended packages:

  thermald

The following NEW packages will be installed:

  debugedit liblua5.2-0 librpm3 librpmbuild3 librpmio3 librpmsign1

  linux-generic-lts-xenial linux-headers-4.4.0-103

  linux-headers-4.4.0-103-generic linux-headers-generic-lts-xenial

  linux-image-4.4.0-103-generic linux-image-extra-4.4.0-103-generic rpm

  rpm-common rpm2cpio snapd squashfs-tools systemd

The following packages will be upgraded:

  linux-image-generic-lts-xenial
```

magiconair:
gpg is at 1.4 and the keyring I've created is for 2.2. The format has changed. Thinking...

magiconair:
The test should work with both gpg 1.4, 2.0 and >= 2.1. Otherwise, `make ci` will break on other machines.

magiconair:
@caarlos0 tests pass and I've added another test for the AddChecksum to make the codecov happy. The test is sufficient for my standards. I'll leave it up to you to merge.

magiconair:
I'll work on the test a bit more.

magiconair:
I've simplified the code itself a bit more and extended the test to cover more cases. Assuming CI passes this should be good to go.

magiconair:
You have a strange PR process. For the last couple of things I've only been fighting your pipeline.

For example:

The way you use `gometalinter` complains about this if `err` is already defined:

```go
if err := foo(); err != nil {
    t.Fatal(err)
}
```

The main reason I'm writing code like this is so that `err` in the `if` statement is different than the `err` defined somewhere else because it is in a different block. With your `gometalinter` check I now have to write this as either

```go
if err = foo(); err != nil {
    t.Fatal(err)
}
```
or 
```go 
if err2 := foo(); err2 != nil {
    t.Fatal(err)
}
```
which both feels weird. 

`gometalinter` also complained about cyclomatic complexity of a test function which meant that I had to arbitrarily break up the test helper to make the CI step happy. 

Also, I had to fight with the test coverage tool which failed the PR because the test coverage didn't meet its standard of some number derived from the rest of the code. 

In the end I've opted to remove code that would have been helpful to notify the user if he/she shot him/herself in the foot but was not easy to test. The code itself was obvious but mocking even more stuff for just these code paths didn't seem worth it since someone else has to understand what I'm testing there. 

While I think that automating things is good I think the current config is pushing things a bit too far, complaining about minor things that can easily be determined in code review.

magiconair:
Point in case:

```
codecov/patch — 93.1% of diff hit (target 93.67%)
codecov/project — 93.65% (-0.03%) compared to 3ea8bc0
```

caarlos0:
Hi @magiconair, sorry for the delay!

> The way you use gometalinter complains about this if err is already defined:
> 
> if err := foo(); err != nil {
>     t.Fatal(err)
> }
> The main reason I'm writing code like this is so that err in the if statement is different than the err defined somewhere else because it is in a different block. With your gometalinter check I now have to write this as either
> 
> if err = foo(); err != nil {
>     t.Fatal(err)
> }
> or
> 
> if err2 := foo(); err2 != nil {
>     t.Fatal(err)
> }
> which both feels weird.

yeah, you were shadowing the previous `err`... 

> gometalinter also complained about cyclomatic complexity of a test function which meant that I had to arbitrarily break up the test helper to make the CI step happy.

guess we could disable gocyclo on test packages...

> Also, I had to fight with the test coverage tool which failed the PR because the test coverage didn't meet its standard of some number derived from the rest of the code.

I'm ok merging something without the coverage required if it can't be tested... guess this isn't clear anywhere... also don't know where to make it evident...

> In the end I've opted to remove code that would have been helpful to notify the user if he/she shot him/herself in the foot but was not easy to test. The code itself was obvious but mocking even more stuff for just these code paths didn't seem worth it since someone else has to understand what I'm testing there.

IMHO you can put it back without the coverage if you want...

> While I think that automating things is good I think the current config is pushing things a bit too far, complaining about minor things that can easily be determined in code review.

The whole purpose of the automation is to make the code review easier. I'm mostly maintaining this in my free time so I'm OK having a little too much automation if it saves me some time.

Either way, I sorry if the experience wasn't good for you :/

I guess we can discuss this in another issue if you want...

caarlos0:
Will merge and release this for now... if you want to revert that change feel free to open another PR (or let me know and I revert the commit).

this will be released as the next minor.

caarlos0:
Oh, and thanks again for the PR and the effort 🍺

magiconair:
Thx for merging this. I get the motivation and I understand the "maintenance during free time" aspect well :) 

Some of the checks are good and they helped me with the patch. 

The fact that the test coverage breaks the build is a bit too strict for my taste. Since you're going to have a look anyway you can ask people to write more tests. The indication that test coverage is low should also be enough for most contributors.

Some other things come down to style like the "shadowing of a variable". I'm doing this on purpose since this is easier to read for me. However, your project, your rules. 

I am really looking forward to releasing my own projects with goreleaser since this will make it so much easier. 

Thanks for the great tool.

Now for #449 which is the last thing I need before I can us goreleaser.

caarlos0:
Thanks for understanding =)

About #449, I'll try to work on it today or tomorrow. I have a wip patch for it but there are still things to fix.

I'll let you know when it's done :)

PR #515, opened 2018-01-19
Docker major minor etc

### LINKED ISSUE #461: Docker alias for major and minor versions
When we use semantic versions and create docker images we end up with versions like this:
- v1.2.4, v1.2.5
- v1.3.1
- v2.2.1
In some cases, people will not like to use the latest version and will get a specific version but this will be a patch version like v1.2.4.
I would like to set alias for just the major and minor, and every time I make a new tag these tags are updated to the newest image
example:
- `fooo:1` => v1.3.1
- `fooo:1.2` => v1.2.5
- `fooo:2` => v2.2.1

The minor tag is easier to implement because it's very improbable to tag a version less than the actual. But for the major version, we need to verify if the current tag is greater than other tags, example:
The major tags for `fooo:1` is the tag v1.3.1, if we add a new tag `v1.2.6` we will not update the alias for `fooo:1`

This makes sense or worth it?

caarlos0:
closes #461 

## Types of changes

* [ ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -57,6 +57,12 @@
   version = "v1.0.0"
 
 [[projects]]
+  name = "github.com/masterminds/semver"
+  packages = ["."]
+  revision = "15d8430ab86497c5c0da827b748823945e1cf1e1"
+  version = "v1.4.0"
+
+[[projects]]
   name = "github.com/mattn/go-colorable"
   packages = ["."]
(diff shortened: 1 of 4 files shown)


GitCop:

 * Commit: efdd35805d4a150cdfcc9ed64125f1b267104fe7
  * Invalid type. Valid types are **feat, fix, docs, style, refactor, perf, test, chore, revert, add, remove, move, bump, update, release**

---

Issue #496, opened 2018-01-01
Generalize this tool for multiple languages


andygrunwald:
### Prerequisites

* [X] Are you running the latest version?
* [X] Did you perform a search to see if this problem was already reported?

### Description

This tool is great and is a big help for everyone who writes golang apps. It automates an important task and enables everyone to ship early, ship often.

The basic idea of this issue is to **discuss** to generalize this tool for multiple languages (e.g. PHP, C, C++, Ruby, etc.). 
The root of this idea [was a tweet](https://twitter.com/andygrunwald/status/947456582161502208):

> Is there anything like #goreleaser (https://github.com/goreleaser/goreleaser …) for other programing languages? In short: A tool that automate your releases to platforms like github, etc. by configuring this via a file (e.g. yaml) to ease a "Ship early, ship often" culture.

In detail i was looking for something like goreleaser for #PHP. For this i asked the [TYPO3-Project](https://typo3.org/) [via twitter here](https://twitter.com/andygrunwald/status/947555491747713024).
They wrote a similar tool at https://github.com/bmack/typo3-darth.
Maybe with a different focus and other tasks, but i think if there is a docker image, a upload to github, etc. this would help the project as well.

If this proposal is not welcomed, feel free to close it.
I believe that such an approach can change a lot of tools to the better.

### Implementation

I can think of two possible implementations right now.

#### `language` like TravisCI

TravisCI uses a `language` keyword. See https://github.com/goreleaser/goreleaser/blob/master/.travis.yml#L3
Based on the language we can run the process.

####  Every language as a single Pipeline

Every language can be implemented as a single pipeline like the upload to Artifactory or Github right now. This would have a few advantages:

* We are able to implement multiple pipelines per language (e.g. different package managers per language e.g. composer for PHP or different compilers e.g. in C)
* We are able to support multiple compile / pipeline steps to run a PHP packaging of a cms + compile / transpile of javascript via webpack or so

### Discussion

This is a lot of work and required time / effort (and of course, this is a limiting factor).
And this might need a breaking change to shift the default behaviour of go compile.
But this would also have the chance to get more contributor and a bigger outreach.

I am sure there are multiple pro and cons about this.
But at first i want to get the discussion started, getting feedback or a voice at all.
Let me know what you / other people think about this.

andygrunwald:
Tonight i started working on this with @mre and @timglabisch.
[...]

PR #520, opened 2018-01-20
Prototype: Multiple language support with Rust

andygrunwald:
## Types of changes

* [X] New feature (non-breaking change which adds functionality)
* [X] I have read the **CONTRIBUTING** document.
* [X] `make ci` passes on my machine.
* [X] My change requires a change to the documentation.
* [X] I have updated the documentation accordingly.

## Why is it useful?

This is a prototype version of #496.
This version is far from complete, but it should start a first discussion or feedback round.
The feedback round is important to let me / us know if it is useful to put more effort into it or to drop the idea at all.

This PR shows how this can be done for Rust.
Testing would be welcomed.

## What it does?

It compiles a rust app with `cargo build` and adds it to the Artifacts.

## How can you test it?

1. Install Rust (e.g. via https://rustup.rs/)
2. Kickstart a new project via `cargo new --bin hello`
3. Place a new `.goreleaser.yaml` with this content into the project
```
rust:
- target:
  - x86_64-apple-darwin
  binary: hello

archive:
  format: tar.gz
  name_template: "{{ .ProjectName }}_{{ .Version }}_{{ .Os }}"
```
4. Tag it and release it like every other tool

## What is missing?

* [ ] Unit tests
* [ ] Proper testing with other pipes

## Challenges

* [ ] Goos, Goarch and Goarm are spreaded accross the codebase and used in a specific way (e.g. for templating). In the Rust context this adds a challenge, because targets are not splitted into the specific parts. They are represented like one string (e.g. `x86_64-apple-darwin` for Mac) known from LLVM
* [ ] Copy / Paste code: Right now we just copied many of the code parts. Reusage would make sense here

### DIFF
--- a/pipeline/rust/rust.go
+++ b/pipeline/rust/rust.go
@@ -40,43 +40,30 @@ func (Pipe) Run(ctx *context.Context) error {
 	return nil
 }
 
-/*
 // Default sets the pipe defaults
 func (Pipe) Default(ctx *context.Context) error {
-	for i, build := range ctx.Config.Builds {
-		ctx.Config.Builds[i] = buildWithDefaults(ctx, build)
+	// only set defaults if there are rusts in the config file.
+	if len(ctx.Config.Rust) == 0 {
+		return nil
 	}
-	if len(ctx.Config.Builds) == 0 {
-		ctx.Config.Builds = []config.Build{
-			buildWithDefaults(ctx, ctx.Config.SingleBuild),
-		}
+
+	for i, rust := range ctx.Config.Rust {
+		ctx.Config.Rust[i] = buildWithDefaults(ctx, rust)
 	}
 	return nil
 }
-*/
-/*
-func buildWithDefaults(ctx *context.Context, build config.Build) config.Build {
-	if build.Binary == "" {
-		build.Binary = ctx.Config.Release.GitHub.Name
+
+func buildWithDefaults(ctx *context.Context, rust config.Rust) config.Rust {
+	if rust.Binary == "" {
+		rust.Binary = ctx.Config.Release.GitHub.Name
 	}
-	if build.Main == "" {
-		build.Main = "."
-	}
-	if len(build.Goos) == 0 {
-		build.Goos = []string{"linux", "darwin"}
-	}
-	if len(build.Goarch) == 0 {
-		build.Goarch = []string{"amd64", "386"}
-	}
-	if len(build.Goarm) == 0 {
-		build.Goarm = []string{"6"}
-	}
-	if build.Ldflags == "" {
-		build.Ldflags = "-s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}}"
-	}
-	return build
+
+	// One idea would be, if no rust.Target are defined
+	// to add the default target for the current OS.
+	// E.g. hit `rustup show`. The default host is shown.
+	// For an up to date Mac is will show "Default host: x86_64-apple-darwin"
+	return rust
 }
-*/
 
 func runPipeOnBuild(ctx *context.Context, rust config.Rust) error {
 	if err := runHook(ctx, rust.Env, rust.Hooks.Pre); err != nil {
--- a/pipeline/rust/rust.go
+++ b/pipeline/rust/rust.go
@@ -143,10 +143,17 @@ func doBuild(ctx *context.Context, build config.Rust, target string) error {
 		Type: artifact.Binary,
 		Path: binary,
 		Name: binaryName,
-		// TODO This is a hack right now
+
+		// TODO Check and refactor this
+		// This is a Go specific part. Rust doesn't
+		// have a "Goos" or "Goarch" or "Goarm".
+		// It has targets like x86_64-apple-darwin.
+		// For now a workaround is to apply the target of Rust
+		// as Goos and the arch as "rust"
 		Goos:   target,
-		Goarch: "goarch",
+		Goarch: "",
 		Goarm:  "",
+
 		Extra: map[string]string{
 			"Binary": build.Binary,
 			"Ext":    ext,
--- a/pipeline/rust/rust.go
+++ b/pipeline/rust/rust.go
@@ -32,6 +32,10 @@ func (Pipe) Run(ctx *context.Context) error {
 	}
 
 	for _, rust := range ctx.Config.Rust {
+		if len(rust.Target) == 0 {
+			return pipeline.Skip("rust section is not configured properly (missing targets)")
+		}
+
 		log.WithField("rust", rust).Debug("building")
 		if err := runPipeOnBuild(ctx, rust); err != nil {
 			return err
--- a/pipeline/rust/rust.go
+++ b/pipeline/rust/rust.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 
 	"github.com/apex/log"
 	"github.com/pkg/errors"
@@ -77,11 +78,10 @@ func buildWithDefaults(ctx *context.Context, build config.Build) config.Build {
 */
 
 func runPipeOnBuild(ctx *context.Context, rust config.Rust) error {
-	/*
-		if err := runHook(ctx, build.Env, build.Hooks.Pre); err != nil {
-			return errors.Wrap(err, "pre hook failed")
-		}
-	*/
+	if err := runHook(ctx, rust.Env, rust.Hooks.Pre); err != nil {
+		return errors.Wrap(err, "pre hook failed")
+	}
+
 	sem := make(chan bool, ctx.Parallelism)
 	var g errgroup.Group
 	for _, target := range rust.Target {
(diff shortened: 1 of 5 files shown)


caarlos0:
I like the idea but I'm not sure we should do it this way.

goreleaser already does too much (as someone else already stated in twitter).

I'm thinking that maybe we should have some sort pluggable pipes...

examples:

- a `piper` that build for rust (or maybe in this case it could even be a `builder` instead of a `piper`
- a `piper` that publishes new releases to twitter
- a `piper` that generates the `| sh` shell scripts

I don't yet know how to make all this though.

I like the way terraform works, for example, and I think it's a good place to start studying the implementation...

caarlos0:
what do you think?

caarlos0:
thinking about a `builder` interface, we could have an interface like this:

```go
type Builder interface {
	Build(ctx *context.Context, build config.Build, target buildtarget.Target) error
}
```

and then on the build pipe, something like:

```go
var builder Builder
switch lang {
case "go":
	builder = gobuilder.Default
case "rust":
	builder = rustbuilder.Default
}
```

or maybe even better: builders register themselves:

```go
package gobuilder

func init() {
	build.Register("go", Default)
}
```

and we n do something like this in the build pipe:

```go
build.ForLang(lang).Build(/* ... */)
```

what do you think?

caarlos0:
I've prototyped something on #522 

check it out @andygrunwald !

caarlos0:
about the goos, goarch and goarm on the artifact struct...

I think we can change them to uname-relatives. They are used to filter artifacts on other pipes (eg: docker, fpm, brew).

question is: can we break the rust `target` into os + arch + arm?

caarlos0:
> question is: can we break the rust target into os + arch + arm reliably?

looking into the results of `rustc --print target-list`, seems like we can!

pipeline/rust/rust.go
```diff
@@ -0,0 +1,213 @@
...
+
+func buildWithDefaults(ctx *context.Context, rust config.Rust) config.Rust {
+	if rust.Binary == "" {
+		rust.Binary = ctx.Config.Release.GitHub.Name
+	}
+
+	// One idea would be, if no rust.Target are defined
+	// to add the default target for the current OS.
+	// E.g. hit `rustup show`. The default host is shown.
+	// For an up to date Mac is will show "Default host: x86_64-apple-darwin"
```
caarlos0: I think we could default this to the same targets as go:

- x86_64-apple-darwin
- x86_64-unknown-linux-gnu
- i686-apple-darwin
- i686-unknown-linux-gnu

andygrunwald:
> goreleaser already does too much (as someone else already stated in twitter).

I agree here. When we continue this road, it will become hard to maintain.

I like your idea of the "builder" and "piper" examples.
To generalize it even more: What about languages that are target independent? I think about JavaScript, Python (e.g. egg files), PHP (composer, phar), Java, Kotlin, Scala (JVM).
They have only one target.

> I like the way terraform works, for example, and I think it's a good place to start studying the implementation...

Never looked deep into it. So at the moment, i am not a help here :(

> about the goos, goarch and goarm on the artifact struct...
> question is: can we break the rust target into os + arch + arm reliably?
> looking into the results of rustc --print target-list, seems like we can!

Those targets are not targets created by rust. Those are targets from LLVM.
I didn't found a document what describes how to parse this, but there seems to be some kind of rule. A few examples:

aarch64-linux-android
aarch64-unknown-freebsd
aarch64-unknown-fuchsia
aarch64-unknown-linux-gnu
aarch64-unknown-linux-musl
asmjs-unknown-emscripten
i586-unknown-linux-gnu

Here are some slides that named it "target-triple" with `Target Triple: <arch><sub>-<vendor>-<sys>-<abi>`: https://llvm.org/devmtg/2014-04/PDFs/LightningTalks/2014-3-31_ClangTargetSupport_LighteningTalk.pdf

Maybe @mre or @timglabisch knows more?

pipeline/rust/rust.go
andygrunwald: We thought about this as well. Like If MacOSx than target this. If linux, then this and so on. But we have to take the architecture and even the linker-system into account (when it comes to GNU libc or Musl libc). And there it is getting hard to decide what is the default

mre:
Yup, target triples. More on that [here](https://github.com/rust-lang-nursery/rust-forge/blob/master/platform-support.md) and [here](https://github.com/japaric/rust-cross). 

> Triples are usually formatted as follows: {arch}-{vendor}-{sys}-{abi}. 
> For example, the triple arm-unknown-linux-gnueabihf refers to the systems that share these characteristics:
>  *  architecture: arm.
>  * vendor: unknown. In this case, no vendor was specified and/or is not important.
>  * system: linux.
>  * ABI: gnueabihf. gnueabihf indicates that the system uses glibc as its C standard library (libc) implementation and has hardware accelerated floating point arithmetic (i.e. an FPU).

The attentive reader might have noticed, that triples are not necessarily *triples*. 

Also - contrary to Go - Rust can not cross compile to all platforms on every platfrom. For example, it's [quite tricky to compile to Mac from a Linux machine](https://github.com/japaric/rust-cross#i-want-to-build-binaries-for-linux-mac-and-windows-how-do-i-cross-compile-from-linux-to-mac). Most people use external services like Travis or AppVeyor for that. There's also [trust](https://github.com/japaric/trust), which provides templates for these services to make cross compiling easier, but it still requires manual steps.

So, I guess if we want to support Rust, we should be a bit more conservative with the build targets and only use the host triple as a default. We can always add more targets later if possible.

pipeline/rust/rust.go
mre: Won't be easy. See comment [here](https://github.com/goreleaser/goreleaser/pull/520#issuecomment-359570387).

caarlos0:
so I guess we can default to the current "triplet" only...

andygrunwald:
@mre @caarlos0 How do we continue here?
#522 is merged (🎉 ), this was the basement for this, right?

Should we (i, @mre, @timglabisch) finish this (with your help)?
Would you (@caarlos0) build the first implementation to show how it is done?

How did we align with the triplet defaults? Or with the native `cargo` cmd?
@mre You mentioned something related to "Build Rust inside a Docker container" for Linux, right? Can you elaborate on this or show an example?
And do you see this as a must-have? Or do you consider this as a user problem?

mre:
> @mre You mentioned something related to "Build Rust inside a Docker container" for Linux, right? Can you elaborate on this or show an example?

There's [cross](https://github.com/japaric/cross) and [cargo-docker](https://github.com/DenisKolodin/cargo-docker), which both use Docker for cross-compiling Rust code.

> And do you see this as a must-have? Or do you consider this as a user problem?

It's completely optional, I guess. At a future stage, we could piggy-back on the above tools for cross-compling, if they are installed. Alternatively, we could integrate a Docker pipeline to be completely independent. A sort of middle-ground would be having our own public (Rust) images that can be used for cross-compling.

andygrunwald:
> Alternatively, we could integrate a Docker pipeline to be completely independent. A sort of middle-ground would be having our own public (Rust) images that can be used for cross-compling.

You mean a `docker-rust`? or a `docker` lang impl? this would be smart.

mre:
> You mean a docker-rust? or a docker lang impl? this would be smart.

Was thinking about a pure, language-agnostic `docker` pipeline. The user would define, which image will be used for compiling. Mostly, people might want to use the official images like [rust](https://hub.docker.com/_/rust/) or [go](https://hub.docker.com/_/golang/) for cross-compiling. Yes, even for Go it makes sense because not everybody might want to install the golang toolchain on their build server. Same for PHP + composer or Javascript + Yarn.
Just an idea. 😜

caarlos0:
> #522 is merged (🎉 ), this was the basement for this, right?

yeap =D



> Should we (i, @mre, @timglabisch) finish this (with your help)?
> Would you (@caarlos0) build the first implementation to show how it is done?

I think you we can kind of do it together: change this to be a builder impl, we can change what else needs to be changed while doing it.

Meanwhile, I'll experiment with plugins and etc

caarlos0:
> Was thinking about a pure, language-agnostic docker pipeline. The user would define, which image will be used for compiling. Mostly, people might want to use the official images like rust or go for cross-compiling. Yes, even for Go it makes sense because not everybody might want to install the golang toolchain on their build server. Same for PHP + composer or Javascript + Yarn.
> Just an idea. 😜

like this idea as well =D

we could have a `rust` lang and a more generic `docker` lang (or `docker-go`, `docker-rust` etc)

mre:
> we could have a rust lang and a more generic docker lang (or docker-go, docker-rust etc)

Exactly. A pre-configured rust toolchain using cargo that covers most use-cases and a language-agnostic Docker toolchain for total control, which can be used for cross-compiling.

caarlos0:
or maybe we can do the plugin thing later (hopefully go will have native plugins for all platforms soon enough?)?

guidiego:
Some news about use goreleaser + cargo?

caarlos0:
> @caarlos0 @mre What do you think?

no problem on my side

caarlos0:
> Some news about use goreleaser + cargo?

not at the moment :/

andygrunwald:
From my side as well no progress so far.
Time is quite limited at the moment, to less time to wrap my head around :(
Sorry.

caarlos0:
> From my side as well no progress so far.
> Time is quite limited at the moment, to less time to wrap my head around :(
> Sorry.

no problem, take your time :)

caarlos0:
will close this for now.

I'll try to work on a plugin system in the next few days.

Thanks!

PR #544, opened 2018-02-08
feat: Adds template support to binary property

ewilde:
I added support for using the Go template engine with the binary
property. This is needed for generating binaries that include
the current version number in their file name. For example
`terraform-provider-NAME_vX.Y.Z`.

## Types of changes

### DIFF
--- a/docs/050-build.md
+++ b/docs/050-build.md
@@ -17,6 +17,13 @@ builds:
     main: ./cmd/main.go
 
     # Name of the binary.
+    # This is parsed with the Go template engine and the following variables
+    # are available:
+    # - Date
+    # - Commit
+    # - Tag
+    # - Version (Git tag without `v` prefix)
+    # Date format is `2006-01-02_15:04:05`.
     # Default is the name of the project directory.
     binary: program
 
--- a/pipeline/build/build.go
+++ b/pipeline/build/build.go
@@ -18,6 +18,9 @@ import (
 
 	// langs to init
 	_ "github.com/goreleaser/goreleaser/internal/builders/golang"
+	"time"
+	"bytes"
+	"text/template"
 )
 
 // Pipe for build
@@ -58,6 +61,7 @@ func buildWithDefaults(ctx *context.Context, build config.Build) config.Build {
 	if build.Binary == "" {
 		build.Binary = ctx.Config.Release.GitHub.Name
 	}
+
 	return builders.For(build.Lang).WithDefaults(build)
 }
 
@@ -95,6 +99,13 @@ func runHook(ctx *context.Context, env []string, hook string) error {
 
 func doBuild(ctx *context.Context, build config.Build, target string) error {
 	var ext = extFor(target)
+
+	binary, err := binary(ctx, build)
+	if err != nil {
+		return err
+	}
+
+	build.Binary = binary
 	var name = build.Binary + ext
 	var path = filepath.Join(ctx.Config.Dist, target, name)
 	log.WithField("binary", path).Info("building")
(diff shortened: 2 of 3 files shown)


ewilde:
@caarlos0 not sure why the travis build is failing here, noticed that you had similar issues recently with the same tests. Any pointers :smile:  ?

caarlos0:
> @caarlos0 not sure why the travis build is failing here, noticed that you had similar issues recently with the same tests. Any pointers 😄 ?

seems like snapcraft is broken, will look into it 👍

caarlos0:
I'm curious on what is your use case on this... can you explain?

ewilde:
This is needed for generating binaries that include
the current version number in their file name. For example
`foo_vX.Y.Z`. My particular use case is for creating terraform plugins. Terraform uses the version number in the file name to determine the plugins version see: https://github.com/ewilde/terraform-provider-runscope/issues/23

My current work around before this PR is to `envsubst` the `goreleaser.yml` : https://github.com/ewilde/terraform-provider-runscope/blob/0efe91927152e2d4ddb7591a316eabb4fe4f196b/scripts/release.sh#L5

caarlos0:
can't you release it using the `binary` format in `archive`?

https://goreleaser.com/#archive

ewilde:
@caarlos0 if you release it using `binary` format in `archive`, then I get the desired binary name; however it's then not package as a `zip` or `tar.gz` which is still wanted/needed

ewilde:
@caarlos0 any thoughts on my [last comment](https://github.com/goreleaser/goreleaser/pull/544#issuecomment-364465595)?

caarlos0:
yeah, sorry I forgot to reply.

I'm thinking about what can break by doing this.

godownloader will surely not work on this kind of release as-is.

another point is that all other properties that are template-able are name `name_template` or something like that. This one is not. Changing it will break the compatibility with previous versions...

I also think that it's not very common to do releases with the version in the binary inside the archive... so there is that as well.

Regarding your release script, you could also do `goreleaser -f <(envsubst '$VERSION'  < ./goreleaser.yml.template)` to "simplify" it a little.

I guess what I'm saying is: let me think about this for a while, as Solomon Hykes once said, "no is temporary, yes is forever." ([ref](https://twitter.com/solomonstre/status/715277134978113536))

ewilde:
@caarlos0 agreed all valid points.

> godownloader will surely not work on this kind of release as-is.

I've not used `godownloader` looks neat.

> another point is that all other properties that are template-able are name name_template or something like that

I guess options would be to: 1. break the current convention, or 2: make a new property `binary_template` which takes precedence over `binary`? Making a breaking change and renaming `binary` to `binary_template` doesn't seem worth it

> I also think that it's not very common to do releases with the version in the binary inside the archive... so there is that as well.

Agreed, also I don't think it's common to do releases with the version as part of the `binary name` whether or not it's part of an archive. My use case with `terraform plugins` is probably not that common

> Regarding your release script, you could also do goreleaser -f <(envsubst '$VERSION' < ./goreleaser.yml.template) to "simplify" it a little.

Thanks for the release script simplification, pretty much solves my immediate problem. I didn't realise that command substituting the contents of a file into an argument which accepts the file name would work!

Happy to make this change work with `godownloader` or  alternative close this PR and just go with `envsubst`.

caarlos0:
well, seems like more people need this, I think we can break the convention, another field `binary_template` will probably be confusing, renaming it also doesn't worth it, so let's proceed as is :)

caarlos0:
releasing v0.50.0

Issue #387, opened 2017-10-04
scoop.sh App Manifest Generation


Southclaws:
### Prerequisites

* [x] Did you perform a search to see if this suggestion already existed?
* [x] Do you think GoReleaser is awesome?

### Description

Hi! Absolutely loving GoReleaser so far! I actually found the tool from [this blog post](https://medium.com/from-the-couch/adventures-in-go-ba217c29b51d) which talked about cross-platform deployment and installation! Great, just what I needed, it builds .deb and .rpm, does homebrew, everything except... Windows!

So, I took a look at scoop.sh and really like how minimal it is. I think, since GoReleaser already has my GitHub token for the release creation why not allow it to generate a scoop.sh [App Manifest](https://github.com/lukesampson/scoop/wiki/App-Manifests) and push it to a specified [Bucket](https://github.com/lukesampson/scoop/wiki/Buckets)?

I could possibly take a crack at this myself in a few days if the maintainer(s) are busy.

Let me know your thoughts! I did a quick search for scoop.sh in the issues and found #131 but I think scoop.sh is a nice minimal complement to Chocolatey and the implementation would be super simple from what I can see (generate json>push to repo>done!)

caarlos0:
Scoop seems awesome!

I don't have a windows machine to test all that, unfortunately. I can though help/guide you if are willing to work on this :D

Southclaws:
Alright then, since I have a windows machine I'll have a crack at this
tomorrow or Friday hopefully!

caarlos0:
Nice, thanks @Southclaws =D

lemme know if you need anything, you can also join goreleaser on the gophers slack :)

Southclaws:
Working on this in https://github.com/Southclaws/goreleaser

I'm basically copying the Brew pipeline as it's pretty much the same process:

- Generate a file based on yaml loaded into `ctx.Config`
- Push that file to a repo the user has access to

Southclaws:
I'm wondering what's the best way to get the binary URLs from the GitHub release pipeline so I can add those URLs as binary downloads for the scoop manifest: https://github.com/Southclaws/goreleaser/blob/master/pipeline/scoop/scoop.go#L83

jorinvo:
@Southclaws do I understand it correctly, scoop needs an URL directly to a binary?

Southclaws:
[...]

PR #547, opened 2018-02-10
Scoop.sh App Manifest

Southclaws:
Finished the Scoop.sh manifest generator for Windows releases!

See #387

## Types of changes

### DIFF
--- a/pipeline/scoop/scoop.go
+++ b/pipeline/scoop/scoop.go
@@ -101,6 +101,7 @@ type Resource struct {
 
 func buildManifest(ctx *context.Context, client client.Client, artifacts []artifact.Artifact) (result bytes.Buffer, err error) {
 	var githubURL = "https://github.com"
+	// TODO: Make this default global
 	if ctx.Config.GitHubURLs.Download != "" {
 		githubURL = ctx.Config.GitHubURLs.Download
 	}
--- a/pipeline/scoop/scoop.go
+++ b/pipeline/scoop/scoop.go
@@ -1,0 +1,165 @@
+// Package scoop provides a Pipe that generates a scoop.sh App Manifest and pushes it to a bucket
+package scoop
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/goreleaser/goreleaser/config"
+	"github.com/goreleaser/goreleaser/context"
+	"github.com/goreleaser/goreleaser/internal/artifact"
+	"github.com/goreleaser/goreleaser/internal/client"
+	"github.com/goreleaser/goreleaser/pipeline"
+)
+
+// ErrNoWindows when there is no build for windows (goos doesn't contain windows)
+var ErrNoWindows = errors.New("scoop requires a windows build")
+
+// Pipe for build
+type Pipe struct{}
+
+func (Pipe) String() string {
+	return "creating Scoop Manifest"
+}
+
+// Run the pipe
+func (Pipe) Run(ctx *context.Context) error {
+	client, err := client.NewGitHub(ctx)
+	if err != nil {
+		return err
+	}
+	return doRun(ctx, client)
+}
+
+// Default sets the pipe defaults
+func (Pipe) Default(ctx *context.Context) error {
+	if ctx.Config.Scoop.CommitAuthor.Name == "" {
+		ctx.Config.Scoop.CommitAuthor.Name = "goreleaserbot"
+	}
+	if ctx.Config.Scoop.CommitAuthor.Email == "" {
+		ctx.Config.Scoop.CommitAuthor.Email = "goreleaser@carlosbecker.com"
+	}
+	return nil
+}
+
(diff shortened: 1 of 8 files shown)


pipeline/scoop/scoop_test.go
```diff
@@ -0,0 +1,423 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			for _, a := range tt.artifacts {
+				tt.args.ctx.Artifacts.Add(a)
+			}
+			err := doRun(tt.args.ctx, tt.args.client)
+			if tt.wantErr {
+				assert.Error(t, err)
```
caarlos0: should be `tt` here

pipeline/scoop/scoop_test.go
```diff
@@ -0,0 +1,423 @@
...
+			} else {
+				assert.NoError(t, err)
```
caarlos0: should be `tt` here as well

pipeline/scoop/scoop_test.go
```diff
@@ -0,0 +1,423 @@
...
+				Description: "A run pipe test formula",
+			},
+			false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotResult, err := buildManifest(tt.args.ctx, tt.args.client, tt.args.artifacts)
+			if tt.wantErr {
+				assert.Error(t, err)
```
caarlos0: in this method as well, should be `tt` instead of `t`

pipeline/scoop/scoop_test.go
```diff
@@ -0,0 +1,423 @@
...
+		{
+			"custom",
+			args{&context.Context{Version: "1.0.0", Config: config.Project{Release: config.Release{GitHub: config.Repo{Owner: "user", Name: "repo"}}}}, "https://git.my.company.com", "file.tar.gz"},
+			"https://git.my.company.com/user/repo/releases/download/1.0.0/file.tar.gz",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotURL := getDownloadURL(tt.args.ctx, tt.args.githubURL, tt.args.file)
+			assert.Equal(t, tt.wantURL, gotURL)
```
caarlos0: here too

pipeline/scoop/scoop.go
```diff
@@ -0,0 +1,146 @@
...
+// Resource represents a combination of a url and a binary name for an architecture
+type Resource struct {
+	URL string `json:"url"` // URL to the archive
+	Bin string `json:"bin"` // name of binary inside the archive
+}
+
+func buildManifest(ctx *context.Context, client client.Client, artifacts []artifact.Artifact) (result bytes.Buffer, err error) {
+	var githubURL = "https://github.com"
+	if ctx.Config.GitHubURLs.Download != "" {
+		githubURL = ctx.Config.GitHubURLs.Download
```
caarlos0: can you add a TODO here to make this default in some defaulter? (it's probably like this in more places, so let me fix it in another PR 👍 )

pipeline/scoop/scoop_test.go
Southclaws: `tt` is a struct, not a `*testing.T`. `t` is shadowed by the inline function, the test generator does this.

GitCop:

 * Commit: 1e0ff86d14f20bfdee662c091f428d4f6e976e0f

---

pipeline/scoop/scoop_test.go
caarlos0: oh, my bad, read it wrong, sorry :D

caarlos0:
Nice!

I think the only thing missing now is the docs, can you add that?

Also: what do I need to do to publish goreleaser to scoop? We should probably document how to install goreleaser with scoop as well in that case =D

Southclaws:
Yep, I can work on some docs tonight! I'm still trying to get a few error branches to trigger in coverage. They are actually hit during tests (I tested via step-through debugging) but for some reason, coverage marks the branches as not hit... I'll keep going at it though.

caarlos0:
repo created -> https://github.com/goreleaser/scoop-bucket :D

Southclaws:
On the topic of unit tests, I noticed that codecov actually reports correct coverage and it all seems 100% fine - the Go vscode plugin seem to be the problem on my end. The only issue is this line https://codecov.io/gh/goreleaser/goreleaser/src/a40353674f6abd1d84e9d825bc4f7d4b656691c5/pipeline/scoop/scoop.go#L133 which seems impossible to trigger given that `Manifest` is entirely made of valid JSON types, there seems no way to force an error case here. So it seems the only way to keep codecov happy for this pull would be to adjust the requirements or modify `Run` to support a dummy client (which would require work across the entire codebase).

But other than that, if you're happy with the docs, this looks ready to merge!

caarlos0:
LGTM

PR #575, opened 2018-02-21
refactor: unifying snapshot with skip-validate and skip-publish

caarlos0:
the `--skip-validate` and `--skip-publish` were implemented before snapshoting was introduced. I don't think they're still relevant, so I'm removing them in favor of snapshot.

The idea is also to improve the docs about this and the `--help` itself.

## Types of changes

* [ ] I have read the **CONTRIBUTING** document.


basgys:
@caarlos0 What will be the new behaviour of `--snapshot`? According to my understanding, it is still impossible to publish snapshots.

caarlos0:
> @caarlos0 What will be the new behaviour of --snapshot? According to my understanding, it is still impossible to publish snapshots.

yes, that was always the idea.
publishing snapshots requires a whole lot of behavior that goreleaser doesnt have (eg; overriding snapshots, etc)

basgys:
@caarlos0 It actually makes sense to keep the same behaviour for `--snapshot`. However, my initial request was to allow rolling releases either by overriding previous releases (not ideal) or by using another workflow (e.g. `--rolling-release`), but I understand that it would be a lot of work.

My problem now is that `goreleaser` won't work for me anymore now that `--skip-validate` is gone.

caarlos0:
@basgys I never seen this rolling release workflow on github.. can you point me out some repos that do that? 

maybe we can find a way of doing that without too many changes...

basgys:
@caarlos0 There is no rolling release workflow on Github. Rolling release is just a concept of continuous delivery. Probably nightly build is more correct. It is a "pointer" to a release.
The `latest` tag on Docker registry is a good example, it just points to the latest container pushed to the registry.

One of the solution for Github is to have a `latest` tag that would be assigned to the latest commit on each `gorelease`. Then the `latest` release has to be edited (text/description updated and new binaries uploaded). Note that I'm not sure about the feasibility of this solution.
Another solution is to check whether the plugin (Github, Docker, Artifactory, Homebrew, ...) supports nightly builds and skip those that don't.

I have the feeling I'm requesting a feature that might not fit into the `goreleaser` scope. As I mentioned, it would be really handy, but I can also use goreleaser for regular releases and roll my own solution for my continuous delivery pipeline.

caarlos0:
> I have the feeling I'm requesting a feature that might not fit into the goreleaser scope. As I mentioned, it would be really handy, but I can also use goreleaser for regular releases and roll my own solution for my continuous delivery pipeline.

yeah, nothing on goreleaser was think of in a rolling release way... maybe we could do that, but I'm not sure if we should...

Issue #578, opened 2018-02-25
Support short Git commit hashes


dnaka91:
### Describe the feature and how you think it should work.

Currently the full length commit hash is used everywhere. For file names (if `--snapshot` is used) as well as during the build process for later use at runtime.

In general the short hash is enough to identify a commit and thus a user should be able to configure to optionally use the short commit hash instead:
```yaml
git:
  short_hash: true
```

I already prepared the changes in my fork for personal use cases and can open a pull request if this feature request becomes accepted.

caarlos0:
Hi,

that's great, feel free to open the PR :)

caarlos0:
closed on #581

Issue #466, opened 2017-12-18
document how to run the sign pipe tests locally


caarlos0:
running it locally I got errors, as pointed out here -> https://github.com/goreleaser/goreleaser/pull/463#issuecomment-352288085

maybe @magiconair can help on that?

PS: CI passes as expected.

magiconair:
This works for me. I can have a look.

caarlos0:
I still need to have a look at this

PR #581, opened 2018-02-27
feat: Configuration to use short git hashes

dnaka91:
Add a git configuration to control to either use long commit hashes
(current default) or use the short version with only the first 7
characters of the hash.

See #578

## Types of changes

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -8,7 +8,7 @@ import (
 	"os"
 
 	"github.com/apex/log"
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
 )
 
 // GitHubURLs holds the URLs to be used when using github enterprise
@@ -212,6 +212,11 @@ type EnvFiles struct {
 	GitHubToken string `yaml:"github_token,omitempty"`
 }
 
+// Git config
+type Git struct {
+	ShortHash bool `yaml:"short_hash,omitempty"`
+}
+
 // Project includes all project configuration
 type Project struct {
 	ProjectName   string        `yaml:"project_name,omitempty"`
@@ -231,6 +236,7 @@ type Project struct {
 	Dist          string        `yaml:",omitempty"`
 	Sign          Sign          `yaml:",omitempty"`
 	EnvFiles      EnvFiles      `yaml:"env_files,omitempty"`
+	Git           Git           `yaml:",omitempty"`
 
 	// this is a hack ¯\_(ツ)_/¯
 	SingleBuild Build `yaml:"build,omitempty"`
--- a/pipeline/git/git_test.go
+++ b/pipeline/git/git_test.go
@@ -173,3 +173,19 @@ func TestSnapshotDirty(t *testing.T) {
 	ctx.Snapshot = true
 	assert.NoError(t, Pipe{}.Run(ctx))
 }
+
+func TestShortCommitHash(t *testing.T) {
+	_, back := testlib.Mktmp(t)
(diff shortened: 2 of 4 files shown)


dnaka91:
Seems that the `prettier` wants to reformat some Markdown files, but actually I didn't touch them.
Any suggestions?

Also couldn't run the `make ci` task on my machine (Mac), always failing with the error the `gpg-agent` is not running.

caarlos0:
> Also couldn't run the make ci task on my machine (Mac), always failing with the error the gpg-agent is not running.

this is a known issue #466 

will look into prettier, probably they change something 👍

caarlos0:
prettier fixed on master in theory

dnaka91:
I will add a test case for the new setting later today. Didn't know the code coverage is not allowed to go down (which is a good check actually).

Also if you like, I will add a section in the docs for this setting. First I thought, this setting is so minor, it's not really worth the documentation. But there is no reason not to document it, so better add it.

caarlos0:
yes, please =)

a lot of things are minors and are documented anyway, so... yeah hehe

about the coverage, yes, its a good check :)

config/config.go
```diff
@@ -8,7 +8,7 @@ import (
 	"os"
 
 	"github.com/apex/log"
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
```
caarlos0: neat 💅

docs/150-git.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Git
```
caarlos0: can we put this section under the environment section?

there are already some github-related docs there, maybe it makes sense for them to be together?

pipeline/git/git_test.go
```diff
@@ -173,3 +173,19 @@ func TestSnapshotDirty(t *testing.T) {
...
+	testlib.GitCommit(t, "first")
+	var ctx = context.New(config.Project{
+		Snapshot: config.Snapshot{
+			NameTemplate: "{{.Commit}}",
+		},
+	})
+	ctx.Snapshot = true
+	ctx.Config.Git.ShortHash = true
+	assert.NoError(t, Pipe{}.Run(ctx))
+	assert.Regexp(t, `^[a-z0-9]{7}$`, ctx.Version)
```
caarlos0: would probably be easier to use `assert.Len`

caarlos0:
Looks great overall!

A couple of minors that maybe make sense, let me know what you think!

(and thanks for the PR :D )

pipeline/git/git_test.go
dnaka91: First I thought so too, but then there is no guarantee that `{{ .Commit }}` really contains a Git hash or something else that just happens to be 7 characters long.

But I guess that would be another test, to check that we picked up an actual hash and not something else from the CLI when calling Git. For this specific feature though, `assert.Len` should be enough.

I will change it later.

docs/150-git.md
dnaka91: Hm. To me, the environment area looks like the bare setup to get started and further customization comes below.

On the other hand it's a really small setting and is no really worth its own section.

I will move it over then.

caarlos0: thanks :)

pipeline/git/git_test.go
caarlos0: makes sense indeed, thanks!

caarlos0:
thanks!

PR #597, opened 2018-03-07
feat: Custom brew tap name

ernoaapa:
Add optional Name field to Brew configuration to allow
overriding the name of the final Brew tap recipe.

closes #595

## Types of changes

* [ ] Bug fix (non-breaking change which fixes an issue)
* [x] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [x] I have read the **CONTRIBUTING** document.
* [ ] `make ci` passes on my machine.
* [x] My change requires a change to the documentation.
* [x] I have updated the documentation accordingly.
* [x] I have added tests to cover my changes.

PR #614, opened 2018-03-25
fix: Fix Brew formula class name when using custom name

ernoaapa:
The #597 added support for customizing the final Brew formula name.
The generated *.rb file still were using the project name as
class name and installing with custom name didn't work.
Fixed the issue by using the `name` field as an input for
building the class name, which defaults to project name
if no name given.

## Types of changes

* [x] Bug fix (non-breaking change which fixes an issue)


caarlos0:
good catch, thanks!

PR #617, opened 2018-03-28
feat: Add source pipeline

dominikschulz:
This PR adds a new pipeline step to build source packages. Those can be e.g. used to build proper homebrew packages.

## Types of changes

### DIFF
--- a/pipeline/source/source.go
+++ b/pipeline/source/source.go
@@ -1,0 +1,170 @@
+package source
+
+import (
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/apex/log"
+	"github.com/goreleaser/goreleaser/context"
+	"github.com/goreleaser/goreleaser/internal/artifact"
+)
+
+// SourceNameTemplate is the default value for the source name template
+const SourceNameTemplate = "{{.Binary}}-{{.Version}}"
+
+// Pipe for archive
+type Pipe struct{}
+
+// Description of the pipe
+func (Pipe) String() string {
+	return "Creating source archives"
+}
+
+// Default sets the defaults
+func (Pipe) Default(ctx *context.Context) error {
+	if ctx.Config.Source.NameTemplate == "" {
+		ctx.Config.Source.NameTemplate = SourceNameTemplate
+	}
+	return nil
+}
+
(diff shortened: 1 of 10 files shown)


GitCop:

 * Commit: 0d00619079f37a57d2b80f03857371565ebc44e5

---

pipeline/source/source.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	args := make([]string, 0, len(ctx.Config.Source.Excludes)+4)
+	for _, ex := range ctx.Config.Source.Excludes {
+		args = append(args, "--exclude="+ex)
+	}
+	args = append(args, fmt.Sprintf("--transform=s|/|/%s-%s/|", ctx.Config.Builds[0].Binary, ctx.Version))
+	args = append(args, "-Pczf")
+	args = append(args, path)
+	args = append(args, ".")
+	cmd := exec.Command("tar", args...)
```
caarlos0: can we do this using `archive/tar` package?

this makes `tar` a dependency of goreleaser, and since there are different versions of tar, gnu vs bsd and all that... I prefer to use plain go as much as possible...

dominikschulz:
I'll have a look at `archive/tar` and add try to add tests and docs.

pipeline/source/source.go
dominikschulz: You're perfectly right, any external dependency will eventually break ;)

dominikschulz:
@caarlos0 I switched to `archive/tar` and added tests and docs. Let me know what you think.

caarlos0:
I went into checking [`hugo` formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/hugo.rb) too see how it works

It uses the github archive URL:

```
  url "https://github.com/gohugoio/hugo/archive/v0.37.1.tar.gz"
```

Can't we just do the same thing?

For example, https://github.com/goreleaser/goreleaser/archive/v0.63.1.tar.gz seems to work..

dominikschulz:
We can certainly pull the archive from GitHub :+1:
For our usecase I'll need to find a way to inject the exact tag and commit which the source was built from but maybe we could work around that somehow.

caarlos0:
> For our usecase I'll need to find a way to inject the exact tag and commit which the source was built from but maybe we could work around that somehow.

hmm, maybe you can get that from the `version` field on the formula?

caarlos0:
anyway, will close this in favor of #625

PR #619, opened 2018-03-28
feat: Read release body template from config

dominikschulz:
This PR makes it possible to override the relase notes template from the config to e.g. avoid the ugly footer.

## Types of changes

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -115,6 +115,7 @@ type Release struct {
 	Draft        bool   `yaml:",omitempty"`
 	Prerelease   bool   `yaml:",omitempty"`
 	NameTemplate string `yaml:"name_template,omitempty"`
+	BodyTemplate string `yaml:"body_template,omitempty"`
 }
 
 // NFPM config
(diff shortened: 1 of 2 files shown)


GitCop:

 * Commit: 2a5812bb36839780509c0827de32806afceb5dab

---

caarlos0:
> void the ugly footer.

didn't know it was ugly 🤔

Don't know if I like this though, we can maybe just remove the footer and then the user can pass anything with the `--release-notes` flag...

https://goreleaser.com/#release.custom_release_notes

dominikschulz:
I'm sorry if my choice of words was misleading. Let's say we don't like the footer. I pretty much like this project and I'm happy to refer to it, but not in every release.

Passing release notes in with the `--release-notes` flag is nice, but this requires us to write release notes before. With this change, especially combined with #622, we're able to create pretty nice release notes without any additional work.

caarlos0:
OK, no problem.

I think it is better to just remove the footer. Then the `--release-notes` flag can solve this problem pretty easily :)

dominikschulz:
Ok, that would work as well.
I'll respect whatever you prefer, but is there a technical reason why you don't like this approach. IMHO it's very unobtrusive and clean.

caarlos0:
> Ok, that would work as well.
> I'll respect whatever you prefer, but is there a technical reason why you don't like this approach. IMHO it's very unobtrusive and clean.

no technical reason at all, just want to keep things as clean as possible...

caarlos0:
@dominikschulz check #624

PR #621, opened 2018-03-28
fix: Improved homebrew formula

### LINKED ISSUE #342: Brew tap not accepted in homebrew-core
See: https://github.com/Homebrew/homebrew-core/pull/16943

dominikschulz:
This PR improves the homebrew formula generation pipeline step to include additional fields that will allow to build "proper" formula from source files.

Fixes #342 

## Types of changes


GitCop:

 * Commit: a0a7ee191f2d71b54d039ddc952dbaf0c11a1d16

---

internal/artifact/artifact.go
```diff
@@ -11,6 +12,27 @@ import (
 //go:generate stringer -type=Type
 type Type int
 
+func (t Type) String() string {
```
caarlos0: this should not be needed, `go generate` should create this using `stringer`.

see line 12.

dominikschulz: Ah, now I see. This didn't work after a go get, of course. I'll remove it.

dominikschulz:
I'll try to address your comments and update the branch.

dominikschulz:
@caarlos0 Should be OK for review now.

docs/120-homebrew.md
```diff
@@ -78,6 +83,12 @@ brew:
     system "#{bin}/program --version"
     ...
 
+  # Specify any additional formula content in the special section.
+  # Default is empty.
+  special: |
```
caarlos0: this is not a homebrew thing, right?

the idea is to use this to add arbitrary instructions to the formula?

if yes, I think this is the indication of other missing features...

caarlos0:
check #625

docs/120-homebrew.md
dominikschulz: Yes, that's correct. We used to use this in the past but actually don't use it anymore. So I think for our use case we don't really need it.

dominikschulz:
Thanks, #625 looks pretty complete at first sight. I'll give it a try in the next days :+1:

caarlos0:
ok, so I'll close this in favor of #625 as well

PR #625, opened 2018-04-01
feat: homebrew: build from source

caarlos0:
This PR improves the homebrew formula generation pipeline step to include additional fields that will allow building proper formula from source files than can be accepted in the core repository.

Fixes #342
Closes #621
Closes #617 

## Types of changes

* [ ] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] I have read the **CONTRIBUTING** document.
* [ ] `make ci` passes on my machine.
* [ ] My change requires a change to the documentation.
* [ ] I have updated the documentation accordingly.
* [ ] I have added tests to cover my changes.

PR #542, opened 2018-02-06
fix: update brew template to use tilde heredoc syntax

ob-stripe:
brew recently deprecated the use of `<<-EOS.undent` in favor of `<<~EOS` (cf. https://github.com/Homebrew/brew/pull/3694/files#diff-d667e35be86b1a2b5d1f95ff39b4a878R3). This PR updates the brew template and test files to use the new syntax.

## Types of changes

* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [x] I have read the **CONTRIBUTING** document.
* [x] `make ci` passes on my machine.
* [ ] My change requires a change to the documentation.
* [ ] I have updated the documentation accordingly.
* [x] I have added tests to cover my changes.

PR #629, opened 2018-04-05
Improve generated Homebrew formulae caveats

elyscape:
Pull request #625 made some changes to the way that caveats are included in Homebrew formulae. Previously, the string value of the caveats config field was included directly into the formula. Now, surrounding whitespace is trimmed, it is split by line, and the lines are included into the formula with some leading whitespace for readability reasons. That's all sensible, but this leads to an unintended problem: if `cfg.Caveats` is the empty string (as will be the case if it's not present in the config file), `split()` will return a slice containing the empty string. This is considered truthy by the template processor, resulting in it including the caveats section with a blank line. We deal with this by adding a check in `split()` for this scenario and returning an empty slice instead.

Relatedly, #542 switched from using the deprecated `<<-EOS.undent` in Homebrew formulae to `<<~EOS` but, when #625 moved to using a heredoc for the caveats section, this subtlety was missed, and `<<-EOS.undent` was used instead.

### DIFF
--- a/pipeline/brew/brew.go
+++ b/pipeline/brew/brew.go
@@ -186,7 +186,11 @@ func dataFor(ctx *context.Context, client client.Client, artifact artifact.Artif
 }
 
 func split(s string) []string {
-	return strings.Split(strings.TrimSpace(s), "\n")
+	strings := strings.Split(strings.TrimSpace(s), "\n")
+	if len(strings) == 1 && strings[0] == "" {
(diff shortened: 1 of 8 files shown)


caarlos0:
Thanks @elyscape for the fixes, sorry for the trouble. Will merge and release asap!

caarlos0:
released, thanks!

PR #634, opened 2018-04-07
docs: fix insecure pipe to bash

jpillora:


caarlos0:
good catch, thank you!

Issue #499, opened 2018-01-06
Ensure tag has been pushed to github


wxio:
I've just started playing with goreleaser and managed to push a release to github, but I hadn't pushed the tag to origin. The source zip is therefore out of sync with the release as there were missing commits.

Is this a valid feature request?
i.e. the validation should (optionally) check that the github has the tag and commit being built. 

Or is there something about git I am misunderstanding.

caarlos0:
I remember that there was some discussion about this some time ago... I think the problem is to do this with sane defaults... like.. what should we do if there are multiple remotes and the tag is pushed to just one of them?

Not saying we shouldn't do this... just that I don't know the best way to do it...

millergarym:
I think that In the context of a github release there are some sane default.
That is the source and the release both refer to the same repository.
Github automatically creates a zip and tar of the source related to the release (I assume the source archive is at the same tag). Now if the commit haven't yet been pushed that source archive can't match.  The release change log shows this inconsistency, as there are commits with no link.

Is there even a way to build from one repo and release to another?

embano1:
Good discussion. Having the same mixed feelings here.

@caarlos0 wrt multiple remotes
IIRC, .goreleaser.yaml only accepts one destination repo. So one could check whether remote and src repo (branch) are in sync (i.e. no diff). May be make it a default flag, where one can opt out.

Just brainstorming here...

millergarym:
@caarlos0 we already have a GIT TOKEN else it wouldn't be possible to create a release. It might be that the token needs different access right, but this isn't a big issue.

@embano1 isn't it enough to check that the head hashs are the same?

caarlos0:
> @caarlos0 we already have a GIT TOKEN else it wouldn't be possible to create a release. It might be that the token needs different access right, but this isn't a big issue.

yeah, but the user might not be using github...

PR #643, opened 2018-04-13
feat: Create new tags for GitHub release

holladay-io:
If applied, this commit will create a new tag if it does not already exist in the remote repository. If the tag already exists, the provided target commitish will not be used. This will not impact users following the existing workflow (tag, push, release).

This commit allows the user to perform `git tag v1.0.0 && goreleaser` without pushing the tags to the remote repository. The git account used for compiling the code may not have different permissions than the token used to create the release. Also, currently `git tag v1.0.0 && goreleaser` without the tags pushed will run successfully but create an invalid release pointing to the latest master commit.

One caveat, this change has no effect if a short hash is used. 

From the GitHub docs [:](https://developer.github.com/v3/repos/releases/)

> target_commitish: Specifies the commitish value that determines where the Git tag is created from. Can be any branch or commit SHA. Unused if the Git tag already exists. Default: the repository's default branch (usually master).

### DIFF
--- a/internal/client/github.go
+++ b/internal/client/github.go
@@ -94,11 +94,12 @@ func (c *githubClient) CreateRelease(ctx *context.Context, body string) (int64, 
 		return 0, err
 	}
 	var data = &github.RepositoryRelease{
-		Name:       github.String(title),
-		TagName:    github.String(ctx.Git.CurrentTag),
-		Body:       github.String(body),
-		Draft:      github.Bool(ctx.Config.Release.Draft),
-		Prerelease: github.Bool(ctx.Config.Release.Prerelease),
+		Name:            github.String(title),
+		TagName:         github.String(ctx.Git.CurrentTag),
+		TargetCommitish: github.String(ctx.Git.Commit),
+		Body:            github.String(body),
+		Draft:           github.Bool(ctx.Config.Release.Draft),
+		Prerelease:      github.Bool(ctx.Config.Release.Prerelease),
 	}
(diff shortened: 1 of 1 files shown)


caarlos0:
so, in theory, this fixes #499 ?

I'm thinking that what will happen is that I'll create a tag locally, run goreleaser, and github will create the same tag on github, but I'm not sure what happens when I try to sync back up locally (same tag name, different objects)...

will try it out

holladay-io:
I tried to test the scenario you described. I created a local tag, ran `goreleaser` with the commitish, then synced the tags. It seems that after syncing an annotated tag becomes a lightweight tag. 

Here's how I tested:

```sh
$ git tag -a v0.0.0-testing -m "Creating a release from d275c96"

$ git for-each-ref refs/tags
4c7b9a43b88abdc17313fc401fb38abd3148e4a8 tag	refs/tags/v0.0.0-testing

$ git ls-remote --tags

$ ./goreleaser

$ git ls-remote --tags
From git@[REPO]
d275c967313d894dd4e04b9dcf2eb21340fd0a3d	refs/tags/v0.0.0-testing

$ git fetch --tags
From [REPO]
 t [tag update]      v0.0.0-testing -> v0.0.0-testing

$ git for-each-ref refs/tags
d275c967313d894dd4e04b9dcf2eb21340fd0a3d commit	refs/tags/v0.0.0-testing
```

caarlos0:
LGTM, thanks!

PR #645, opened 2018-04-18
feat: nfpm scripts

### LINKED ISSUE #635: Support for package scripts on nfpm
This was already added to nfpm on tag v0.7.0 (https://github.com/goreleaser/nfpm/issues/19), we just need to enable its config through goreleaser as well.

tympanix:
Resolves #635

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -137,6 +137,15 @@ type NFPM struct {
 	Bindir       string            `yaml:",omitempty"`
 	Files        map[string]string `yaml:",omitempty"`
 	ConfigFiles  map[string]string `yaml:"config_files,omitempty"`
+	Scripts      NFPMScripts       `yaml:"scripts,omitempty"`
+}
+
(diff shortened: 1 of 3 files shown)


tympanix:
You recently simplified the PR template (e3003ded3fd9ceb6f28f2e2b91d24d105698b673). Let me know if you need more info. Also `make ci` runs on my machine (macOS 10.13.4) except:
https://github.com/goreleaser/goreleaser/blob/42fe5761f74198782928f21f678a3f7bcf319670/Makefile#L47 
which results in:
```
xargs: prettier: No such file or directory
```
Experienced this before?

caarlos0:
> Experienced this before?

hmm, I think make setup is not installing prettier, you can install it with `npm install -g prettier`... but the build passed, so this part should be ok :)

caarlos0:
This looks good, thanks!

PR #648, opened 2018-04-20
feat: Add asmflags and gcflags fields

### LINKED ISSUE #638: Add templated asmflags and goflags fields
It would be nice to be able to use environment variables in the `flags` field. My use case is that I'd like to be able to pass `-gcflags all=-trimpath={{.Env.GOPATH}}` and `-asmflags all=-trimpath={{.Env.GOPATH}}` to strip my `GOPATH` from the various source filenames that are recorded in the binary. At present, the `ldflags` field is treated as a template, but the `flags` field is not, making this impossible.

elyscape:
This adds support for templated `asmflags` and `gcflags` fields to the config file to match the `ldflags` field. While less often useful than the `ldflags` field, these fields are sometimes useful, especially when set to `all=-trimpath={{.Env.GOPATH}}`.

Resolves #638.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -80,18 +80,20 @@ type IgnoredBuild struct {
 
 // Build contains the build configuration section
 type Build struct {
-	Goos    []string       `yaml:",omitempty"`
-	Goarch  []string       `yaml:",omitempty"`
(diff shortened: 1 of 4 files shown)


caarlos0:
thanks!

Issue #640, opened 2018-04-12
Allow distincts nfpm "name_template" per package format


bpineau:
The canonical package file names (word separator, architecture names) differs slightly depending on the package format and target distribution:
```
wget-1.18-3.fc25.x86_64.rpm        # CentOS, RHEL, Fedora
wget_1.18-5ubuntu1.3_amd64.deb     # Debian, Ubuntu
```

If I'm not mistaken, goreleaser allow for only one nfpm `name_template` (and arch `replacements`), shared by both deb and rpm.

It would be nice if goreleaser would provide some way to generate package names in the native/canonical format.

(originaly reported as a [nfpm issue](https://github.com/goreleaser/nfpm/issues/28))

caarlos0:
in your examples, there are `fc25` and `ubuntu` in the names. We don't have this information at the build time... what is the sane thing to do?

bpineau:
Well the `fc25` and `5ubuntu1.3` are different beast really, and the semantics differs.

For Debian/Ubuntu: the `debian_revision` part of the version number indicate changes in the packaging as opposed to changes in the upstream version ; mostly used when Debian want to publish a new package version for the same upstream version. It's optional, not mandatory:  https://www.debian.org/doc/debian-policy/#version .

So in the Debian case, and gorelaser/nfpm being maintained on the upstream side, this is only useful if you want to push a new release of the same software version after having tweaked the nfpm/goreleaser config. Maybe an optional `debian_revision` variable in goreleaser would be nice, but that's the icing on the cake ;).

On Fedora & co, this is provided by the `%{dist}` macro. This indicates the origin of the package, and is mandated by most guidelines. See https://fedoraproject.org/wiki/Packaging:DistTag for instance. Here I think the correct course of action would be just to replace in `nfpm/rpm/rpm.go` :
`Relase: 1`
by
`Release: 1%{?dist}`

The value of %{dist} is provided by /usr/lib/rpm/macros.d/macros.dist from rpmbuild, can be overided by the user (with ~/.rpmmacros), and by using the '1{?dist}' format we'll have '1' if dist is not defined anyway. Like for Debian, the "1" part here indicate the version of the package itself (by op. to the upstream software version), and it may be useful to increase that, so a `rpm_release` variable in goreleaser could be nice. Or maybe just a common `extra_version` variable used by both packaging systems?

caarlos0:
I don't know when I'll have some free time to look into this, mind opening a PR?

Cheers

[...]

PR #649, opened 2018-04-21
feat: Support per-format NFPM overrides

bpineau:
Support per-format NFPM overrides (as recently implemented in NFPM),
plus a pair or Goreleaser specific overrides, for "Replacements"
and "NameTemplate" (to address #640).

This keeps backward compatibility with existing config files:
settings at the root of the nfpm config section will still work as
previously (which is also useful to define common/shared settings
that we don't want to override anyway).

See #640

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -123,23 +123,16 @@ type Release struct {
 
 // NFPM config
 type NFPM struct {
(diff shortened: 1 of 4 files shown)


caarlos0:
thanks!

PR #653, opened 2018-04-25
docs: starting to re-work the docs site

caarlos0:
it will then be more similar to godownloader site :)


caarlos0:
i'll put this in production, let's see how it goes :)

PR #522, opened 2018-01-21
feat: support multiple build systems

caarlos0:
support multiple build langs by having a "pluggable" system of builders.

for now, we can have them all inside goreleaser itself. Maybe later we can move them out (?) (or no).

What do you guys think?

~~PS: extremely wip, tests wont pass and misses docs.~~

refs #520 
closes #496 

## Types of changes

* [ ] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] I have read the **CONTRIBUTING** document.
* [ ] `make ci` passes on my machine.
* [ ] My change requires a change to the documentation.
* [ ] I have updated the documentation accordingly.
* [ ] I have added tests to cover my changes.

PR #654, opened 2018-04-25
feat: Add new language builder for PHP

andygrunwald:
Since #522 goreleaser supports multiple build systems.
This commit enables the basic version of a PHP builder.
The PHP builder triggers a composer install, the
defactor standard package manager in the PHP world.

refs #520, #496

## Requirements

To test this PR you need to have [Composer](https://getcomposer.org/download/) installed.

## How to test this?

I prepared a small repository that contains a _very_ simple PHP project.
See https://github.com/andygrunwald/goreleaser-test-repo.

Clone the repo, tag a new version and execute a standard goreleaser release

## A really basic version ...

This is a really basic version and I don't think that this should be merged this way.
It doesn't have tests at all.
Basic functionality is not configurable (e.g. a `composer.phar`)

This PR is more to get feedback from you @caarlos0 what do you think about the implementation.
Especially at the file packaging part at https://github.com/andygrunwald/goreleaser-test-repo/blob/master/.goreleaser.yml#L15
We don't package a complete folder here or all files. We enter a list of files + subdirs.
I kind of like this system and don't saw a need to add a "add a complete folder" here.
The disadvantage is that the log output will be huge, due to thousands of files added.

## But the hooks ...

To be honest, you can do this 100% with the before hooks.
Because in the before hooks we could trigger a `composer install`.
But then the build target would not be triggered at all.

## A couple of hacks ...

The `config.Build` property contains tons of golang specific parts.
For this PR we kept it this way and "highjacked" only the relevant parts.
Furthermore, in PHP we don't have targets, because it runs natively on all supported platforms where PHP is supported. For this, we set a dummy target on linux and amd64.

## Just feedback

As I wrote, lets first collect a little bit of feedback here and iterate over it.
I implemented this before finishing #520, because for the PHP part I really have a use case right now. And I am on vacation, so I had fun to do it (hacked from [Hotel room 404 with such a sick view from the terrace, no kidding](https://twitter.com/andygrunwald/status/987736663987314688).
The #520 part I will finish once I have the next hack night with @mre

### DIFF
--- a/internal/artifact/artifact.go
+++ b/internal/artifact/artifact.go
@@ -26,6 +26,8 @@ const (
 	Checksum
 	// Signature is a signature file
 	Signature
+	// Folder is a complete folder
+	Folder
 )
 
 // Artifact represents an artifact and its relevant info
--- a/internal/builders/php/build.go
+++ b/internal/builders/php/build.go
@@ -1,0 +1,111 @@
+package php
+
+import (
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/apex/log"
+	api "github.com/goreleaser/goreleaser/build"
+	"github.com/goreleaser/goreleaser/config"
+	"github.com/goreleaser/goreleaser/context"
+	"github.com/goreleaser/goreleaser/internal/artifact"
+	"github.com/pkg/errors"
+)
+
+// Default builder instance
+var Default = &Builder{}
+
+func init() {
+	api.Register("php", Default)
+}
+
+// Builder is php builder
+type Builder struct{}
+
+// WithDefaults sets the defaults for a php build and returns it
+func (*Builder) WithDefaults(build config.Build) config.Build {
+	// We re use the config.Build.Main field here
+	// An alternative would be to alter the config.Build structure
+	// and a a "Composer" property, but i think the `func main()`
+	// part reflects nearly the same as the composer.json
+	// But then in my head there is a conflict.
+	// How to deal with `Gopkg.toml` files (dep package manager)?
+	if build.Main == "" {
+		build.Main = "composer.json"
+	}
+
+	// We add a simply dummy value here
+	// to enther the target loop in pipeline/build/build.go
+	// to trigger the PHP builder at all
+	if len(build.Targets) == 0 {
+		build.Targets = []string{"linux_amd64"}
+	}
+
+	return build
+}
+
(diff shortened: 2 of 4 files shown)


caarlos0:
Hi Andy, Yes, I've seen it, sorry for not commenting anything there.. I'm changing jobs so things are kind of crazy...

I Don't know about the php builder... I don't know anything about php... maybe we should first land the plugin infrasctructure (to load them from third parties)? GoReleaser scope is pretty big already, adding more builders may be just too much to maintain (and I don't know nothing about php haha)...

About the plugin structure itself:[go native plugins now support linux and macos](https://golang.org/pkg/plugin/)... The terraform way is quite complex but feasible. Either way, we may decide to support plugins only on the same platforms Go supports... it's not ideal, but its good enough for most cases I believe... we can discuss that...

Thoughs?

andygrunwald:
> I'm changing jobs so things are kind of crazy...
Good luck! Those things have priority. No need to say sorry.

> maybe we should first land the plugin infrastructure (to load them from third parties)
I agree here. One question comes to my mind:
Even if we get the plugins working and extract external language builders into plugins, how to distribute those to the end users while maintaining the "single binary" benefit or the `brew install goreleaser` way?

If i get it right, every language builder will be a `*.so` file. E.g. `builder_php.so`.
Those can be then loaded with goreleaser.
Is it possible with homebrew to distribute those `.so` files and goreleaser checks a standard folder which plugins are installed?
Or would it be necessary to start goreleaser always like `goreleaser --plugin-dir=/foo/plugins/`?

caarlos0:
> Is it possible with homebrew to distribute those .so files and goreleaser checks a standard folder which plugins are installed?

What I had in mind is:

- add a `golib` builder to goreleaser
- add a `plugins` section to config, and make goreleaser download from that to some folder (maybe even borrow some parts of dep to define versions and etc)

so, as an user, I would:

```yaml
plugins:
  'user/goreleaser-php-builder': '^1.2.3'
build:
  language: php
```

and goreleaser would take care of the rest...

what do you think?

andygrunwald:
Sounds like a doable plan!

caarlos0:
I created a project to try to organize all that https://github.com/goreleaser/goreleaser/projects/1

caarlos0:
will close this for now.

I'll try to work on a plugin system in the next few days.

Thanks!

PR #674, opened 2018-05-15
test: Use multiple calls to assert.NotEmpty()

elyscape:
`assert.NotEmpty()` appears to have been used here as if it would check multiple arguments. [In actuality][docs], it checks the second argument and, if it determines that it is empty, fails the test with an error message containing all subsequent arguments. This changes the test to call `assert.NotEmpty()` multiple times.

Feel free to close this PR without merging if I am mistaken as to the intent of the code in question.

[docs]: https://godoc.org/github.com/stretchr/testify/assert#NotEmpty


caarlos0:
good catch! 

Thanks!

PR #675, opened 2018-05-15
feat: Convert flag fields into lists

### LINKED ISSUE #668: List formats for flags, ldflags, asmflags, and gcflags
[As of Go 1.10][go1.10]:
>The `go build` `-asmflags`, `-gcflags`, `-gccgoflags`, and `-ldflags` options now apply by default only to the packages listed directly on the command line. For example, `go build -gcflags=-m mypkg` passes the compiler the `-m` flag when building `mypkg` but not its dependencies. The new, more general form `-asmflags=pattern=flags` (and similarly for the others) applies the flags only to the packages matching the pattern.

In other words, you can pass different flag sets to the compiler based on the package you're passing. Unfortunately, this has to be done with multiple uses of `-gcflags` et al. For example, given the following code:
```shellsession
elyscape:flagdemo eliyoung$ pwd
/Users/eliyoung/go/src/github.com/elyscape/flagdemo
elyscape:flagdemo eliyoung$ cat main.go
package main

import (
	"fmt"

	"github.com/elyscape/flagdemo/sub"
)

var (
	one = ""
)

func main() {
	fmt.Println("one:", one)
	sub.Print()
}
elyscape:flagdemo eliyoung$ cat sub/sub.go
package sub

import (
	"fmt"
)

var (
	two = ""
)

func Print() {
	fmt.Println("two:", two)
}
```
Here are the source file paths included in the resulting binary:
```shellsession
elyscape:flagdemo eliyoung$ go build -a .
elyscape:flagdemo eliyoung$ strings flagdemo | grep 'elyscape.*go$'
/Users/eliyoung/go/src/github.com/elyscape/flagdemo/main.go
/Users/eliyoung/go/src/github.com/elyscape/flagdemo/sub/sub.go
/Users/eliyoung/go/src/github.com/elyscape/flagdemo/sub/sub.go
/Users/eliyoung/go/src/github.com/elyscape/flagdemo/main.go
```
Suppose that, for some reason, I want to set different trimpaths for different subpackages. Trying to put both in a single `-gcflags` value doesn't work:
```shellsession
elyscape:flagdemo eliyoung$ go build -a -gcflags '.=-trimpath=/Users/eliyoung/go/src ./sub=-trimpath=/Users/eliyoung/go/src/github.com' .
# github.com/elyscape/flagdemo
./sub=-trimpath=/Users/eliyoung/go/src/github.com:0:0: open ./sub=-trimpath=/Users/eliyoung/go/src/github.com: no such file or directory
```
By contrast, using `-gcflags` multiple times does work:
```shellsession
elyscape:flagdemo eliyoung$ go build -a -gcflags '.=-trimpath=/Users/eliyoung/go/src' -gcflags './sub=-trimpath=/Users/eliyoung/go/src/github.com' .
elyscape:flagdemo eliyoung$ strings flagdemo | grep 'elyscape.*go$'
github.com/elyscape/flagdemo/main.go
elyscape/flagdemo/sub/sub.go
elyscape/flagdemo/sub/sub.go
github.com/elyscape/flagdemo/main.go
```
Additionally, some of the flags one might want to specify in the `flags` field, such as `-tags` (which is given in the example), take space-separated arguments, and there's currently no way to represent that.

It would be nice if the `ldflags` et al. took arrays, each of which was then processed as a template and then passed individually to the compiler. To demonstrate with a slightly modified version of example above, this:
```yaml
flags:
- -a
- tags
- tagone tagtwo
gcflags:
- .=-trimpath={{.Env.GOPATH}}
- ./sub=-trimpath=/Users/eliyoung/go/src/github.com
```
Would be turned into something like this (with a bunch of extraneous quotes and newlines for clarity):
```shell
go build \
	'-a' \
	'-tags' \
	'tagone tagtwo' \
	'-gcflags=.=-trimpath=/Users/eliyoung/go/src' \
	'-gcflags=./sub=-trimpath=/Users/eliyoung/go/src/github.com' \
	.
```
Of course, making a field in the YAML be either a string or an array is complicated in Go. Probably the best way to do this is to use the approach provided [in this comment][StringArray]; that is, mark those fields as a new type called `StringArray`, which is basically just an alias for `[]string` but which has a custom YAML unmarshalling method that converts bare strings into one-element arrays.

For `ldflags`, `gcflags`, and `asmflags`, this would be a drop-in fix with no further ramifications, but `flags` is complicated by the fact that it currently runs the value through `strings.Fields()`, and we'd probably want to preserve that behavior to avoid breaking anything. The best way to deal with that would probably be to have it have its own custom type (`FlagsArray`?) and YAML unmarshaller that, when provided with a bare string, runs it through `strings.Fields()` and returns the result of that, instead of just wrapping the bare string in an array.

Instead of making the various fields automatically upconvert into arrays, we could instead deprecate them and add new fields like `ldflags_list` which are natively arrays. Alternatively, we could do that only for `flags`, which is the only problematic field.

[go1.10]: https://golang.org/doc/go1.10#build
[StringArray]: https://github.com/go-yaml/yaml/issues/100#issuecomment-324964723

elyscape:
This allows users to specify different sets of compiler, assembler, and/or linker flags for different packages. It also makes it possible to specify generic flags that contain spaces. It does this while maintaining compatibility with the old format of the fields in question by up-converting bare strings into single-element lists.

Resolves #668

### DIFF
--- a/internal/builders/golang/build.go
+++ b/internal/builders/golang/build.go
@@ -44,8 +44,8 @@ func (*Builder) WithDefaults(build config.Build) config.Build {
 	if len(build.Goarm) == 0 {
 		build.Goarm = []string{"6"}
 	}
-	if build.Ldflags == "" {
-		build.Ldflags = "-s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}}"
+	if len(build.Ldflags) == 0 {
+		build.Ldflags = []string{"-s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}}"}
 	}
 	if len(build.Targets) == 0 {
 		build.Targets = matrix(build)
@@ -59,31 +59,29 @@ func (*Builder) Build(ctx *context.Context, build config.Build, options api.Opti
 		return err
 	}
 	cmd := []string{"go", "build"}
-	if build.Flags != "" {
-		cmd = append(cmd, strings.Fields(build.Flags)...)
-	}
 
-	if build.Asmflags != "" {
-		flags, err := processField(ctx, build.Asmflags, "asmflags")
-		if err != nil {
-			return err
-		}
-		cmd = append(cmd, "-asmflags="+flags)
-	}
+	cmd = append(cmd, build.Flags...)
 
-	if build.Gcflags != "" {
-		flags, err := processField(ctx, build.Gcflags, "gcflags")
-		if err != nil {
-			return err
-		}
-		cmd = append(cmd, "-gcflags="+flags)
-	}
-
-	flags, err := processField(ctx, build.Ldflags, "ldflags")
+	asmflags, err := processFlags(ctx, build.Asmflags, "asmflags", "-asmflags=")
 	if err != nil {
 		return err
 	}
-	cmd = append(cmd, "-ldflags="+flags, "-o", options.Path, build.Main)
+	cmd = append(cmd, asmflags...)
+
+	gcflags, err := processFlags(ctx, build.Gcflags, "gcflags", "-gcflags=")
+	if err != nil {
+		return err
+	}
+	cmd = append(cmd, gcflags...)
+
+	ldflags, err := processFlags(ctx, build.Ldflags, "ldflags", "-ldflags=")
+	if err != nil {
+		return err
+	}
+	cmd = append(cmd, ldflags...)
+
+	cmd = append(cmd, "-o", options.Path, build.Main)
+
 	target, err := newBuildTarget(options.Target)
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


internal/builders/golang/build.go
```diff
@@ -44,8 +44,8 @@ func (*Builder) WithDefaults(build config.Build) config.Build {
 	if len(build.Goarm) == 0 {
 		build.Goarm = []string{"6"}
 	}
-	if build.Ldflags == "" {
-		build.Ldflags = "-s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}}"
+	if len(build.Ldflags) == 0 {
+		build.Ldflags = append(build.Ldflags, "-s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}}")
```
caarlos0: it is not needed to call `append` here right? just `build.Ldflags = []string{"-s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}}"}`.

I think it is more clear, as we are not really appending to an array with items...

caarlos0:
just a single minor up there...

if you could also improve the test coverage a bit (the config pkg coverage lowered a bit) it would be awesome!

also, thank you very much for the pull request and all the effort 🍻

caarlos0:
> Comment addressed and tests enhanced.

awesome @elyscape! Thank you very much!

PR #678, opened 2018-05-17
fix: Remove broken documentation link

srishanbhattarai:
If applied, this commit will remove a broken link to the docs (https://github.com/goreleaser/goreleaser/blob/master/docs) on the README.

The `docs` folder seems to have moved and hence the link is broken. [This](https://github.com/goreleaser/goreleaser.github.io) is the best I could find and it isn't ideal to be in the README.


caarlos0:
thanks!

PR #684, opened 2018-05-28
fix: snapcraft: confinement: strict

caarlos0:
I think we can use the strict confinement... 

@elopio any inputs?


elopio:
Hey @caarlos0. I don't think so, because you are not shipping the go compiler together with the goreleaser snap, AFAIR.
But it would be nice to make this a strict snap, if possible. I'm no longer working on this, so my friends @Wimpress and @popey will be able to help you moving forward.

PR #688, opened 2018-06-01
feat: support simple http put artifacts uploading

lalloni:
Supports doing simple http put requests for uploading artifacts.
Heavily based con Artifactory pipeline.
In archive mode uploads all archives and the checksums file.
In binary mode uploads just the binaries.
If applied this PR will add a pipeline named httpupload (modeled closely after the artifactory pipeline) which does simple http puts for pulishing release artifacts.
This change is made to support publishing artifacts to private instances of [Nexus Repository Manager](https://www.sonatype.com/nexus-repository-sonatype) but general enough to be capable of publishing to any http server allowing PUT requests.


lalloni:
Should I solve the code duplication issue raised by the CI pipeline?
In that case, where would be appropriate to put code shared by pipelines?
At first I thought of using a new package inside github.com/goreleaser/goreleaser/internal but I wasn't sure if that would be the right place for that.

caarlos0:
Hi @lalloni, looks very good!

Can we remove the artifactory pipeline? Maybe when the config says artifactory we use the httpupload pipe underneath or something... do you think we can do that?

lalloni:
@caarlos0 the big difference is that artifactory responds with a JSON document with checksums which are usable to verify the upload was correct... so for not losing that feature I would keep the artifactory pipeline besides the simpler httpupload.

To reuse common code the approach you propose could work... artifactory pipeline could use httpupload functionality providing the latter somehow returns the http response to the caller. I should give it a try if sounds good for you.

Is it ok around here creating a dependency from one pipeline to another at the code level (i.e. import from the other pipeline package)?

caarlos0:
> To reuse common code the approach you propose could work... artifactory pipeline could use httpupload functionality providing the latter somehow returns the http response to the caller. I should give it a try if sounds good for you.

I'm OK with that :)

caarlos0:
> Is it ok around here creating a dependency from one pipeline to another at the code level (i.e. import from the other pipeline package)?

We don't have a strict rule about that, but so far as far as I remember pipelines dont depend on other pipelines... anyway, if we would like to keep it that way, we can just move the common code to another package... (`internal/httpupload` for example).

lalloni:
@caarlos0 please check the new modifications

pipeline/artifactory/artifactory.go
```diff
@@ -2,24 +2,13 @@
...
-	"strings"
 
-	"github.com/apex/log"
-	"github.com/pkg/errors"
-	"golang.org/x/sync/errgroup"
-
-	"github.com/goreleaser/goreleaser/config"
 	"github.com/goreleaser/goreleaser/context"
-	"github.com/goreleaser/goreleaser/internal/artifact"
+	internal "github.com/goreleaser/goreleaser/internal/httpupload"
```
caarlos0: why not call it `httpupload`? `internal` doesn't seem to be a descriptive name..

pipeline/httpupload/httpupload.go
```diff
@@ -0,0 +1,51 @@
+package httpupload
```
caarlos0: missing the package godoc...

caarlos0:
@lalloni a few minors, but overall LGTM!

@andygrunwald, can you give a look as well?

Once this is all good, we should also update the docs :D

pipeline/artifactory/artifactory.go
lalloni: Yep...
It all started when needed to import internal/httpupload from pipeline/httpupload, there the alias was required, calling it internal seemed like a good choice... from there it just spilled over pipeline/artifactory because it seemed more uniform/readable to call it the same from the two client packages.
I must say I would really like to toss the "httpupload" name completely but couldn't come up with something better.
I also believe it would be best to call the internal package something different to the pipeline packages to indicate the difference in generality, but not just "http" because collides with "net/http"...

Maybe something like this:



Although I didn't put too much thought on it, I never found good names for the internal and the  pipeline http packages.

Ideas?

caarlos0:
I spoke with @andygrunwald and he is traveling, will be back on sunday, so let's wait a few more days, since he wrote the artifactory pipeline and is the only user I'm aware of.

If anyone else uses the artifactory pipe and can give this a try, I'll be glad as well :D

caarlos0:
the build failure is on me, let me fix it on master...

lalloni:
I was just adding some tests for `internal/http` and thinking about adding configuration for customizing which artifacts to upload.
But as it is now, it's already usable for our use case.

pipeline/artifactory/artifactory.go
caarlos0: not really :/

caarlos0:
BTW: I believe most of the artifactory test suite is now irrelevant as things are tested on the internal http upload package, right? Maybe we could also clean that up (can be later of course)?

lalloni:
Not really... as it is now, tests exist for both pipelines and no test exist for `internal/http`, hence my previous comment.
I didn't want to modify/remove artifactory pipeline tests to ensure the behavior was kept unmodified... that's what tests are for, right? :smile: 
Note that although no test exist for `internal/http` it transitively gets coverage because it gets called from both pipelines tests.

caarlos0:
oh, I've misread the code... shame on me haha

yeah, makes sense to me, let's keep it that way for now.. the later I meant is like another PR on another release.. but I wasn't clear, sorry about that as well 😄

lalloni:
@caarlos0 I believe this is ready to go now, please check it out and apply if you agree.

we're using goreleaser internally so a release including this PR would significatively ease development environment setup.

caarlos0:
let's do it!

andygrunwald:
Sorry to be so late to the game.
Looks good from my side. We didn't upgrade yet, but I am sure this will work out.
Thanks for your effort @lalloni!

Issue #686, opened 2018-05-29
Wrong sha256 used in Brew Forumula generation


TylerBrock:
**Describe the bug**

goreleaser is using the `darwin_amd64.tar.gz` sha256 instead of the one for the source.

**To Reproduce**

```yaml
builds:
  - binary: saw
    goos:
      - darwin
      - linux
      - freebsd
    goarch:
      - amd64
      - arm
      - arm64
    goarm:
      - 6
      - 7

archive:
  format: tar.gz
  files:
    - LICENSE

brew:
  name: saw

  github:
    owner: TylerBrock
    name: homebrew-saw

  download_strategy: GitHubPrivateRepositoryReleaseDownloadStrategy

  commit_author:
    name: TylerBrock
    email: <redacted>

  folder: Formula
  homepage: "https://github.com/TylerBrock/saw"
  description: "Fast, multipurpose tool for AWS CloudWatch Logs"

  build_dependencies:
    - go
    - dep

git:
  short_hash: true
```

I'm generating a formula for my project called Saw. It works fine then I add the tap and install the formula but then it says it expects the sha256 to be the darwin one instead of the one for the source tarball.

```sh
$ brew tap TylerBrock/saw
...
$ brew install saw
==> Installing saw from tylerbrock/saw
==> Downloading https://github.com/TylerBrock/saw/archive/v0.1.2.tar.gz
==> Downloading from https://codeload.github.com/TylerBrock/saw/tar.gz/v0.1.2
######################################################################## 100.0%
Error: SHA256 mismatch
Expected: f7398ea54c80befe994f898010c21ef9a62c880c2f25a9ac2bf84765a09b5313
Actual: bfdf5988554276327544eb8fea3943792ac63bfbe0d558bc1ef65d105b6764f3
Archive: /Users/tbrock/Library/Caches/Homebrew/saw-0.1.2.tar.gz
To retry an incomplete download, remove the file above.
```

This is what the generated formula looks like:

```rb
class Saw < Formula
  desc "Fast, multipurpose tool for AWS CloudWatch Logs"
  homepage "https://github.com/TylerBrock/saw"
  url "https://github.com/TylerBrock/saw/archive/v0.1.2.tar.gz"
  head "https://github.com/TylerBrock/saw.git"
  version "0.1.2"
  sha256 "f7398ea54c80befe994f898010c21ef9a62c880c2f25a9ac2bf84765a09b5313"
  depends_on "go" => :build
  depends_on "dep" => :build

[...]

PR #687, opened 2018-05-29
WIP: fix: use the right sha256 sum on archive releases

caarlos0:
closes #686

I particularly don't like the way I did this... will think mroe about it and give another try later.

PR #700, opened 2018-06-20
fix: breaking: remove the build dependencies from the brew pipe

caarlos0:
Removes the `build_dependencies` option from the Homebrew pipe.

The idea of that pipe always was to generate a tap, not a core-ready formula.

Anyway, we gave it a try, and it doesn't really work. To make it work, we would need to know the SHA256 of the tag, which will make tests difficult. 

I also noticed that people were using this feature by mistake.

Out in the wild, it seems that core formulas have a bot to update them (see  https://github.com/Homebrew/homebrew-core/pulls?q=is%3Apr+sort%3Aupdated-desc+goreleaser+is%3Aclosed ), so maybe this doesn't make sense indeed.

Anyone agains removing this?

Closes: #686 and #687


caarlos0:
OK, 2 days, no one asked to keep this, will remove and see how it goes :D

PR #706, opened 2018-06-29
chore: bump to go 1.10.3, also test on tip

caarlos0:


caarlos0:
don't really need to test on tip for now...

Issue #707, opened 2018-07-02
Expand 'goos' and 'goarch' combinations in 'dockers'.


1138-4EB:
I am trying to reproduce the compact syntax of the 'build' field in order to generate multiple docker images, each one corresponding to a built binary, and all of them based on the same Dockerfile and sharing the same image name.

Given then following configuration file, I'd like multiple docker images to be created, each of them with two tags.

``` yaml
project_name: filebrowser

build:
  env:
    - CGO_ENABLED=0 
  main: cmd/filebrowser/main.go
  binary: filebrowser
  goos:
    - darwin
    - linux
    - windows
    - freebsd
    - netbsd
    - openbsd
    - dragonfly
    - solaris
  goarch:
    - amd64
    - 386
    - arm
    - arm64
  goarm:
    - 5
    - 6
    - 7
  ignore:
    - goos: darwin
      goarch: 386
    - goos: openbsd
      goarch: arm
    - goos: freebsd
      goarch: arm
    - goos: netbsd
      goarch: arm
    - goos: solaris
      goarch: arm

archive:
  name_template: "{{.Os}}-{{.Arch}}{{if .Arm}}v{{.Arm}}{{end}}-{{ .ProjectName }}"
  format: tar.gz
  format_overrides:
    - goos: windows
      format: zip

release:
  disable: true

dockers:
  goos: linux
  goarch:
    - amd64
    - arm
    - arm64
  goarm:
    - 7
    - 6
  image: "{{ .ProjectName }}/{{ .ProjectName }}"
  skip_push: true
  tag_templates:
    - "{{ .Arch }}{{if .Arm}}v{{.Arm}}{{end}}-{{ .Tag }}"
    - "{{ .Arch }}{{if .Arm}}v{{.Arm}}{{end}}-latest"
  extra_files:
    - Docker.json
```

The problems I find are:
- Each of `goarch` and `goarm` in `dockers` must have a single value.
- `{{ .ProjectName }}` cannot be used to set the dockers `image`.
- Neither `{{ .Arch }}` nor `{{ .Arm }}` can be used to set `tag_templates`.

I've considered setting each image explicitly:

``` yaml
dockers:
  -
    goos: linux
    goarch: amd64
    goarm: ''
    image: filebrowser/filebrowser
    skip_push: true
    tag_templates:
      - "amd64-{{ .Tag }}"
      - "amd64-latest"
    extra_files:
      - Docker.json
  -
    goos: linux
    goarch: arm64
[...]

PR #709, opened 2018-07-03
fix: allow more fields into docker tag_template

caarlos0:
just use the internal nametemplate package.

Also supporting Minor, Major and Patch fields in other
namtemplates

refs #707


caarlos0:
closed in favor of #716

PR #723, opened 2018-07-11
Proposal: Support monorepos

tsenart:
[Vegeta](https://github.com/tsenart/vegeta) hosts both a Go library and a CLI in a single repo. In that sense, it is a mono-repo, with multiple projects that ought to be versioned separately.

This version split only happened recently with [this PR](https://github.com/tsenart/vegeta/pull/309). When releasing, I found goreleaser didn't have a way of selecting which tags to use for a release.

I had to unblock myself and release Vegeta so I hacked my way in go-releaser. This proposal is not so much to bring in this code which is not robust enough in all likelihood. It's about discussing the idea of supporting monorepos which contain multiple independent components that ought to be versioned separately.

In my case, as convention, I decided to prefix all tags of a given component with the component's identifier (i.e. `cli/v8.1.0`, `lib/v9.0.0`).

### DIFF
--- a/internal/tmpl/tmpl.go
+++ b/internal/tmpl/tmpl.go
@@ -82,7 +82,7 @@ func (t *Template) Apply(s string) (string, error) {
 		return "", err
 	}
 
-	sv, err := semver.NewVersion(t.fields[tag].(string))
+	sv, err := semver.NewVersion(t.fields[version].(string))
 	if err != nil {
 		return "", errors.Wrap(err, "tmpl")
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -266,7 +266,8 @@ type EnvFiles struct {
 
 // Git config
 type Git struct {
-	ShortHash bool `yaml:"short_hash,omitempty"`
+	TagsPrefix string `yaml:"tags_prefix,omitempty"`
+	ShortHash  bool   `yaml:"short_hash,omitempty"`
 }
 
 // Before config
--- a/pipeline/git/git.go
+++ b/pipeline/git/git.go
@@ -63,7 +63,7 @@ func getGitInfo(ctx *context.Context) (context.GitInfo, error) {
 	if err != nil {
 		return context.GitInfo{}, errors.Wrap(err, "couldn't get current commit")
 	}
-	tag, err := getTag()
+	tag, err := getTag(ctx.Config.Git.TagsPrefix)
 	if err != nil {
 		return context.GitInfo{
 			Commit:     commit,
@@ -86,7 +86,10 @@ func setVersion(ctx *context.Context) error {
 		return nil
 	}
 	// removes usual `v` prefix
-	ctx.Version = strings.TrimPrefix(ctx.Git.CurrentTag, "v")
+	ctx.Version = strings.TrimPrefix(
+		strings.TrimPrefix(ctx.Git.CurrentTag, ctx.Config.Git.TagsPrefix),
+		"v",
+	)
 	return nil
 }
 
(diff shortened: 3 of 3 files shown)


GitCop:

 * Commit: 7993da9df46fb3916cbd44cdf4c4bea15dfbb438

---

internal/tmpl/tmpl.go
```diff
@@ -82,7 +82,7 @@ func (t *Template) Apply(s string) (string, error) {
 		return "", err
 	}
 
-	sv, err := semver.NewVersion(t.fields[tag].(string))
+	sv, err := semver.NewVersion(t.fields[version].(string))
```
caarlos0: this is a good change anyways, can you open a separate PR with just this? Thanks!

caarlos0:
To be honest, that's the first time I've seen two projects in the same git project with different tags... as far as I remember at least... 

Not sure if that's a common use case.

Either way, we need to fix the packaging step, which if you have multiple builds, always archive all of them in the same archive - this is to allow people to release two projects, with the same version, on separated artifacts...

caarlos0:
BTW: Vegeta is awesome, thanks for the great work on the project!!

tsenart:
> To be honest, that's the first time I've seen two projects in the same git project with different tags... as far as I remember at least... Not sure if that's a common use case.

That will always be the case with monorepos if people need to release its components independently. Good links on monorepos: https://gomonorepo.org/

jacekjagiello:
@caarlos0 I'll also think that support for monorepos is a good thing.

Monorepo in many cases, allows developers(and even whole teams) to work together faster. It simplifies things like dependency management, build and release configuration, CI pipelines or deployment process. Of course, structuring project in monorepo has it's cons too, but it's up to team to decide whenever or not they want to go with monorepo.

Nevertheless, more and more projects decide to go with monorepo structure. In Go I think monorepos are quite common because of the lightweight package system. Beside awesome Vegeta, I can highlight projects like:
- [rook](https://github.com/rook/rook)
- [perkeep](https://github.com/perkeep/perkeep)
- [prometheus](https://github.com/prometheus/prometheus)

Note the _cmd_ directory that is used to separate source code to build binaries.

Other projects for management Go source code, already adapted to monorepo structured projects. For example in [realize](https://github.com/oxequa/realize) you can define in single configuration file multiple binaries to compile when source code changes.

I'm happy to help by contributing to this feature :)

caarlos0:
Yeah, I didn't disagree about the usage of monorepos (which goreleaser already support), I just had never seen before things being tagged separately inside a monorepo... so, don't know how popular that is... 

Either way, if we decide to do this thing with tags as proposed, I'll need some help, probably haha :D

tsenart:
> Either way, if we decide to do this thing with tags as proposed, I'll need some help, probably haha :D

What do you need? :-)

caarlos0:
I think that if are gonna support this tagging thing, we need at least it.

Meanwhile, I'll create a twitter poll to see if I can find out what people usually do. :)

dpastoor:
Perhaps an issue or a wiki article where known monorepos are to use as a reference and a place to look at how they are currently releasing version(s)?

caarlos0:
Link to the research, FWIW: https://twitter.com/caarlos0/status/1031516398315552769

sidh:
Any status update on this? Any help needed with it?

I'm currently using goreleaser patched with proposed PR and it seems to work just fine.

caarlos0:
only that we will move on with the tag prefixes thing, but I have other stuff on line before working on this.

sidh:
Some activity so it won't get closed.

PR #731, opened 2018-07-21
feat: add use_env_vars for s3

mgrachev:
Hi there!

I add option `use_env_vars` to s3 configuration. This option tells s3 to take credentials from environment variables, such as: `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` instead of using `~/.aws` config file. I consider that this will be handy while using goreleaser on a CI-server.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -276,11 +276,12 @@ type Before struct {
 
 // S3 contains s3 config
 type S3 struct {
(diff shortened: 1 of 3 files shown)


caarlos0:
don't think we need the config, just the chain credential provider...

PR #732, opened 2018-07-21
feat: add acl option for s3

mgrachev:
Hi there!

By default, goreleaser uploads archives to s3 as private files. So I add option `acl` to s3 configuration which let upload an archive as a public file.


caarlos0:
thanks!

PR #735, opened 2018-07-24
feat: add s3 as an option to use in a brew formula

mgrachev:
Hi there!

`goreleaser` can publish archives to s3, but can't use s3-links in `brew` formulas. This solution fixes that.

### DIFF
--- a/internal/tmpl/tmpl.go
+++ b/internal/tmpl/tmpl.go
@@ -33,10 +33,11 @@ const (
 	timestamp   = "Timestamp"
 
 	// artifact-only keys
-	os     = "Os"
-	arch   = "Arch"
-	arm    = "Arm"
-	binary = "Binary"
+	os           = "Os"
+	arch         = "Arch"
+	arm          = "Arm"
+	binary       = "Binary"
+	artifactName = "ArtifactName"
 )
 
 // New Template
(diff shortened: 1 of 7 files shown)


caarlos0:
Maybe it would be better to add a `url_template` field?

Then, it would work for other things as well (e.g.: http upload pipe).

What do you think?

mgrachev:
@caarlos0 I've done.

caarlos0:
Hi, sorry for the delay... I don't think we need to add those fields into the tmpl package, the user could just write the url template, eg. `http://github.mycompany.com/foo/bar/releases/{{.Tag}}/etc` (and we have the default github url as default so it doesn't break anything). S3 users could also write the URL the way they need, and people using other upload strategies could do the same...

tmpl is supposed to be generic for a lot of places...

internal/tmpl/tmpl.go
```diff
@@ -64,6 +65,7 @@ func (t *Template) WithArtifact(a artifact.Artifact, replacements map[string]str
 	t.fields[arch] = replace(replacements, a.Goarch)
 	t.fields[arm] = replace(replacements, a.Goarm)
 	t.fields[binary] = bin
+	t.fields[archive] = a.Name
```
caarlos0: I think the key should probably be `Artifact` or `ArtifactName`, as it can be other kind of artifacts, not only an archive...

caarlos0:
Just a small comment, and then we're good.

The build failure you can fix by adding a version to the context being used on the test failing - or you can leave it for me and I fix it later :)

Thanks for the PR 🚀

mgrachev:
@caarlos0 Done. Ready for review.

caarlos0:
Thanks @mgrachev !

PR #749, opened 2018-08-08
feat: Add default command for snapcraft

mgrachev:
Hi there!

I have a snap, the name of which differs from the name of the binary build. So, I propose to use the snap name as the app name instead of the name of the binary build.

### DIFF
--- a/internal/pipeline/snapcraft/snapcraft.go
+++ b/internal/pipeline/snapcraft/snapcraft.go
@@ -160,6 +160,13 @@ func create(ctx *context.Context, arch string, binaries []artifact.Artifact) err
 			return err
 		}
 	}
+
+	if _, ok := metadata.Apps[metadata.Name]; !ok {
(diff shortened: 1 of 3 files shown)


caarlos0:
I think this would break when you have multiple build/binaries (I should add a test for that) 

/remind me add a test for this in 8 hours

mgrachev:
I've already tested this solution. It works if we have a build or several builds.

If we have a build then `snap.yml` looks like this:
```yml
name: test_snap
apps:
  build_1:
    command: build_1
  test_snap:
    command: build_1
```

I we have several build then `snap.yml` looks like this:
```yml
name: test_snap
apps:
  build_1:
    command: build_1
  build_2:
    command: build_2
  test_snap:
    command: build_1
```

caarlos0:
cool, can you also add a test for that?

caarlos0:
Thanks, sorry for the delay!

Issue #754, opened 2018-08-10
Support profile and assume an IAM role with MFA prompting for token code on stdin when uploading to S3


kgrodzicki:
**Is your feature request related to a problem? Please describe.**
I have a need to use profile with assume role and MFA when uploading artifacts to S3.
So upload will work with configuration like below:
~/.aws/config
[profile cloudformation@flowlab-development]
role_arn=arn:aws:iam::11111111111111:role/service-role/cloudformation
source_profile=user
mfa_serial=arn:aws:iam::22222222222:mfa/user

~/.aws/credentials
[user]
aws_access_key_id = IWA0WZQ1QJ2I8I1ALW64
aws_secret_access_key = zcK4QQegvYwVGJaBm2E6k20WRLIQZqHOKOPP2npT

**Describe the solution you'd like**
Have already implemented fallback to create session with enabled AssumeRoleTokenProvider, SharedConfigState and profile in case when there is no credentials from providers chain. 
Can create PR if thats sounds good

caarlos0:
Hi @kgrodzicki , sounds good to me ;)

kgrodzicki:
@caarlos0 it is a bit tricky to write proper tests for this code. What about extracting all code responsible for creating a s3 client and testing it outside of `upload` function ?

caarlos0:
@kgrodzicki I agree, I think there is already a TODO regarding some of its behavior... if you could work on that it would be awesome!

caarlos0:
oh, yes, thanks!

PR #767, opened 2018-08-20
feat: Add support for profile and assume an IAM role with MFA prompting

kgrodzicki:
for token code on stdin when uploading to s3

In case when there are no credentials from provider chain session for S3
client is created with support for profile and assume an IAM role with
MFA prompting for token code on stdin

#754


caarlos0:
Awesome @kgrodzicki !

Could you also document that on `s3.md`?

caarlos0:
test error seems to not be related... I'll take a look later :)

PR #777, opened 2018-08-25
feat: Allow to get http username from env

lalloni:
If applied, this commit will allow to get http username (used from "artifactory" and "put" pipelines) from env var if not configured in goreleaser configuration.

This change is useful to avoid committing to vcs the username in goreleaser configuration file when it needs to change based on the actual user which will run goreleaser to upload files.


caarlos0:
thanks!

Issue #432, opened 2017-12-04
Support Google cloud builder CI


gedw99:
### Prerequisites

* [ ] Are you running the latest version?
* [ ] Did you perform a search to see if this problem was already reported?

For more information, see the `CONTRIBUTING` guide.

### Description

[Description of the bug or feature]

### Environment

Please paste the output of:

- `go version`
- `uname -a`
- `goreleaser --debug`

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behavior:** [What you expected to happen]

**Actual behavior:** [What actually happened]

https://wildfish.com/blog/2017/03/29/google-cloud-container-builder-introduction/

I played with it and it's a concise way to do your CI. It's also alot faster than others.

caarlos0:
in theory you should be able to call goreleaser and it should work... 

I don't know what is the issue here exactly...

gedw99:
None really.
Just more thinking me and others need examples of how to integrate into
different CI and CAN systems.

I can send you the script when done.

You also might be interested in the bitrise golang code for integration
into goreleaser.
It builds mobiles. If you want I can send you that script too.

caarlos0:
@gedw99 understood.

well, feel free to open a PR adding it to the docs if you want to :) other user would probably appreciate :P 

Since there is nothing I can do here, I'll close the issue. Feel free to reopen if needed.

PR #780, opened 2018-08-27
Add instructions for using with Google CloudBuild.

ncabatoff:
Add instructions for running goreleaser in Google CloudBuild.

See also issue #432.

### DIFF
--- a/www/content/ci.md
+++ b/www/content/ci.md
@@ -124,3 +124,61 @@ pipeline:
     when:
       event: tag
 ```
+
+## Google CloudBuild
+
+CloudBuild works off a different clone than your github repo: it seems that
+your changes are pulled to a repo like 
+source.developers.google.com/p/YourProjectId/r/github-YourGithubUser-YourGithubRepo, and that's what
+you're building off.  
+
+This repo does not have your tags, which prevents goreleaser from running at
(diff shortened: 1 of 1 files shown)


caarlos0:
thanks @ncabatoff !

question: does the changelog work? If you don't have the tags, and as far as I understand, you re-create the current tag on that repo, you won't have the previous one, so the changelog will be all commits... is that right?

ncabatoff:
Ok, so I tried removing my explicit 'git tag $TAG_NAME' from the repo where I developed these instructions, and it failed as it had in the past with "error=git doesn't contain any tags. Either add a tag or use --snapshot".  I tried add a git show-ref to the command and it only had master.  So even though the source repo mirror that's used to do the build *does* contain tags, the checkout they build from *does not*.

Having reverted my change, I can now answer your original question: it seems that goreleaser in this scenario (single tag, produced by explicit tagging with $TAG_NAME as in my PR example), creates a changelog with a single entry containing the commit message from the last commit.

GitCop:

 * Commit: cdbb01eb57dace1f105301dcb81622e0f8f6eafe

 * Commit: 3563d61d7b1c6ec5e73b1edf85789229e73b43d9

---

caarlos0:
merged, thanks!

PR #782, opened 2018-08-30
wip: feat: better snapshots

### LINKED ISSUE #667: GitHub pre-releases on alpha, beta, rc versions
When the version tag is not stable; e.g. `1.0.0-alpha`, ` 1.0.0-beta` or `1.0.0-rc` (see [https://semver.org/#spec-item-11](https://semver.org/#spec-item-11)) I would like the github release to be a `pre-release`.

In the github API set the `prerelease: true` option (see [https://developer.github.com/v3/repos/releases/#create-a-release](https://developer.github.com/v3/repos/releases/#create-a-release))

 * Always use prereleases for `beta`, `alpha`, `rc`
 * Let the user choose whether to use prereleases or not in the configuration file

**Additional context**
Example:
image

### LINKED ISSUE #727: Add a publish snapshots configuration to the docker build
We use goreleaser to publish nightly snapshots. We have recently added docker images that are created using the nightlies. The s3 section will copy the snapshot to s3, but the docker push is skipped because snapshot disables publishing.

A configuration option such as `publish_snapshots` that would enable pushing a specific image when snapshot mode is enabled.

Our workaround is just to call `docker push` after calling `goreleaser`, but it's not ideal because then we have to duplicate the image names.

### LINKED ISSUE #740: clarify snapshot docs
see #737 

maybe we need to change it entirely though...

caarlos0:
starting to work/experiment with snapshots.

closes #667
closes #740
closes #727

----

Problems found so far: if the idea is to publish a pre-release on github, it must have a tag - and for goreleaser to work after that, it must be a valid semver tag...

So, the snapshot feature will have to be deeply changed... 

Tests are failing and most things are just wrong, this is just a PoC.


caarlos0:
so, what I have in mind:

- snapshots require a semver tag with an extension, eg `v1.2.3-beta1`
- to build locally without publishing anywhere and without any tags or whatever, use `--skip-validate` and `--skip-publish` flags

---

Let's suppose we are on v2.3.1, to do a snapshot:

1. tag `v2.3.2-beta1` (or something like that)
2. `goreleaser --snapshot`

You can control wether to publish snapshots or not using the `snapshot.publish` option.

To do a local test build:

1. `goreleaser --skip-validate --skip-publish`

I think with that the `--snapshot` flag will make more sense than it does today.

What do you folks think?

nathanielc:
We are using the snapshot feature to build and publish snapshots on every commit to master. Requiring a semver tag for snapshots would break our workflow.

caarlos0:
@nathanielc so, you don't create a github release with the snapshot, right?

nathanielc:
Correct, I have 

```
release:
    disable: true
```

in my config.

soupdiver:
We use the `--snapshot` to test build for all platforms. Our binary is cross-platform and if you develop on macOS you don't immediately recognise if the windows build fails. Goreleaser builds _all_ binaries for _all_ platforms on every commit to see if things at least compile.

caarlos0:
@soupdiver this PR was mostly an experiment... I don't think I'll move on with it... 

the initial purpose of `--snapshot` was indeed to build everything to see if it works... this was me trying to make it better, but it didn't work as I was expecting.

So, the pre-release thing will always require a tag too, the only difference being that it has a prerelease semver "part" (eg, `v1.2.3-beta1`) and it shouldn't affect the snapshot feature at all.

soupdiver:
> I don't think I'll move on with it

gotcha

For me _personally_ `snapshot` does something only locally, I don't expect anything to be uploaded anywhere. Just build/archive/package everything locally and stop before uploading anything.
But it seems some ideas conflict in scope and maybe lead to the situation where you can achieve the same goal with different flags/commands. Probably makes sense to streamline this a bit then.

caarlos0:
> Probably makes sense to streamline this a bit then.

yes, I think that will be the goal on this from now on...

Issue #783, opened 2018-08-31
Issue with dockers pipeline


namreg:
**Describe the bug**

`Dockers` pipeline is not working when I run goreleaser in the docker container. 
```
CREATING DOCKER IMAGES
      • tempdir: /tmp/goreleaserdocker590848700
   ⨯ release failed after 90.76s error=failed to link dockerfile: link cmd/server/Dockerfile /tmp/goreleaserdocker590848700/Dockerfile: invalid cross-device link
```

**To Reproduce**
Try to make a release with something like that command:
```
docker run --rm --privileged -v $PWD:/go/src/$PACKAGE -v /var/run/docker.sock:/var/run/docker.sock -v $(shell which docker):/usr/bin/docker -w /go/src/$PACKAGE -e GITHUB_TOKEN goreleaser/goreleaser release 
```
```yaml
project_name: storage
git:
  short_hash: true
release:
  name_template: "v{{ .Version }}"
  prerelease: true 
builds:
  -
    main: ./cmd/cli/main.go
    binary: cli
    env:
      - CGO_ENABLED=0
    goos:
      - linux
      - darwin
    goarch:
      - amd64
      - 386
  -
    main: ./cmd/server/main.go
    binary: server
    env:
      - CGO_ENABLED=0
    goos:
      - linux
      - darwin
    goarch:
      - amd64
      - 386
archive:
  replacements:
    darwin: Darwin
    linux: Linux
    386: i386
    amd64: x86_64
checksum:
  name_template: 'checksums.txt'
dist: build/dist
changelog:
  filters:
    exclude:
    - '^docs:'
    - '^test:'
    - Merge pull request
    - Merge branch
dockers:
  -
    goos: linux
    goarch: amd64
    binary: server
    image: namreg/server
    dockerfile: cmd/server/Dockerfile
    tag_templates:
      - "{{ .Tag }}"
      - latest
```

**Expected behavior**

I expect that goreleaser push built docker images.

**Environment (please complete the following information):**

- OS: mac
- OS version: `Darwin iMac-user.onliner 16.0.0 Darwin Kernel Version 16.0.0: Mon Aug 29 17:56:20 PDT 2016; root:xnu-3789.1.32~3/RELEASE_X86_64 x86_64`
- GoReleaser Version `0.84.0, commit 93a0055d036200749a25b5a3454737b0de6af2d3, built at 2018-08-28T14:53:26Z`

**Additional context**
[...]

PR #788, opened 2018-09-04
fix: entrypoint

namreg:
It seems entrypoint script is slightly incorrect :( sorry for that, my mistake. I always forget bash syntax...

refs https://github.com/goreleaser/goreleaser/issues/783


caarlos0:
also didnt notice haha

thanks

PR #789, opened 2018-09-06
feat: add hash to scoop manifest

mgrachev:
Hi there!

When I install an application via scoop I always see the warning:
> Warning: No hash in manifest...

So, I've added the hash key to the scoop manifest.


caarlos0:
Thanks @mgrachev !

PR #793, opened 2018-09-06
feat: support setting trusted x509 certificates for tls clients

lalloni:
This PR will allow users to specify a certificate chain on configuration which will be used to validate HTTPS server certificates when using Put and Artifactory pipelines.
This allows to publish artifacts to corporate servers using private/custom CA emitted certificates.


lalloni:
Hey @caarlos0, please take a look at this PR. We need this change so goreleaser can publish to our internal Nexus repository server which uses a TLS X.509 certificate emitted by a corporate CA.
I decided to add trusted_certificates at the root of the config instead of adding it to the Put instances, but you may disagree.
My reasoning was to allow the user to reuse the certificates for all the Put add Artifactory instances so the config doesn't get cluttered when many of these pipelines need to be used.
I also made this choice so the certificate chain could be added at the end of the configuration file improving readability of the other parts.
Note that while in this PR the certificates are used just for Put & Artifactory pipelines, they might be used for other pipelines that could need to have custom trusted certificates.

pkg/config/config.go
```diff
@@ -317,6 +317,7 @@ type Project struct {
 	EnvFiles      EnvFiles  `yaml:"env_files,omitempty"`
 	Git           Git       `yaml:",omitempty"`
 	Before        Before    `yaml:",omitempty"`
+	TrustedCerts  string    `yaml:"trusted_certificates,omitempty"`
```
caarlos0: Maybe this config should be on the HTTP put section?

www/content/put.md
```diff
@@ -102,13 +126,20 @@ puts:
...
-    target: http://some.server/some/path/example-repo-local/{{ .ProjectName }}/{{ .Version }}/
+    target: https://some.server/some/path/example-repo-local/{{ .ProjectName }}/{{ .Version }}/
     # User that will be used for the deployment
     username: deployuser
     # Upload checksums (defaults to false)
     checksum: true
     # Upload signatures (defaults to false)
     signature: true
+# Certificate chain used to validate server certificates
+trusted_certificates: |
```
caarlos0: missing a tab here

internal/http/system_win.go
```diff
@@ -0,0 +1,42 @@
...
+
+package http
+
+import (
+	"crypto/x509"
+	"syscall"
+	"unsafe"
+)
+
+func loadSystemRoots() (*x509.CertPool, error) {
```
caarlos0: Doesn't `x509.SystemCertPool()` work on windows as well?

lalloni: It doesn't. See from https://github.com/golang/go/issues/16736 to a lot of discussion about it.
Using that function on windows just returns an error as you can see in [its source code](https://golang.org/src/crypto/x509/cert_pool.go?s=1328:1368#L45).

lalloni:
> Hi @lalloni! Thanks for the PR!!
> 
> I've made a few comments.
> 
> About the config, leaving it on the root will imply that it is valid for all pipes, and it is not... a little repetition may be better and make the config easier to reason about...
> 
> So... I think its better to have it in the http put section, if we the user don't want to repeat that much, they can use YAML aliases for that...

Ok. I'll go by your call. Particularly because of the nice YAML alias trick (which I didn't know about).

internal/http/system_win.go
caarlos0: so... I don't know if I wan't to maintain specific windows code... I haven't touch a windows machine in years and I don't have any idea about how right/secure the code you wrote is... nor will I be able to maintain it if needed :(

lalloni: Well, I seldom used windows for many years too, but I'm targeting some 50 windows developers with this changes on goreleaser.

I guess we can leave it out (avoiding the call to x509.SystemCertPool() if GOOS="windows") but that would mean those users will be second-class citizens, needing to add certificates in their's configs instead of relying on the globally installed certificates on Windows or on Windows group policy installed certificates which are managed by the corporate security and support teams.

And since the same projects will be worked on by linux and windows users that will mean that linux users will have to see/use certificates in the configuration too.

Any other ideas?

lalloni:
ok @caarlos0, I believe this is ready to go.

internal/http/http.go
```diff
@@ -280,21 +302,21 @@ type targetData struct {
...
-func resolveTargetTemplate(ctx *context.Context, artifactory config.Put, artifact artifact.Artifact) (string, error) {
+func resolveTargetTemplate(ctx *context.Context, put *config.Put, artifact artifact.Artifact) (string, error) {
 	data := targetData{
 		Version:     ctx.Version,
 		Tag:         ctx.Git.CurrentTag,
 		ProjectName: ctx.Config.ProjectName,
 	}
 
-	if artifactory.Mode == ModeBinary {
+	if put.Mode == ModeBinary {
```
caarlos0: 👍

lalloni:
@caarlos0 done.

caarlos0:
Thanks you very much @lalloni !

lalloni:
I must say... it hurts me when I see those negative codecov stats... but TBH, the lines still uncovered are mostly err returns which doesn't make sense or are just impossible to test.

Issue #790, opened 2018-09-06
Valid PUT ends up in error


jbaruch:
**Describe the bug**

An HTTP PUT successful with CURL ends up with `error=put: upload failed: getting http response location: http: no Location header in response` in goreleaser

**To Reproduce**

```yaml
puts:
  - name: bintray
    mode: archive
    target: https://api.bintray.com/content/your.org/your.repo/{{ .ProjectName }}/{{ .Version }}/
    username: username
```

**Expected behavior**

Looks like the put is valid in CURL. Should work in Goreleaser as well.

**Environment (please complete the following information):**

- OS: mac
- OS version: Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64 x86_64
- GoReleaser Version [goreleaser --version]

**Additional context**
I am trying to produce and document a "standard" Bintray PUT example.

caarlos0:
fixed on #794

jbaruch:
It still fails intermediately on 0.85.2:

```RELEASING WITH HTTP PUT
      ⨯ put: upload failed        error=unexpected http response status: 400 Bad Request instance=bintray username=jbaruch
      ⨯ put: upload failed        error=unexpected http response status: 400 Bad Request instance=bintray username=jbaruch
      ⨯ put: upload failed        error=unexpected http response status: 400 Bad Request instance=bintray username=jbaruch
      • uploaded successful       instance=bintray mode=archive
      • uploaded successful       instance=bintray mode=archive
      • uploaded successful       instance=bintray mode=archive
      • uploaded successful       instance=bintray mode=archive
      • uploaded successful       instance=bintray mode=archive
      • uploaded successful       instance=bintray mode=archive
      • uploaded successful       instance=bintray mode=archive
      • uploaded successful       instance=bintray mode=archive
      • uploaded successful       instance=bintray mode=archive
      • uploaded successful       instance=bintray mode=archive
      • uploaded successful       instance=bintray mode=archive
      • uploaded successful       instance=bintray mode=archive
   ⨯ release failed after 23.29s error=put: upload failed: unexpected http response status: 400 Bad Request
```

caarlos0:
uh :/ 

maybe we should print the body when we err (at least)?

lalloni:
Does goreleaser have some kind of verbose mode in which we can print all http requests being made and http responses being received?
[...]

PR #794, opened 2018-09-08
fix: remove location header requirement on http put responses

lalloni:
This PR removes the requirement for HTTP servers to return a Location header when are used as targets for publishing artifacts with the Put and Artifactory pipelines.

This is needed because Nexus repository server doesn't include a Location header on responses of PUT requests.

This change doesn't lose any functionality since the location wasn't used at all before the patch.


lalloni:
I've just noticed this PR fixes #790.

caarlos0:
thanks @lalloni !

PR #797, opened 2018-09-11
docs: Fix misplaced 'sort' for release

Jaskaranbir:
The docs for `release` show that the `sort` field should a member of `filter`, which is incorrect syntax.


caarlos0:
Thanks @Jaskaranbir ! 🚀

PR #805, opened 2018-09-17
feat: Add option to publish snapshots

nathanielc:
This change enables snapshots to be published.

This is helpful for building and publishing nightly artifacts.


caarlos0:
this will be handled by #782.

BTW: I need feedback on that one :D

caarlos0:
will close this as #782 will be used for that...

PR #808, opened 2018-09-19
fix github.com/Masterminds/semver import path

h8liu:
If applied, this commit will imports `github.com/Masterminds/semver` instead of `github.com/masterminds/semver`. This fixes case-sensitive collisions when other repos are also importing it but with a different report path.

`github.com/Masterminds/semver` seems to prefer the uppercase-leading one, see the usages in its readme.

https://raw.githubusercontent.com/Masterminds/semver/master/README.md

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -181,7 +181,7 @@
 
 [[projects]]
   digest = "1:b856d8248663c39265a764561c1a1a149783f6cc815feb54a1f3a591b91f6eca"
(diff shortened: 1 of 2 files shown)


h8liu:
build breakage was found when trying to build with other repos on [gopkgs.io](https://gopkgs.io):

https://gopkgs.io/builds/20180919-144821-ac2e1d6b

caarlos0:
goimports failed, I'll fix locally!

Thanks

Issue #807, opened 2018-09-19
Road to v1.0.0


caarlos0:
So, I think it is time we can start to plan v1.0.0.

What I have in mind so far is:

- [x] Support multiple archives
- [x] Clarify snapshots
- [x] Monorepo support with tag prefixes ( #723 )
- [x] split the build/packaging from the publishing phase (#816)
- [x] gitlab support ( #1038 )
- [x] wasm builds ( #1017 )

I think that once this is done and we are stable for a couple of weeks, we can cut the v1 and maybe start thinking on v2!

What do you think?

sidh:
Monorepo support would be welcome.

soupdiver:
Willing to help ... trying to get an overview atm

caarlos0:
> Willing to help ... trying to get an overview atm

awesome, let me know if I can help on anything or if something can be better for new people wanting to contribute 🚀

soupdiver:
>  let me know if I can help on anything

As said in the other PR, it's not completely obvious what is the direction right now.
A lot of issues/PRs have `won't fix`, are closed or were just experiments for features.

_If_ this ticket is still up-to-date then I think the mentioned PRs need some clarification.
#787 for example. It's still in the todo list here but the PR is closed and the branch is deleted.

So, making it easier to grasp _where_ exactly help is needed would be beneficial for new people to start contributing I think.

robdefeo:
@caarlos0 would you consider moving https://github.com/goreleaser/goreleaser/issues/1017 to v1.0.0 milestone?

caarlos0:
I think we can, yes.

robdefeo:
Amazing!!! 🍰 This will help a lot we are looking to get a wasm build in place to support browser implementation.

caarlos0:
cc/ @robdefeo https://github.com/goreleaser/goreleaser/pull/1159

caarlos0:
wasm support is on master.

karuppiah7890:
@caarlos0 Any updates on the mono repo support? Is anyone working on it? I see that the existing PR #723 has been closed with not much activity

caarlos0:
> @caarlos0 Any updates on the mono repo support? Is anyone working on it? I see that the existing PR #723 has been closed with not much activity

monorepos work given that you release everything in the monorepo on all tags... otherwise, no, no one working on it afaik.

jphastings:
[...]

PR #809, opened 2018-09-19
feat: Adding GitLab support for releasing, homebrew, and scoop

wburningham:
If applied, this commit will add support for GitLab see #110.

This change is being made to support talking to GitLab for releases, homebrew and scoop. It has several key changes to the internal design of goreleaser. Below is a list/summary of some of the changes you'll see in code:

- Artifacts now have an internal `id` and an `artifacts.SetExtra` function. This is so GitHub and GitLab clients can upload an artifact and then store the returned download URL on the artifact. The previous behavior was to build the download URL using a default template that uses the artifact name. Because [GitLab's uploads API](https://docs.gitlab.com/ce/api/projects.html#upload-a-file) generates a download URL rather than accepting a URL from the consumer (like [GitHub's release asset API](https://developer.github.com/v3/repos/releases/#upload-a-release-asset)) we cannot use the existing a template to build the URL, but have to store the returned download URL so we can use it later (used later in the brew and scoop pipelines).
- The `client.CreateRelease` method now returns a string for a release ID rather than an int64 since GitLab uses the tag string as the release ID when uploading a file to a project.
- Default `API` and `Download` URLs were previously set in `defaults.go` as the first default set, but at that point environment variables have not been set so it is unknown if a GitHub or GitLab token is set which determines which client to instantiate. The default `API` and `Download` URLs were deferred to when a client is instantiated.
- The `config` pkg now has custom YAML unmarshalers that will read in `github*` or `gitlab*` keys and store them in a agnostic `Repo` key on the config so code accessing the repo and URL information later doesn't need extra logic to determine whether or not to use GitLab or GitHub values. The config package will error if a `gorelease.yaml` file has both GitLab and GitHub options defined.
- A new `client.gitlabClient` exists for interacting with GitLab. Rather than importing another package (like `github.com/xanzy/go-gitlab`) the client uses the stdlib `http` package.
- The  `client.Upload` method returns a string in addition to an error where the string is the path to download the file. This is so homebrew and scoop pipelines can use that as a default URL.

This PR is large and I've tested my changes with a few use cases, but I'm not confident that it will work with other configurations.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -49,7 +49,7 @@ brew:
   test: |
     system "#{bin}/goreleaser -v"
 scoop:
-  bucket:
+  github:
     owner: goreleaser
     name: scoop-bucket
   homepage:  https://goreleaser.com
--- a/internal/artifact/artifact.go
+++ b/internal/artifact/artifact.go
@@ -4,8 +4,10 @@ package artifact
 import (
 	"crypto/sha256"
 	"encoding/hex"
+	"fmt"
 	"io"
 	"os"
+	"strconv"
 	"sync"
 
 	"github.com/apex/log"
@@ -53,6 +55,7 @@ func (t Type) String() string {
 
 // Artifact represents an artifact and its relevant info
 type Artifact struct {
+	id     int
 	Name   string
 	Path   string
 	Goos   string
@@ -76,6 +79,10 @@ func (a Artifact) Checksum() (string, error) {
 		return "", err
 	}
 	return hex.EncodeToString(hash.Sum(nil)), nil
+}
+
+func (a Artifact) ID() string {
+	return strconv.Itoa(a.id)
 }
 
 // Artifacts is a list of artifacts
@@ -116,7 +123,28 @@ func (artifacts *Artifacts) Add(a Artifact) {
 		"path": a.Path,
 		"type": a.Type,
 	}).Info("added new artifact")
+	a.id = len(artifacts.items)
 	artifacts.items = append(artifacts.items, a)
+}
+
+// SetExtra safely sets an extra k/v on the artifact with the given id.
+func (artifacts *Artifacts) SetExtra(id string, k, v string) error {
+	i, err := strconv.Atoi(id)
+	if err != nil {
+		return err
+	}
+	if i > (len(artifacts.items) - 1) {
+		return fmt.Errorf("artifact with id [%s] not found", id)
+	}
+	artifacts.lock.Lock()
+	defer artifacts.lock.Unlock()
+	a := artifacts.items[i]
+	if a.Extra == nil {
+		a.Extra = map[string]string{}
+	}
+	a.Extra[k] = v
+	artifacts.items[i] = a
+	return nil
 }
 
 // Filter defines an artifact filter which can be used within the Filter
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -18,7 +18,18 @@ type Info struct {
 
 // Client interface
 type Client interface {
-	CreateRelease(ctx *context.Context, body string) (releaseID int64, err error)
+	CreateRelease(ctx *context.Context, body string) (releaseID string, err error)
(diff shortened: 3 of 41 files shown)


wburningham:
@caarlos0 As pointed out in the summary of my changes
>Because GitLab's uploads API generates a download URL rather than accepting a URL from the consumer (like GitHub's release asset API) we cannot use the existing a template to build the URL, but have to store the returned download URL so we can use it later (used later in the brew and scoop pipelines).

In the case where a release is disabled, but the brew or scoop pipes are still enabled (the exact case that exists with the TravisCI build for this repo), the code currently fails. What are your thoughts on the following options:

If the release is disabled, but brew or scoop is enabled:
1. throw an error at the time the config is parse indicating that a URLTemplate is required.
2. Default to a URL template that is a local path to the artifact in the dist folder
3. Change the code to be something else (not sure what this would look like)

I'm not sure if ^ make sense so please let me know. Thanks.

caarlos0:
Hi, this is indeed a big PR hehe

I'll take a look in the weekend :) 

Thanks!

wburningham:
@caarlos0 were you able to look at this PR over the weekend?

tylerchr:
It would be great if goreleaser had GitLab support. I'd like to use it to make some internal CLI tools installable via Homebrew but since we don't use GitHub I would need to host the custom taps in our self-hosted GitLab instance. Right now there's not really a viable way to achieve that, at least not one that includes goreleaser. But that could change with a MR like this so I'd be glad to see this merged.

caarlos0:
Sorry, I was prepping for gophercon br, and now I'm here... hopefully next week I can take a look.

What I can say for now is that the PR is very big, probably I'll be cherry-picking stuff into goreleaser master branch... but, for sure, this is a good addition! 

Many thanks again for the PR :)

wburningham:
@caarlos0 you mentioned that you would be cherry-picking stuff. Would be breaking my PR into smaller sections or coding up the PR differently be helpful? Wondering how I can be helpful.

caarlos0:
> @caarlos0 you mentioned that you would be cherry-picking stuff. Would be breaking my PR into smaller sections or coding up the PR differently be helpful? Wondering how I can be helpful.

yes, I think so, I plan to finish the other things on #807 first, as I think they would help to make this more robust... unfortunately I have not had much time to work on those things the last few days :/

alexggordon:
@caarlos0 @wburningham so recently, GitLab added their [releases API](https://docs.gitlab.com/ee/api/releases/index.html), which allows for arbitrary artifact URL's in the release creation process. Is this something you've considered revisiting since this now seems like a significantly smaller change?

caarlos0:
> @caarlos0 @wburningham so recently, GitLab added their [releases API](https://docs.gitlab.com/ee/api/releases/index.html), which allows for arbitrary artifact URL's in the release creation process. Is this something you've considered revisiting since this now seems like a significantly smaller change?

it sure seems a simpler change than this one... interested in working on a PR for that?

alexggordon:
Yeah--assuming this is something you don't mind supporting, I'm more than happy to do the initial work. 

You have any thoughts on things I need to be aware of?

caarlos0:
Since now they have the releases support, I think we just need to add gitlab client and that should be enough to do the release part...

brew and etc will require extra work, but could be done afterwards IMHO 🤔

mavogel:
@caarlos0 I'd also like to contribute, especially to add gitlab support. But one question first would it make sense to help with #807 first or to rewrite this PR with regards to

> Since now they have the releases support, I think we just need to add gitlab client and that should be enough to do the release part...

As mentioned in https://github.com/goreleaser/goreleaser/issues/807#issuecomment-441237466 we'd need some clarification on which issue to help first :)

caarlos0:
hey @mavogel 

I think that using the new gitlab releases makes more sense and would probably be easier, hopefully also less changes than this one... 🤔

I failed to try to merge this in, sorry about that @wburningham :/

caarlos0:
will close this in favor of #1038 


thanks for the work, sorry it didn't get merge in :(

PR #818, opened 2018-10-03
fix customization link

sjansen:
Repro:
 - Open https://goreleaser.com/ in Chrome or Firefox
 - Click on "customize" in the second paragraph
Expected:
 - The page jumps to "Customization"
Actual:
 - The page doesn't change because no matching `name` or `id` is found


caarlos0:
Thanks!

PR #831, opened 2018-10-05
feat: image names can ontain template variables

flyinprogrammer:

Docker image names can now have template variables in them. This is super helpful for use cases where we want to reuse the same `.goreleaser.yml` file for all our builds so they all follow the same conventions.

Because convention > configuration


caarlos0:
Awesome! Mind to add that to the `docker.md` file as well (docs 🚀 )?

caarlos0:
Thanks for the PR! 🚀

Issue #832, opened 2018-10-05
Docker build + multiple registries.


flyinprogrammer:
**Is your feature request related to a problem? Please describe.**

We build 1 container with the same content, and tags, but we have to publish it to more than 1 registry. Today that involves creating more than 1 docker block, and then that means I loose sane defaults:  https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/docker/docker.go#L47

**Describe the solution you'd like**

I'd like to specify more than 1 registry in a docker block, just as an additional tagging/processing step within a single docker block.

**Describe alternatives you've considered**
I could do this processing outside of goreleaser.
We could just best-effort our sane defaults for all docker blocks.

**Additional context**
I work for a bigger company that has insane processes for builds, so I full appreciate that I'm probably alone in having this problem. I'm also more than willing to write the PR, I just wanted community input first.

caalberts:
I could give this a shot, taking #834 approach to test it.

caarlos0:
fixed on #840!

PR #840, opened 2018-10-13
feat: Add support for publishing to additional docker registries

caalberts:
Created a list of registries based on the docker image name and
additional registries. Image is then tagged with all permutations
of registry and tags.

See #832


flyinprogrammer:
LGTM

caarlos0:
Hi, many thanks for the PR! 🎉

maybe it would be more flexible if we had a `image_templates`? I'm thinking that maybe the user wants to push some tags to a given registry and other just to another set of registries or something like that, so the config would look like:

```yaml
# .goreleaser.yml
dockers:
  # You can have multiple Docker images.
  -
    # GOOS of the built binary that should be used.
    goos: linux
    # GOARCH of the built binary that should be used.
    goarch: amd64
    # GOARM of the built binary that should be used.
    goarm: ''
    # Name of the built binary that should be used.
    binary: mybinary
    # Skips the docker push. Could be useful if you also do draft releases.
    # Defaults to false.
    skip_push: false
    # Path to the Dockerfile (from the project root).
    dockerfile: Dockerfile
    image_templates:
    - "foo.bar/myimage/latest"
    - "myuser/myimage/latest"
    - "myuser/myimage/{{ .Tag }}"
    - "myuser/myimage/{{ .Tag }}-{{ .Env.GO_VERSION }}"
    - "myuser/myimage/v{{ .Major }}"
    # Template of the docker build flags.
    build_flag_templates:
    - "--label=org.label-schema.schema-version=1.0"
    - "--label=org.label-schema.version={{.Version}}"
    - "--label=org.label-schema.name={{.ProjectName}}"
    - "--build-arg=FOO={{.ENV.Bar}}"
    # If your Dockerfile copies files other than the binary itself,
    # you should list them here as well.
    extra_files:
    - config.yml
```

the config would become a little bit more verbose, but it would be more flexible...

what do you think?

Cheers!

caalberts:
This is definitely more flexible, but existing config file with `image` would break, wouldn’t it?

caarlos0:
> This is definitely more flexible, but existing config file with `image` would break, wouldn’t it?

yes, we have a system to deprecate things (basically warns on the output and points to the documentation), so we could deprecate that and remove after a few months.

Example -> https://github.com/goreleaser/goreleaser/pull/759

caalberts:
@caarlos0 do you intend to deprecate both `image` and `tag_templates` and replace them with just `image_templates`?

caarlos0:
> @caarlos0 do you intend to deprecate both `image` and `tag_templates` and replace them with just `image_templates`?

yes, I think that make sense... config may become more repetitive, but it would be more flexible, so I think it is a good tradeof...

caalberts:
I updated the PR with the changes for `image_templates`.

In summary, `image_templates` is the main configuration for docker now. For backward compatibility,`image` + `tag_templates` are still supported. However, `image_templates` and `image` + `tag_templates` cannot be used at the same time.

I assumed that at least 1 `image_template` is given, so there is no default value for that.

caarlos0:
awesome! Thanks for the PR!

Issue #846, opened 2018-10-24
docker images are not listed in the release notes anymore


caarlos0:
**Describe the bug**

Once upon a time we listed the published docker images within changelog:
![image](https://user-images.githubusercontent.com/245435/47436875-c45b4600-d77d-11e8-9d1d-52bf83a57d78.png)

Since v0.90.0 it stopped working, probably due to the publishing (#827) refactory.

**To Reproduce**

Just release a project with docker images.

**Expected behavior**

Same as before

jonathanlloyd:
I'd like to take a look at this :slightly_smiling_face:

caarlos0:
Great, please go ahead! :D

PR #847, opened 2018-10-25
fix: Bring back docker release notes

jonathanlloyd:
Docker image names were not being added to the release notes
because the Github release step was happening earlier in the
pipeline than the docker release itself (so no artifacts had
been added to the context). This patch re-orders the pipeline
and adds a comment to warn against this happening again.

Fixes: #846

### DIFF
--- a/internal/pipe/publish/publish.go
+++ b/internal/pipe/publish/publish.go
@@ -38,11 +38,13 @@ var publishers = []Publisher{
 	s3.Pipe{},
 	put.Pipe{},
 	artifactory.Pipe{},
-	release.Pipe{},
 	brew.Pipe{},
 	scoop.Pipe{},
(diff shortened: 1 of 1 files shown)


jonathanlloyd:
Hmm... looks like the build is erroring because it is an external PR and I do not have access to the appropriate encrypted secrets

jonathanlloyd:
I've tested this here: https://github.com/jonathanlloyd/goreleasertest/releases/tag/v2.0.0
It seems to work.

I may also have a look at adding a regression test if I can find a good way of doing it...

Happy Hacktoberfest :slightly_smiling_face:

caarlos0:
Looks good! Once I’m on my computer I’ll merge it! Thanksss 🎉🎉

PR #853, opened 2018-10-26
bump: go-homedir to stable version

jpo-joyent:
github.com/mitchellh/go-homedir has a stable release version now. Using master causes dep to be unable to solve for a suitable version when goreleaser is depended on in other projects which also (transiently) depend on the released version. This fixes that, and brings API stability guarantees.


jpo-joyent:
Note: go-homedir's master HEAD & v1.0.0 are currently the same.

caarlos0:
That's great! Thank you!

Issue #796, opened 2018-09-11
template support for goos and goarch


jeffreystoke:
**Is your feature request related to a problem? Please describe.**

I was trying to start a native build with CGO_ENABLED=1, and cross build with CGO_ENABLED=0, so I wrote the config like this

```yaml
builds:
  # native build (cgo enabled)
  - binary: config-server
    main: cmd/server/main.go
    env:
    - CGO_ENABLED=1
    - GIN_MODE=release
    ldflags:
    - -s -w -X main.version={{.Version}}
    - -X main.commit={{ .Commit }} -X main.date={{ .Date }}
    goos:
    - "{{ .Env.GOHOSTOS }}"
    goarch:
    - "{{ .Env.GOHOSTARCH }}"
  # cross build (cgo disabled)
  - binary: config-server
    main: cmd/server/main.go
    env:
    - CGO_ENABLED=0
    - GIN_MODE=release
    ldflags:
    - -s -w -X main.version={{.Version}}
    - -X main.commit={{ .Commit }} -X main.date={{ .Date }}
    goos:
    - darwin
    - linux
    goarch:
    - amd64
    - 386
    - arm
    - arm64
    goarm:
    - 5
    - 6
    - 7
    ignore:
    # ignore native build
    - goos: "{{ .Env.GOHOSTOS }}"
      goarch: "{{ .Env.GOHOSTARCH }}"
    - goos: darwin
      goarch: 386
```

then I started build as

```bash
GOHOSTOS=$(go env GOHOSTOS) \
GOHOSTARCH=$(go env GOHOSTARCH) \
GOVERSION=$(go version | awk '{print $3;}') \
goreleaser --snapshot --rm-dist --debug
```

but the output (build section)

```
   • BUILDING BINARIES
      • building                  build={[{{ .Env.GOHOSTOS }}] [{{ .Env.GOHOSTARCH }}] [6] [] [] cmd/server/main.go [-s -w -X main.version={{.Version}} -X main.commit={{ .Commit }} -X main.date={{ .Date }}] [] config-server { } [CGO_ENABLED=1 GIN_MODE=release] go [] []}
[...]

PR #855, opened 2018-10-28
feat: Add support for templates in build targets

PSanetra:
This PR fixes https://github.com/goreleaser/goreleaser/issues/796 by introducing template support for build targets and ignored golang targets.

### DIFF
--- a/internal/builders/golang/build.go
+++ b/internal/builders/golang/build.go
@@ -7,7 +7,6 @@ import (
 	"go/token"
 	"os"
 	"os/exec"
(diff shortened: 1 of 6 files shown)


caarlos0:
sorry for the delay, I'm just not sure about this yet... will let it open and think for some time..

caarlos0:
I rather not merge this in.

For the use case you need, I think its better to use something like `envsubst` to replace something in the goreleaser file.

Thanks for the PR anyway.

PR #860, opened 2018-10-31
docs: Adding Semaphore CI to GoReleaser documentation

CommanderK5:
Added Semaphore to CI docs

### DIFF
--- a/www/content/ci.md
+++ b/www/content/ci.md
@@ -182,9 +182,9 @@ steps:
 
 ```
 
-## SemaphoreCI 2.0
+## Semaphore 2.0
 
-In [SempahoreCI 2.0](https://semaphoreci.com) each project starts with the
+In [Sempahore 2.0](https://semaphoreci.com) each project starts with the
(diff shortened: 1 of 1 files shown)


www/content/ci.md
```diff
@@ -181,3 +181,88 @@ steps:
...
+
+```yml
+apiVersion: v1alpha
+kind: Secret
+metadata:
+  name: goreleaser
+data:
+  env_vars:
+    - name: GITHUB_TOKEN
+      value: "4afk4388304hfhei34950dg43245"
```
caarlos0: this is a fake token right?

caarlos0:
thanks!

www/content/ci.md
CommanderK5: Sorry, yes

caarlos0: no problem, just asked in case it was by mistake... I tested it and it wasn't valid, so I merged it :) 

thanks again for the PR :)

PR #876, opened 2018-11-21
feat: add custom_require to brew formula generation

fifitrixabelle:
Due to the below linked changes in homebrew deprecating "unused" download
strategies and the suggested workaround of adding a custom
"require_relative" I have added the ability to specify a ruby filename
in the brew specification

My use case is to continue using the GitHubPrivateRepositoryDownloadStrategy
without deprecation warnings on every brew install.

brew deprecation:
https://github.com/Homebrew/brew/commit/599ecc9b5ad7951b8ddc51490ebe93a976d43b29#diff-ad9201cdd3b6f948345629ad812cd5bd

suggested workaround:
https://github.com/Homebrew/brew/issues/5074#issuecomment-429274286

### DIFF
--- a/internal/pipe/brew/template.go
+++ b/internal/pipe/brew/template.go
@@ -14,9 +14,13 @@ type templateData struct {
 	Dependencies     []string
 	Conflicts        []string
 	Tests            []string
+	CustomRequire    string
 }
 
-const formulaTemplate = `class {{ .Name }} < Formula
+const formulaTemplate = `{{ if .CustomRequire -}}
+require_relative "{{ .CustomRequire }}"
+{{ end -}}
+class {{ .Name }} < Formula
   desc "{{ .Desc }}"
   homepage "{{ .Homepage }}"
   url "{{ .DownloadURL }}"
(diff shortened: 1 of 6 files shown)


internal/pipe/brew/template.go
```diff
@@ -14,9 +14,11 @@ type templateData struct {
 	Dependencies     []string
 	Conflicts        []string
 	Tests            []string
+	CustomRequire    string
 }
 
-const formulaTemplate = `class {{ .Name }} < Formula
+const formulaTemplate = `{{- if .CustomRequire }}require_relative "{{ .CustomRequire }}"{{- else}}# No require_relative{{- end }}
```
caarlos0: I think we can drop the `# no require_relative` comment...

fifitrixabelle: I was hoping you wouldn't say that :smile_cat: it will leave an empty line at the top of the file unless you know a good way to trim it?

fifitrixabelle: I'll play with this today and fix - think I've found how

fifitrixabelle:
That removes the comment - is that okay?

fifitrixabelle:
Thank you! :balloon:

PR #880, opened 2018-11-25
feat: support overriding files by Goos

### LINKED ISSUE #637: Support Platform Specific Files for generating the archives
I have a use case where our core binary needs to be packaged with dependencies. Since dependencies are pre-built binaries specific to each platform, we need a way to specify them in archive section. Today GoReleaser supports adding all files to release archives but not in a platform specific manner.

caarlos0:
this closes #637 

not sure it is a good idea though... maybe multiple archives is better?

### DIFF
--- a/internal/pipe/archive/archive.go
+++ b/internal/pipe/archive/archive.go
@@ -116,7 +116,7 @@ func create(ctx *context.Context, binaries []artifact.Artifact) error {
 	var a = NewEnhancedArchive(archive.New(archiveFile), wrap)
 	defer a.Close() // nolint: errcheck
 
-	files, err := findFiles(ctx)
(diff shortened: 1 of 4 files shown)


soupdiver:
> not sure it is a good idea though

I think the idea itself can be valuable. Just make sure it's not making things too complicated.

> maybe multiple archives is better

What you mean by that? Another way to specify which archives are created than Arch/OS combination?

caarlos0:
@soupdiver yeah, multiple archives would be more verbose, but also more clear.

we could also remove the format override I guess...

I still not sure how the API should look like though... probably would have to link to a specific build/list of builds somehow

caarlos0:
I'll close this in favor of multiple archives...

PR #884, opened 2018-11-29
Converted current docker images to Alpine

### LINKED ISSUE #882: Use Alpine as Docker image base
Current use of `golang:1.11` creates a _very_ bulky Debian version of the final image (1.15GB) vs what could be a smaller image if they were swapped to use Alpine.

sgnn7:
This should significantly reduce the image size of the Docker images
needed to use Goreleaser

```
Current:
goreleaser        1.15GB 
goreleaser-cgo    1.15GB

Alpine-based:
goreleaser        595MB
goreleaser-cgo    741MB
```

Resolves #882

I didn't have a big enough project to test everything that goreleaser supports in this manner but the medium-size one seems to work fine.

PS: `mercurial` is added since `go mod download` is supposed to support mercurial-backed plugins which some k8s go plugins use.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,11 @@
-FROM golang:1.11
+FROM golang:1.11-alpine
 
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends apt-transport-https \
-                                               curl \
-                                               git \
-                                               gnupg2 \
-                                               rpm \
-                                               software-properties-common && \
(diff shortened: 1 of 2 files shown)


sgnn7:
CC: @caarlos0

caarlos0:
Awesome, thanks @sgnn7 !

sgnn7:
@caarlos0 Anytime :)

PR #889, opened 2018-12-02
docs(ci): change snapcraft package with snapd

pierreboissinot:
Fix package name required for snap pipe in documentation.


caarlos0:
thanks!

Issue #894, opened 2018-12-10
Snapcraft tests are broken


caarlos0:
tests started to fail recently, probably some change on underlying `snap` utility.

PR #895, opened 2018-12-13
fix: snapshot version handling

caarlos0:
Make sure there is always a `ctx.Version` when snapshotting.

if `config.snapshot.name_template` was empty, the default was being set after the template (which was empty) was already evaluated, leaving the user with an empty `ctx.Version`.

For some reason, snapcraft just now started to complain about that.


caarlos0:
closes #894

PR #905, opened 2018-12-29
use retry package to repeat uploads

kamilsk:
it's just a suggestion to simplify code, any feedback is welcome


caarlos0:
There is a test failing, other that that looks good to me!

kamilsk:
fixed

caarlos0:
thanks!

PR #913, opened 2019-01-07
Replace deprecated .Commit var with .ShortCommit

sagikazarmark:
Replace deprecated .Commit var with .ShortCommit


caarlos0:
thanks!

PR #934, opened 2019-01-21
feat: allow s3 bucket name customization

mmatur:
### Description

This PR allow user to use custom value for S3 Bucket name


caarlos0:
Thanks!

PR #941, opened 2019-01-26
fix docker always push

guilhem:
Current docker implementation doesn't look at `ctx.SkipPublish`.

I also fix tests which was silently failing.

### DIFF
--- a/internal/pipe/docker/docker.go
+++ b/internal/pipe/docker/docker.go
@@ -92,6 +92,9 @@ func (Pipe) Run(ctx *context.Context) error {
 
 // Publish the docker images
 func (Pipe) Publish(ctx *context.Context) error {
+	if ctx.SkipPublish {
+		return pipe.ErrSkipPublishEnabled
+	}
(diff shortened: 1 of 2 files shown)


caarlos0:
this is done on a single place for all publishers: https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/publish/publish.go#L52

the tests should indeed be fixed/removed though, thanks for reporting!

guilhem:
@caarlos0 I do something like that instead: https://github.com/barpilot/goreleaser/commit/d4da9160798de330358238ed7c9c24b6a297434a

My main problem is that tests are using `Pipe{}.Publish(ctx)` and this function doesn't test `SkipPublish`.

If you remove ability to disable publishing, can we remove https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/docker/docker_test.go#L572 ?
But in that case we never test publishing docker images.

Right now, tests are just failing.

caarlos0:
> I begin to work on a version with “copy” instead... (this is by far more robust).

because there is no standard way to copy a file - and no native Go function for that, so I just link instead...

Issue #705, opened 2018-06-27
Create Multiple Archives


tdmanv:
**Is your feature request related to a problem? Please describe.**
We would like to release a server/client from a single repo using goreleaser. 

**Describe the solution you'd like**
Ideally we'd be able to create multiple archives - one for the server and one for the client. When installing the client, it shouldn't be required to get the server binary as well. We should also be able to limit the number of os/arch for client/server independently.

**Describe alternatives you've considered**
We're currently using multiple builds and the client and server get packaged in every archive. 

----

Edit by @caarlos0:

- [x] add multiple archives (with everything else using `archives[0]`)
- [x] add multiple nfpm support
- [x] add multiple brew support
- [x] add multiple scoop support
- [x] add multiple snap support
- [x] add multiple http/artifactory support
- [x] add multiple s3/blob support

bep:
Hugo really needs this ... I'm in semi-manual build mode at the moment, and that is a pain when we're doing a fair amount of releases.

caarlos0:
I was thinking in doing something similar to the docker image:


```yaml
archives:
  - binaries: [] # empty means all
    format: zip
    name_template: full
  - binaries: ['foo']
    name_template: foo
  - binaries: ['bar']
    name_template: bar
```

what do you folks think?

bep:
>what do you folks think?

Not sure. In the "Hugo use case", we have a set of builds producing `hugo` (name of binary), which should be packaged in different archives. 

```
{{.ProjectName}}_{{.Version}}_{{.Os}}-{{.Arch}}
```

The above is is the current archive name template. The above works because we have 2 goreleaser config: One with `project_name` `hugo`, one with `hugo_extended`.

The creator of this issue has probably a simpler requirement that fits your outline better, e.g. binaries `myapp-server` and `myapp-client`.

But I think that in the Go world where plugins are hard, I think having a way to package different bundles of the same app would be really useful for many. And you would want these to be named the same so they could be "drop-in-replacements".

caarlos0:
@bep hmm

so, yeah, my example maybe wasn't good, in hugo use case it would look like:

```yaml
archives:
  - binaries: ['hugo']
    name_template: 'hugo_{{.Version}}_{{.Os}}-{{.Arch}}'
  - binaries: ['hugo_extended']
    name_template: 'hugo_extended_{{.Version}}_{{.Os}}-{{.Arch}}'
```

[...]

PR #942, opened 2019-01-27
feat: multiple archives

caarlos0:
this is just the first step... not all features will work...

the idea is to have `ids` on things...  unfortunately this does not look natural on YAML, and would probably look better on HCL...

anyway, for now, you can have several builds:

```yaml
builds:
- id: foo
- id: bar
```

and then the archives reference the builds:

```yaml
archives:
- id: 'foobar' 
  builds: ['foo', 'bar']
```

later, when we have multiple everything-else, the user will reference the archive ids.

By default, all ids are set to `default`, this still requires some work though...

refs #705


syntaqx:
Would be cool to see backwards-compatibility in the form of many/one marshalability:

```go
type Archives []Archive

func (a *Archives) UnmarshalYAML(unmarshal func(interface{}) error) error {
	var archives []Archive
	if err := unmarshal(&archives); err != nil {
		var archive Archive
		if err := unmarshal(&archive); err != nil {
			return err
		}
		archives = []Archive{archive}
	}

	*a = archives
	return nil
}
```

Then you can allow for the `archive` or `archives` key to be used as the same value, providing people flexibility in their usage. If they only want a single archive, being able to use the singular verb without an array is nice, as well as having both keys behave identically.

caarlos0:
FWIW: I plan to merge and release this next weekend.

It will be a soft release, so nothing should be affected. 

After that I'll start working on the other pipes to support this, and once I'm done, finally tag and release 1.0.0 :)

frapposelli:
This is great, thanks @caarlos0

PR #943, opened 2019-01-29
feat: ability to skip Homebrew Tap update for preleases

ahamilton55:
If applied, this commit will allow for pre-releases to not be pushed to a Homebrew tap.

 Why is this change being made? When working on a tool that uses a Homebrew tap it would be helpful to be able to upload new releases to GitHub so that others can pull these down and test them manually but to help us prevent breaking the usage by other users that just want the current working version of the tool.

- Updates Homebrew config to make "SkipUpload" a string so that it can
  contain more than true or false values. New available options are
  "true", "false" and "auto".
- Adds a new check in Homebrew publish to see if it should skip due to a
  prerelease and "SkipUpload" set to auto.
- Adds a new test to make sure that tap publishing is skipped when set
  to "auto" and a prerelease semver is provided.
- Updates documents to add information about the new "auto" option for
  "SkipUpload".

### DIFF
--- a/internal/pipe/brew/brew.go
+++ b/internal/pipe/brew/brew.go
@@ -122,7 +122,7 @@ func doRun(ctx *context.Context, client client.Client) error {
 		return err
 	}
 
-	if ctx.Config.Brew.SkipUpload {
+	if strings.TrimSpace(ctx.Config.Brew.SkipUpload) == "true" {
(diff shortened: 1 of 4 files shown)


bkeroackdsc:
I was just affected by this issue where I inadvertently updated a stable homebrew formula with a prerelease, so it would be great if this can get merged.

Issue #898, opened 2018-12-19
Support additional checksum options specifically sha512


jspaleta:
**Is your feature request related to a problem? Please describe.**
I have a need to output sha512 checksum file from goreleaser instead of the default sha256 as part of CI deployment.

**Describe the solution you'd like**
I'd like to be able to specify the checksum algorithm supported by golang crypto package in the goreleaser.yml 
Naively, I would think this is a matter introducing a new yml configuration option by which to specify the desired checksum algorithm,, and then using a switch statement in artifact.Checksum to choose matching crypto package checksum function in artifact.Checksum, taking sha256 as the fallback default.

**Describe alternatives you've considered**
Currently i use a travis post deploy hook and a little shell script to generate the sha512 and then upload it into a github release. But I would really like to be able to do this inside goreleaser as a much cleaner solution.

caarlos0:
The only problem I could think about this is that brew requires sha256... not sure what's the best way to solve it though...

jspaleta:
can the config support multiple checksum files maybe?
a list of yaml dictionaries that describe each  checksum file
Strawman config:
```
checksum:
  - name_template: " {{ .ProjectName }}_{{ .Version }}_{{.Algorithm }}-checksums.txt"
    algorithm: "sha256"
  - name_template: " {{ .ProjectName }}_{{ .Version }}_{{.Algorithm }}-checksums.txt"
    algorithm: "sha512"
```
Here i'm using a new template variable to reference the checsum algorithm string.
You can even have the default name_template encode the algorithm when the checksum attribute contains a list.

If the checksum isn't a list, config parser behaves like it does now.. just using the sha256 as the fallback.

Strawman multiple files is effectively what I'm doing now.. adding the algo name into the checksum filename to make it clear which file is which in the github release. I'm just producing the sha512 outside of goreleaser which is just hacky enough to be distasteful to even myself.

jspaleta:
ohnoes  stalebot marked this as wont fix?
hey @caarlos0 
Is that just the bot being aggressive about labeling or is that a judgement as to suitability?

caarlos0:
yes, bot is agressive, sorry :)

IdanAdar:
Would also love to see support for sha1sum.

PR #951, opened 2019-02-04
feat: support different checksum algorithms

caarlos0:
closes https://github.com/goreleaser/goreleaser/issues/898

### DIFF
--- a/internal/artifact/artifact.go
+++ b/internal/artifact/artifact.go
@@ -2,8 +2,12 @@
 package artifact
 
 import (
+	"crypto/sha1"
 	"crypto/sha256"
+	"crypto/sha512"
 	"encoding/hex"
+	"fmt"
+	"hash"
 	"io"
 	"os"
 	"sync"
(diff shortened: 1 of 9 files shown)


internal/artifact/artifact.go
```diff
@@ -78,20 +82,30 @@ func (a Artifact) ExtraOr(key string, or interface{}) interface{} {
...
+	log.Debugf("calculating checksum for %s", a.Path)
 	file, err := os.Open(a.Path)
 	if err != nil {
 		return "", errors.Wrap(err, "failed to checksum")
 	}
 	defer file.Close() // nolint: errcheck
-	var hash = sha256.New()
-	_, err = io.Copy(hash, file)
+	var h hash.Hash
+	switch algorithm {
```
syntaqx:
```suggestion
	switch algorithm {
		case "crc32":
			hash = crc32.NewIEEE()
		case "md5":
			hash = md5.New()
		case "sha224":
			hash = sha256.New224()
		case "sha384":
			hash = sha512.New384()
```

Might as well add some of the other common types as well. I realize the use case is minimal, but it's almost 0 cost.

caarlos0: yeah, why not haha

jspaleta:
:smile:

Issue #952, opened 2019-02-05
GPG signed commits break hash parsing


sublimino:
**Describe the bug**

Git output for GPG signed commits is not parsed properly and breaks a build.

**To Reproduce**

When running this command against a GPG-signed HEAD commit:

```
$ goreleaser --snapshot --rm-dist --debug
```
The response contains the git CLI's signing output before the commit: 

```
-X main.commit=gpg: Signature made Fri 01 Feb 2019 19:42:11 GMT 
```

(which should read: `-X main.commit=cbf6bdcebf61c3a5b9e7407865621aa73e461c69`)

as part of

```
   • BUILDING BINARIES                                                                                                                                                                                                                                                                   
      • building                  build={[darwin linux] [amd64] [6] [darwin_amd64 linux_amd64] [{darwin 386 }] ./cmd/kubesec [-s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}}] [] kubesec { } [CGO_ENABLED=0] go [] []}                             
      • building                  binary=dist/linux_amd64/kubesec                                                                                                                                                                                                                        
      • building                  binary=dist/darwin_amd64/kubesec                                                                                                                                                                                                                       
      • running                   cmd=[go build -ldflags=-s -w -X main.version=v0.0.0-next -X main.commit=gpg: Signature made Fri 01 Feb 2019 19:42:11 GMT -X main.date=2019-02-02T13:21:13Z -o dist/linux_amd64/kubesec ./cmd/kubesec] env=[CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GOARM=]
      • running                   cmd=[go build -ldflags=-s -w -X main.version=v0.0.0-next -X main.commit=gpg: Signature made Fri 01 Feb 2019 19:42:11 GMT -X main.date=2019-02-02T13:21:13Z -o dist/darwin_amd64/kubesec ./cmd/kubesec] env=[CGO_ENABLED=0 GOOS=darwin GOARCH=amd64 GOARM
=]                                                                                                                                                                                                                                                                                       
      • failed                    cmd=[go build -ldflags=-s -w -X main.version=v0.0.0-next -X main.commit=gpg: Signature made Fri 01 Feb 2019 19:42:11 GMT -X main.date=2019-02-02T13:21:13Z -o dist/linux_amd64/kubesec ./cmd/kubesec] env=[CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GOARM=]
[...]

PR #953, opened 2019-02-05
fix: parse commit hash of GPG signed commit

sublimino:
This change will support parsing the commit hash from GPG-signed commits. It does so by disabling signing output from the `git` CLI using `git -c log.showSignature=false` prior to the git subcommand.

@controlplaneio security policy mandates GPG signed commits, and we'd like to use this tool.

Closes https://github.com/goreleaser/goreleaser/issues/952

### DIFF
--- a/internal/testlib/git.go
+++ b/internal/testlib/git.go
@@ -48,6 +48,7 @@ func fakeGit(args ...string) (string, error) {
 		"-c", "user.name='GoReleaser'",
 		"-c", "user.email='test@goreleaser.github.com'",
 		"-c", "commit.gpgSign=false",
+		"-c", "log.showSignature=false",
 	}
 	allArgs = append(allArgs, args...)
 	return git.Run(allArgs...)
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -18,6 +18,10 @@ func IsRepo() bool {
 // Run runs a git command and returns its output or errors
 func Run(args ...string) (string, error) {
 	// TODO: use exex.CommandContext here and refactor.
+	var extraArgs = []string{
+		"-c", "log.showSignature=false",
+	}
+	args = append(extraArgs, args...)
 	/* #nosec */
 	var cmd = exec.Command("git", args...)
 	log.WithField("args", args).Debug("running git")


caarlos0:
wouldn't this work too? https://github.com/goreleaser/goreleaser/compare/master...quiet?expand=1

internal/testlib/git.go
```diff
@@ -48,6 +48,7 @@ func fakeGit(args ...string) (string, error) {
 		"-c", "user.name='GoReleaser'",
 		"-c", "user.email='test@goreleaser.github.com'",
 		"-c", "commit.gpgSign=false",
+		"-c", "log.showSignature=false",
```
caarlos0: this is not needed I believe...

sublimino:
> wouldn't this work too? [goreleaser/goreleaser/compare/master...quiet?expand=1](https://github.com/goreleaser/goreleaser/compare/master...quiet?expand=1)

Not on latest Debian with git version 2.20.1. The flag in this PR is needed. 

`-q` can be added to suppress further lines (although it seems they're discarded anyway).

```
$ git show --format='%H' HEAD
gpg: Signature made Fri 11 Jan 2019 16:50:03 CET
gpg:                using RSA key 7DCF335A3BB2EC440C516D4C7B9EA6DD705D5F09
gpg: Good signature from "Andrew Martin <sublimino@gmail.com>" [ultimate]
e0754ec9a9cfc8cd9fe2de6b918235b44da916d1

diff --git a/x b/y
similarity index 100%
rename from x
rename to y
```

```
$ git show --format='%H' HEAD -q
gpg: Signature made Fri 11 Jan 2019 16:50:03 CET
gpg:                using RSA key 7DCF335A3BB2EC440C516D4C7B9EA6DD705D5F09
gpg: Good signature from "Andrew Martin <sublimino@gmail.com>" [ultimate]
e0754ec9a9cfc8cd9fe2de6b918235b44da916d1
```

```
$ git -c log.showSignature=false show --format='%H' HEAD
e0754ec9a9cfc8cd9fe2de6b918235b44da916d1

diff --git a/x b/y
similarity index 100%
rename from x
rename to y
```

```
$ git -c log.showSignature=false show --format='%H' HEAD -q
e0754ec9a9cfc8cd9fe2de6b918235b44da916d1
```

caarlos0:
oh, ok, thanks for testing it... I couldn't reproduce this locally, even with signed commits... anyway, will merge this in.

thanks for the PR :)

sublimino:
Ah yes, that's because my `~/.gitconfig` is set to show signatures. Thanks for the merge :pray: 

```toml
[log]              
  showSignature = true   
```

caarlos0:
hmmm, that makes sense... anyway, it safer to prevent that =D

Issue #946, opened 2019-02-01
An additional parameter not to compress bundled files in archive. Just archive the binaries as it is.


warrensbox:
**Is your feature request related to a problem? Please describe.**
I am trying to just publish the binaries, 
appbin_0.1.438_darwin_amd64.tar.gz 
appbin_0.1.438_linux_amd64.tar.gz 
appbin_0.1.438_darwin_386.tar.gz 
appbin_0.1.438_linux_386.tar.gz 

**Describe the solution you'd like**
appbin_0.1.438_darwin_amd64 
appbin_0.1.438_linux_amd64
appbin_0.1.438_darwin_386 
appbin_0.1.438_linux_386 
appbin_0.1.438_checksums.txt

I do not want to compress it, I just need the binary in archives

**Describe alternatives you've considered**
Manually building and pushing it.

**Additional context**
An additional parameter not to compress bundled files in archive. Just archive the binaries as it is.

IdanAdar:
Can be done by specifying: 

```
archive:
  format: binary
  name_template: "certificate-manager-{{.Os}}-{{.Arch}}-{{.Version}}"
```

caarlos0:
that's true, I thought the user wanted to have an archive in "archive mode" (without compression).

I was probably drunk when I read this.

caarlos0:
thanks for the help @IdanAdar !

warrensbox:
@IdanAdar  @caarlos0  In the documentation it say the following:
```
 # Archive format. Valid options are `tar.gz`, `zip` and `binary`.
  # If format is `binary`, no archives are created and the binaries are instead uploaded directly.
  # In that case name_template and the below specified files are ignored.
  # Default is `tar.gz`.
  format: zip
```

but in the `name_template` section of the archieve, it says:
```
  # Archive name template.
  # Defaults:
  # - if format is `tar.gz` or `zip`:
  #   - `{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}`
  # - if format is `binary`:
  #   - `{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}`
  name_template: "{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}"
```

I thought `name_template` are ignored when the format is `binary`?
[...]

PR #954, opened 2019-02-06
Make a few things clearer in quickstart docs

WheresAlice:
If applied, this commit will update the quickstart docs to make a few things clearer

This change is being made to clear up a few things that I struggled with on my first time using GoReleaser.

#946 could potentially have been avoided by making it clearer in the quickstart docs that you can change the format to release just the binaries not in any archive.


caarlos0:
Awesome! Thanks!

PR #963, opened 2019-02-19
Support custom file upload for S3

jondot:
Add support for custom S3 upload. This is useful when releasing to S3 and not Github. Github contains an implicit feature -- "latest release". When using S3 as a release destination, there is no way to express "this is the latest release version", unless you write a custom service to provide that by scanning S3 objects.

On the other hand, if we upload all assets, and then upload a custom-built manifest that states the latest build number (or in my case, an updated `install.sh` script), we have feature parity with releasing through Github in a sense.

Apologies for not adding tests. I set everything up, began writing them, and saw that the current test don't really test for correctness, just for failure. I assume testing is done manually, so saw no point in doing that (i'm using this feature in production on my product).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ coverage.txt
 goreleaser
 debug.test
 snap.login
+.vscode/
--- a/internal/pipe/s3/s3.go
+++ b/internal/pipe/s3/s3.go
@@ -84,34 +84,66 @@ func upload(ctx *context.Context, conf config.S3) error {
 	}
 
 	var g = semerrgroup.New(ctx.Parallelism)
-	for _, artifact := range ctx.Artifacts.Filter(
-		artifact.Or(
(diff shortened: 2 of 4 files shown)


caarlos0:
I'll try to review this over the weekend...

.gitignore
```diff
@@ -5,3 +5,4 @@ coverage.txt
 goreleaser
 debug.test
 snap.login
+.vscode/
```
caarlos0: this should be on your `~/.gitignore`

caarlos0:
I'm sorry, I rather not add this in right now... it seems confusing to use and I haven't seen people asking for this (except for you 😇)

this is not a definitive answer though, but for now its a no.

thanks for the effort anyways 🚀

jondot:
Sure, im using my own binary for the moment

PR #972, opened 2019-03-02
update retry package version

kamilsk:
I improved my [package](https://github.com/kamilsk/retry) and updated its major version here.


kamilsk:
Also, I suggest to wrap [this](https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/release/release.go#L120) by [safe.Close](https://github.com/kamilsk/platform/blob/master/pkg/safe/close.go#L5-L26) instead of use `nolint: errcheck`. It could be look like:

```go
defer safe.Close(file)
// or
defer safe.Close(file, func (err error) { log.Println(err) }) // etc
```

caarlos0:
thanks!

kamilsk:
@caarlos0, what do you about this https://github.com/goreleaser/goreleaser/pull/972#issuecomment-468935544 proposal? Can I add a dependency on my another package to use https://github.com/kamilsk/platform/blob/master/pkg/safe/close.go? Or it is not interesting to you. Or I can copy-paste it to your project.

caarlos0:
no need for now, thanks for the proposal though! :)

kamilsk:
ok, thank you for your great project. goreleaser saved me a lot of time!

PR #976, opened 2019-03-03
Handle repository URLs without owner

hansmi:
When a repository has a remote URL such as "user@host:foobar.git" the
code extracting the owner panics. With this pull request such URLs are
handled more gracefully.

### DIFF
--- a/internal/pipe/release/release.go
+++ b/internal/pipe/release/release.go
@@ -36,6 +36,11 @@ func (Pipe) Default(ctx *context.Context) error {
 		ctx.Config.Release.GitHub = repo
 	}
 
+	if (ctx.Config.Release.GitHub.Owner == "" || ctx.Config.Release.GitHub.Name == "") &&
+		!(ctx.Snapshot || ctx.Config.Release.Disable) {
+		return errors.New("Releases require a repository owner and name")
+	}
+
 	// Check if we have to check the git tag for an indicator to mark as pre release
 	switch ctx.Config.Release.Prerelease {
 	case "auto":
--- a/internal/pipe/release/release_test.go
+++ b/internal/pipe/release/release_test.go
@@ -314,6 +314,38 @@ func TestDefaultGitRepoWithoutRemote(t *testing.T) {
 	assert.Empty(t, ctx.Config.Release.GitHub.String())
 }
 
+func TestDefaultGitRepoWithoutSlash(t *testing.T) {
+	_, back := testlib.Mktmp(t)
+	defer back()
+	testlib.GitInit(t)
+	testlib.GitRemoteAdd(t, "user@host:myrepo.git")
+
+	var ctx = context.New(config.Project{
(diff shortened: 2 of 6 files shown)


caarlos0:
I need to double check if there are other implications on that change... probably it's fine, but will test it over the weekend anyway...

thanks for the contribution 🚀

caarlos0:
I think there are too many things depending on both user and repo... not sure it's safe to merge this 🤔

hansmi:
> I think there are too many things depending on both user and repo

In my very limited usecase I resorted to using a remote URL of `user@host:./repo.git` (note the `./`) before I submitted this PR. Not every environment uses a `user/repo` format.

What usecases can't work with an empty user string?

caarlos0:
> What usecases can't work with an empty user string?

github and brew releases will break for sure - on those cases we shouldn't accept a repo without both user and name...

idk, seems risky to add this... 

btw, what's your usecase? you want to only build things with goreleaser without pushing them anywhere of just don't want to push to github?

hansmi:
>  you want to only build things with goreleaser without pushing them anywhere of just don't want to push to github?

I'm using my own server where I store Git repositories in the home directory of a dedicated user. No third-party infrastructure involved. The resulting packages are published via reprepro.

caarlos0:
in that case, I believe you can just add 

```
release:
  github:
    owner: nope
    name: repo-name
```

which will make goreleaser not try to extract it from the git URL...

hansmi:
@caarlos0, explicitly setting the GitHub owner and name seems to work indeed. I only had the following:

```
release:
  disable: true
```

Even then a panic is not the nicest error message. I will update this pull request to return an error from `extractRepoFromURL` when the format doesn't contain a slash.

caarlos0:
awesome, thanks! :d

hansmi:
@caarlos0, while making that change I realized that what was actually missing was a merely a detection of repositories without owner/name for GitHub releases. I'll push an update which once again goes into the same direction as my originally proposed change, but with more tests. I hope you don't mind. Could you please have another look?

caarlos0:
I still think it's better to just validate both owner and repo are provided and put on a fake owner on the config...

hansmi:
> put on a fake owner on the config

Why require that in every configuration where GitHub releases are disabled?

caarlos0:
> Why require that in every configuration where GitHub releases are disabled?

that same repo/url is used in more places... goreleaser is kind of hardcoded with github right now... your patch will probably result in hard to understand errors in some configurations... I rather see it erroring before doing anything...

hansmi:
@caarlos0 Which parts of this PR do you want to retain?

caarlos0:
I think I only want to "not panic if repo does not have user and repo"

the other parts maybe just better documentation... AFAIK most of time goreleaser is being used with github repos so I don't want to add this right now...

Issue #977, opened 2019-03-05
Pre-releases are pushed to Docker hub


rvolosatovs:
**Describe the bug**

Docker images are pushed to Docker hub on pre-releases - that should not happen.
https://github.com/TheThingsNetwork/lorawan-stack/releases/tag/v3.0.0-rc1 produced updates here: https://hub.docker.com/r/thethingsnetwork/lorawan-stack/tags

**To Reproduce**

Relevant(docker) config:
https://github.com/TheThingsNetwork/lorawan-stack/blob/bb43c8d9a6bb6762373bdf7ac4eda78ff14e6977/.goreleaser.yml#L120-L132

**Expected behavior**

No Docker image push on pre-release

**Environment (please complete the following information):**

- OS: linux
- OS version: Ubuntu 14.04.5 LTS
- GoReleaser Version: v0.101.0

PR #978, opened 2019-03-05
ability to skip Docker update for pre-releases (#977)

rvolosatovs:
If applied, this commit will skip docker pushes on `skip_publish` builds, draft builds and on pre-releases if `skip_push` is set to `auto`

Closes https://github.com/goreleaser/goreleaser/issues/977

I directly followed approach introduced in https://github.com/goreleaser/goreleaser/pull/943
So:
- Change `config.Docker.SkipPush` to `string`
- Respect `skip_publish` setting
- Respect `draft` setting
- Skip `docker push` if `skip_push` is set to `auto` and version is a pre-release


caarlos0:
thanks!

PR #987, opened 2019-03-19
feat: linux compatible arch for ppc64/ppc64le

pdemonaco:
If applied, this commit will allow goreleaser to generate rpm packages for ppc64 and ppc64le targets which include the correct architecture in the SPEC definition.

### DIFF
--- a/internal/linux/arch.go
+++ b/internal/linux/arch.go
@@ -16,6 +16,10 @@ func Arch(key string) string {
 		return "armel"
 	case strings.Contains(key, "arm7"):
 		return "armhf"
+	case strings.Contains(key, "ppc64le"):
+		return "ppc64le"
+	case strings.Contains(key, "ppc64"):
+		return "ppc64"
 	}
 	return key
 }


caarlos0:
Hi, thanks for the PR!

I think this should be added [here](https://github.com/goreleaser/nfpm/blob/master/rpm/rpm.go#L30)

the method you changed if I recall correctly is only used on snapcraft (maybe we should refactor that even)...

pdemonaco:
Interesting. I was going to make my change in nfpm at that exact location initially but it was unclear how to actually test it. I'm working on adding ppc64/ppc64le to another project which is built with goreleaser so I've been using that project to test.

I can confirm that making the change here does work and is passed to nfpm. In this case, I added the target architectures to an existing projects .goreleaser.yml and performed a test release using my locally built version of goreleaser with this change.

If I were to add it to that table in nfpm what values should I map? linuxppc64 and linuxppc64le?

caarlos0:
> If I were to add it to that table in nfpm what values should I map? linuxppc64 and linuxppc64le?

yes, you can start with that, we can change more later... there are some acceptance tests that you can play with to check the changes...

pdemonaco:
OK. I've added a replacement PR against [goreleaser/nfpm](https://github.com/goreleaser/nfpm/pull/47) which should deliver the same functionality.

caarlos0:
actually, this PR is needed indeed... the other on the nfpm side wasn't required... 

will merge this and do the needed things on nfpm, thanks!

PR #988, opened 2019-03-20
test: validate arch map for linuxpp64/ppc64le

pdemonaco:
If applied, this commit will validate the goarch to linux compatible architectures for the two ppc64 cases.

I didn't notice the go testing framework when I made the first pull
request so these were left out. Not a big deal either way but we might
as well have complete tests.

Related to #987.


caarlos0:
thanks!

PR #989, opened 2019-03-21
Set GoCenter as GOPROXY to speed up go builds

rimusz:
Use https://gocenter.io as GOPROXY.

Using https://gocenter.io as GOPROXY will speed modules download and will ensure that modules referenced in go.mod are always the same, doesn't matter if some modules get deleted/changed/etc at source e.g. GitHub and etc, they get cached for live at https://gocenter.io  :-)


caarlos0:
awesome, thanks!

rimusz:
any time :-)

PR #1000, opened 2019-04-08
feat: add flag to wrap binary in bin directory on archives for unix

SimonBarendse:
Add a flag to the archive configuration that makes it possible to add the binary in a bin directory.

This makes it possible to extract the archive in a location outside the `PATH` and add the `bin` directory to the `PATH`. We don't want to add the root of the archive, as we don't want the README, LICENSE and other included files to be in the `PATH`.

This is also how [`Golang`](https://golang.org/dl/) structures their archive.

### DIFF
--- a/internal/pipe/archive/archive.go
+++ b/internal/pipe/archive/archive.go
@@ -126,8 +126,12 @@ func create(ctx *context.Context, binaries []artifact.Artifact) error {
 		}
 	}
 	for _, binary := range binaries {
-		if err := a.Add(binary.Name, binary.Path); err != nil {
-			return fmt.Errorf("failed to add %s -> %s to the archive: %s", binary.Path, binary.Name, err.Error())
+		dest := binary.Name
+		if ctx.Config.Archive.BinDirectory {
+			dest = filepath.Join("bin", binary.Name)
+		}
+		if err := a.Add(dest, binary.Path); err != nil {
+			return fmt.Errorf("failed to add %s -> %s to the archive: %s", binary.Path, dest, err.Error())
 		}
 	}
 	ctx.Artifacts.Add(artifact.Artifact{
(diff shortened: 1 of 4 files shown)


caarlos0:
I think you can already do that using `wrap_in_directory: bin` on your archive config...

florisvdg:
The `wrap_in_directory` setting is there to rename the untarred artifact top directory. 
So with `wrap_in_directory` set to `bin`, untarring `my-app.tar.gz` would result into:
```
bin
   README.md
   LICENSE
   my-app
```
Which is not that useful, and even confusing. With the proposed `bin_directory` flag however, the untarred directory name is kept as is (i.e. the name of the artifact) and inside of it only the binary is wrapped in a `bin` directory:
```
my-app
   README.md
   LICENSE
   bin
      my-app
```

SimonBarendse:
I've updated the PR to have scoop and brew take into account the `bin_directory` flag.

I've submitted #1001 as an alternative, which requires less changes in existing pipes.

caarlos0:
> The `wrap_in_directory` setting is there to rename the untarred artifact top directory.
> So with `wrap_in_directory` set to `bin`, untarring `my-app.tar.gz` would result into:
> 
> ```
> bin/
>    README.md
>    LICENSE
>    my-app
> ```
> Which is not that useful, and even confusing. With the proposed `bin_directory` flag however, the untarred directory name is kept as is (i.e. the name of the artifact) and inside of it only the binary is wrapped in a `bin` directory:
> 
> ```
> my-app/
>    README.md
>    LICENSE
>    bin/
>       my-app
> ```

oh yes, you're right.

I've looked into #1001 and liked it better, lets roll with that one?

SimonBarendse:
> I've looked into #1001 and liked it better, lets roll with that one?

:+1: I'll close this PR.

PR #1001, opened 2019-04-09
fix: use the filename of the binary in snapcraft

SimonBarendse:
When builds.binary is a path, only use the filename as the name
of the snapcraft app and as the command, instead of the full path.

As an alternative to #1000 we could have the binary field of the builds section be a path:

```yaml
builds:
  binary: bin/app
```

This works for most pipes, but fails on the snapcraft pipe. This PR fixes this usecase for the snapcraft pipe.

### DIFF
--- a/internal/pipe/snapcraft/snapcraft.go
+++ b/internal/pipe/snapcraft/snapcraft.go
@@ -152,11 +152,12 @@ func create(ctx *context.Context, arch string, binaries []artifact.Artifact) err
 	}
 
 	for _, binary := range binaries {
+		_, name := filepath.Split(binary.Name)
 		log.WithField("path", binary.Path).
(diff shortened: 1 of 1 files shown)


caarlos0:
I like this one better! 

Can you also add something on the docs about it?

--- 

about the build, it's my fold, will fix it.

SimonBarendse:
Hi @caarlos0, I've added a small note on the build docs. Is there any other place that needs updating?

caarlos0:
> Hi @caarlos0, I've added a small note on the build docs. Is there any other place that needs updating?

I think that's good! No need to add in more places.

I merged master into your branch btw

SimonBarendse:
> I merged master into your branch btw

Thnx! :smile: That should fix the build issues

PR #1003, opened 2019-04-10
Fix signature path and upload

hamstah:
The signature Path was assumed to be at the root of dist without
taking into account the target of the original artifact. This made the
upload of signature fails with file not found errors.

The signature Name was the actual signature filename and didn't take
into account the target and would lead to conflicts when flattening
the paths for upload. This change derives the Name from the original
artifact Name instead.

Fix the Path and Name of Signature artifacts

The current version does not work when uploading multiple signatures for the same build with multiple goarch/goos:
- The Path is incorrect and points to `dist/filename.sig` instead of `dist/target/filename.sig`
  ```
     ⨯ release failed after 42.20s error=GitHub Releases: failed to publish artifacts: failed to upload aws-dump.asc after 10 retries: open dist/aws-dump.asc: no such file or directory
  ```
- The Name is set to the `filename.sig` and ends up conflicting on release upload when 2 signature for the same binary for 2 different targets are uploaded.
  ```
     ⨯ release failed after 40.16s error=GitHub Releases: failed to publish artifacts: failed to upload aws-dump.asc after 10 retries: POST https://uploads.github.com/repos/hamstah/awstools/releases/16661863/assets?name=aws-dump.asc: 422 Validation Failed [{Resource:ReleaseAsset Field:name Code:already_exists Message:}]
  ```

Here is a minimal yaml file that reproduces the issue for me.

```
before:
  hooks:
    - go mod download
archive:
  format: binary
sign:
  signature: "${artifact}.asc"
  artifacts: all
release:
  draft: true
builds:
  - env:
      - CGO_ENABLED=0
    main: ./aws/dump/
    binary: aws-dump
    goos:
      - linux
    goarch:
      - amd64
```

### DIFF
--- a/internal/pipe/sign/sign.go
+++ b/internal/pipe/sign/sign.go
@@ -62,30 +62,23 @@ func (Pipe) Run(ctx *context.Context) error {
 }
 
 func sign(ctx *context.Context, artifacts []artifact.Artifact) error {
-	// nolint:prealloc
-	var sigs []string
+
 	for _, a := range artifacts {
-		sig, err := signone(ctx, a)
+		artifact, err := signone(ctx, a)
 		if err != nil {
 			return err
 		}
-		sigs = append(sigs, sig)
-	}
-	for _, sig := range sigs {
-		ctx.Artifacts.Add(artifact.Artifact{
-			Type: artifact.Signature,
-			Name: sig,
-			Path: filepath.Join(ctx.Config.Dist, sig),
-		})
+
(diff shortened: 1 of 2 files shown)


caarlos0:
awesome, thanks @hamstah 
can you add a test case for it, please?

thanks!

hamstah:
@caarlos0 done

caarlos0:
awesome @hamstah, thanks!

hamstah:
🎉  Thanks for the quick merge!

PR #1011, opened 2019-04-18
feat: add base key to snapcraft

mgrachev:
Hi there!

I've added the base key to the snapcraft pipe.


caarlos0:
awesome, thank you! :D

PR #1020, opened 2019-05-17
fix: add env expansion within signage

### LINKED ISSUE #1009: sign: should use env variables
**Describe the bug**

Within the sign step the environment variables set in the env section are not considered for expanding.

**To Reproduce**

Use custom env variable within ags of sign step. For e.g.:

```
sign:
  signature: "${artifact}.sig"
  cmd: openssl
  args: ["dgst", "-sha256", "-sign", "${SIGN_KEY}", "-out", "${signature}", "${artifact}"]
  artifacts: al
```

**Expected behavior**

I would expect that goreleaser expands the content with the variable set in the environment section.

- OS: Linux
- OS version: Linux Volta 4.4.0-18362-Microsoft #1-Microsoft Mon Mar 18 12:02:00 PST 2019 x86_64 GNU/Linux
- GoReleaser Version: latest

**Additional context**
I created a custom build which enables this by adding all environment variables from `ctx.Env` to the map which is used to expand variables. 

I've added the follow snippet at this [location](https://github.com/goreleaser/goreleaser/blob/d8818c2683daf01cde35ee194215cfc38a0da550/internal/pipe/sign/sign.go#L89). I would like to receive comments if this is considered as it should work ins this way, then I would create a PR.

```Go
for _, e := range ctx.Env.Strings() {
	item := strings.SplitN(e, "=", 2)
	env[item[0]] = item[1]
}
```

Thanks in advance.

tobiaskohlbau:
Add support for environment variables
within sign arguments. For e.g. the private key for signing.

Fixes #1009

### DIFF
--- a/internal/pipe/sign/sign.go
+++ b/internal/pipe/sign/sign.go
@@ -77,9 +77,8 @@ func sign(ctx *context.Context, artifacts []artifact.Artifact) error {
 func signone(ctx *context.Context, a artifact.Artifact) (*artifact.Artifact, error) {
 	cfg := ctx.Config.Sign
 
-	env := map[string]string{
-		"artifact": a.Path,
-	}
+	env := ctx.Env
+	env["artifact"] = a.Path
 	env["signature"] = expand(cfg.Signature, env)
 
 	// nolint:prealloc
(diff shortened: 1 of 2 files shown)


caarlos0:
awesome, thanks!

Issue #1029, opened 2019-05-28
Customization in bintray upload: file filters


apoorvam:
Firstly, a great library!

I wanted to upload my packages to Bintray and referred this [documentation](https://goreleaser.com/customization/#Bintray). I want to selectively upload files of particular type only out of all the things generated by goreleaser. Is there an option to customize this?

caarlos0:
have you checked http put docs?

https://goreleaser.com/customization/#HTTP%20Put

you should be able to filter artifacts by id (which is kind of new in goreleaser, so make sure you're on latest).

apoorvam:
Checked it. Are IDs of artifacts same as the filename of artifact? If not, where are they assigned?

caarlos0:
no, several artifacts have an `id` field (check archives for instance), so the id filter will filter for all artifacts with that given id...

apoorvam:
I used `id` with put, but it has no support for .OS and .Arch information. But I would need this info which is part of my target url.

caarlos0:
hmm, can you provide the full `.goreleaser.yaml` of what you're trying to do? just to make sure I understood it correctly...

apoorvam:
[Here](https://github.com/leopardslab/dunner/blob/deb-package/.goreleaser.yml) is my goreleaser.yaml file.

* We currently generate binaries, rpm, deb from goreleaser
* Now we want to push deb packages to bintray and rpm packages to separate bintray repo. 

I gave different ids for each of rpm and deb, so i can use it in http put. However, http put url also needs os type, arch type and version info. All of it is not available in `archive` mode and only available in binary mode.

caarlos0:
hmm, got it, this is indeed lacking, and, unfortunately, not too easy to add. 

I'll try to work on it over the next days, but can't promise anything.

caarlos0:
maybe if you could filter by arch on nfpm this would be doable too, you could create several nfpm configs (one for each arch) with different ids and several http put configs too, but still lacking things :(

apoorvam:
Yes, looks like that's way to do it with goreleaser for now. Otherwise, will try to write a script for this task. 

Thanks!

apoorvam:
Since, travis does not have support `rpmbuild >= 4.13`, how is goreleaser released for rpm builds? I tried this, but latest version available with travis is 4.11.

apoorvam:
[...]

PR #1030, opened 2019-05-29
fix: nfpm id

caarlos0:
missed it on #1025


caarlos0:
also refs https://github.com/goreleaser/goreleaser/issues/1029

PR #1031, opened 2019-05-29
Skip arm5 snaps

SimonBarendse:
Skip arm5 snaps.

Snapcraft returns the following error when trying to upload arm5 snaps:
```
The store was unable to accept this snap.
  - architectures: Invalid architecture specified in the manifest: arm5.
```


caarlos0:
hi, thanks for the PR, this was fixed by https://github.com/goreleaser/goreleaser/pull/1027

goreleaser was wrongly assuming arm6 as armel, but arm6 is armhf and armel is arm5, so this shouldn't be needed I think...

let me know otherwise!

thanks

SimonBarendse:
thanks! That indeed solves our issue, so this change isn't needed.

PR #1036, opened 2019-06-02
feat: add support for pushing artifacts to cloud storage provider( S3, Azure Blob, GCS)

C123R:
### Types of changes
- [ ]  Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ]  Breaking change (fix or feature that would cause existing functionality to change)
- [X]   I have read the CONTRIBUTING document.
- [ ]  make ci passes on my machine.
- [X]  My change requires a change to the documentation.
- [X]   I have updated the documentation accordingly.
- [X]  I have added tests to cover my changes.

### Comments
make ci don't pass on my machine, because of failing **TestSignArtifacts** unit tests.
Documentation: Only a first draft based on S3 documentation.
### Description
If applied, this commit will add support for pushing artifacts to Cloud Storage Providers:
- AWS S3
- Azure Blob
- Google Cloud Storage.

and future providers which will be supported by [Go cloud](https://github.com/google/go-cloud).

Currently, blob artifactory supports authentication only using ENVIRONMENT variables.

Below is the list of ENV variable required:

## [S3 Provider](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html)

- AWS_ACCESS_KEY
- AWS_SECRET_KEY
- AWS_DEFAULT_REGION

## [Azure Blob Provider](https://docs.microsoft.com/en-us/azure/storage/common/storage-azure-cli#set-default-azure-storage-account-environment-variables)

- AZURE_STORAGE_ACCOUNT
- AZURE_STORAGE_KEY

## [GCS Provider](https://cloud.google.com/docs/authentication/production)

- GOOGLE_APPLICATION_CREDENTIALS

### .goreleaser.yaml
Add support in goreleaser configuration to specify Go Cloud bucket provider:

```bash

blob:
  -
    provider: azblob
    bucket: goreleaser-bucket
  -
    provider: gs
    bucket: goreleaser-bucket
  -
    provider: s3
    bucket: goreleaser-bucket
```

### Why is this change being made?
Since we(as in my org) currently using Azure Cloud and wanted to have support for Azure Blob storage which can be used for storing our command line tools or so. Instead of creating a feature to support Azure Blob like the one already present for S3, I thought of using GO SDK which can support GCS, S3, and Azure Blob and in future may be more if required.

### Feedback
I would love to hear feedback from you about the implementation. This is just a first draft, and I am sure this feature would be helpful.


pkg/config/config.go
```diff
@@ -314,29 +320,29 @@ type Put struct {
...
+	Release       Release   `yaml:",omitempty"`
+	Brew          Homebrew  `yaml:",omitempty"`
+	Scoop         Scoop     `yaml:",omitempty"`
+	Builds        []Build   `yaml:",omitempty"`
+	Archive       Archive   `yaml:",omitempty"` // TODO: remove this
+	Archives      []Archive `yaml:",omitempty"`
+	NFPM          NFPM      `yaml:",omitempty"` // TODO: remove this
+	NFPMs         []NFPM    `yaml:"nfpms,omitempty"`
 	Snapcraft     Snapcraft   `yaml:",omitempty"` // TODO: remove this
-	Snapcrafts    []Snapcraft `yaml:",omitempty"`
```
caarlos0: this line should not have been deleted...

caarlos0:
This is awesome! 🚀 

Can we add the other options from S3 too (folder, profile, etc)?

https://goreleaser.com/s3/

if so, maybe we can remove that pipe and leave only the blob one in the future....

thanks for the PR, once CI is green I'll review it, but it does look ok at first glance 👍

C123R:
> Can we add the other options from S3 too (folder, profile, etc)?
> 
> https://goreleaser.com/s3/
> 
> if so, maybe we can remove that pipe and leave only the blob one in the future....

@caarlos0 For sure, yeah I just wanted to add a basic feature to support blob storage, I will go through S3 implementation. Currently struggling with PR checks ;)

pkg/config/config.go
C123R: done

pkg/config/config.go
```diff
@@ -332,6 +338,7 @@ type Project struct {
 	Artifactories []Put       `yaml:",omitempty"`
 	Puts          []Put       `yaml:",omitempty"`
 	S3            []S3        `yaml:"s3,omitempty"`
+	Blob          []Blob      `yaml:"blob,omitempty"`
```
caarlos0: I would just renamed it to `Blobs` since its plural..

internal/pipe/blob/openbucket.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// Import the blob packages we want to be able to open.
+	_ "gocloud.dev/blob/azureblob"
+	_ "gocloud.dev/blob/gcsblob"
+	_ "gocloud.dev/blob/s3blob"
+)
+
+// OpenBucket is the interface that wraps the BucketConnect and UploadBucket method
+type OpenBucket interface {
+	BucketConnect(ctx *context.Context, bucketURL string) (*gocdk.Bucket, error)
```
caarlos0: maybe just `Connect`

internal/pipe/blob/openbucket.go
```diff
@@ -0,0 +1,111 @@
...
+	UploadBucket(ctx *context.Context, bucketURL string) error
```
caarlos0: maybe just `Upload`

internal/pipe/blob/openbucket.go
```diff
@@ -0,0 +1,111 @@
...
+				if errorContains(err, "NoSuchBucket", "ContainerNotFound", "notFound") {
+					return fmt.Errorf("(%v) provided bucket does not exist", bucketURL)
+				}
+				return fmt.Errorf("failed to write to bucket : %s", err)
+			}
+			if err = w.Close(); err != nil {
+				// Invalid AWS Keys
+				switch {
+
+				case errorContains(err, "InvalidAccessKeyId"):
```
caarlos0: why can't we just return `errors.Wrap("failed to upload", err)`?

internal/pipe/blob/blob.go
```diff
@@ -0,0 +1,61 @@
...
+
+// Default sets the pipe defaults
+func (Pipe) Default(ctx *context.Context) error {
+	for i := range ctx.Config.Blob {
+		blob := &ctx.Config.Blob[i]
+		if blob.Bucket == "" {
+			continue
+		}
+		if blob.Provider == "" {
+			blob.Provider = "azblob"
```
caarlos0: I think it's best to leave this as empty by default, and just err earlier `if bucket != "" && provider == ""`

C123R: yeah, you are right, even setting bucket to "" is not good as Go CDK gonna complain anyway.
`azureblob.OpenBucket: containerName is required` or `s3blob.OpenBucket: bucketName is required`

I would go for something like this then:
```bash
	if blob.Bucket == "" || blob.Provider == "" {
			return fmt.Errorf("bucket or provider cannot be empty")
	}

```
Agree?

caarlos0: agree!

C123R: resolve in 59c8e4

internal/pipe/blob/openbucket.go
C123R: resolve in 59c8e4

internal/pipe/blob/openbucket.go
C123R: resolve in 59c8e4

caarlos0:
LGTM, is there anything you else you want to do in this PR or should we merge it in?

C123R:
> LGTM, is there anything you else you want to do in this PR or should we merge it in?

yeah, let's do it. I think we are good to go 👍

caarlos0:
thanks!

caarlos0:
I wonder if we can replace the s3 pipe with this one now... what do you think?

C123R:
> I wonder if we can replace the s3 pipe with this one now... what do you think?

yeah, even I think so 👍, I have added support for the folder as well. But still there are few things which blob are not supporting over S3 implementation:
- Profile ( I would prefer env, as it is CI friendly)
- ACL
- minio

Let's deprecate S3 and try to implement all features of S3 here in Blob.

caarlos0:
yeah, profile seems like something that will for sure hurt us... I rather use ENVs too..

the ACL is tricky, maybe the user can change the ACL of the entire folder instead? I guess that would work too, right? cc/ @mgrachev 

minio is used for integration tests, as it is s3-compatible, so it should be fine...

C123R:
> the ACL is tricky, maybe the user can change the ACL of the entire folder instead? I guess that would work too, right? cc/ @mgrachev

yeah, need to check this.. as I know there were some issues with Go Cloud for the ACL. I will check this if I find something will create new PR with updates.

Btw @caarlos0 Thanks for approving it and being so nice 💯.

caarlos0:
> yeah, need to check this.. as I know there were some issues with Go Cloud for the ACL. I will check this if I find something will create new PR with updates.

awesome! :D 

> Btw @caarlos0 Thanks for approving it and being so nice 💯.

hehe thank you for the all the work 💯

PR #1039, opened 2019-06-06
docs: fix snap install command

felicianotech:
This PR simply fixes the snapcraft install command listed on the website.

1. Almost always the user will need to prefix the command with `sudo`.
2. The GoReleaser snap is a classically confined snap which requires the flag `--classic` to install.


caarlos0:
thanks!

PR #1054, opened 2019-06-19
Add Cross compile with CGO and GoReleaser

robdefeo:
This commit will help users of GoReleaser to understand how to cross compile using CGO.


caarlos0:
thanks!

PR #1046, opened 2019-06-11
feat: Add support for specifying SSE on S3 objects

rafaelmagu:
If applied, this commit will add support for specifying Server-Side Encryption (SSE) on uploaded S3 objects, both with default encryption and with custom-specified keys.

**Note**: I added this to the now-deprecated `s3` handler. I could not find a good example of how to set SSE for the `blob` handler.

There is currently no support for SSE on uploaded S3 objects.

Closes #1045

PR #1056, opened 2019-06-25
feat: blob kms support

caarlos0:
refs #1046


pkg/config/config.go
```diff
@@ -292,20 +292,21 @@ type Before struct {
...
 
 // Blob contains config for GO CDK blob
 type Blob struct {
-	Bucket   string
-	Provider string
-	Folder   string
+	Bucket   string   `yaml:",omitempty"`
+	Provider string   `yaml:",omitempty"`
+	Folder   string   `yaml:",omitempty"`
+	KMSKey   string   `yaml:",omitempty"`
```
rafaelmagu: @caarlos0 I'm keen to test it, but I'm using `AES256` and not a custom KMS key. I had a look at the code but don't see a way to specify it.

rafaelmagu: Also bumped into another issue. `blob` provider does not support the default authentication mechanisms provided by the AWS SDK, so IAM roles have no effect. That's a no-go for us at this stage, as our deployment environment does not have any long-living credentials available.

caarlos0: got it... let me think about it some more ;)

caarlos0: thanks for the review btw

C123R: @caarlos0  Sorry for not being so active on this PR(vacation time). That's a really nice feature to have.
As I know,  S3/GCS/AZBLOB supports Server-side encryption just providing your key or respective KMS key reference. But go cloud is kind of blocker for this and really liked the way you wrapped it up by uploading encrypted data by our own. 👍 
I will test it as soon as possible.

@rafaelmagu 
> Also bumped into another issue. `blob` provider does not support the default authentication mechanisms provided by the AWS SDK, so IAM roles have no effect.

Are you talking about the IAM Roles for EC2 Instances? I have to check if there is a way to get access keys from such IAM role to make AWS calls.

C123R: I guess [this](https://github.com/aws/aws-sdk-go/blob/master/aws/credentials/ec2rolecreds/ec2_role_provider.go#L86) is how AWS GO SDK reads credentials from EC2 instances. 
@caarlos0 do you think it makes sense to have something like this? We can define an order for retrieving credentials in our environment check function.

For instance:
- IAM Role
- Environment Variable.

caarlos0: yeah I think it makes sense...

on the s3 repo, we read from role, env and ~/.aws/profile if I'm not mistaken...

if this is doable with gocloud too it would be awesome!

caarlos0: > As I know, S3/GCS/AZBLOB supports Server-side encryption just providing your key or respective KMS key reference. But go cloud is kind of blocker for this and really liked the way you wrapped it up by uploading encrypted data by our own. 👍

thanks! Was the only way I found to do it...  haha

caarlos0: > Sorry for not being so active on this PR(vacation time). Btw this is a really nice feature to have.

no problem at all! :)

C123R: @caarlos0 
> yeah I think it makes sense...
> 
> on the s3 repo, we read from role, env and ~/.aws/profile if I'm not mistaken...
> 
> if this is doable with gocloud too it would be awesome!

I had a quick look in gocloud authentication. It's using AWS SDK which means it supports authentication mechanisms(IAM Roles, credentials file and ENV variables)by default.

Reason why it's **not** working now: 

As we discussed earlier, i was forcing the blob provider to use only environment variable as of now: 
`case "s3":
		return checkEnv("AWS_ACCESS_KEY", "AWS_SECRET_KEY", "AWS_REGION")`

[reference](https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/blob/util.go#L16)

I just tried skipping this env variable check and used aws credentials file instead. It worked, did not test with IAM Role though, but I should work as well 👍 

We could skip this check but then we need to deal with errors from gocloud something like this:

`⨯ release failed after 8.66s error=Blob: failed to publish artifacts: failed to close Bucket writer: blob (code=Unknown): NoCredentialProviders: no valid providers in chain. Deprecated.
	For verbose messaging see aws.Config.CredentialsChainVerboseErrors`

**Worst in case of Azure**: if you don't specify AZURE_STORAGE_KEY gocloud won't complain and at the end it will throw Azure HTTP response for ResourceNotFound.

Any ideas, suggestion :) ?

caarlos0: maybe just wrap the error with something like:

```go
errors.Wrapf(err, "failed to upload, check credentials and access to bucket %s", bucket)
```

?

caarlos0: not very descriptive, but would in theory be correct 🤔

rafaelmagu: As much as possible, the CDK should trust the chain provided by the AWS SDK. If it has support for that already, then this becomes a lot easier to test and covers more deployment scenarios.

If you can wrap the error but still surface the exception class to the user, that _should_ be enough context.

C123R: @caarlos0 @rafaelmagu

Yeah. I will come up with new PR for this. 👍

caarlos0: so, the way this is pr is implemented won't work, correct?

caarlos0: afaik there is no option for that https://gocloud.dev/howto/secrets/crypt/

rafaelmagu: So I managed to test this and it works with aliases, but no `AES256` support. For now it's good to go.

caarlos0: awesome, will merge to master :)

eventually I'll need to figure out a way to cover this package with tests... several things are not covered... don't know how to do it yet though :(

caarlos0:
thanks @C123R and @rafaelmagu  for helping on this!

PR #1060, opened 2019-06-25
fix: non-templateable field as default build.id

### LINKED ISSUE #1059: Templating in binary field of build
**Describe the bug**

Not sure if this is really a bug or simply a change in behavior between versions. However, in version 0.85.1 you were able to use templating for binary names.

Given the following yaml:
```yaml
builds:
- binary: "{{ .ProjectName }}_{{ .Tag }}"
```

In version 0.85.1, the binary ended up in the following path:
```
dist/darwin_386/terraform-provider-healthchecksio_v1.3.0
```

In version 0.110.0, the binary ends up in a path suggesting templating is no longer supported or broken:
```
dist/{{ .ProjectName }}_{{ .Tag }}_darwin_386/terraform-provider-healthchecksio_v1.3.0
```

I would prefer to keep the old style. Is there a way to configure the final dist path (including the directory name and binary name) in v 0.110.0 and above?

**To Reproduce**

```yaml
# This is an example goreleaser.yaml file with some sane defaults.
# Make sure to check the documentation at http://goreleaser.com
before:
  hooks:
    - go mod download
builds:
- binary: "{{ .ProjectName }}_{{ .Tag }}"
  env:
  - CGO_ENABLED=0
  - GO111MODULE=on
archives:
- replacements:
    darwin: Darwin
    linux: Linux
    windows: Windows
    386: i386
    amd64: x86_64
checksum:
snapshot:
  name_template: "{{ .Tag }}-next"
changelog:
  sort: asc
  filters:
    exclude:
    - '^examples:'
    - '^docs:'
    - '^test:'
```

**Expected behavior**

I expect the binary field to support templating as it did in version 0.85.1 or other option to replicate the old behavior. 

- OS: mac
- OS version: darwin
- GoReleaser Version: 0.110.0

caarlos0:
fixes #1059

### DIFF
--- a/internal/pipe/build/build.go
+++ b/internal/pipe/build/build.go
@@ -63,7 +63,7 @@ func buildWithDefaults(ctx *context.Context, build config.Build) config.Build {
 		build.Binary = ctx.Config.ProjectName
 	}
 	if build.ID == "" {
-		build.ID = build.Binary
+		build.ID = ctx.Config.ProjectName
 	}
 	for k, v := range build.Env {
 		build.Env[k] = os.ExpandEnv(v)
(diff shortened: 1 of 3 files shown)


caarlos0:
this may be a breaking change if the user has several build without any ID...

PR #1061, opened 2019-06-27
fix: blob provider authentication issues for default aws mechanism

C123R:
fixes issues with the default AWS authentication mechanism.


internal/pipe/blob/util.go
```diff
@@ -1,41 +1,9 @@
...
-	}
-
-	if len(missingEnv) != 0 {
-		return fmt.Errorf("missing %v", strings.Join(missingEnv, ","))
-	}
-	return nil
-}
-
 // Check if error contains specific string
 func errorContains(err error, subs ...string) bool {
```
caarlos0: maybe we can merge this file with blob.go since now it only has one function ?

caarlos0:
awesome, thank you for all the effort! :)

PR #1062, opened 2019-06-28
docs: fix typos in blob doc

C123R:
fixing typos in blob authentication doc.


caarlos0:
I didn't catch that before, thanks!

caarlos0:
0a8478b should prevent that in the future ;D

PR #1074, opened 2019-07-08
Fixed token required if release is disabled

### LINKED ISSUE #1073: Fails if no GitHub/GitLab token and release disabled
Since the introduction of GitLab release in #1038 either a GitHub or GitLab token is required for the build to succeed even if releases are disabled via

```
release:
  disable: true
```

The error message is
```
release failed after 6.92s error=GitHub/GitLab Releases: failed to publish artifacts: client is not yet implemented:
```

fheidenreich:
If applied, this commit will fix #1073

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -2,7 +2,6 @@
 package client
 
 import (
-	"fmt"
 	"os"
 
 	"github.com/goreleaser/goreleaser/pkg/config"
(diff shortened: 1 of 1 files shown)


mavogel:
Thank you for pointing this out. I wrongly *assumed* everyone would at least do a release. We should probably add some tests for a whole pipeline walkthrough so this pops out earlier. My bad

caarlos0:
thanks!

PR #1077, opened 2019-07-12
Fix opencollective links in README

rkravchik:
Subj


caarlos0:
good catch, no idea how it got that way

thanks!

PR #1079, opened 2019-07-12
feat: write brew formulae before publishing

dawidd6:
I think it would be nice to be able to take a look at generated Homebrew Formulae without running actual publish task, for debugging purposes at least.

This commit changes the place where and when writing Formulae to file happens (before every publishing task + check if publishing is disabled).

Why is this change being made?
To be able to debug Formulae before actual publishing.

### DIFF
--- a/internal/pipe/brew/brew.go
+++ b/internal/pipe/brew/brew.go
@@ -37,6 +37,53 @@ func (Pipe) String() string {
 	return "homebrew tap formula"
 }
 
+func getFilters(brew config.Homebrew) []artifact.Filter {
+	var filters = []artifact.Filter{
+		artifact.Or(
+			artifact.ByGoos("darwin"),
+			artifact.ByGoos("linux"),
+		),
+		artifact.ByFormats("zip", "tar.gz"),
+		artifact.ByGoarch("amd64"),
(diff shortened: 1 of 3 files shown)


caarlos0:
can I have a different take on this?

caarlos0:
AFAIK it already works this way.

if you set `brew` `skip_upload` to `true` it will write the file to `dist` but will not upload it...

caarlos0:
unless you pass the `--skip-publish` flag, which I guess is your case...

caarlos0:
I understand your pain, but I'm not sure this is a good abstraction, let me think about it for a while :)

thanks for the PR and bringing this up anyway 👍

dawidd6:
Well, if I don't pass `--skip-publish`, then I get this:
```sh
error=missing both GITHUB_TOKEN and GITLAB_TOKEN
```

My point here is that I want to get homebrew formula without publishing, while not being the project owner, but just a regular contributor.

I mean it would be nice to also have the formula written to `dist/` as well. The reason why I brought this up is that I wanted to test if changing deprecated `brew` to `brews` will generate formula supporting linux and it did (after applying above patch). So I raised this PR, cause it can be useful IMO :)

PR #1083, opened 2019-07-16
fixes env usage description for docker images

mavogel:
### If applied, this commit will...
fix the docs for using environment variables in the `.goreleaser.yml`

###  Why is this change being made?
I tested as described in the docs but the map does not have a key for `.ENV` but has for `.Env` as the case matters

### Provide links to any relevant tickets, URLs or other resources
See context struct:  https://github.com/goreleaser/goreleaser/blob/master/pkg/context/context.go#L55


caarlos0:
thanks!

PR #1038, opened 2019-06-04
feat: add gitlab for releases

mavogel:
Closely to #809 but with the latest `github.com/xanzy/go-gitlab` client. I'll try to keep the changes/adaptions small (cuz refactorings will be done in #807) and probably re-use some of the code of the mentioned PR. 

- If applied, this commit will add the possibility to release to `gitlab`
- Why is this change being made?  To support `gitlab`
- Provide links to any relevant tickets, URLs or other resources: my playground https://gitlab.com/mavogel/release-testing/releases

Open questions for @caarlos0 
- should it be possible to release to either gitlab or github and the validation will be done in config parsing?
- should it be possible in the first step to be only able to release to brew and scoop if it's a github release and then later after #807 enable it for gitlab as well? This would keep the changes in this PR smaller

Left todo (incl. tests)
- [x] add factory to client to load depending on tokenType
- [x] adapt release pipe according
- [x] add yaml parsing and errors for invalid
- [x] skip brew and scoop by when gitlab is used for releasing
- testing with demo repo and fixing errors
   - [x] fixed changelog generation
   - [x] testing without given owner and repo and default extraction
   - [x] testing with private gitlab
   - [x] testing with docker images and gitlab
   - [x] test with gitlab ci: [release](https://gitlab.com/mavogel/release-testing/-/releases) 0.3.1 and [job](https://gitlab.com/mavogel/release-testing/pipelines/65747428)
- [x] update docs ee4e299535e7c83e3d69b42b2f46e0f8fb1fcaea and d00105b3290442f0a206db41914ffae5b0dde948
- [x] merge latest master and resolve conflicts

PR #1084, opened 2019-07-16
feat: full gitlab support for brew and scoop

mavogel:
enabled the usage of gitlab for brew and scoop. This was still left open in #1038

I would like to use gitlab as repo for my code for releasing to brew and scoop as well

- #1038
- #110

## Steps
- [x]  add brew for public and private gitlab
- [x]  add scoop for public and private gitlab
- [x] update tests
- [x] update docs (brew, scoop, gitlab ci)

## open questions 
for @caarlos0  :)
- Am I right in the assumption that it is possible to *push* to brew and scoop with skipping the release pipe?
- I am not 100% happy by using the `Extra` field of `artifact.Extra["GitLabFileUploadHash"]` to pass around the hash I need to check if the file was already uploaded to the repo to reuse it in the brew/scoop pipe. (Yes github makes it easier and different) Would you suggest a new variable for this?


mavogel:
ping @caarlos0 for some clarifications :) Then I'll move on in the next days

internal/pipe/brew/brew.go
```diff
@@ -222,12 +216,27 @@ func dataFor(ctx *context.Context, cfg config.Homebrew, artifacts []artifact.Art
...
+					ctx.Config.Release.GitHub.Name,
+				)
+			case context.TokenTypeGitLab:
+				//https://gitlab.com/mavogel/release-testing/uploads/22e8b1508b0f28433b94754a5ea2f4aa/release-testing_0.3.7_Darwin_x86_64.tar.gz
+				cfg.URLTemplate = fmt.Sprintf(
+					"%s/%s/%s/releases/uploads/%s/{{ .ArtifactName }}",
+					ctx.Config.GitLabURLs.Download,
+					ctx.Config.Release.GitLab.Owner,
+					ctx.Config.Release.GitLab.Name,
+					artifact.Extra["GitLabFileUploadHash"],
```
caarlos0: maybe use https://godoc.org/github.com/goreleaser/goreleaser/internal/artifact#Artifact.ExtraOr with some fake string?

in theory it will never happen, but will be easier to debug in case it does (better have a url with `fake-hash-should-not-happen` than just `nil` 🤔

or maybe handle it and fail earlier

caarlos0:
> * Am I right in the assumption that it is possible to _push_ to brew and scoop with skipping the release pipe?

shouldn't be I think... if it allows it, its a bug we should probably handle 🤔

> * I am not 100% happy by using the `Extra` field of `artifact.Extra["GitLabFileUploadHash"]` to pass around the hash I need to check if the file was already uploaded to the repo to reuse it in the brew/scoop pipe. (Yes github makes it easier and different) Would you suggest a new variable for this?

I think the extra thing is OK for now...

mavogel:
Unfortunately, adding the `GitlabUploadHash` to the `artifact` does not work properly. I think because a copy is passed and so `artifact.SetUploadHash(fileUploadHash)` is not reflected. I tried it with pointer but it did not work either.

Somehow I do not fully get what happens here: https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/release/release.go#L111. Why `artifact := artifact`?

Good news are, that the brewfile generation works (https://gitlab.com/mavogel/homebrew-tab/blob/master/Formula/release-testing-program.rb) and scoop will be a piece of cake once the hash-passing-issue is solved :tada:

caarlos0:
> Somehow I do not fully get what happens here: https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/release/release.go#L111. Why `artifact := artifact`?

its to force a new variable, if you remove that line and run `make lint` it will give you a better explanation...

> Good news are, that the brewfile generation works (https://gitlab.com/mavogel/homebrew-tab/blob/master/Formula/release-testing-program.rb) and scoop will be a piece of cake once the hash-passing-issue is solved 🎉

awesome!

caarlos0:
> Unfortunately, adding the `GitlabUploadHash` to the `artifact` does not work properly. I think because a copy is passed and so `artifact.SetUploadHash(fileUploadHash)` is not reflected. I tried it with pointer but it did not work either.

and with the `Extra` thing?

mavogel:
> and with the Extra thing?

Tried it but the key in the map is not set and I get a panic. I assume that the artifact in the `context` cannot be modified cuz we pass it as a copy in the release pipe (where I actually set the new attribute). I still don't get it completely but maybe I'll introduce another var in the context to store the hash. But this is more a workaround. Any ideas how to do this in an elegant way?

caarlos0:
hmm, maybe we need to change this:

```go
type Artifacts struct {
	items []Artifact
	lock  *sync.Mutex
}
```

to

```go
type Artifacts struct {
	items []*Artifact
	lock  *sync.Mutex
}
```

?

Probably pass a pointer everywhere as well 🤔

mavogel:
> Probably pass a pointer everywhere as well

Yep did the trick 👍 Will add scoop and merge with master ;)

mavogel:
### Update
- Scoop got successfully added
- Verified the release to private brew and scoop bucket :tada:

internal/artifact/artifact.go
```diff
@@ -67,13 +67,20 @@ func (t Type) String() string {
...
-	Goarm  string
-	Type   Type
-	Extra  map[string]interface{}
+	Name       string
+	Path       string
+	Goos       string
+	Goarch     string
+	Goarm      string
+	Type       Type
+	UploadHash string
```
caarlos0: can we move this into `Extra`?

mavogel: sure, I can do that :) Makes sense because it's a gitlab only thing. Maybe there will be more in the future for other vcs.

caarlos0: Yeap... its not documented anywhere, but my idea was to use extra for things that not common between all things...

caarlos0:
@mavogel can I fix the conflicts or you rather do it yourself? anyway is fine by me :D 

(just trying to make myself useful here hehe)

mavogel:
no worries, already on it...

mavogel:
when I check the report I don't get why there was a negative change in `internal/client/github.go` 🤔

caarlos0:
> when I check the report I don't get why there was a negative change in `internal/client/github.go` 🤔

me neither, maybe some bug 🤔

mavogel:
> awesome work @mavogel! Thanks so much!

You're welcome :) This tool simplifies our automated release process so much and I learned quite some stuff on the way. So thanks to you!

> what do you say we merge this and leave it on master for a few days?

Fine for me. Merge it and then the others can continue based on the latest master :tada:

caarlos0:
Awesome, merged!

Hopefully people help testing :D

Issue #1097, opened 2019-08-02
Docker build order


tdmanv:
**Is your feature request related to a problem? Please describe.**
I build multiple docker containers. One of them is used as the FROM field for others

**Describe the solution you'd like**
If `--parallelism=1`, then build the docker images in the order specified in the yaml. This will let me choose the order of images.

**Describe alternatives you've considered**
I could potentially build the base image in a prehook, but I'd prefer to use the existing docker support.

**Additional context**
This is the project: https://github.com/kanisterio/kanister/blob/master/.goreleaser.yml#L42

caarlos0:
this was fixed on #1100 

closing.

PR #1100, opened 2019-08-05
fix: honor order of functions in semerrgroup w/ parallelism == 1

tdmanv:
## Overview
* Honor total order of methods added to `semerrgroup.Group` when `parallelism == 1`
* `semerrgroup.Group.Go(...)` will block if the number of concurrently executing goroutines in the group is at the maximum. 

## Reason

This change is being made to so that project authors may include dependencies between build phases by ordering them in their project's `goreleaser.yml`

## Issues
* https://github.com/goreleaser/goreleaser/issues/1097

### DIFF
--- a/internal/semerrgroup/sem.go
+++ b/internal/semerrgroup/sem.go
@@ -52,6 +52,9 @@ type serialGroup struct {
 
 // Go execs one function at a time.
 func (s *serialGroup) Go(fn func() error) {
+	if s.err != nil {
+		return
+	}
 	if err := fn(); err != nil {
 		s.err = err
 	}
--- a/internal/semerrgroup/sem.go
+++ b/internal/semerrgroup/sem.go
@@ -2,11 +2,7 @@
 // size, so you can control the number of tasks being executed simultaneously.
 package semerrgroup
 
-import (
-	"sync"
-
-	"golang.org/x/sync/errgroup"
-)
+import "golang.org/x/sync/errgroup"
 
 // Group is the Semphore ErrorGroup itself
 type Group interface {
@@ -51,22 +47,17 @@ func (s *parallelGroup) Wait() error {
 var _ Group = &serialGroup{}
 
 type serialGroup struct {
-	lock sync.Mutex
-	err  error
+	err error
 }
 
 // Go execs one function at a time.
 func (s *serialGroup) Go(fn func() error) {
-	s.lock.Lock()
-	defer s.lock.Unlock()
-	if err := fn(); err != nil && s.err == nil {
+	if err := fn(); err != nil {
 		s.err = err
 	}
 }
 
 // Wait waits for Go to complete and returns the first error encountered.
 func (s *serialGroup) Wait() error {
-	s.lock.Lock()
-	defer s.lock.Unlock()
 	return s.err
 }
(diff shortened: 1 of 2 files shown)


caarlos0:
maybe just skip the errgroup entirely when `parallelism == 1` is clearer?

tdmanv:
> maybe just skip the errgroup entirely when `parallelism == 1` is clearer?

Happy to, but please clarify what you mean. Are you thinking this should be done outside the `internal/semerrgroup` package? Should every caller be modified to first check the parallelism before using a Group?

caarlos0:
I was thinking maybe have `semerrgroup .Group` as an interface, and on `semerrogroup.New` check if parallelism is 1 and return the "no semaphore" implementation or the current one...

what do you think?

tdmanv:
Makes sense to me! Let me put that together

internal/semerrgroup/sem.go
```diff
@@ -30,6 +44,29 @@ func (s *Group) Go(fn func() error) {
...
+
+type serialGroup struct {
+	lock sync.Mutex
+	err  error
+}
+
+// Go execs one function at a time.
+func (s *serialGroup) Go(fn func() error) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
```
caarlos0: do we need to lock here?

since we don't have any errgroup or whatever, it seems like it won't really do anything...

tdmanv: it serializes access to err in case this is called in parallel. if it's not called in parallel anywhere, I can remove this. What do you think?

caarlos0: you can remove it yes... :)

internal/semerrgroup/sem.go
```diff
@@ -30,6 +40,24 @@ func (s *Group) Go(fn func() error) {
...
+
+var _ Group = &serialGroup{}
+
+type serialGroup struct {
+	err error
+}
+
+// Go execs one function at a time.
+func (s *serialGroup) Go(fn func() error) {
+	if err := fn(); err != nil {
```
caarlos0: maybe here we should check if `s.err != nil` first?

if so, ignore the subsequent runs? 

The idea is to make the behavior similar to errgroup which afaik works that way 🤔

tdmanv: done

tdmanv:
@caarlos0 is there any way to retrigger this? the failure looks unrelated
```

909--- FAIL: TestUploadCustomBucketID (5.20s)
910    s3_test.go:228: waiting for minio to be healthy
911    s3_test.go:225: minio is healthy
912    s3_test.go:195: creating test bucket
913    require.go:794: 
914        	Error Trace:	s3_test.go:202
915        	            				s3_test.go:139
916        	Error:      	Received unexpected error:
917        	            	RequestError: send request failed
918        	            	caused by: Put http://127.0.0.1:39448/test: http: server closed idle connection
919        	Test:       	TestUploadCustomBucketID
920FAIL
```

caarlos0:
thanks for the contribution! will release this shortly :)

tdmanv:
thank you for the reviews!

PR #1105, opened 2019-08-08
fix: &base always not nil

exfly:
Is it a bugs?

```go
type A interface {
	A()
}

func TestA(t *testing.T) {
	var a A
	assert.NotEqual(t, nil, &a)
	a = nil
	assert.NotEqual(t, nil, &a)
}
```
// this test is passed. So this pr is working?

### DIFF
--- a/internal/client/github.go
+++ b/internal/client/github.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


caarlos0:
thanks!

caarlos0:
and sorry for the delay :)

PR #1113, opened 2019-08-13
feat: support arm for linuxbrew

### LINKED ISSUE #1052: linuxbrew: support arm
Continue #927 adding ARM support.

mgrachev:
Hi there!

I've added support architecture arm for linuxbrew formulas.
I've used that formula as example https://github.com/Homebrew/linuxbrew-core/blob/c7f81028802006ba26b5e079556d1617b38ca495/Formula/openssl.rb#L58

/cc @caarlos0 @sethkor What do you think about it?

Close #1052

### DIFF
--- a/internal/pipe/brew/template.go
+++ b/internal/pipe/brew/template.go
@@ -16,6 +16,8 @@ type templateData struct {
 	CustomBlock      []string
 	MacOS            downloadable
 	Linux            downloadable
+	Arm              downloadable
+	Arm64            downloadable
 }
 
 type downloadable struct {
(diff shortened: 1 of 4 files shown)


internal/pipe/brew/template.go
```diff
@@ -36,15 +38,31 @@ class {{ .Name }} < Formula
...
+    {{- if .DownloadStrategy }}, :using => {{ .DownloadStrategy }}{{- end }}
     sha256 "{{ .MacOS.SHA256 }}"
     {{- end }}
   elsif OS.linux?
-    {{- if .Linux.DownloadURL }}
-    url "{{ .Linux.DownloadURL }}"
-	{{- if .DownloadStrategy }}, :using => {{ .DownloadStrategy }}{{- end }}
-    sha256 "{{ .Linux.SHA256 }}"
-    {{- end }}
+    if Hardware::CPU.arm?
```
caarlos0: maybe add an `if or .Arm.DownloadURL .Arm64.DownloadURL` to prevent printing this if when no arm/arm64 build exist?

mgrachev: Done

caarlos0:
Awesome, thanks!

dawidd6:
It does not seem to work, see:
- https://github.com/jesseduffield/homebrew-lazydocker/blob/master/lazydocker.rb
- https://github.com/goreleaser/homebrew-tap/blob/master/Formula/goreleaser.rb

PR #1114, opened 2019-08-13
feat: gitea release support

### LINKED ISSUE #1106: Gitea release support
I want to upload the build artifacts to my [Gitea](https://gitea.io/) instance, but currently only GitHub and gitlab.com are supported.

I want to be able to specify [my Gitea instance](https://git.dtluna.net/) as the destination for uploading artifacts in `release` section.

dtluna:
Add support for Gitea in release pipe

I (and probably many others) want to create releases and upload artifacts to their Gitea instances.

Gitea: https://gitea.io/en-us/
Gitea Go SDK: https://godoc.org/code.gitea.io/sdk/gitea

closes #1106


dtluna:
Please tell me if you would like me to add more tests for the features I introduced!

internal/client/gitea.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+type giteaClient struct {
+	client *gitea.Client
+}
+
+// NewGitea returns a gitea client implementation
+func NewGitea(ctx *context.Context) (Client, error) {
+	token := ctx.Token
+	client := gitea.NewClient(ctx.Config.Release.Gitea.InstanceURL, token)
```
mavogel: is there also an option for `SkipTlsVerify` like the other clients have?

internal/client/gitea.go
```diff
@@ -0,0 +1,161 @@
...
+// or updates the file if it exists
+func (c *giteaClient) CreateFile(
+	ctx *context.Context,
+	commitAuthor config.CommitAuthor,
+	repo config.Repo,
+	content []byte,
+	path,
+	message string,
+) error {
+	// doesn't have to be implemented for Gitea
```
mavogel: We would need this method for `brew` and `scoop`. I just checked the sdk has only has a `GetFile` method, whereas the API offers CRUD. Filed an issue at gitea: https://gitea.com/gitea/go-sdk/issues/187

internal/client/gitea.go
```diff
@@ -0,0 +1,161 @@
...
+	return release, nil
+}
+
+func (c *giteaClient) getExistingRelease(owner, repoName, tagName string) (*gitea.Release, error) {
+	releases, err := c.client.ListReleases(owner, repoName)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, release := range releases {
```
mavogel: could become quite expensive for a lot of repositories. Probably using https://gitea.com/gitea/go-sdk/src/branch/master/gitea/release.go#L28 suites better

pkg/config/config.go
```diff
@@ -41,6 +41,20 @@ func (r Repo) String() string {
 	return r.Owner + "/" + r.Name
 }
 
+// GiteaRepo represents repo hosted on Gitea instance
+type GiteaRepo struct {
```
mavogel: Fine for me atm, because I guess we will unify the repo in another PR to one  *repo* struct

internal/client/gitea.go
dtluna: Maybe if we use https://godoc.org/code.gitea.io/sdk/gitea#NewClientWithHTTP

internal/client/gitea.go
dtluna: So we should wait until they introduce the CRUD features into the SDK?

internal/client/gitea.go
dtluna: You need a release id for that. Where would you get it from?

pkg/config/config.go
dtluna: We could, but then the users would have to know how to construct the API URLs for their instance and then in the client creation code we would have to parse that URL to pass to 
https://godoc.org/code.gitea.io/sdk/gitea#NewClient

internal/client/gitea.go
mavogel: Ah I see, you're right :) Well [repoListReleases](https://try.gitea.io/api/swagger#/repository/repoListReleases) also has pagination, but I assume adding no or a negative value gives you all releases right?

dtluna: I will look into it!

dtluna: Well, the SDK doesn't allow for pagination https://godoc.org/code.gitea.io/sdk/gitea#Client.ListReleases

internal/client/gitea.go
mavogel: Yes and then make another PR for brew and scoop. Could you add that those 2 are not supported yet for gitea please

internal/client/gitea.go
mavogel: okay I see, the sdk would need some updates. Then we keep it the way you implemented it

internal/client/gitea.go
dtluna: Sure, I will

pkg/config/config.go
mavogel: Well for public github/gitlab we defined the default urls but if a user host its enterprise/private one, then he/she should know how to build the full url. That's why we have the `api` and `download` fields. 

Imagine they host their vcs behind a LoadBalancer or Proxy which expects the api at `/foo/bar` and rewrites to `/api/v1/` where the api acually ist. We don't know the user's configuration, so we give them their freedom to completely define it

internal/client/gitea.go
mavogel: Yes, that's what I meant

pkg/config/config.go
dtluna: So you propose to parse the API URL, say `https://gitea.dtluna.net/api/v1` and construct `https://gitea.dtluna.net` for [gitea.NewClient](https://godoc.org/code.gitea.io/sdk/gitea#NewClient) manually in the code?

mavogel: Yes. A good point for the validation could be parsing the url. ATM we only do it at `client/github.go`. What do you think @caarlos0 ?

caarlos0: sounds good @mavogel :)

dtluna:
@mavogel OK, I am finally done with unit tests

www/content/environment.md
```diff
@@ -47,7 +47,7 @@ If none are set, they default to GitHub's public URLs.
 
 ## GitLab Enterprise or private hosted
 
-You can use GoReleaser with GitHub Enterprise by providing its URLs in
+You can use GoReleaser with GitLab Enterprise by providing its URLs in
```
mavogel: Thanks for catching this one 👍

mavogel:
nice @dtluna! The unit test look very neat :tada: I'll update the gitlab ones as well in another PR.

I just tested you code locally by setting up local gitea and using my dummy [release-testing](https://gitlab.com/mavogel/release-testing#testing-against-gitea) project. I encountered errors while uploading the artifacts to the release. 
```console
 • failed to upload artifact, will retry artifact=release-testing_0.3.34_Windows_i386.tar.gz try=9
         • uploading to release      file=dist/release-testing_0.3.34_Darwin_i386.tar.gz name=release-testing_0.3.34_Darwin_i386.tar.gz
         • failed to upload artifact, will retry artifact=release-testing_0.3.34_Darwin_i386.tar.gz try=9
         • uploading to release      file=dist/release-testing_0.3.34_Windows_x86_64.tar.gz name=release-testing_0.3.34_Windows_x86_64.tar.gz
         • failed to upload artifact, will retry artifact=release-testing_0.3.34_Windows_x86_64.tar.gz try=9
   ⨯ release failed after 27.55s error=GitHub/GitLab/Gitea Releases: failed to publish artifacts: failed to upload checksums.txt after 10 retries: File type is not allowed: text/plain; charset=utf-8
```

- Did you test your code against the public gitea instance and your private one? At a first glimpse, something with the `File type is not allowed: text/plain; charset=utf-8`. 
- Feel free to use the dummy release project to debug locally and have fast **uploads**.
- @caarlos0 : I'am currently thinking of a more advanced intergation test setup where we can test this manual stuff in an automatic way. Let's discuss in slack later that week

dtluna:
@mavogel 
> File type is not allowed: text/plain; charset=utf-8

That is a setting on the specific instance, not an error in `goreleaser`

mavogel:
Alright good to know. I used the default settings, then we'd document the which properties should be set to make it work with goreleaser. Because people will run into this issue, when using the default. How did you configure it on your instance? Which leads me to the question:

> Did you test your code against the public gitea instance and your private one?

caarlos0:
> * @caarlos0 : I'am currently thinking of a more advanced intergation test setup where we can test this manual stuff in an automatic way. Let's discuss in slack later that week

agreed! I have a repo I do some tests some times, but it is not ideal for sure...

dtluna:
@mavogel I've tested the feature on my instance at git.dtluna.net
> How did you configure it on your instance?

```conf
ALLOWED_TYPES = image/jpeg|image/png|application/zip|application/gzip|application/x-gzip|application/x-gtar|application/x-tgz|application/x-compressed-tar|text/plain
```
```

**Comment by dtluna** [comment]
@mavogel I've tested it on my instance git.dtluna.net, but there's [no way currently to allow uploading checksums](https://github.com/go-gitea/gitea/issues/7882) short of allowing all types with `*/*`

**Comment by mavogel** [comment]
thank you @dtluna for poking this on the gitea side :+1:
- Well, as far as I can see only `checksums.txt` fails right? Other artifacts can be successfully uploaded?
- The other files work fine if the `ini` file has the correct `ALLOWED_TYPES` as you denoted in https://github.com/go-gitea/gitea/issues/7882#issuecomment-522071788 ?

They probably won't fix this in the future right, because the issue was closed without answering your question...

@caarlos0 any idea how to work around this in goreleaser? Maybe skip the upload of the checksums or put them into a `tgz` as well?

**Comment by dtluna** [comment]
@mavogel 

> Well, as far as I can see only checksums.txt fails right? Other artifacts can be successfully uploaded?

Correct.

> The other files work fine if the ini file has the correct ALLOWED_TYPES as you denoted in go-gitea/gitea#7882 (comment) ?

Yes.

**Comment by caarlos0** [comment]
> @caarlos0 any idea how to work around this in goreleaser? Maybe skip the upload of the checksum or compress it into a `tgz` as well?

we could skip it on gitea client... or warn that it might fail and point to the required config changes... 

not sure which option is better though

**Comment by mavogel** [comment]
> we could skip it on gitea client... or warn that it might fail and point to the required config changes... not sure which option is better though

It feels like we could skip it in the gitea client for `text/plain` release artifact files. So we would have the specific logic there with a todo to allow it if it will be possible in the future. And we would not have to touch the artifact compression logic.  What do you think @dtluna ?

**Comment by dtluna** [comment]
@mavogel I say it's completely a matter on the side of Gitea. Goreleaser is working fine as is, it's Gitea that needs proper configuration

**Comment by mavogel** [comment]
> it's Gitea that needs proper configuration

I fully agree on that.

So either we have a temporary workaround here and could make releases to gitea **or** we wait until they fix it on their side and sdk client. 

As we all would like to have gitea support, we could do a skip for the `c.client.CreateReleaseAttachment(...)` call in the `Upload` method for the `checksums.txt` file. And I was curious if this would be an option.

**Comment by dtluna** [comment]
@mavogel I don't understand why you think that any action is required. Checksums are uploaded by goreleaser on my instance. Everything works.

**Comment by mavogel** [comment]
Then I misunderstood. Sorry for that. I'll retest later on a local and public instance

**Comment by dtluna** [comment]
@mavogel You'll need to have
```conf
[attachment]
ALLOWED_TYPES = */*
```
in Gitea `app.ini` for now for uploads to work.

That's what I have at https://git.dtluna.net , you can test there (be sure to delete it later tho please :))

mavogel:
Successfully tested on the local instance :)
Two small things remain to fix then we can merge: 
- gitea [1.9.2](https://github.com/go-gitea/gitea/milestone/49?closed=1) will fix the filetype issue, but we'd add a note in the release docs that the `[attachment]` section in the `app.ini` should contain `text/plain; charset=utf-8` or `*/*`, because it's not in their defaults.
- the line breaks are not correctly interpreted in the release section. Or maybe another configuration issue on my side again. See screenshot:

Screenshot 2019-08-21 at 18 09 33

dtluna:
@mavogel I've added docs for proper Gitea configuration needed

mavogel:
Nice :) did those line break issues with multiple commits for a tag in the release section also occur on your instance?

dtluna:
@mavogel Yeah, I do here: https://git.dtluna.net/dtluna/testproject/releases
изображение

Also, what's up with these strange filenames?

dtluna:
@mavogel If you look at the contents of release description, it's not surprising it's one line:
изображение

mavogel:
> Also, what's up with these strange filenames (no testproject in names of archives)?

Probably you'd set a `project_name: testproject` line in `goreleaser.yaml`. Then it will work. But IMHO it should parse it from git. If not it's probably a bug and we'd investigate in another PR.

> it's not surprising it's one line

I also had issues with the gitlab markdown (see: https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/changelog/changelog.go#L59). Maybe gitea needs a special treatment as well?

dtluna:
> Probably you'd set a project_name: testproject line in goreleaser.yaml

I don't remember having to do that for any projects I used with goreleaser before though :confused: 

> Maybe gitea needs a special treatment as well?

It's not "special" treatment, that's how vanilla Markdown works, but yes, we need two newlines.

caarlos0:
Thanks! 🙌

Issue #1133, opened 2019-09-03
Docker images listed twice in release


Luit:
**Describe the bug**

Since v0.115.0 and caused by #1110 the Release body contains the full listing of Docker images twice.

**To Reproduce**

Release with some Docker images present (see this repo's releases).

**Expected behavior**

Each docker image should be mentioned only once. The change in #1110 caused duplicate DockerImage artifacts to be present after Publish, causing double listing. As the change made the artifacts all pointers, the artifact passed to the dockerPush function will be edited in place, as well as being added after the edit.

I'll make a pull request for this, which will copy the Artifact before adding it.

PR #1134, opened 2019-09-03
fix: copy into DockerImage Artifact

Luit:
Since #1110 the artifacts are passed by reference, causing the previous
code to edit the Artifact in place, _and_ add it to the list.

Fixes: #1133


caarlos0:
oh, thanks!

PR #1135, opened 2019-09-04
feat: go 1.13

caarlos0:
latest and greatest


dawidd6:
`golang.org/x/xerrors` needs to be updated to latest revision.

PR #1136, opened 2019-09-04
fix: install only docker-cli instead of docker

Luit:
While working on a GitLab CI setup I was copying the behavior of the Dockerfile and entrypoint.sh, so that I had control over the Go version used in CI, but still use goreleaser. When I was looking at the CI output I noticed there was 303 MB of packages installed when I did `apk add --no-cache bash curl docker git`. The list of packages included `containerd`, `libseccomp`, `docker-engine` and many more that seem unnecessary. When I changed it to `apk add --no-cache bash curl docker-cli git` it reduced this to 97 MB.

I think that's something the goreleaser/goreleaser Docker image can benefit from as well, so here's that same change. If there's a need for a full Docker engine instead of just the CLI, feel free to ignore this pull request.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM golang:1.12-alpine
 RUN apk add --no-cache bash \
                        bzr \
                        curl \
-                       docker \
+                       docker-cli \
                        git \
                        mercurial \
                        rpm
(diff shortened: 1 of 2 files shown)


caarlos0:
I'm not sure if we need the full docker to use dind (as we build docker images inside a docker image)... 🤔

Luit:
When doing `dind` in Gitlab (my use-case) and I think with this project's own setup in Travis, it runs as a service, so the daemon is a different container. All you need in the build container is the client.

caarlos0:
Awesome!

Thanks for the PR @Luit 🙏

Issue #1148, opened 2019-09-18
Can't combine brew and multiple goarm versions


SimonBarendse:
**Describe the bug**

When building multiple `goarm` versions and releasing on brew, an error is returned:
```
homebrew tap formula: failed to publish artifacts: one tap can handle only 1 linux and 1 macos archive
```

**To Reproduce**
We encounter this issue when releasing with [this `.goreleaser.yml`](https://github.com/secrethub/secrethub-cli/blob/v0.28.0/.goreleaser.yml). I expect anything using brew and having multiple goarm versions returns this error.

**Expected behavior**
I'd expect either one of these things:
1. Nothing is added to the brew formula for ARM architectures (as was the case in goreleaser `v0.117` and earlier).
2. An entry is added to the brew formula for each of the goarm versions and the formula determines which to use.

**Environment**
GoReleaser Version: `v0.118.0` (works as expected in `v0.117.2`)

**Additional context**
The error first occurs after #1140

The error is returned [here](https://github.com/goreleaser/goreleaser/blob/v0.118.0/internal/pipe/brew/brew.go#L278)

SimonBarendse:
I can't add it anymore, but this should be tagged [![](https://i.imgur.com/JflcH01.png)](../labels/bug)

mavogel:
Thanks for pointing out @SimonBarendse. We will keep track of the issue in #1146

Issue #1146, opened 2019-09-16
error=homebrew tap formula: failed to publish artifacts: one tap can handle only 1 linux and 1 macos archive


gamelife1314:
**Describe the bug**

error=homebrew tap formula: failed to publish artifacts: one tap can handle only 1 linux and 1 macos archive

[travis build error information](https://travis-ci.org/bigfile/bigfile/builds/585422830)

**To Reproduce**

Nothing changes, it goes wrong suddenly.

> .goreleaser.yml

```yaml
project_name: bigfile

env:
  - GO111MODULE=on

before:
  hooks:
    - go mod tidy

builds:
  -
    id: bigfile
    main: ./artisan/bigfile.go
    binary: bigfile
    goos:
      - darwin
      - freebsd
      - linux
      - windows
    goarch:
      - 386
      - amd64
      - arm
      - arm64
    goarm:
      - 6
      - 7

archives:
  -
    format_overrides:
      - goos: windows
        format: zip
    files:
      - LICENSE
      - README.md

checksum:
  name_template: "{{ .ProjectName }}_checksums.txt"
  algorithm: sha256

changelog:
  sort: asc
  filters:
    exclude:
      - '^docs:'
      - '^test:'
      - Merge pull request
      - Merge branch

dockers:
    -
      goos: linux
      goarch: amd64
      image_templates:
        - "bigfile/bigfile:latest"
        - "bigfile/bigfile:{{ .Tag }}"
        - "bigfile/bigfile:v{{ .Major }}"
      skip_push: false
      dockerfile: Dockerfile
      binaries:
        - bigfile
      build_flag_templates:
        - "--label=org.label-schema.schema-version=1.0"
        - "--label=org.label-schema.version={{.Version}}"
        - "--label=org.label-schema.name={{.ProjectName}}"
        - "--label=com.github.actions.name={{.ProjectName}}"
        - "--label=com.github.actions.description=a file transfer system that supports http, rpc and ftp protocols"
        - "--label=com.github.actions.icon=terminal"
        - "--label=com.github.actions.color=blue"
        - "--label=repository=https://github.com/bigfile/bigfile"
        - "--label=homepage=http://bigfile.site"
        - "--label=maintainer=bigfilefu <bigfilefu@gmail.com>"
brews:
  -
    github:
      owner: bigfile
      name: homebrew-tap
    folder: Formula
    homepage:  https://bigfile.site
    caveats: start mysql before starting bigfile
[...]

PR #1153, opened 2019-09-23
fix: brew multiple archives on one tap

mavogel:

Hotfixes #1146 for backward compatibility

Atm is possible to build multiple `arm` artifacts with different `goarm` versions (v5, v6 and v7) `brew` can only handle one arm version for the formula, so if multiple versions are present brew will use the latest.

I'm still looking for a better error msg of 
```go
var ErrMultipleArchivesSameOS = errors.New("one tap can handle only 1 linux and 1 macos archive")
```

- #1148 
- #1140


caarlos0:
I'm not sure about this 🤔 

its not like arm7 replaces arm6... its more or less like saying amd64 replaces i386... 

This is good as a hotfix, but I think we need to make it work for "all" given arm artifacts 🤔

maybe someone from the homebrew team can help with that - I don't even know if homebrew is actually supported on those envs... the site mentions something about a ["best effort basis"](https://docs.brew.sh/Homebrew-on-Linux#arm).

Maybe we should really just drop arm support from the brew pipe as well 🤔

mavogel:
Unfortunately, I couldn't find either a brew formula supporting multiple non-64-bit arm versions. I only found [armv8 or armv6](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/hardware.rb#L147). 

> maybe someone from the homebrew team can help with that

Do they have slack or all goes via gh issues?

caarlos0:
> Do they have slack or all goes via gh issues?

afaik github issues, but I may be dead wrong

mavogel:
Let's see what happens: https://discourse.brew.sh/t/multiple-arm-32-bit-binaries-build-with-golang-which-to-use-in-formula/5757

mavogel:
ping @caarlos0 : what you think regarding the answers? 

> Generally homebrew takes the upstream default for stuff so I’d say 6

shall we go with `armv6`  if it is present, if not go with the latest? 

Or wait here and extend homebrew to support multiple arm versions but I have no clue how to implement this atm

> it should technically be possible but the easiest way might be to let golang detect the processor

caarlos0:
Another option would be to let the user choose which goarm they want, defaulting to 6 otherwise...

what do you think?

mavogel:
Yes, sounds good to me. I'll provide a `goarm` property though in the brew section

mavogel:
@caarlos0 I implemented it the way that `goarm` filters over all artifacts. If there are multiple builds, each with multiple 32-bit arm versions and in the brew section no ids are specified, hence we still have multiple 32-bit arm versions, then the section still fails. I improved the `ErrMultipleArchivesSameOS` therefore

caarlos0:
awesome, thanks!

mavogel:
shall we release a patch?

caarlos0:
yes, will do it now :)

PR #1152, opened 2019-09-23
fix: cleanup/tidy go.mod deps

tpounds:
Cleanup/tidy `go.mod` dependencies. Should fix similar issues in downstream projects using `go get` or pinning a specific versions with `go.mod` and a `tools.go` file.

see:
#1145
https://github.com/goreleaser/godownloader/issues/133
https://github.com/golangci/golangci-lint/issues/730

PR #1155, opened 2019-09-24
fix: GOPROXY and dependency issues

tpounds:
Updates Makefile to use official GOPROXY (proxy.golang.org) to detects issues in which third party module proxies differ in checksum behavior for broken/invalid packages. In this case the `proxy.golang.org` less permissive than `gocenter.io`.

The issue can easily be reproduced with the following commands:

Works:
```       
$ go clean -modcache
$ GOPROXY=gocenter.io go get code.gitea.io/gitea@v1.10.0-dev.0.20190711052757-a0820e09fbf7
```

Broken:
```
$ go clean -modcache
$ GOPROXY=proxy.golang.org go get code.gitea.io/gitea@v1.10.0-dev.0.20190711052757-a0820e09fbf7
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,8 @@ TEST_OPTIONS?=
 
 export PATH := ./bin:$(PATH)
 export GO111MODULE := on
-export GOPROXY := https://gocenter.io
+# enable consistent Go 1.12/1.13 GOPROXY behavior.
+export GOPROXY = https://proxy.golang.org
 
 # Install all the build and lint dependencies
 setup:
--- a/go.mod
+++ b/go.mod
@@ -27,3 +27,6 @@ require (
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
 	gopkg.in/yaml.v2 v2.2.2
 )
+
+// Fix invalid pseudo-version: revision is longer than canonical (6fd6a9bfe14e)
+replace github.com/go-macaron/cors => github.com/go-macaron/cors v0.0.0-20190418220122-6fd6a9bfe14e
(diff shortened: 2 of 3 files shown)


tpounds:
Mirrors a similar pr made to godownloader.

see: https://github.com/goreleaser/godownloader/pull/135

tpounds:
@caarlos0 This partially reverts the removal of the go-macaron/cors replace directive made in https://github.com/goreleaser/goreleaser/pull/1152. Unfortunately, this wasn't caught due to differences in the way `proxy.golang.org` and `gocenter.io` due version validation.

caarlos0:
thanks :)

ankushchadha:
GoCenter does validation before a go module version is served to the users. As part of the validation, a `go get `operation is executed. For module versions where a go get failed, those are not served by GoCenter and the status is reflected in the UI.

When this module was included in GoCenter, there was no Go 1.13 and a `go get` operation works with Go 1.12 for github.com/go-macaron/cors v0.0.0-20190309005821-6fd6a9bfe14e9. 

Since we honor immutability and availability requirement, GoCenter will continue to serve this module version. There might be bugs in Go 1.12 but atleast that was the actual behavior in Go 1.12 which is what we honor.

tpounds:
@ankushchadha Thanks for the info! I didn't investigate too deeply but figured it had to do with slight differences between how Go 1.12 and 1.13 handled modules. Unfortunately, this change is somewhat necessary since lots of folks will be transparently migrating to the default proxy when they upgrade to Go 1.13.

kamilsk:
default is

```
$ go env GOPROXY
https://proxy.golang.org,direct
```

PR #1158, opened 2019-09-27
ci: extend GOPROXY list

kamilsk:
I just added `https://gocenter.io` as a reserve proxy because `https://proxy.golang.org` may not be available.

Any comments are welcome.

refs  #1155


kamilsk:
default `GOPROXY` value is

```
$ go env GOPROXY
https://proxy.golang.org,direct
```

PR #1159, opened 2019-09-27
feat: wasm support

### LINKED ISSUE #1017: Support releasing WASM builds
Right now the `GOARCH=wasm GOOS=js ` combination is ignored by GoReleaser, but it would be nice if we could also release `.wasm` files packaged and ready to use.

caarlos0:
adds JS WASM build target.

closes #1017


caarlos0:
Not sure if this is enough, as I don't really use wasm et al... it does work though 🤔

PR #1188, opened 2019-10-10
ci: add golangci-lint configuration file

### LINKED ISSUE #1185: improve golangci-lint config
Right now we are running `golangci-lint --enable-all --{disable several things}`.

It would be better to put it in a yaml config file and share it across all repos in this org.

mstruebing:
This commit adds  a configuration for `golangci-lint` to not need to add a bunch of cli parameters.

closes #1185

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,14 @@
+# This file contains all available configuration options
+# with their default values.
+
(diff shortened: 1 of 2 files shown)


caarlos0:
build issue is unrelated, thanks!

PR #1204, opened 2019-10-19
feat: pipe for Sonatype Nexus Raw repositories

antonienko:
Enables Goreleaser to upload artifacts to [Sonatype Nexus](https://www.sonatype.com/product-nexus-repository) Raw repositories.


antonienko:
I needed this feature for my own selfish goals, and I figured I could contribute it to the project. Since it's my first contribution, I'm open to receive feedback and suggestions on how to improve this pull request.

internal/pipe/nexus_raw/nexus_raw.go
```diff
@@ -0,0 +1,145 @@
...
+		body, err := ioutil.ReadAll(res.Body)
+		if err == nil {
+			bodyContents = string(body)
+		}
+		return pipe.Skip(fmt.Sprintf("bad status: %s, body: %s", res.Status, bodyContents))
+	}
+	return nil
+}
+
+func createForm(w *multipart.Writer, artifPath, artifName, nexusDir string) error {
```
caarlos0: errors inside this function shouldn't be `skip errors`.

skip errs are used only to skip a pipe if some configuration is missing, other that that, regular errors should be used.

pkg/config/config.go
```diff
@@ -340,6 +340,14 @@ type Put struct {
 	Signature      bool     `yaml:",omitempty"`
 }
 
+type NexusRaw struct {
+	Url        string `yaml:",omitempty"`
```
caarlos0:
```suggestion
	URL        string `yaml:",omitempty"`
```

pkg/config/config.go
```diff
@@ -340,6 +340,14 @@ type Put struct {
 	Signature      bool     `yaml:",omitempty"`
 }
 
+type NexusRaw struct {
```
caarlos0: missing struct docs here...

internal/pipe/nexus_raw/nexus_raw.go
```diff
@@ -0,0 +1,145 @@
+package nexus_raw
```
caarlos0: missing package docs

internal/pipe/nexus_raw/nexus_raw.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func (Pipe) Publish(ctx *context.Context) error {
+	if len(ctx.Config.NexusRaws) == 0 {
+		return pipe.Skip("sonatype nexus section is not configured")
+	}
+	for _, instance := range ctx.Config.NexusRaws {
+		if skipErr := checkConfig(&instance); skipErr != nil {
+			return pipe.Skip(skipErr.Error())
+		}
```
caarlos0: could be just:

```go
for _, instance := range ctx.Config.NexusRaws {
		instance := instance
		if err := checkConfig(&instance); err != nil {
			return err
		}
}
```

caarlos0:
Just a couple of comments. :) 

Also would be nice if you could add the docs as well 🚀 

Thanks for the PR 👏

antonienko:
I have added some basic documentation based on the artifactory one.

Since the functionality of this pipe is so simple, I haven't go into details of the different configuration parameters. Let me know if you need anything else at this stage, or we can evolve the documents at the same time that this pipe grows.

antonienko:
@caarlos0 Is there anything else I need to do for the pull request to be reviewed?

caarlos0:
> @caarlos0 Is there anything else I need to do for the pull request to be reviewed?

no, I just haven't make the time to look into it yet... sorry for the delay :(

internal/pipe/nexusraw/nexusraw.go
```diff
@@ -0,0 +1,148 @@
+// Package nexus_raw provides a Pipe that push artifacts to a Sonatype Nexus repository of type 'raw'
```
caarlos0: `nexus_raw` -> `nexusraw`

internal/pipe/nexusraw/nexusraw.go
```diff
@@ -0,0 +1,148 @@
...
+		return fmt.Errorf("cannot fill directory field contents: '%s'", err.Error())
+	}
+	filenameField, err := w.CreateFormField("raw.asset1.filename")
+	if err != nil {
+		return fmt.Errorf("cannot create filename field: '%s'", err.Error())
+	}
+	if _, err = io.Copy(filenameField, strings.NewReader(artifName)); err != nil {
+		return fmt.Errorf("cannot fill filename field contents: '%s'", err.Error())
+	}
+	return w.Close()
```
caarlos0: shouldn't this be on a `defer`?

internal/pipe/nexusraw/nexusraw_test.go
```diff
@@ -0,0 +1,210 @@
...
+	assert.ElementsMatch(t, cs.Items, []item{
+		{"testDirectory/artifact.tar.gz"},
+		{"testDirectory/artifact2.tar.gz"}},
+	)
+}
+
+func randomListen(t *testing.T) string {
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
+	require.NoError(t, err)
+	listener.Close()
```
caarlos0: shouldn't it be in a defer?

www/content/nexusraw.md
```diff
@@ -0,0 +1,35 @@
...
+and your credentials (`username` and `password`):
+
+```yaml
+nexus_raws:
+  -
+    url: https://<Your-Instance>:<Your-Port>
+    repository: example-repo
+    directory: example-folder
+    username: your-username
+    password: your-password
```
caarlos0: we should probably get the password from an environment variable?

antonienko: In order to do this, shoud I use the `EnvFiles` struct in the `config/config.go` file together with the loadEnv function?

Seems like it is very tailor-made to the git release process.

Should I load the env variable in my pipe instead?

caarlos0: take a look on the artifactory pipe, it uses the internal `http` package, which maybe you can try to use here as well, which should handle secrets and etc

https://goreleaser.com/artifactory/

antonienko: The problem I see with it is that the internal http package has a hardcoded PUT method, so in order to use it I would need to refactor it to be able to use POST, which is the verb that Nexus accepts for uploads.

caarlos0: I think we can refactor that first then? Maybe in another PR even?

Basically adding a `method` option which defaults to `PUT`, right? Or are there more things?

antonienko: I've been checking out the http package, to assess the refactor to include POST, and I see that PUT is highly hardcoded in it, it wouldn't be a trivial refactor.

I'm sorry to say I don't have time enough to attack this challenge at the moment. 

Would you mind if I finish this Nexus feature loading the environment variable from inside my pipe, and I can hopefully tackle the refactoring of the http package later on?

caarlos0: Sorry, I rather not.

I can try to take a look at the `http` package as well...

caarlos0: @antonienko would https://github.com/goreleaser/goreleaser/pull/1246 work?

antonienko: On a first quick review, it looks great. As soon as it is merged I'll try and complete this PR using it.

caarlos0: merged to master :)

Issue #1206, opened 2019-10-19
Error when releasing a tag


bodgit:
**Describe the bug**

I'm trying to set up GoReleaser with Github actions. It seems to build commits fine, but when I push a tag that should be published, it errors with `git is currently in a dirty state, please check in your pipeline what can be changing the following files`, and then doesn't print any files. I've run `go mod tidy` locally and it's not producing any changes.

**To Reproduce**

Here's my goreleaser.yml:

```yaml
---
before:
  hooks:
    - go mod tidy
builds:
  - main: ./cmd/torrent/main.go
    env:
      - CGO_ENABLED=0
    goarch:
      - amd64
      - arm
    goos:
      - darwin
      - linux
      - windows
      - freebsd
      - openbsd
      - solaris
checksum:
  name_template: 'checksums.txt'
release:
  prerelease: auto
```

**Expected behavior**

I was expecting to get some releases built and published.

**Environment (please complete the following information):**

- OS: Ubuntu (whatever is running Github Actions)
- GoReleaser Version 0.120.0

**Additional context**

Here's what I get in the logs:

```
Run goreleaser/goreleaser-action@v1
  with:
    version: latest
    args: release
  env:
    GOROOT: /opt/hostedtoolcache/go/1.13.3/x64
    GITHUB_TOKEN: ***
✅ GoReleaser version found: v0.120.0
⬇️ Downloading https://github.com/goreleaser/goreleaser/releases/download/v0.120.0/goreleaser_Linux_x86_64.tar.gz...
📦 Extracting GoReleaser...
/bin/tar xz -C /home/runner/work/_temp/dcce10ee-f711-4d9e-b0ff-4d795f0c1f8e -f /tmp/goreleaser-TO2Hcm/goreleaser_Linux_x86_64.tar.gz
✅ v1.0.0 tag found
🏃 Running GoReleaser...
/home/runner/work/_temp/dcce10ee-f711-4d9e-b0ff-4d795f0c1f8e/goreleaser release

   • releasing using goreleaser 0.120.0...
   • loading config file       file=.goreleaser.yml
   • RUNNING BEFORE HOOKS     
      • running go mod tidy      
   • LOADING ENVIRONMENT VARIABLES
   • GETTING AND VALIDATING GIT STATE
      • releasing v1.0.0, commit a838c78fcd1e006745a2c30d7f33b60d62c6c127
[...]

PR #1208, opened 2019-10-20
Revert "feat: git diff when dirty (#1178)"

caarlos0:
`git diff` does not work when we have untracked files.


caarlos0:
refs #1206

PR #1210, opened 2019-10-21
docs: add GITHUB_TOKEN env to github action file

nadoo:
We will get the following error without it: `error=missing GITHUB_TOKEN, GITLAB_TOKEN and GITEA_TOKEN`, and the release action will be failed.


caarlos0:
thanks 🙌

PR #1217, opened 2019-10-27
chore(deps): cleanup go.mod file

chenrui333:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,3 @@ require (
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
 	gopkg.in/yaml.v2 v2.2.4
 )
-
-// Fix invalid pseudo-version: revision is longer than canonical (b0274f40d4c7)
-replace github.com/go-macaron/cors => github.com/go-macaron/cors v0.0.0-20190925001837-b0274f40d4c7
--- a/go.sum
+++ b/go.sum
@@ -131,7 +131,7 @@ github.com/go-logfmt/logfmt v0.4.0/go.mod h1:3RMwSq7FuexP4Kalkev3ejPJsZTpXXBr9+V
 github.com/go-macaron/binding v0.0.0-20160711225916-9440f336b443/go.mod h1:u+H6rwW+HQwUL+w5uaEJSpIlVZDye1o9MB4Su0JfRfM=
 github.com/go-macaron/cache v0.0.0-20151013081102-561735312776/go.mod h1:hHAsZm/oBZVcY+S7qdQL6Vbg5VrXF6RuKGuqsszt3Ok=
 github.com/go-macaron/captcha v0.0.0-20190710000913-8dc5911259df/go.mod h1:j9TJ+0nwUOWBvNnm0bheHIPFf3cC62EQo7n7O6PbjZA=
-github.com/go-macaron/cors v0.0.0-20190925001837-b0274f40d4c7/go.mod h1:r98dHFBQZg2Y3YbiqaQ5YkEJTkYrLSkxWsunlpDddGs=
+github.com/go-macaron/cors v0.0.0-20190309005821-6fd6a9bfe14e9/go.mod h1:utmMRnVIrXPSfA9MFcpIYKEpKawjKxf62vv62k4707E=
 github.com/go-macaron/csrf v0.0.0-20180426211211-503617c6b372/go.mod h1:oZGMxI7MBnicI0jJqJvH4qQzyrWKhtiKxLSJKHC+ydc=
 github.com/go-macaron/i18n v0.0.0-20160612092837-ef57533c3b0f/go.mod h1:MePM/dStkAh+PNzAdNSNl4SGDM2EZvZGken+KpJhM7s=
 github.com/go-macaron/inject v0.0.0-20160627170012-d8a0b8677191/go.mod h1:VFI2o2q9kYsC4o7VP1HrEVosiZZTd+MVT3YZx4gqvJw=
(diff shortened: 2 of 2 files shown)


caarlos0:
so, downgrade the cors lib?

chenrui333:
`github.com/go-macaron/cors` is indirect dependedncy, and I did not see the code reference in the repo.

To my memory, that was the issue when doing `go mod tidy` before (and the dependency problem with `code.gitea.io/gitea`)

Also, I think it is indrect dependency of `code.gitea.io/gitea@v1.10.0-dev`

caarlos0:
oh yah, you're right.

thanks! 🚀

caarlos0:
@chenrui333 check #1218

chenrui333:
Interesting, I did not have the same error in my local though.

chenrui333:
I switched back to go1.13, and the issue does pop up. I will submit a dependabot issue on this matter.

chenrui333:
@caarlos0 Just reported an issue.

caarlos0:
Thanks @chenrui333 ! :)

ldez:
It's still the same problem that I fixed in https://github.com/goreleaser/goreleaser/pull/1138

caarlos0:
IMHO if it works on go1.13 only, its fine.

we just need to change the docs https://github.com/goreleaser/goreleaser/blob/master/CONTRIBUTING.md

ldez:
The `replace` allow to use go1.12 and go1.13.

So, for me this PR must be reverted.

chenrui333:
I am little bit confused on why need to support both go 1.12 and go 1.13? This is more like a developer-oriented tool and we should constantly keep it using the latest go offering.

caarlos0:
> The `replace` directive is required by go1.13

does not seem to be, hence the build passing on go 1.13 🤔

chenrui333:
I think the build passed, right?
https://travis-ci.org/goreleaser/goreleaser/builds/603598848

The dependabot PR build issue is indeed concerning, but I think that sounds like they need to do the environment update on their end, still waiting for their response on this though.

ldez:
The `replace` directive is required by go1.13: to be precise `go mod tidy` in go1.13 required the `replace` directive

ldez:
The CI need to validate the `go.mod` and `go.sum`

A quick way to validate that:

```
go mod tidy
git diff --exit-code go.mod
git diff --exit-code go.sum
go mod download
```

caarlos0:
CI does that already: 

- https://github.com/goreleaser/goreleaser/blob/master/.travis.yml#L17
- https://github.com/goreleaser/goreleaser/blob/master/Makefile#L52
- https://github.com/goreleaser/goreleaser/blob/master/Makefile#L45-L49

chenrui333:
@ldez I believe it passed CI, it is just the dependabot environment needs to be updated to use latest golang. I works on go v1.13.3 but not v1.13 though.

ldez:
I use go 1.13.13 and I need to set the `replace` directive to get an error:

```console
$ go version
go version go1.13.3 linux/amd64

$ go mod tidy
go: code.gitea.io/gitea@v1.10.0-dev.0.20190711052757-a0820e09fbf7 requires
        github.com/go-macaron/cors@v0.0.0-20190309005821-6fd6a9bfe14e9: invalid pseudo-version: revision is longer than canonical (6fd6a9bfe14e)
```

chenrui333:
let me re-check. :)

ldez:
@caarlos0 I find the tricks:

```
export GOPROXY = https://proxy.golang.org,https://gocenter.io,direct
```

I think that `https://gocenter.io` give the invalid hash.

chenrui333:
Good finding. Isn't the default goproxy goes to `https://proxy.golang.org`?

ldez:
It's a fallback system: https://proxy.golang.org > https://gocenter.io > direct

chenrui333:
That is right

ldez:
I confirm: the problem come from https://gocenter.io

https://proxy.golang.org/github.com/go-macaron/cors/@v/v0.0.0-20190309005821-6fd6a9bfe14e9.info

https://search.gocenter.io/github.com~2Fgo-macaron~2Fcors/info?version=v0.0.0-20190309005821-6fd6a9bfe14e9

caarlos0:
cc/ @ankushchadha

elioengcomp:
Version `v0.0.0-20190309005821-6fd6a9bfe14e9` works fine to previous versions of Go even though it has 13 digits for the commit hash part instead of the expected 12. Go 1.12 seems not to care

elioengcomp:
Version `v0.0.0-20190309005821-6fd6a9bfe14e9` works fine to previous versions of Go even though it has 13 digits for the commit hash part instead of the expected 12. Go 1.12 seems not to care about it:

```
→ go get github.com/go-macaron/cors@v0.0.0-20190309005821-6fd6a9bfe14e9

go: finding github.com/go-macaron/cors v0.0.0-20190309005821-6fd6a9bfe14e9
go: finding github.com/go-macaron/inject v0.0.0-20160627170012-d8a0b8677191
go: finding gopkg.in/ini.v1 v1.42.0
```

ldez:
go1.12 don't use `https://proxy.golang.org` as `GOPROXY` by default.
go1.13 use `https://proxy.golang.org` as `GOPROXY` by default.

If you use go1.12 and `GOPROXY=https://proxy.golang.org` the behavior is very close.

The problem comes because the date `20190309005821` is not the valid date for `6fd6a9bfe14e9` and the hash size problem (13 instead 12)

elioengcomp:
> This PR do the inverse: use v0.0.0-20190309005821-6fd6a9bfe14e9 instead of a valid pseudo version.

Sorry, my mistake. We should move to the latest version of the dependency instead.

ldez:
`github.com/go-macaron/cors` is a dependency of  'code.gitea.io/sdk/gitea` that does not contain any semver version.

I fixed the problem.

Issue #1132, opened 2019-09-02
Invalid dependency version in `go.sum`


rvolosatovs:
**Describe the bug**

Version 0.117.1 contains invalid(with go 1.13) dependency version:
https://github.com/goreleaser/goreleaser/blob/ebe510a894357f7e5d8895a934d5db5dd3c7ecb4/go.sum#L134 should be
```
github.com/go-macaron/cors v0.0.0-20190418220122-6fd6a9bfe14e/go.mod h1:utmMRnVIrXPSfA9MFcpIYKEpKawjKxf62vv62k4707E=
```

Otherwise, dependency resolution fails with `410 Gone` (see https://github.com/golang/go/issues/33346)
You can validate this by opening https://proxy.golang.org/github.com/go-macaron/cors/@v/v0.0.0-20190309005821-6fd6a9bfe14e9.info in your browser of choice.

(you can validate the version I suggest by using the same method: https://proxy.golang.org/github.com/go-macaron/cors/@v/v0.0.0-20190418220122-6fd6a9bfe14e.info )

**To Reproduce**

`go run github.com/goreleaser/goreleaser` on Go 1.13

**Expected behavior**

No error

PR #1220, opened 2019-10-29
 fix: update code.gitea.io/sdk/gitea

ldez:
Allow to build with go1.13.

Fixe invalid pseudo version for `github.com/go-macaron/cors`
Update `code.gitea.io/sdk/gitea`

Related #1132 #1138 #1217


caarlos0:
missing a go mod tidy maybe? 🤔

caarlos0:
Thanks for the PR btw! 🚀

elioengcomp:
Any specific reason for removing GoCenter from the proxy list? It would be valuable feedback for us. Tks!

ldez:
I removed GoCenter due to https://github.com/goreleaser/goreleaser/pull/1217#issuecomment-547150158

Since GoCenter can give a positive answer to an "invalid" pseudo version, it does not provide enough guarantees for a CI.

From my point of view, a CI must guarantee a strict check for the dependencies.

From my point of view, as https://proxy.golang.org is the default in Go, the other Go proxies must find a way to guarantee the same behavior to allow all the users, whatever the proxy used, to obtain a reproducible environment, it's something really important for an open source project.

For now, the PR #1217 prove that GoCenter don't provide those guarantees, but I hope that it will evolve or that I missed something. 

PS: I'm not a part of the goreleaser team, it's just my point of view.

caarlos0:
@elioengcomp  basically what @ldez said.

It needs to be consistent, otherwise we can get hard-to-debug issues like the ones refs by the OP on the PR description.

elioengcomp:
I disagree when you say that it does not provide enough guarantees for a CI and I actually think it is the other way around. 

A central repository like GoCenter has two basic premisses that need to be followed in order for users and CI jobs to achieve reproducible builds:
- Immutability: The content once served can never change
- Availability: The content once served can never go away or not be served anymore

Those two premisses need to be followed regardless the environment in which the central repository is being used, since the conditions in which builds happened are unknown and, if the same conditions happen again, the build needs to provide the same results. That is what reproducible builds mean. 

Also keep in mind that if a module is available in GoCenter that means it was requested and fetched by a user that is depending on that to run a build. If the user was fetching the content from GitHub or GoCenter and decide to switch to Google's proxy, the build will break. 

So it seems to me that, by keeping serving that content regardless of the Go command version being used, GoCenter is following the two basic premisses of a central repository and promoting reproducible builds.

elioengcomp:
@caarlos0 I agree that it need to be consistent, but it needs to be consistent with the past. If we change the central repository behavior over time, old builds will never be able to be reproduced. And like we say in Brazil, reproducible builds are like pregnancy, there is no partially reproducible builds. :)

caarlos0:
I get it. 

I think it should also be consistent with the default proxy, otherwise a build may pass with gocenter and fails with the default proxy, which may cause confusion...

elioengcomp:
I agree. That is the goal of the interoperability between proxies provided by the Go modules download protocol and is probably true for the majority of the cases. However, we are not willing to chase that when Google's proxy opinion goes against the two premises and breaks the previous behavior observed by users. Remember that, even though it is the default one, Google's proxy was created after users were already fetching their dependencies from source and even after GoCenter.

PR #1224, opened 2019-11-01
docs: Update CircleCI example

felicianotech:
Updating the CircleCI example.

1. We now have config v2.1 so using that for the example.
2. The workflow name was changed to main as that's more common than one called "release". Which is also confusing since the job is called the same thing.
3. Updated the version of Go to the latest (because why not).
4. Placed the workflow section first.
5. Added a comment explaining what the filters are doing.


caarlos0:
thanks!

PR #1232, opened 2019-11-07
feat: add basic string function to template

gchaincl:
This commit introduces a set of functions from the `strings` package to the templates.

It might be useful to modify the available template predefined variables.

In my case, the version format we use for tags is not compatible with docker tag format so I needed to modify it a bit


caarlos0:
awesome, thanks!

Issue #1221, opened 2019-10-29
Building a dll in windows using c-shared buildmode generates a dll with .exe extension


qmuntal:
**Describe the bug**

When building a binary using `-buildmode=c-shared` in windows go generates a dll, not an executable, but goreleaser always add the extension `.exe` to all the binaries generated with goos=windows, at least this is what I can see in the [source code](https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/build/build.go#L131).

**To Reproduce**

```yaml
builds:
  - id: "testdll"
    env:
      - CGO_ENABLED=1
      - CC=x86_64-w64-mingw32-gcc
      - CXX=x86_64-w64-mingw32-g++
    goos:
      - windows
    goarch:
      - amd64
    flags: -buildmode=c-shared
    ldflags: -s -w
    main: .
    binary: test.dll
archives:
  - format: binary
    builds:
      - testdll
```

**Expected behavior**

I'm expecting a binary archive named `test.dll`, but I get a `test.dll.exe` archive.

**Environment (please complete the following information):**

- OS: windows
- OS version: windows 10 x64
- GoReleaser Version: 0.120.3

caarlos0:
goreleaser was built thinking on building binaries only... 

["Yes is forever"](https://twitter.com/solomonstre/status/715277134978113536), so, I rather not do this - or at least not right now.

Thanks for the suggestion either way.

qmuntal:
I understand that this can add complexity to the code base and that would require effort to maintain, but a `dll` file is also a binary, so you are not breaking the foundation goals. In fact, goreleaser can be used to build shared libraries, the only problem is that the binary file name is wrong.

Would you at least consider fixing this issue if I create a PR with a possible solution?

Thanks!

caarlos0:
> I understand that this can add complexity to the code base and that would require effort to maintain, but a `dll` file is also a binary, so you are not breaking the foundation goals. In fact, goreleaser can be used to build shared libraries, the only problem is that the binary file name is wrong.

yeah, you're right about that.


maybe if we add a custom `extension` config field (which defaults to exe on windows)?

qmuntal:
Great! The custom `extension` is a good solution but tricky to define, as the file extension depends on the `goos` + `goarch` combination, and I can't think on an easy way to represent this in yaml (I'm not an expert neither).

[...]

PR #1243, opened 2019-11-15
feat: Add support for c-shared and c-archive build modes in windows

qmuntal:
This PR modifies the default extension of windows that are build using the flag `-buildmode=c-shared` and `-buildmode=c-archive` from `.exe` to `.dll` and `.lib`, respectively.

Even though go allows other non-`.exe` build modes (archive, shared, pie and plugin), these are not supported in windows.

More info about build modes: https://golang.org/cmd/go/#hdr-Build_modes

This commit is based on this issue:
#1221

### DIFF
--- a/internal/pipe/build/build.go
+++ b/internal/pipe/build/build.go
@@ -132,7 +132,8 @@ func extFor(target string, flags config.FlagArray) string {
 		for _, s := range flags {
 			if s == "-buildmode=c-shared" {
 				return ".dll"
-			} else if s == "-buildmode=c-archive" {
+			}
+			if s == "-buildmode=c-archive" {
 				return ".lib"
 			}
 		}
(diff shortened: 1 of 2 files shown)


internal/pipe/build/build.go
```diff
@@ -127,8 +127,15 @@ func doBuild(ctx *context.Context, build config.Build, target string) error {
 	})
 }
 
-func extFor(target string) string {
+func extFor(target string, flags config.FlagArray) string {
 	if strings.Contains(target, "windows") {
+		for _, s := range flags {
+			if s == "-buildmode=c-shared" {
+				return ".dll"
+			} else if s == "-buildmode=c-archive" {
```
caarlos0: the else here is not needed 💅

caarlos0:
closes #1221

caarlos0:
Awesome, thanks!

PR #1246, opened 2019-11-15
feat: http POST 

caarlos0:
This would allow to send HTTP POSTs as well instead of only PUTs.

refs https://github.com/goreleaser/goreleaser/pull/1204


internal/http/http.go
```diff
@@ -264,7 +293,7 @@ func newUploadRequest(target, username, secret string, headers map[string]string
 	return req, err
 }
 
-func getHTTPClient(put *config.Put) (*h.Client, error) {
+func getHTTPClient(put *config.Upload) (*h.Client, error) {
```
lalloni: Since we're renaming `put` variables as `upload` everywhere, this probably needs to renamed too.

internal/http/http.go
```diff
@@ -288,7 +317,7 @@ func getHTTPClient(put *config.Put) (*h.Client, error) {
 }
 
 // executeHTTPRequest processes the http call with respect of context ctx
-func executeHTTPRequest(ctx *context.Context, put *config.Put, req *h.Request, check ResponseChecker) (*h.Response, error) {
+func executeHTTPRequest(ctx *context.Context, put *config.Upload, req *h.Request, check ResponseChecker) (*h.Response, error) {
```
lalloni: Same here about renaming `put` to `upload`.

internal/http/http.go
```diff
@@ -334,7 +363,7 @@ type targetData struct {
 // resolveTargetTemplate returns the resolved target template with replaced variables
 // Those variables can be replaced by the given context, goos, goarch, goarm and more
 // TODO: replace this with our internal template pkg
-func resolveTargetTemplate(ctx *context.Context, put *config.Put, artifact *artifact.Artifact) (string, error) {
+func resolveTargetTemplate(ctx *context.Context, put *config.Upload, artifact *artifact.Artifact) (string, error) {
```
lalloni: And here.

PR #1247, opened 2019-11-15
refactor: remove some deprecated options

caarlos0:
removed some options that were deprecated for 6 months or more:

- singular nfpm
- singular archive
- singular snapcraft

on the next few months a few more will be removed.

Make sure to run `goreleaser check` every now and then to ensure your config is still valid.

We try hard to avoid those changes, but sometimes we can't help it. Some things must change so the project can continue to evolve.

### DIFF
--- a/internal/http/http.go
+++ b/internal/http/http.go
@@ -343,9 +343,9 @@ func resolveTargetTemplate(ctx *context.Context, put *config.Put, artifact *arti
 
 	if put.Mode == ModeBinary {
 		// TODO: multiple archives here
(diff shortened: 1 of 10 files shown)


mavogel:
I'll review it when the build is green. I just took a quick look at the error and let me see if I find time to fix it on the weekend.

mavogel:
@caarlos0 is this PR still relevant? Then I'd take a look

caarlos0:
yes, will split in smaller prs

Issue #1237, opened 2019-11-09
Homebrew release is coupled with github release


combor:
Hi guys,

Do I understand this correctly that if the github release is disabled than homebrew tap is not gonna be pushed also?

https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/brew/brew.go#L169

We have another switch to controll if the homebrew tap is pushed there: https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/brew/brew.go#L162

So maybe the github release and homebrew tap shouldn't be coupled?

My use case is that I'm not releasing to github but would like to push the tap only.

caarlos0:
the homebrew tap needs a release to somewhere (github or gitlab) to work, because it uses those urls to download the artifacts.

that's why one depends on the other.

right now I don't have planned on adding support to other "backends". 

That being said, closing this :)

combor:
But I don't need it to publish anywhere, I just need it to generate the brew tap only, this was definitely possible several releases back and was changed in this commit: https://github.com/goreleaser/goreleaser/commit/2da118a296a0dd8cb5d9c591abbaad845b15c900. It stopped working when I updated goreleasrer.

caarlos0:
understood, makes sense, definitely a bug then

thanks for reporting :)

caarlos0:
@combor please check #1252 :)

lumaxis:
@caarlos0 Unfortunately, I think this was broken again by #1374 😕

caarlos0:
hey @lumaxis, do you have a reproducible so I can take a look?

lumaxis:
Sure!

1. Clone https://github.com/cli/cli
2. Run `goreleaser release --snapshot` or `goreleaser release --np-publish`
3. Observe that no Homebrew Formula is produced in `dist/`

caarlos0:
will take a look, thanks @lumaxis 🙏

caarlos0:
check https://github.com/goreleaser/goreleaser/pull/1425

lumaxis:
Nice! Looks good from what I can tell.
And thanks for the quick fix 🙂

PR #1252, opened 2019-11-20
fix: brew: do not really depends on release

caarlos0:
The user might provide another `url_template`.

This is not ideal, but I guess its OK for now.

closes https://github.com/goreleaser/goreleaser/issues/1237


combor:
Thanks @caarlos0 . Looks good

Issue #1223, opened 2019-10-31
Support custom endpoint for blob providers


jonfriesen:
**Is your feature request related to a problem? Please describe.**
Blobs don't support custom endpoints (the deprecated s3 support does) for s3 compliant providers like DigitalOcean and Minio

**Describe the solution you'd like**
Add an endpoint property to the config.Blob to override the providers default endpoint

**Describe alternatives you've considered**
Using deprecated s3 support, adding another step in ci/cd to copy the created packages over

**Additional context**
I'm using an CI/CD environment where I would like to keep the endpoint in an environment variable or the goreleaser configuration if possible.

caarlos0:
I was looking into the code, it seems to me that passing the URL in the `bucket` field will solve it. Just need to remove the `s3://` prefix.

Can you verify @jonfriesen ?

caarlos0:
@jonfriesen I think #1253 will fix this :)

jonfriesen:
Looks like it will! Thanks @caarlos0  :raised_hands:

PR #1253, opened 2019-11-20
feat: fully support all s3 pipe feats on blob

caarlos0:
- added blobs.endpoint config
- make bucket template-able as the docs states
- copied s3 pipe's tests into blob package, all passing
- updated gocloud lib

refs #1223


jonfriesen:
fixes #1223 
thank-you!

PR #1258, opened 2019-11-23
Support for running a subset of builds, targets, archives

amdprophet:
I'm not sure if this is the best approach to accomplish what I am hoping for but I figured I'd open this PR to get some feedback.

Add support for running a subset of builds, targets, and archives configured in a goreleaser config. It also adds support for skipping archives all together.

We're using goreleaser as a primary source for both releases & building development binaries across our organization. For example:

* The docker containers that our web developers trigger a goreleaser build with `--no-publish --skip-validate` each time the containers are rebuilt -- the containers only need `linux/amd64` binaries however.

* We build our Windows MSI packages on AppVeyor and we only need to produce the `windows/386` and `windows/amd64` binaries as a result. We're currently maintaining two goreleaser configs, one for Windows and one for our other platforms to solve this for now.

These changes will significantly cut down time spent where goreleaser is building binaries for every platform we support when we only need one or two.

### DIFF
--- a/internal/pipe/archive/archive.go
+++ b/internal/pipe/archive/archive.go
@@ -16,6 +16,7 @@ import (
 	"github.com/goreleaser/goreleaser/internal/artifact"
 	"github.com/goreleaser/goreleaser/internal/deprecate"
 	"github.com/goreleaser/goreleaser/internal/ids"
+	"github.com/goreleaser/goreleaser/internal/pipe"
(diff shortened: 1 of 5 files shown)


caarlos0:
Hi, sorry for the late reponse. 

I think its best to just have multiple goreleaser config files... 

Thanks for the suggestion either way.

PR #1271, opened 2019-12-12
feat: support tar.zst archives

horjulf:
This adds support for `tar.zst` archives.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ goreleaser
 debug.test
 snap.login
 card.png
+
+# Golang vendor folder
+/vendor/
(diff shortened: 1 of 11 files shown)


horjulf:
@caarlos0 what do you think, anything that I might have missed ?

caarlos0:
curious: is it common to use this compression for this usecase? I don't recall seeing it before for this 🤔

horjulf:
Its getting well supported, its way faster than gzip while having better compression ratios.
For example tar officially supports it now, Archlinux also added support recently for distro packages.

Mostly at the moment is a very good option for internal packages or cases where you want to pull and manipulate archives programmatically due to the speed/size advantages.

caarlos0:
not stale

horjulf:
@caarlos0 any plans on merging this or can we drop it ?

caarlos0:
for now I'm waiting if more people want/need this... if there isn't many I probably won't merge it in (at least not now), because it would be another format to test and maintain and etc... and once it is in we can't get it out without breaking people...

Issue #1251, opened 2019-11-18
Upload custom components to Github releases


schollz:
I have a hook that generates a vendored source base for my code. I'd like to upload this tarball of code along with the binaries generated from goreleaser.  I have a hook that generates this a tarball that I need into the `dist` directory, but I'd like a way that I could specify to goreleaser to upload this extra piece to Github too. 

If this already exists, please let me know! Right now I'm using goreleaser to generate the tarball (via hooks) and then manually uploading it.

caarlos0:
> If this already exists, please let me know! Right now I'm using goreleaser to generate the tarball (via hooks) and then manually uploading it.

It doesn't. 

Thanks for the suggestion :)

schollz:
Hi @caarlos0 . Thanks for your amazing work on this great tool. I really appreciate your code and also how well your engagement has helped to make goreleaser integrated into so many Go workflows (mine and others)!

I have figured out a way to do this (I run a script do upload via the Github API after goreleaser finishes) so feel free to close this.

caarlos0:
thanks @schollz :) 

will leave it here, maybe more people need it... 

if not, it will autoclose in a couple of days :)

Cheers!

leighmcculloch:
I was just trying to figure out how to do this. I'm needing to upload an additional file to the GitHub release at the end, and I can write a curl command to upload it, but I can't figure out how to get the GitHub Release ID from goreleaser to use in that upload. Any tips @schollz ?

schollz:
@leighmcculloch Here's what I did and its not elegant. I added to my `goreleaser.yml` a hook to generate the file to upload. Then after runnning `goreleaser release` I manually run [a script to upload it](https://github.com/schollz/croc/blob/master/src/install/upload-src-tarball.sh). 

The "Github Relase ID" is just the version. In my script above the version comes from grepping my source code, but you could easily do `git describe`. And you can follow the rest of that script to identify the right file, repo, etc. Similar to `goreleaser` the `GITHUB_TOKEN` environmental variable is used, so that doesn't need to be set in that script.

This script would be pretty easy to integrate into goreleaser, and I could take a stab at it if there is interest. I would suggest making a new class called "extras" which specifies the filenames of extra files to upload. The files themselves could be generated as the user specifies in "hooks" (which is what I currently do).

leighmcculloch:
[...]

Issue #962, opened 2019-02-16
Hacking support for Rust (not related to previous efforts)


jondot:
Hi all,
Just wanted to let you know that I managed to hack support for rust, without modifying the goreleaser code or doing anything that's mentioned in any of https://github.com/goreleaser/goreleaser/pull/520, https://github.com/goreleaser/goreleaser/issues/496, https://github.com/goreleaser/goreleaser/pull/522.

Here's my configuration:

```yaml
project_name: cep
builds:
  -
    main: dummy.go
    goarch:
      - amd64
    binary: cep
    hooks:
      post: sh build-cross-release.sh
archive:
  replacements:
    darwin: Darwin
    linux: Linux
    windows: Windows
    386: i386
    amd64: x86_64
checksum:
  name_template: 'checksums.txt'
release:
  disable: true
```

By simply providing it with a `dummy.go` and doing all the hard Rust build work in `build-cross-release.sh`, which includes:

```
mkdir -p dist/darwin_amd64 && cp target/x86_64-apple-darwin/release/cep dist/darwin_amd64
mkdir -p dist/linux_amd64 && cp target/x86_64-unknown-linux-gnu/release/cep dist/linux_amd64
```

And so I kind of "cheat" GoReleaser into thinking it did those builds. It will then carry on with the rest of the pipeline happily.

Hope to provoke that there isn't really a very intimate build support needed for Rust, or for that matter any language -- just the ability to provide a shell script, and to copy artifacts in a predictable manner.

caarlos0:
Oh my, this is great! Thanks for hacking around this!

PR #1273, opened 2019-12-17
feat: specify (sub)directory with go code when building

tjanez:
This will enable using GoReleaser in multi-language projects where Go code (and `go.mod`/`go.sum`) is not in project's root directory.

An example of such project is [Oasis Core](https://github.com/oasislabs/oasis-core) (where we are planning on using GoReleaser to do releases).

### DIFF
--- a/internal/builders/golang/build.go
+++ b/internal/builders/golang/build.go
@@ -7,6 +7,7 @@ import (
 	"go/token"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	"github.com/apex/log"
@@ -32,6 +33,9 @@ type Builder struct{}
 
 // WithDefaults sets the defaults for a golang build and returns it
 func (*Builder) WithDefaults(build config.Build) config.Build {
+	if build.Dir == "" {
+		build.Dir = "."
+	}
 	if build.Main == "" {
 		build.Main = "."
 	}
@@ -111,7 +115,7 @@ func (*Builder) Build(ctx *context.Context, build config.Build, options api.Opti
 	cmd = append(cmd, processedLdFlags)
 
 	cmd = append(cmd, "-o", options.Path, build.Main)
-	if err := run(ctx, cmd, env); err != nil {
+	if err := run(ctx, cmd, env, build.Dir); err != nil {
 		return errors.Wrapf(err, "failed to build for %s", options.Target)
 	}
 	ctx.Artifacts.Add(artifact)
@@ -138,11 +142,12 @@ func joinLdFlags(flags []string) string {
 	return ldflagString.String()
 }
 
-func run(ctx *context.Context, command, env []string) error {
+func run(ctx *context.Context, command, env []string, dir string) error {
 	/* #nosec */
 	var cmd = exec.CommandContext(ctx, command[0], command[1:]...)
 	var log = log.WithField("env", env).WithField("cmd", command)
 	cmd.Env = env
+	cmd.Dir = dir
 	log.Debug("running")
 	if out, err := cmd.CombinedOutput(); err != nil {
 		log.WithError(err).Debug("failed")
@@ -181,6 +186,9 @@ func checkMain(build config.Build) error {
 	var main = build.Main
 	if main == "" {
 		main = "."
+	}
+	if build.Dir != "" {
+		main = filepath.Join(build.Dir, main)
 	}
 	stat, ferr := os.Stat(main)
 	if ferr != nil {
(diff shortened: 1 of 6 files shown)


internal/builders/golang/build.go
```diff
@@ -182,6 +187,9 @@ func checkMain(build config.Build) error {
 	if main == "" {
 		main = "."
 	}
+	if build.Dir != "" {
+		main = filepath.Join(build.Dir, main)
```
caarlos0: I'm not sure why this is necessary... can you explain?

internal/pipe/build/build.go
```diff
@@ -113,11 +113,16 @@ func doBuild(ctx *context.Context, build config.Build, target string) error {
 
 	build.Binary = binary
 	var name = build.Binary + ext
-	var path = filepath.Join(
-		ctx.Config.Dist,
-		fmt.Sprintf("%s_%s", build.ID, target),
-		name,
+	path, err := filepath.Abs(
```
caarlos0: not sure why we need this either...

www/content/build.md
```diff
@@ -21,6 +21,10 @@ builds:
     # Defaults to the project name.
     id: "my-build"
 
+    # Path to project's (sub)directory containing Go code.
+    # Default is `.`.
+    dir: go
```
caarlos0: in the end this will be the build working directory, maybe name it as `cwd` or explain this better on docs? what do you think?

caarlos0:
in this case, I think it would probably be simpler to just `cd folder && goreleaser`, or, we could add a `--cwd` flag to the `goreleaser` binary...

any thoughts?

internal/builders/golang/build.go
tjanez: This is necessary to enable GoReleaser to search for the main function in the right directory.

For example, if Go code is in project's `go/` subdirectory and main is in `foo/main.go`, then without this, the next line:
https://github.com/goreleaser/goreleaser/blob/468401bc54d3a31aebcc1526ba803760a80b51c3/internal/builders/golang/build.go#L185

would wrongly stat the `foo/main.go` file and return:
```
error=stat ./foo/main.go: no such file or directory
```

With this change, the code correctly stats the `go/foo/main.go` file.

internal/pipe/build/build.go
tjanez: To make this explanation easier, let's say this relative path here is: `dist/foo_linux_amd64/foo`.

This gets passed to the golang build builder (via `options`):
https://github.com/goreleaser/goreleaser/blob/468401bc54d3a31aebcc1526ba803760a80b51c3/internal/builders/golang/build.go#L57

which sets the output binary path to:
https://github.com/goreleaser/goreleaser/blob/468401bc54d3a31aebcc1526ba803760a80b51c3/internal/builders/golang/build.go#L113

In our example, that would be `-o dist/foo_linux_amd64/foo`.

Since the gobuild is now run in a different directory, e.g. `go/` subdirectory, this would create the `foo` binary in `go/dist/foo_linux_amd64/foo` instead of `dist/foo_linux_amd64/foo`.

And creating the release archive would fail with:
```
error=failed to add dist/foo_linux_amd64/foo -> foo to the archive: open dist/foo_linux_amd64/foo: no such file or directory
```
because that expects to find the `foo` binary in `dist/foo_linux_amd64/foo` (instead of `go/dist/foo_linux_amd64/foo`).

Hence, converting the relative path to an absolute path here seemed like a good solution.

www/content/build.md
tjanez: Yes, this could be explained better in the docs. I've changed this part to:

```yaml
    # Path to project's (sub)directory containing Go code.
    # This is the working directory for the Go build command(s).
    # Default is `.`.
    dir: go
```
---

I'm slightly more in favor of using `dir` (or `working_dir`) instead of `cwd` since this configuration key would read as _build's directory_ or _build's working directory_ whereas using `cwd` would read as _build's current working directory_ which is a bit awkward in my opinion.

tjanez:
> in this case, I think it would probably be simpler to just `cd folder && goreleaser`

I also though about this approach but I think GoReleaser could actually be quite useful for a whole (multi-language) project.

One could use it to build code in other languages (see #962 for an example) and/or upload custom project's artifacts to GitHub releases (see #1251).

Also, I would rather have `.goreleaser.yml` in project's top-level directory since it creates a GitHub release of the whole project and not just the Go part.

caarlos0:
Thanks for the PR @tjanez, sorry for the delay to merge it :)

tjanez:
> Thanks for the PR @tjanez, sorry for the delay to merge it :)

No worries, thanks for merging this :)!

Issue #1287, opened 2020-01-04
Add support for GOMIPS


rs:
The GOMIPS env is used to generate hard/soft float and mips revisions. It would be nice to be able to build variants with different values of this env in the same way as `goarm`.

caarlos0:
hey @rs, can you check #1288?

rs:
With both soft and hardfloat I get the following error:

```
   ⨯ release failed after 12.70s error=failed to add /Users/rs/Development/nextdns/dist/nextdns_linux_mips_softfloat/nextdns -> nextdns to the archive: file nextdns already exists in the archive
```

PR #1288, opened 2020-01-04
feat: support to gomips

caarlos0:
adds support to set `GOMIPS` and `GOMIPS64` when building one of the `mips` `GOARCH`es.

closes https://github.com/goreleaser/goreleaser/issues/1287


internal/builders/golang/build.go
```diff
@@ -163,9 +164,12 @@ func newBuildTarget(s string) (buildTarget, error) {
 	}
 	t.os = parts[0]
 	t.arch = parts[1]
-	if len(parts) == 3 {
+	if t.arch == "arm" && len(parts) == 3 {
```
caarlos0: note to self: should actually check for prefix here..

PR #1305, opened 2020-01-18
docs: invite people to use docker --pull

greut:
Add the flag `--pull` to the `goreleaser` releaser process itself as well as the Docker documentation.

By default, docker won't refresh the images when performing a `docker build`, thus when building new releases, it might reuse an old base image, e.g. `alpine:3` (in my case) or `golang:1.13-alpine` in this case.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -36,9 +36,10 @@ dockers:
   - goreleaser
   build_flag_templates:
   - "--pull"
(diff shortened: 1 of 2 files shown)


greut:
also, label-schema.org has been deprecated, http://label-schema.org/rc1/

caarlos0:
the CI issue is unrelated, I'll have a look soon

PR #1306, opened 2020-01-19
docs: run only on new tag note on github actions

crazy-max:
refs https://github.com/goreleaser/goreleaser-action/issues/61#issuecomment-569286582

### DIFF
--- a/www/content/actions.md
+++ b/www/content/actions.md
@@ -46,7 +46,29 @@ jobs:
 > **IMPORTANT**: note the `Unshallow` workflow step. It is required for the
 > changelog to work correctly.
 
-> If you want to run GoReleaser only on new tag, check [this comment][only-new-tag].
+If you want to run GoReleaser only on new tag, you can use this event:
+
+```yaml
+on:
+  push:
+    tags:
+      - '*'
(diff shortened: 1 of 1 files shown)


www/content/actions.md
```diff
@@ -79,3 +81,4 @@ command and a private key without a passphrase.
 [goreleaser-action]: https://github.com/goreleaser/goreleaser-action
 [actions]: https://github.com/features/actions
 [syntax]: https://help.github.com/en/articles/workflow-syntax-for-github-actions#About-yaml-syntax-for-workflows
+[only-new-tag]: https://github.com/goreleaser/goreleaser-action/issues/61#issuecomment-569286582
```
andreynering: I think that actually adding the example (see below) with a short description to the README would be better than linking the issue. That's not a strong opinion, though. 🙂 

```yaml
on:
  push:
    tags:
      - '*'
```

caarlos0: I agree.. 🤔

crazy-max: Done :)

caarlos0:
thanks 🚀

PR #1327, opened 2020-01-31
Use CI environment variables to figure out tag

### LINKED ISSUE #1163: getTag return incorrect tag, if we have more than 1 tag for 1 commit
**Describe the bug**
**To Reproduce**

1. Create new repo
```
mkdir test_goreleaser && git init
touch test.txt
git commit -m 'test'
```
than create few tags:
```
$ git tag 0.10.1-rc.3
$ git tag 0.10.1-rc.4
$ git tag 0.10.1-rc.5
$ git log
commit 278932bfb8d28c6346354e887b29d01eb4cb25d1 (HEAD -> master, tag: 0.10.1-rc.5, tag: 0.10.1-rc.4, tag: 0.10.1-rc.3)
Author: Andrii Soldatenko <andrii.soldatenko@gmail.com>
Date:   Tue Oct 1 16:27:42 2019 +0300

    init commit
```

if you run goreleaser it will run under the hood (`git describe --tags --abbrev=0`):
release tag `0.10.1-rc.3`

**Expected behavior**
goreleaser will release latest tag `0.10.1-rc.5`

- OS: all
- GoReleaser Version: master

More details you can find in this SO question: https://stackoverflow.com/questions/14273531/how-to-sort-git-tags-by-version-string-order-of-form-rc-x-y-z-w

### LINKED ISSUE #1311: Allow specifying which tag to build
We currently have our CLI inside a monorepo with the server. Our deploy process creates tags associated with the environment (eg. `prod-timestamp`). We are using semantic version tags for releasing the CLI. 
We've run into issues where the commit has already been tagged with the environment. Tagging the commit with a semantic version (eg. `v0.3.2`) causes GitHub Actions to run goreleaser. However, goreleaser picks the environment tag and fails because it's not a semantic version.
It looks like this issue has been raised before: #1163.

A few options are:
- Allow setting which tag to build using an environment variable, like `GORELEASER_TAG`.
- Allow specifying a custom [`getTag()`](https://github.com/goreleaser/goreleaser/blob/449ed95bbe73fb88159e061352fcc786987d7d7d/internal/pipe/git/git.go#L124-L126) command. 
  - Then we could use `git describe --tags --match "v*" --abbrev=0` to pick the semantic version tag.
- Swapping to `git describe --tags --match "v*" --abbrev=0` to pick the tag. 
  - However, that would break users who don't prefix with `v`. 
    - Though maybe there could be an environment variable or setting to enable that mode, (eg. `GORELEASER_ENABLE_V_SEMVAR=true`? 🤷‍♂)
    - Or if `git describe --tags --match "v*" --abbrev=0` doesn't yield a tag, we fallback to the current command.
      - However, that wouldn't work if someone started using `v*` tags, but later dropped the prefix.

We currently have a bash script to loop through all the local tags and delete them until it reaches one that matches `v*`. While that works, it feels a bit hacky.

**Additional context**
I'm willing to help implement this if there's interest.

aeneasr:
This patch detects CI environments and uses the available tag information when
collecting the git tag.

This resolves issues where one commit has multiple tags.

Closes #1163
Closes #1311

### DIFF
--- a/www/content/build.md
+++ b/www/content/build.md
@@ -152,6 +152,6 @@ GOVERSION=$(go version) goreleaser
 
 ## Define Build Tag
 
-Goreleaser uses `git describe` to get the build tag. You can set
+GoReleaser uses `git describe` to get the build tag. You can set
 a different build tag using the environment variable `GORELEASER_CURRENT_TAG`.
 This is useful in scenarios where two tags point to the same commit.
--- a/www/content/build.md
+++ b/www/content/build.md
@@ -149,3 +149,9 @@ GOVERSION=$(go version) goreleaser
  ```
 
  [hook]: /hooks
+
+## Define Build Tag
+
+Goreleaser uses `git describe` to get the build tag. You can set
+a different build tag using the environment variable `GORELEASER_CURRENT_TAG`.
+This is useful in scenarios where two tags point to the same commit.
--- a/www/content/release.md
+++ b/www/content/release.md
@@ -137,6 +137,12 @@ changelog:
       - (?i)foo
 ```
 
+### Define Previos Tag
+
+Goreleaser uses `git describe` to get the previous tag used for generating the Changelog.
+. You can set a different build tag using the environment variable `GORELEASER_PREVIOUS_TAG`.
+This is useful in scenarios where two tags point to the same commit.
+
 ## Custom release notes
 
 You can specify a file containing your custom release notes, and
--- a/www/content/release.md
+++ b/www/content/release.md
@@ -140,7 +140,7 @@ changelog:
 ### Define Previous Tag
 
 GoReleaser uses `git describe` to get the previous tag used for generating the Changelog.
-. You can set a different build tag using the environment variable `GORELEASER_PREVIOUS_TAG`.
+You can set a different build tag using the environment variable `GORELEASER_PREVIOUS_TAG`.
 This is useful in scenarios where two tags point to the same commit.
 
 ## Custom release notes
--- a/www/content/release.md
+++ b/www/content/release.md
@@ -137,7 +137,7 @@ changelog:
       - (?i)foo
 ```
 
-### Define Previos Tag
+### Define Previous Tag
 
 Goreleaser uses `git describe` to get the previous tag used for generating the Changelog.
 . You can set a different build tag using the environment variable `GORELEASER_PREVIOUS_TAG`.
--- a/www/content/release.md
+++ b/www/content/release.md
@@ -139,7 +139,7 @@ changelog:
 
 ### Define Previous Tag
 
-Goreleaser uses `git describe` to get the previous tag used for generating the Changelog.
+GoReleaser uses `git describe` to get the previous tag used for generating the Changelog.
 . You can set a different build tag using the environment variable `GORELEASER_PREVIOUS_TAG`.
 This is useful in scenarios where two tags point to the same commit.
 
--- a/internal/pipe/changelog/changelog.go
+++ b/internal/pipe/changelog/changelog.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"regexp"
 	"sort"
@@ -189,6 +190,10 @@ func gitLog(refs ...string) (string, error) {
 }
 
 func previous(tag string) (result string, err error) {
+	if tag := os.Getenv("GORELEASER_PREVIOUS_TAG"); tag != "" {
+		return tag, nil
+	}
+
 	result, err = git.Clean(git.Run("describe", "--tags", "--abbrev=0", fmt.Sprintf("tags/%s^", tag)))
 	if err != nil {
 		result, err = git.Clean(git.Run("rev-list", "--max-parents=0", "HEAD"))
--- a/internal/pipe/changelog/changelog_test.go
+++ b/internal/pipe/changelog/changelog_test.go
@@ -109,6 +109,30 @@ func TestChangelog(t *testing.T) {
 	require.NotEmpty(t, string(bts))
 }
 
+func TestChangelogPreviousTagEnv(t *testing.T) {
+	folder, back := testlib.Mktmp(t)
+	defer back()
+	testlib.GitInit(t)
+	testlib.GitCommit(t, "first")
+	testlib.GitTag(t, "v0.0.1")
+	testlib.GitCommit(t, "second")
+	testlib.GitTag(t, "v0.0.2")
+	testlib.GitCommit(t, "third")
+	testlib.GitTag(t, "v0.0.3")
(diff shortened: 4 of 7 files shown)


caarlos0:
# [Codecov](https://codecov.io/gh/goreleaser/goreleaser/pull/1327?src=pr&el=h1) Report
> Merging [#1327](https://codecov.io/gh/goreleaser/goreleaser/pull/1327?src=pr&el=desc) into [master](https://codecov.io/gh/goreleaser/goreleaser/commit/abd66bf2e9f7df9b212f9e984a4b0fcb61e3033b&el=desc) will **increase** coverage by `0.07%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/goreleaser/goreleaser/pull/1327?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master    #1327      +/-   ##
==========================================
+ Coverage   83.61%   83.69%   +0.07%     
==========================================
  Files          57       57              
  Lines        3278     3294      +16     
==========================================
+ Hits         2741     2757      +16     
  Misses        460      460              
  Partials       77       77              
```



------

[Continue to review full report at Codecov](https://codecov.io/gh/goreleaser/goreleaser/pull/1327?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/goreleaser/goreleaser/pull/1327?src=pr&el=footer). Last update [abd66bf...7b999d3](https://codecov.io/gh/goreleaser/goreleaser/pull/1327?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

caarlos0:
this solves the issue running on CI only... but not locally

maybe we can make this clearer by accepting `GORELEASER_CURRENT_TAG` and `GORELEASER_PREVIOUS_TAG`? 

Then people with these problems can still do `GORELEASER_PREVIOUS_TAG=$MY_CI_ENV_WITH_THE_TAG goreleaser` and locally they can run it as well...

what do you think?

aeneasr:
I think we can do both. Getting this stuff from another env var would definitely be nice. I think for the sake of "it just works", having the CI checks in goreleaser would be nice!

aeneasr:
Just for clarification - what's the difference between `GORELEASER_CURRENT_TAG` and `GORELEASER_PREVIOUS_TAG`?

caarlos0:
> Just for clarification - what's the difference between `GORELEASER_CURRENT_TAG` and `GORELEASER_PREVIOUS_TAG`?

Previous tag is the one used to get make the changelog later on (basically `git log tag...previous_tag`), so maybe we could make it available as well...



> I think we can do both. Getting this stuff from another env var would definitely be nice. I think for the sake of "it just works", having the CI checks in goreleaser would be nice!

Since this changes previous behavior, I prefer to have it as an opt-in... having a custom variable seems like a nice way of doing that 🤔

aeneasr:
I see - that makes sense! Could you point me to the place where `GORELEASER_PREVIOUS_TAG` should be used? I'll update the PR then :)

caarlos0:
I think it will require some refactor...

today the previous tag is gather on `changelog.go`

and the current on `git.go` (the file you changed).

ideally I think they should be in the same place, but you can change these 2 and leave a TODO for now... one day I'll get to it... maybe :D

aeneasr:
Ok, sounds good - I'll update the PR then :)

aeneasr:
changelog and git run fine in tests, not sure why minio fails - maybe some flakiness? Don't think it's related to these changes.

https://travis-ci.org/goreleaser/goreleaser/builds/644465453#L635-L642

www/content/release.md
```diff
@@ -137,6 +137,12 @@ changelog:
       - (?i)foo
 ```
 
+### Define Previos Tag
```
caarlos0: typo:

```suggestion
### Define Previous Tag
```

www/content/release.md
```diff
@@ -137,6 +137,12 @@ changelog:
...
+
+Goreleaser uses `git describe` to get the previous tag used for generating the Changelog.
```
caarlos0:
```suggestion
GoReleaser uses `git describe` to get the previous tag used for generating the Changelog.
```

caarlos0: I would document both environment variables here only.

www/content/build.md
```diff
@@ -149,3 +149,9 @@ GOVERSION=$(go version) goreleaser
  ```
 
  [hook]: /hooks
+
+## Define Build Tag
```
caarlos0: I think we can describe both environments on the release page... or maybe on `environment.md` 🤔

caarlos0:
> changelog and git run fine in tests, not sure why minio fails - maybe some flakiness? Don't think it's related to these changes.

looks like it yes

www/content/build.md
aeneasr: We can add it to `environment.md` but I would keep them separate because they impact different features. If there is a known issues / workarounds file, I would put it there also.

aeneasr:
Thank you for your quick replies and reviews and the great work - do you want me to add the docs to environment too? I would be super happy if we could get this done because it's friday late over here :)

www/content/build.md
caarlos0: hmm, yeah, makes sense... lets then add both to environment as well?

we can just link to the relevant topics 🤔

www/content/build.md
```diff
@@ -149,3 +149,9 @@ GOVERSION=$(go version) goreleaser
...
+
+Goreleaser uses `git describe` to get the build tag. You can set
```
caarlos0:
```suggestion
GoReleaser uses `git describe` to get the build tag. You can set
```

www/content/release.md
```diff
@@ -137,6 +137,12 @@ changelog:
       - (?i)foo
 ```
 
+### Define Previous Tag
+
+GoReleaser uses `git describe` to get the previous tag used for generating the Changelog.
+. You can set a different build tag using the environment variable `GORELEASER_PREVIOUS_TAG`.
```
caarlos0:
```suggestion
You can set a different build tag using the environment variable `GORELEASER_PREVIOUS_TAG`.
```

www/content/build.md
aeneasr: Whichever you prefer :)

caarlos0: lets do it then 🙏

caarlos0:
thank you very much @aeneasr 🚀

aeneasr:
Awesome, thank you for the quick turnaround :) Not sure if you plan any release in the next days but I definitely would appreciate it :)

caarlos0:
I think I'll probably do another release by Monday, sounds good?

aeneasr:
Perfect, thank you!!

PR #1332, opened 2020-02-06
feat: add ability to skip uploads for prereleases

vilmibm:

Expose a new config option for `scoop`, `skip_upload`, that behaves like `brew.skip_upload`

We make heavy use of prereleases over at https://github.com/cli/cli and have come to rely on the
brew pipeline's `skip_upload` setting to avoid shipping prereleases. We were surprised that the
scoop pipeline didn't seem to support a similar feature, so this PR adds it. Please let me know if
there is existing support for this that I just missed!

[the existing brew feature](https://github.com/goreleaser/goreleaser/blob/479798126ea3d066f11fbd9ab67c21aa057ce204/internal/pipe/brew/brew.go#L157)

### DIFF
--- a/internal/pipe/scoop/scoop.go
+++ b/internal/pipe/scoop/scoop.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/goreleaser/goreleaser/internal/artifact"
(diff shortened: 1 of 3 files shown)


caarlos0:
Thanks for the PR!

I'll handle the build issue! :)

PR #1333, opened 2020-02-06
feat: upload extra files to the release

### LINKED ISSUE #1320: Question: Is it possible to update arbitrary files to release?
According to https://goreleaser.com/release/, it appears like only "archives" that can be referred to with their "id"s can be uploaded to release.

However, I have several extra files that I produce (such as `${artifact}.sig`) files that I'm trying to upload as well.

It doesn't appear like there's an option for this? (Something like `additional_files:` that matches to things added to `dist/` would be immensely helpful.)

If this doesn't exist and is a feature request, please label it as such –I wasn't sure.

caarlos0:
closes #1320

cc/ @ahmetb


www/content/release.md
```diff
@@ -45,6 +45,13 @@ release:
   # GitHub.
   # Defaults to false.
   disable: true
+
+  # You can add extra pre-existing files to the release.
+  # The left side is the name you want on the github release and the right
+  # side is the path to the file.
+  # Defaults to empty.
+  extra_files:
+    name_on_github: ./path/to/file.txt
```
ahmetb: have you considered something like doublestar glob syntax?

the equivalent feature on both GitHub workflows and Travis CI offer `files:` field where you can just glob like `**/**.bar`

caarlos0: yes, I'm worried about that because we would have to guess the filename on github side then... 

we could use the last part of the path, but it could end up with duplicated names - which would lead to failed releases 🤔

eg:

```
foo/bar/1.md
foo/lala/1.md
```

```
extra_files:
- foo/**/*.md
```

in this case we would get `1.md` 2 times 🤔

I'm not sure if this is something we can just leave to the user (and accept the globs) or be more restrict about (current impl)...

caarlos0: gently ping @ahmetb

ahmetb: oops sorry.

Most extensible and fail-safe way would be to fail maybe when you get name overlaps.

Similarly you can use something like:

```
extra_files:
- file: foo
- file: bar
- file_glob: foo/**/**.tar.gz
```

this way the user can control the behavior.

caarlos0: yeah I guess it makes sense, done.

ahmetb: I assume you went for extra_files_globs, which isn't exactly what I suggested but if you see that as fit that's ok.

caarlos0:
@ahmetb let me know what you think...

we can later add a syntax like `file=name` if needed

ahmetb:
Honestly if you make the list []struct you can extend with fields later or implement backwards compat. If you do []string like this, you can’t. 

I recommend doing this:

```
extra_files:
- glob: val1
- glob: val2
```

caarlos0:
makes sense @ahmetb 

fixed

ahmetb:
Yeah that's better, now you can nicely deprecate `glob` in the future or implement backfill or additional fields.

Issue #1336, opened 2020-02-10
Ability to to append parameters to HTTP upload TargetURL


sysbot:
When using the HTTP upload feature[1], it's not possible to add additional parameters for the PUT method. 

In order to use HTTP Upload with Artifactory API [2] (i'm aware of the Artifactory features, it currently doesn't do what I need yet) and have the flexibility to manipulate the passing TargetURL, I need to be able to append parameters to the TargetURL in order to properly build the index for the archive. 

Manually Appending to the parameters to end of the TargetURL will not work because the Archive name is added at the end of the TargetURL, where the additional parameters needed to be at the end of the TargetURL.

I'm suggesting exposing the filename as a variable that can be moved around within the TargetURL or having a new variable to pass in map[string]string of the parameters.

[1] https://goreleaser.com/customization/#HTTP%20Upload
[2] https://www.jfrog.com/confluence/display/RTF/Artifactory+REST+API#ArtifactoryRESTAPI-Example-DeployinganArtifact

caarlos0:
fixed in #1338, correct?

sysbot:
Yup! Thanks for merging.

PR #1338, opened 2020-02-10
feat: Add the ability to place artifact name

sysbot:
This commit exposes the Artifact name as a template variable so that the
artifact name can be moved around within the `targetURL`. This enable users to
manipulate the target URL as desired in order to pass in PUT parameters that
doesn't always following the currently defined `targetURL` which append the
artifact name to the end.

This is needed to address[1] by enable the ability to add metadata to
Artifactory REST API [2] as the URL parameters, which need to be after the
artifact names.

[1]:https://github.com/goreleaser/goreleaser/issues/1336
[2]:https://www.jfrog.com/confluence/display/RTF/Artifactory+REST+API#ArtifactoryRESTAPI-Example-DeployinganArtifact


caarlos0:
thanks!

PR #1344, opened 2020-02-13
feat(nfpm): add nfpm release config

dotwoo:
add config.NFPM.Release  to set  nfpm release
the default value read from Env["RELEASE"]

### DIFF
--- a/internal/pipe/nfpm/nfpm.go
+++ b/internal/pipe/nfpm/nfpm.go
@@ -62,6 +62,9 @@ func (Pipe) Default(ctx *context.Context) error {
 				fpm.Builds = append(fpm.Builds, b.ID)
 			}
 		}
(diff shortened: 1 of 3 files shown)


internal/pipe/nfpm/nfpm.go
```diff
@@ -62,6 +62,9 @@ func (Pipe) Default(ctx *context.Context) error {
 				fpm.Builds = append(fpm.Builds, b.ID)
 			}
 		}
+		if fpm.Release == "" {
+			fpm.Release = ctx.Env["RELEASE"]
```
caarlos0: I don't get why this default

no other config fields on goreleaser default to environment variable values

dotwoo:
ok. i will open other pull request

Issue #1347, opened 2020-02-14
Variables in archives.files not respected


ostrowr:
**Describe the bug**

I have three directories that look like

- `extras/linux/executable`
- `extras/windows/executable`
- `extras/darwin/executable`

and I'd like to include each executable in its respective archive; i.e.

```yaml
archives:
-
  files:
      - ./extras/{{.Os}}/executable
 ```

No errors are thrown, but the resulting archive does not contain the executable.

**To Reproduce**

Using

```yaml
archives:
-
  files:
      - ./extras/{{.Os}}/executable
 ```

doesn't work, while

```yaml
archives:
-
  files:
      - ./extras/darwin/executable
 ```

does.

**Expected behavior**

The {{.Os}} variable is passed through properly.

**Environment**

- OS: mac
- OS version: macos 10.15
- GoReleaser Version 0.126.0

**Additional context**
Oddly, if I do something like

```yaml
archives:
-
  files:
      - ./extras/{{.Os}}/executablethatdoesntexist
 ```

then I do get an error. So the globbing is happening but the files just aren't getting sent to the archive.

ostrowr:
I'll add tests to the above PR, just haven't had a chance yet

PR #1350, opened 2020-02-17
feat: support variable substitution in archive.files

ostrowr:
If applied, this commit will allow template variables like {{.Os}} to appear in the archive.files field. 

It can be useful to include different files in an archive depending on os and arch. This will be especially useful if names of files can be customized – I'll push up another PR shortly to get your thoughts on an approach there.

Fixes https://github.com/goreleaser/goreleaser/issues/1347

### DIFF
--- a/internal/pipe/archive/archive.go
+++ b/internal/pipe/archive/archive.go
@@ -131,8 +131,9 @@ func create(ctx *context.Context, archive config.Archive, binaries []*artifact.A
 	var log = log.WithField("archive", archivePath)
 	log.Info("creating")
 
-	wrap, err := tmpl.New(ctx).
-		WithArtifact(binaries[0], archive.Replacements).
+	template := tmpl.New(ctx).
+		WithArtifact(binaries[0], archive.Replacements)
+	wrap, err := template.
 		Apply(wrapFolder(archive))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/pipe/archive/archive.go
```diff
@@ -210,9 +211,13 @@ func skip(ctx *context.Context, archive config.Archive, binaries []*artifact.Art
 	return nil
 }
 
-func findFiles(archive config.Archive) (result []string, err error) {
+func findFiles(template *tmpl.Template, archive config.Archive) (result []string, err error) {
 	for _, glob := range archive.Files {
-		files, err := zglob.Glob(glob)
+		replaced, err := template.Apply(glob)
+		if err != nil {
+			return result, fmt.Errorf("Variable substitution failed for pattern %s: %s", glob, err.Error())
```
caarlos0: errors should all be lowercased...

caarlos0:
```suggestion
			return result, errors.Wrapf(err, "failed to apply template %s", glob)
```

caarlos0:
thanks for the pr!

can you add some tests as well?

thanks!

caarlos0:
continues on #1373

Issue #679, opened 2018-05-18
File placement inside a archive


diegobernardes:
**Is your feature request related to a problem? Please describe.**

With this config:
```yaml
archive:
  files:
    - readme.md
    - LICENSE
    - service/flare/cmd/flare.sample.toml
```
I want to to chose where the files gonna stay inside the archive.

**Describe the solution you'd like**

We could do something like Docker does to volumes and ports. (`path:destination`)

```yaml
archive:
  files:
    - readme.md
    - misc/LICENSE:.
    - service/flare/cmd/flare.sample.toml:flare.sample.toml
```

If the destination is a folder, copy the file to the folder, if it's a file, copy and rename the file.

**Describe alternatives you've considered**

Don't have much alternative besides change the project structure.

oschwald:
I'd also love to see something along these lines. In my case, I am trying to generate files in hooks that have the same name as the template file in the repo. Is there any way to get this issue reopened?

caarlos0:
I think the biggest issue with that is globs, e.g.:

```
/foo/**/*.txt:/bar/foo/test
```

what is the sane thing to be expected to happen in this case?

oschwald:
I wasn't advocating for a particular implementation. My use case involves generated files that differ between platforms (e.g., different paths for Windows and Linux). Right now, I haven't found a good way to include these files in the archive in a fixed location between builds for different OSes. I tried using hooks, but given that the archiving happens after the generation, the files will get overwritten by the next build. If I place the files directly in the `dist` folder from a hook, they get ignored when the archive is created.

If we did go with the suggested syntax, perhaps globs such as `/foo/**/*.txt:/bar/foo/test` could translate to putting the files in the directory `/bar/foo/test`.

caarlos0:
got it, right now there is indeed no way of doing what you need.

I'll keep the issue open, but I don't know a good way of implementing this yet... 

maybe we would have to accept the target path only when no glob is used or something, don't know yet.

qiangli:
Have a similar requirement - packaging binaries depending on platforms. Is it possible to not ignore custom files placed in dist folder or  add a prehook to the archive section?

caarlos0:
> Have a similar requirement - packaging binaries depending on platforms. Is it possible to not ignore custom files placed in dist folder or add a prehook to the archive section?

you can have several archive configs and several build configs... the config file will get repetitive, but it should work.

sorry for the late response, somehow lost this notification...

[...]

PR #1351, opened 2020-02-17
feat: add archive.rewrite_paths option [WIP]

ostrowr:
Allow a user to specify a `source` and `dest` for some (list of) files in the archive
and name the archived file at `source`, `dest`.

If applied, this commit will add a new `archive.rewrite_paths` option to the yaml config. This is a list of `source`, `dest` pairs where `source`s are files that are to be added to the archive, and `dest`s are what we want them to be called within the archive.

This change pairs with 
https://github.com/goreleaser/goreleaser/pull/1350 as it allows you to do something like include a different file per-OS, but called the same thing

and also fixes https://github.com/goreleaser/goreleaser/issues/679.

This is one of many ways that this feature could be implemented. If you're interested in including this, I still need to 

- [ ] write docs for the new feature
- [ ] add tests, especially around moving whole directories
- [ ] determine what should happen if the source doesn't exist
- [ ] improve various error conditions

### DIFF
--- a/internal/pipe/archive/archive.go
+++ b/internal/pipe/archive/archive.go
@@ -146,7 +146,7 @@ func create(ctx *context.Context, archive config.Archive, binaries []*artifact.A
 		return fmt.Errorf("failed to find files to archive: %s", err.Error())
 	}
 	for _, f := range files {
-		if err = a.Add(f, f); err != nil {
+		if err = a.Add(rewritePath(archive, f), f); err != nil {
 			return fmt.Errorf("failed to add %s to the archive: %s", f, err.Error())
 		}
 	}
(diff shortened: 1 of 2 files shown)


ostrowr:
@caarlos0 if this is something you're interested in including, let me know and I'll work on the remaining tasks above! Otherwise, feel free to close this.

caarlos0:
I think a better solution would be to make `archive.Files` have a structure similar to `release.Files`... will see if I can do something like that soon

ostrowr:
I'm not sure how that would work if files need to be renamed/within subdirectories.

Specifically, imagine you want to include a macOS `.app` file (which is really a directory.) Just including the last part of the path will flatten that directory and the app will become invalid.

caarlos0:
> I'm not sure how that would work if files need to be renamed/within subdirectories.
> 
> Specifically, imagine you want to include a macOS `.app` file (which is really a directory.) Just including the last part of the path will flatten that directory and the app will become invalid.

I think this should work fine, can you test it?

ostrowr:
yep! I'll rebase + test by the end of the week.

PR #1356, opened 2020-02-20
docs: Clarify usage of --snapshot in quick start

twexler:
The docs for `--snapshot` in the quick start are confusing, especially since they mention Github releases and `--snapshot` doesn't create releases at all.


twexler:
Fixed the PR title, how does one get the semantic pr check to run again without committing?

caarlos0:
thanks!

PR #1363, opened 2020-02-26
feat(nfpm): add nfpm release config

dotwoo:
add config.NFPM.Release  to set  nfpm release

the rpm install process will compare release value. So to set the nfpm release is necessary .


caarlos0:
thanks!

Issue #1337, opened 2020-02-10
ProjectName empty when releasing to gitlab


acobaugh:
**Describe the bug**

`.ProjectName` appears to be empty when releasing to gitlab. I'm trying to build a reusable CI pipeline, so I don't want to set the `project_name` setting in `.goreleaser.yml` if I can avoid it.

**To Reproduce**

```yaml
env:
  - CGO_ENABLED=0
  - GO111MODULE=on
  - GOPROXY=https://gocenter.io
before:
  hooks:
    #- go mod tidy
builds:
- ldflags:
  - -X git.psu.edu/swe-golang/buildversion.commit={{.ShortCommit}}
  - -X git.psu.edu/swe-golang/buildversion.version={{.Version}}
  - -X git.psu.edu/swe-golang/buildversion.buildDate={{.Date}}
  goos:
  - windows
  - darwin
  - linux
  goarch:
  - amd64
checksum:
  name_template: 'checksums.txt'
snapshot:
  name_template: "snapshot"
changelog:
  sort: asc
  filters:
    exclude:
    - '^docs:'
    - '^test:'
    - '^ci:'
    - Merge pull request
    - Merge branch
gitlab_urls:
  api: https://git.psu.edu/api/v4/
  download: https://git.psu.edu
  skip_tls_verify: false
release:
  gitlab:
```

> Make sure your config file is valid by running
> `goreleaser check -f path-to-config-file`.

**Expected behavior**

.ProjectName is set correctly.

**Environment (please complete the following information):**

- OS: linux
- OS version: golang:1.12-alpine
- GoReleaser Version v0.126.0
- GitLab: Community Edition 12.6.6

**Additional context**
This appears to be caused by https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/project/project.go#L16

caarlos0:
whats the output of your `git remote -v`?

acobaugh:
So, what's interesting is I have two versions of this pipeline, one that runs `goreleaser release --snapshot --rm-dist` and another that does `goreleaser release --rm-dist`. The `--snapshot` pipeline gets the correct `.ProjectName`, but the actual release doesn't.

`git remote -v`:
```
33 $ git remote -v
34 origin	https://gitlab-ci-token:[MASKED]@git.psu.edu/swe-golang/umgctl.git (fetch)
35 origin	https://gitlab-ci-token:[MASKED]@git.psu.edu/swe-golang/umgctl.git (push)
```

[...]

PR #1364, opened 2020-02-26
fix: infer project name from GitLab/Gitea release

craigfurman:
When a GitLab release is configured, the project name will default to the repository name, if not set explicitly. This is already the behaviour with GitHub releases.

This should fix https://github.com/goreleaser/goreleaser/issues/1337.

### DIFF
--- a/internal/pipe/project/project.go
+++ b/internal/pipe/project/project.go
@@ -13,7 +13,14 @@ func (Pipe) String() string {
 // Default set project defaults
 func (Pipe) Default(ctx *context.Context) error {
 	if ctx.Config.ProjectName == "" {
-		ctx.Config.ProjectName = ctx.Config.Release.GitHub.Name
+		switch {
+		case ctx.Config.Release.GitHub.Name != "":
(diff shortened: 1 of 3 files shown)


internal/pipe/project/project.go
```diff
@@ -13,7 +13,11 @@ func (Pipe) String() string {
 // Default set project defaults
 func (Pipe) Default(ctx *context.Context) error {
 	if ctx.Config.ProjectName == "" {
-		ctx.Config.ProjectName = ctx.Config.Release.GitHub.Name
+		if ctx.Config.Release.GitHub.Name != "" {
+			ctx.Config.ProjectName = ctx.Config.Release.GitHub.Name
+		} else if ctx.Config.Release.GitLab.Name != "" {
```
caarlos0: can we also add gitea here please?

craigfurman: Thanks for the quick response @caarlos0! I've added Gitea, and updated the docs. I had to convert the if-else chain to a switch to keep golangci-lint happy.

PR #1366, opened 2020-02-27
feat(config): add json schema to config file

fitzix:

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -1,0 +1,919 @@
+{
+    "$id": "https://github.com/goreleaser/goreleaser/schema.json",
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "goreleaser",
+    "description": "goreleaser config json schema",
+    "definitions": {
+      "goosTypes": {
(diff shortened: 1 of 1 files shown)


caarlos0:
is there a way to generate this from code? 

also, is it supposed to work with yaml as well? I tried on vscode and it didn't work.. maybe I'm missing something 🤔

fitzix:
maybe it can be generated from code, but I haven't found it yet. 
vscode doesn't support yaml schema by default, you need to install plugin [YAML Language Support by Red Hat](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml).
```json
"yaml.schemas": {
    "https://raw.githubusercontent.com/fitzix/goreleaser/master/schema.json": ["/.goreleaser.yml"]
}
```

caarlos0:
I'm sorry, if we can't generate this from code, it would be just too many things to maintain.

Rather not add this right now.

Thanks for the PR anyway, maybe in the future. :)

PR #1369, opened 2020-03-03
docs: improve GitLab CI documentation

craigfurman:
Fix typos that referenced GitHub.

Move config to job level. Users can extract elements to global config if
necessary, but since the snippet has no other jobs the use of global
config added confusion.

Disable shallow clones to fix changelogs.

Update use of GitLab image registry to use the USER/PASSWORD vars and
clarify that they are automatically set on jobs running on GitLab
instances with the image registry enabled.

Document use of `only` to not release every commit (optionally).

Format nested docker goreleaser command multi-line.

Document purpose of GITLAB_TOKEN.

### DIFF
--- a/www/content/ci.md
+++ b/www/content/ci.md
@@ -303,42 +303,64 @@ for more detailed documentation.
 
 ## GitLab CI
 
-To push releases to both GitHub and the **official** Docker registry, add a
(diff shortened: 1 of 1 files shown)


caarlos0:
Awesome, thank you very much 🙏

Issue #1314, opened 2020-01-27
env does not work on blobs pipe


caarlos0:
I think the blob pipe does not use the env from the goreleaser file itself - which may be a bug, but its also safer.

You probably shouldn't have your keys in this file, or commit them to version control (don't know if that is what you're doing...)

anyway, your additional context example is probably the correct way to run it.

will close this as I really don't think we should make the `.goreleaser.yml` `env` be used that way.

thanks for the issue though!

_Originally posted by @caarlos0 in https://github.com/goreleaser/goreleaser/issues/1297#issuecomment-578509408_

radeksimko:
> will close this as I really don't think we should make the .goreleaser.yml env be used that way.

I am inclined to agree, I think it's too risky even if we put a disclaimer in the docs.

Generally I feel there is often need for some mechanism to allow reusing parts of config, but I don't think it has to or should be in the form of _global_ inheritable environment variables.

The common use case for the `env` seems to be variables which are used during building:

https://github.com/goreleaser/goreleaser/blob/da787478753c06aa6137c1cf0ed05699f193b198/.goreleaser.yml#L1-L3

Arguably that could be also passed directly to GoReleaser, but it's nice to see it in context.

Should we introduce `build_env` to allow reducing scope of these variables?

caarlos0:
> Should we introduce `build_env` to allow reducing scope of these variables?

yeah I think something like that makes sense, or just a `builds.env` property... 

the global before hook would also need its own `env` then... 

I think that's actually good. More self-contained and clear, although a bit repetitive...  🤔

caarlos0:
yeah, this should probably not be implemented, please export the envs outside `.goreleaser.yml`.

Closing...

PR #1391, opened 2020-03-16
feat: load envs from config

caarlos0:
closes https://github.com/goreleaser/goreleaser/issues/1314


caarlos0:
not stale

caarlos0:
still thinking about this one

caarlos0:
I think I rather not do this.

Issue #1398, opened 2020-03-22
Gorelease fails to create an archive when the builds > binary field contains a sub-folder


andrewkav:
**Describe the bug**
Goreleaser fails to create an archive when binary contains a sub folder in builds scetion.
```
 error=failed to create directory dist/test/test_1.3.0-test_Linux_x86_64.tar.gz: open dist/test/test_1.3.0-test_Linux_x86_64.tar.gz: no such file or directory
```

**To Reproduce**

```yaml
builds:
  - id: test
    dir: test
    binary: ./test/test
    goos:
      - linux
    goarch:
      - amd64
archives:
  - id: test
    name_template: "{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}"
    builds:
      - test
    replacements:
      linux: Linux
      amd64: x86_64

```

> Make sure your config file is valid by running
> `goreleaser check -f path-to-config-file`.

**Expected behavior**

It should not fail and create a subfolder.

here is a potential fix for `create` function in `internal/pipe/archive/archive.go`
```
	if err := os.MkdirAll(filepath.Dir(archivePath), 0755|os.ModeDir); err != nil {
		return err
	}
```

**Environment (please complete the following information):**

- OS: Ubuntu
- OS version: 19.10
- GoReleaser Version [goreleaser --version]: latest from master

caarlos0:
why are you trying to create the binary in a subfolder?

BTW: `binary` there is the binary name, not the path to the binary... changing this will break a lot of things I believe

andrewkav:
Hi @caarlos0 ,
```
    # Binary name.
    # Can be a path (e.g. `bin/app`) to wrap the binary in a directory.
    # Default is the name of the project directory.
    binary: program
```
the documentation says it's possible to have sub-folders, so the behavior contradicts the docs, unless I missed something.
I have a mono repo so I need this to be able to package resources correctly.
Thanks.

caarlos0:
hmmm, I did't recall that... yeah, in that case, seems like a bug ...

thanks for reporting, will check it out

PR #1399, opened 2020-03-22
fix: build.binary and artifact.extra.binary

caarlos0:
closes https://github.com/goreleaser/goreleaser/issues/1398

3 things:

1. make archive pipe work when `build.binary` is a path
2. `artifact.extra[binary]` should not have the path to the binary, just its name...
3. scoop pipe was using the `artifact.extra[binary]` field instead of just `artifact.name`


internal/pipe/archive/archive.go
```diff
@@ -117,6 +117,9 @@ func create(ctx *context.Context, archive config.Archive, binaries []*artifact.A
 	}
 	archivePath := filepath.Join(ctx.Config.Dist, folder+"."+format)
 	lock.Lock()
+	if err := os.MkdirAll(filepath.Dir(archivePath), 0755|os.ModeDir); err != nil {
+		return err
```
andrewkav: hey, I think you missed `lock.Unlock()` here

caarlos0: thanks!

Issue #1404, opened 2020-03-30
net/http fails on ContentLength vs Body length on uploads


greut:
**Describe the bug**

I'm trying to upload to a Nexus Repository and it fails doing 502 in HTTPS, and complaining about the body size vs the content-length in HTTP.

**To Reproduce**

```yaml
uploads:
  - name: nexus
    mode: archive
    target: http://repo.ch/nexus/repository/devops/{{ .ProjectName }}/{{ .Version }}/
```

**Expected behavior**

A nice upload :)

**Environment (please complete the following information):**

- OS: WSL (so Linux, but Windows)
- OS version: Linux W100154 4.4.0-17763-Microsoft #864-Microsoft Thu Nov 07 15:22:00 PST 2019 x86_64 x86_64 x86_64 GNU/Linux
- GoReleaser Version 0.129.0 (529cfca87d49cd93cdd7f8abd6eea6489c7cbd2e)

**Additional context**

```
      • http upload
         • will upload 4 artifacts
         • generated target url: http://repo.ch/nexus/repository/devops/bender/0.2.5/bender_0.2.5_linux_amd64.tar.gz
         • executing request: GET http://repo.ch/nexus/repository/devops/bender/0.2.5/bender_0.2.5_linux_amd64.deb (headers: map[Authorization:[Basic xxx]])
         ⨯ upload: upload failed     error=Get "http://repo.ch/nexus/repository/devops/bender/0.2.5/bender_0.2.5_linux_amd64.tar.gz": net/http: HTTP/1.x transport connection broken: http: ContentLength=5199803 with Body length 4194304 instance=nexus username=yoan
```

greut:
NB: in binary mode, it worked just fine.

greut:
Binary mode says:

```
      • http upload
         • will upload 0 artifacts

```

greut:
PEBCAK.. it's the GET. I'll do some docs :wink:

PR #1406, opened 2020-03-30
docs: a name key is required

greut:
Change the HTTP upload example to include the `name` key.

I banged my head on it for a while when trying to upload to Nexus using archive, then binary. Apparently it wants to use GET for some reasons.

#1404


caarlos0:
thanks!

PR #1407, opened 2020-03-30
refactor: simplify retries

caarlos0:
make it simpler and depending on one less library.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,6 @@ require (
 	github.com/goreleaser/nfpm v1.2.1
 	github.com/imdario/mergo v0.3.9
 	github.com/jarcoal/httpmock v1.0.5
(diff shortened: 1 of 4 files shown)


caarlos0:
fixes #1410 probably

kamilsk:
:(

the origin problem of the #1410 was 7661587a9e7be869131a6e50a77685169c8222f6 and https://github.com/goreleaser/goreleaser/blob/cca25688d07bba57c810786b9300b305279a55ab/internal/client/client.go#L48

it appeared with https://github.com/goreleaser/goreleaser/commit/0736162d5efc262b324b9416b10750cd50182c41#diff-2c846caea57b2b1d6e49c6417f337625

caarlos0:
yes, but anyway we just do a linear sleep basically... its not personal or anything, just easier to understand this way...

Issue #1328, opened 2020-02-02
build hooks


caarlos0:
check https://github.com/goreleaser/goreleaser/issues/1261 for details

it seems like something useful to have, for example, to run `upx` against the binaries 🤔

radeksimko:
I feel this would also address #1411 

--- 

What do you think about new `each` and `all` configuration options under `after` - that would satisfy the current use case of 1 hook per build ID and also provide some flexibility where folks want to run hook for each binary

```yml
builds:
  -
    id: signable
    targets:
      - darwin_amd64
      - windows_386
      - windows_amd64
    after:
      each:
        - upx {{ .Path }}
      all:
        - zip ./signable_binaries.zip {{ range .Binaries }} {{.Path}}{{ end }}
        - ./codesign_binaries_in_zip -src=./signable_binaries.zip -dst=./signed_binaries.zip
        - unzip ./signed_binaries.zip
```
Technically these could also be combined with the assumption that `each` hooks run in parallel, right after the relevant build and `all` runs after all individual hooks are finished.

Would you be open for a PR that implements something like this? ^

caarlos0:
fixed by #1414 right?

radeksimko:
> fixed by #1414 right?

Yes.

--- 

I feel it would good to open a separate discussion about restricting the inheritance of environment variables. I intentionally didn't make it too restrictive in my implementation to avoid more breaking changes than necessary, but I also feel that this is not ideal from security/isolation perspective.

So I'm thinking this could be considered as a reasonable breaking change for 1.0 - would you like me to open a separate issue for that?

caarlos0:
> So I'm thinking this could be considered as a reasonable breaking change for 1.0 - would you like me to open a separate issue for that?

yes please, I'm also a bit concerned on how it end up... pretty messy... there is even a PR opened for like a month which I'm not sure is a good idea merging :(

caarlos0:
created #1443 

will close this one.

thaanks 🙏

PR #1414, opened 2020-03-31
feat(pipe/build): Add support for flexible build hooks

radeksimko:
This is an early prototype of a solution for #1328 to solicit early feedback.

### Usage

```yml
builds:
  -
    env:
      - CGO_ENABLED=0
    targets:
      - darwin_amd64
      - windows_386
      - windows_amd64
    after:
      all:
        hooks:
          - zip ./signable_binaries.zip {{ range .Builds }} {{ .Path }}{{ end }}
          - echo "Signing every eligible binary..."
          - cmd: ./codesign.sh
            env:
              - ZIP_FILE={{ .ProjectName }}_{{ .Version }}_{{ .Target }}.zip
      each:
        default_dir: "{{ .Path }}"
        hooks:
          - upx {{ .Name }}
```

--- 

I would be happy to add some tests, perhaps some more logging and docs if we agree on the rough direction.

### DIFF
--- a/internal/pipe/build/build_test.go
+++ b/internal/pipe/build/build_test.go
@@ -79,7 +79,9 @@ func TestBuild(t *testing.T) {
 		Version: "1.2.3",
 		Config:  config,
 	}
-	error := doBuild(ctx, ctx.Config.Builds[0], "darwin_amd64")
+	opts, err := buildOptionsForTarget(ctx, ctx.Config.Builds[0], "darwin_amd64")
+	assert.NoError(t, err)
+	error := doBuild(ctx, ctx.Config.Builds[0], *opts)
 	assert.NoError(t, error)
 }
 
@@ -117,9 +119,13 @@ func TestRunFullPipe(t *testing.T) {
 				Binary:  "testing",
 				Flags:   []string{"-v"},
 				Ldflags: []string{"-X main.test=testing"},
-				Hooks: config.Hooks{
-					Pre:  "touch " + pre,
-					Post: "touch " + post,
+				Hooks: config.HookConfig{
+					Pre: []config.BuildHook{
+						{Cmd: "touch " + pre},
+					},
+					Post: []config.BuildHook{
+						{Cmd: "touch " + post},
+					},
 				},
 				Targets: []string{"whatever"},
 			},
@@ -148,9 +154,13 @@ func TestRunFullPipeFail(t *testing.T) {
 				Binary:  "testing",
 				Flags:   []string{"-v"},
 				Ldflags: []string{"-X main.test=testing"},
-				Hooks: config.Hooks{
-					Pre:  "touch " + pre,
-					Post: "touch " + post,
+				Hooks: config.HookConfig{
+					Pre: []config.BuildHook{
+						{Cmd: "touch " + pre},
+					},
+					Post: []config.BuildHook{
+						{Cmd: "touch " + post},
+					},
 				},
 				Targets: []string{"whatever"},
 			},
@@ -166,29 +176,29 @@ func TestRunFullPipeFail(t *testing.T) {
 func TestRunPipeFailingHooks(t *testing.T) {
 	folder, back := testlib.Mktmp(t)
 	defer back()
-	var config = config.Project{
+	var cfg = config.Project{
 		Dist: folder,
 		Builds: []config.Build{
 			{
 				Lang:    "fake",
 				Binary:  "hooks",
-				Hooks:   config.Hooks{},
+				Hooks:   config.HookConfig{},
 				Targets: []string{"whatever"},
 			},
 		},
 	}
 	t.Run("pre-hook", func(t *testing.T) {
-		var ctx = context.New(config)
+		var ctx = context.New(cfg)
 		ctx.Git.CurrentTag = "2.3.4"
-		ctx.Config.Builds[0].Hooks.Pre = "exit 1"
-		ctx.Config.Builds[0].Hooks.Post = "echo post"
+		ctx.Config.Builds[0].Hooks.Pre = []config.BuildHook{{Cmd: "exit 1"}}
+		ctx.Config.Builds[0].Hooks.Post = []config.BuildHook{{Cmd: "echo post"}}
 		assert.EqualError(t, Pipe{}.Run(ctx), `pre hook failed: "": exec: "exit": executable file not found in $PATH`)
 	})
 	t.Run("post-hook", func(t *testing.T) {
-		var ctx = context.New(config)
+		var ctx = context.New(cfg)
 		ctx.Git.CurrentTag = "2.3.4"
-		ctx.Config.Builds[0].Hooks.Pre = "echo pre"
-		ctx.Config.Builds[0].Hooks.Post = "exit 1"
+		ctx.Config.Builds[0].Hooks.Pre = []config.BuildHook{{Cmd: "echo pre"}}
+		ctx.Config.Builds[0].Hooks.Post = []config.BuildHook{{Cmd: "exit 1"}}
 		assert.EqualError(t, Pipe{}.Run(ctx), `post hook failed: "": exec: "exit": executable file not found in $PATH`)
 	})
 }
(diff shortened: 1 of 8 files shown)


caarlos0:
I don't get the all vs each thing 🤔

radeksimko:
> I don't get the all vs each thing

Good question - the idea is that you may want to do something after `all` the builds finish (e.g. ZIP all binaries from that "build group" with the same build ID), or - perhaps more commonly (?) - do something after each build within that group.

I'm not sure how common these scenarios are, so I assume we could ship just `each` initially?

I'm open to ideas.

caarlos0:
yeah I think `each` is more common...

radeksimko:
That is fair - I'm happy to reduce the implementation to just `each` scenario.

Would you also then simplify the configuration - with the risk that adding `all` in the future may require some deprecation? I assume it's low risk and the UX benefit outweighs here?

So we may be able to leverage the existing hierarchy and avoid deprecations:

```yml
builds:
  -
    env:
      - CGO_ENABLED=0
    targets:
      - darwin_amd64
      - windows_386
      - windows_amd64
    hooks:
      post_default_dir: "{{ .Path }}"
      post:
          - upx {{ .Name }}
```

Is there anything else you feel is confusing or needs changing, or would you design it entirely differently? 😃

caarlos0:
maybe we can do something like:

```yaml
builds:
- main: main.go
  hooks:
    pre: ...
    post: ...
  each_hooks:
    # pre: ignore and throw a warning if != ""
    post: ...
```

this way we don't need to change the old hooks and can still add the each behavior...


wdyt?

radeksimko:
Yep, I think we could leverage the existing hierarchy that way. 👍 

The only downside of the current implementation of `hooks` is that it allows just a single command, so we'd probably need to implement custom unmarshaler for `hooks.pre` and `hooks.post` so that we retain the existing functionality of single command given as `string` and a slice of commands, where each can have its own isolated ENV variables and workdir. Certainly possible to implement, as long as we're ok with the UX of `pre` and `post` accepting both array and string.

Relatedly the current implementation is not concerned about workdir and I feel that having a default workdir may be beneficial to config simplicity, instead of repeating it in every command, or calling `cd` repeatedly.

--- 

Isolated environments are IMO better from security perspective, although there isn't really a way of passing through sensitive variables as they'd have to be hard-coded - but I didn't want to bike shed on this as I feel that's a problem that could be discussed separately.

caarlos0:
> The only downside of the current implementation of `hooks` is that it allows just a single command, so we'd probably need to implement custom unmarshaler for `hooks.pre` and `hooks.post` so that we retain the existing functionality of single command given as `string` and a slice of commands, where each can have its own isolated ENV variables and workdir. Certainly possible to implement, as long as we're ok with the UX of `pre` and `post` accepting both array and string.

yes, we'll need a custom unmarshaler... we can also add the cwd config... or maybe its really easier to just add a `EachHooks` or something haha

anyway, multiple commands could be run by passing a path to a shell script instead 🤔 

> Isolated environments are IMO better from security perspective, although there isn't really a way of passing through sensitive variables as they'd have to be hard-code

maybe you can use `{{ .Env.NAME }}` for that?

radeksimko:
> maybe you can use {{ .Env.NAME }} for that?

True, I'm not sure how/why I missed that in the docs!

> anyway, multiple commands could be run by passing a path to a shell script instead

True, it's just a little more plumbing and it hurts config reusability - so I suppose it depends on what the common scenario is and whether we can optimize for it. Will people more often run just a single command, or will they need to wrap things in shell script for common scenarios?

I personally see multiple commands more commonly around me, but I may have biased view 😄

caarlos0:
> True, it's just a little more plumbing and it hurts config reusability - so I suppose it depends on what the common scenario is and whether we can optimize for it. Will people more often run just a single command, or will they need to wrap things in shell script for common scenarios?
> 
> I personally see multiple commands more commonly around me, but I may have biased view 😄

yeah idk about that either... I usually use 0 commands hehe

but... i would probably put in a script so I can use `if` and etc if needed... maybe for simplicity we can assume that?

internal/pipe/build/build_test.go
```diff
@@ -79,7 +79,11 @@ func TestBuild(t *testing.T) {
 		Version: "1.2.3",
 		Config:  config,
 	}
-	error := doBuild(ctx, ctx.Config.Builds[0], "darwin_amd64")
+	opts, err := buildOptionsForTarget(ctx, ctx.Config.Builds[0], "darwin_amd64")
+	if err != nil {
```
oleg-balunenko, line 83: require.NoError could be used from testify lib.
https://pkg.go.dev/github.com/stretchr/testify/require?tab=doc#NoError

radeksimko:
I believe I reflected most feedback in the prototype. It now leverages the existing config option and allows a few different levels of complexity:

```yml
builds:
- main: main.go
  hooks:
    pre: ./script.sh {{ .Path }}
    post: ./script.sh {{ .Path }}
```
```yml
builds:
- main: main.go
  hooks:
    pre:
      - ./script.sh {{ .Path }}
      - ./another-script.sh {{ .Name }}
    post:
      - ./script.sh {{ .Target }}
      - ./another-script.sh {{ .Target }}
```
```yml
builds:
- main: main.go
  hooks:
    pre:
      - cmd: ./script.sh {{ .Path }}
        env:
          - VAR={{ .Env.VAR }}
      - ./another-script.sh
```

I decided to keep the commands-as-array option, as I feel that GoReleaser itself has a great benefit of cross-OS reusability of config (as a result of Go cross-compilation abilities) and it would be IMO shame if users were forced to make hooks Linux/Windows-specific or generally deal with that problem of OS differences.

Should that be a blocker though I can step back from that decision, in the interest of getting _something_ in.

--- 

One important thing to note is that this _could_ in theory be perceived as a breaking change by some users as the existing hooks are executed just once before and after, not for each build target.

caarlos0:
> One important thing to note is that this _could_ in theory be perceived as a breaking change by some users as the existing hooks are executed just once before and after, not for each build target.

this is a problem IMHO, but to be fair, I don't know if anyone uses that...

caarlos0:
I did take a look on the code and it lgtm

radeksimko:
@caarlos0 Thank you for the review, I will try to increase the code coverage a bit, add some docs and mark it as ready by the end of the day.

> this is a problem IMHO, but to be fair, I don't know if anyone uses that...

It sounds gross and a bit too magical in my mind 😬 but I will share that idea here anyway:

We could theoretically make the new behaviour opt-in by assuming that templating syntax wasn't supported before, so we'd only run hook for each target when there's valid Go template syntax (`{{}}`) found in the hook. In theory though people may still want the new behaviour (run per target) even _without_ templating syntax - I can't think of such use case, but there may be some?

I don't have strong opinions on this as I haven't used these hooks before myself, so it's not breaking for me. If you feel strongly this should be postponed for 1.0, or you have a better idea on how to introduce this sensibly without breaking stuff, I'd be happy to implement it.

caarlos0:
for some reason I can't see the logs on travis 🤔

caarlos0:
it looks good to me, thanks @radeksimko for the awesome work on this PR 🚀

caarlos0:
I can't see the build logs though, will merge and fix it if something breaks :)

radeksimko:
@caarlos0 Let me try to rebase it first - that may re-trigger the Travis job?

caarlos0:
@radeksimko sorry, already did hehe

anyway, I trying re-trigging it on travis as well.. doesn't work... happened with other jobs too... seems like something on travis 🤔

caarlos0:
(already rebase but feel free to check if I didn't break anything 🙏)

caarlos0:
BTW now that I did a commit, I can see the logs 🤣

radeksimko:
I will fix the two typos in a sec

> www/content/build.md:144:11: "explicitely" is a misspelling of "explicitly"
> www/content/build.md:194:40: "overriden" is a misspelling of "overridden"

caarlos0:
Thanks!

caarlos0:
Build passed, thanks again @radeksimko for the hard work on this 🙏

PR #1416, opened 2020-04-01
feat: use maximum compression for gzip and zip archives

jftuga:
I have read the CONTRIBUTING guide.

If applied, this commit will use maximum compression levels for gzip and zip archives instead of the default compression level.

This change will speed up uploading archive files created by goreleaser to GitHub and also consume less disk space.

### DIFF
--- a/pkg/archive/targz/targz.go
+++ b/pkg/archive/targz/targz.go
@@ -25,6 +25,7 @@ func (a Archive) Close() error {
 
 // New tar.gz archive
 func New(target io.Writer) Archive {
+	// the error will be nil since the compression level is valid
 	gw, _ := gzip.NewWriterLevel(target, gzip.BestCompression)
 	tw := tar.NewWriter(gw)
 	return Archive{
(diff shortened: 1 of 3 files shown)


pkg/archive/targz/targz.go
```diff
@@ -25,7 +25,7 @@ func (a Archive) Close() error {
 
 // New tar.gz archive
 func New(target io.Writer) Archive {
-	gw := gzip.NewWriter(target)
+	gw, _ := gzip.NewWriterLevel(target, gzip.BestCompression)
```
caarlos0, line 28: should we add a comment saying the error will be nil since the compression level is valid on both usages of this? 

just so its clear why the error is being ignored 🤔

caarlos0:
thanks for the PR, just a single comment above :)

pkg/archive/targz/targz.go
jftuga: I think add a comment is a good idea.  Do I need to submit another PR with a comment?

jftuga: I added a comment about this.

PR #1419, opened 2020-04-02
feat: build.skip option, support for library projects

### LINKED ISSUE #981: Support for library projects?
I've used goreleaser in the past for projects with binaries and I like it. I'm now working on a Go library  (i.e. no `main` function) and I would still like to use goreleaser to produce nice release notes and source zips/tars.

Is there any interest in supporting this use case? Based on #415 and #418, it suggests not. What about a flag in the config that disables builds?

leogr:
This PR introduces the `skip` option to the build configuration. That is mainly relevant for a library project which would need the changelog generation and creating the release, but not the building step.

With this feature, also simple library projects can benefit from `goreleaser`. 

Configuration example for a library project:
```yaml
project_name: mylib-go
build:
  skip: true
release:
  github:
  prerelease: auto
```

Fixes #981

### DIFF
--- a/internal/pipe/build/build.go
+++ b/internal/pipe/build/build.go
@@ -32,6 +32,10 @@ func (Pipe) String() string {
 // Run the pipe
 func (Pipe) Run(ctx *context.Context) error {
 	for _, build := range ctx.Config.Builds {
+		if build.Skip {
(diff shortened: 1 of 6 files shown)


caarlos0:
This looks great! Thanks!

PR #1429, opened 2020-04-09
test: Allow tests to run in parallel without failing

radeksimko:
Despite the comment in code I don't believe there is anything Travis-specific about the nature of the failure. I managed to reproduce (and fix) the failure locally on my MacOS.

All tests share the same environment (hence variable namespace too) and setting & reading the same variables has lead to race conditions which can be fixed by using different variables in each test.


caarlos0:
build failure seems unrelated, will look later

thanks for the pr though :)

PR #1439, opened 2020-04-12
docs: Update signing section to reflect build hooks

### LINKED ISSUE #1411: More flexible signing interface
I'm attempting to sign MacOS & Windows binaries by sending them over to a separate (centralised) CI pipeline for signing and downloading signed ones back from there.

I admit this may not be a common scenario, but I thought I'd open a conversation here to see if you were interested in supporting it somehow, also because I feel that some of the problems described here apply to people who sign locally with `gon` anyway.

--- 

There is a few problems I ran into while attempting to implement this, all related to the current signing interface (as represented by `signs` section of the config):

1. It is assumed that signature will be stored in a separate file such as `<artifact>.sig` - this is common for GPG-based signing, but not suitable for MacOS or Windows signing, where the binary is basically modified in place. I suppose this can be overwritten via `signature: "${artifact}"` but it seems hacky, or non-intuitive at the least as GoReleaser doesn't seem to have a notion of `SignedBinary` or `SignedArchive` artifact type, just `Signature`.
2. While there is filtering of artifacts there is no way to filter _just archives_ (to avoid notarizing checksum files) - the only options available are `none`, `all` and `checksums`
3. It is assumed that signing takes place _after_ archival, so the only artifacts available are actually archives, rather than binaries
4. It is assumed that checksum calculation takes place before signing, so checksums won't match for archives which have binaries replaced with signed ones.

radeksimko:
Closes #1411

I'm marking this as a draft for now as I still need to test this example actually works (e.g. if gon actually interpolates ENV variables in that section of config), but I'm sharing it early in case there's any feedback on the wording itself.


caarlos0:
Just read it, lgtm so far! Thanks 🤟

radeksimko:
I realized the interpolation isn't supported in `source` currently:
https://github.com/mitchellh/gon/issues/20

So we could recommend just hardcoding `source = ["./dist/ID_TARGET/expected-binary-name"]` for now, until it is supported. It's not ideal as it exposes implementation details of GoReleaser, but it is what gon itself [does](https://github.com/mitchellh/gon/blob/master/.gon.hcl#L1) today anyway and at least the guide would become a bit clearer 😄

caarlos0:
yeah sounds good to me!

thanks again @radeksimko

caarlos0:
thanks 🙏

PR #1445, opened 2020-04-15
Login to GitHub/GitLab Docker Registry automatically

### LINKED ISSUE #1431: Login to GitHub/GitLab Docker Registry automatically
Both GitHub and GitLab offer their own public Docker Registry:
https://github.com/features/packages
https://docs.gitlab.com/ee/user/packages/container_registry/

Gitea doesn't seem offer one [yet](https://github.com/go-gitea/gitea/issues/2316).

It would be great if GoReleaser could just automatically authenticate with the given registry on hostname/image match if it already has all it needs for authentication (token), e.g.

```yml
dockers:
  -
    goos: linux
    goarch: amd64
    image_templates:
      - "docker.pkg.github.com/org/repo/image-name:latest"
      - "docker.pkg.github.com/org/repo/image-name:{{ .Tag }}"
      - "docker.pkg.github.com/org/repo/image-name:v{{ .Major }}"
```
would automatically trigger logging into GitHub's Docker Registry if `GITHUB_TOKEN` is available - which is high chance, as uploading to GitHub's Registry from GitLab or Gitea seems like unlikely scenario.

Logging in as a separate GitHub Action step:

```yml
...
jobs:
  release:
    runs-on: ubuntu-latest
    steps:
...
      -
        name: Login to Docker Registry
        run: |
          echo "${{ secrets.GITHUB_TOKEN }}" | \
          docker login docker.pkg.github.com -u docker --password-stdin
      -
        name: Release
        uses: goreleaser/goreleaser-action@v1
        with:
          version: latest
          args: release --rm-dist
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

**Additional context**

C123R:
If applied this will fix #1431


caarlos0:
I was thinking in a simpler alternative: 

- if there is a github token, `docker login` in the github registry
- if there is a gitlab token, `docker login` in the gitlab registry 

maybe made it optional though...

C123R:
> I was thinking in a simpler alternative:
> 
> * if there is a github token, `docker login` in the github registry
> * if there is a gitlab token, `docker login` in the gitlab registry
> 
> maybe made it optional though...
> 
> wdyt?

Sorry, I didn't get you. You mean if there is a GITHUB_TOKEN env set, log in to GitHub registry?
```yaml
dockers:
  - image_templates:
      - docker.pkg.github.com/c123r/test:latest
      - registry.example.com/c123r/test:latest
```
if that's the case, then technically it's only possible to push the docker image to `docker.pkg.github.com` and `registry.example.com` won't work.

With current commit, it's possible to push it to Github as well using DOCKER_USERNAME and DOCKER_PASSWORD to push it to third party registry.

I think it's a bit complicated for now but would be helpful as we are supporting multiple images.

C123R:
> > I was thinking in a simpler alternative:
> 
> > 
> 
> > * if there is a github token, `docker login` in the github registry
> 
> > * if there is a gitlab token, `docker login` in the gitlab registry
> 
> > 
> 
> > maybe made it optional though...
> 
> > 
> 
> > wdyt?
> 
> 
> 
> Sorry, I didn't get you. You mean if there is a GITHUB_TOKEN env set, log in to GitHub registry?
> 
> ```yaml
> 
> dockers:
> 
>   - image_templates:
> 
>       - docker.pkg.github.com/c123r/test:latest
> 
>       - registry.example.com/c123r/test:latest
> 
> ```
> 
> if that's the case, then technically it's only possible to push the docker image to `docker.pkg.github.com` and `registry.example.com` won't work.
> 
> 
> 
> With current commit, it's possible to push it to Github as well using DOCKER_USERNAME and DOCKER_PASSWORD to push it to third party registry.
> 
> 
> 
> I think it's a bit complicated for now but would be helpful as we are supporting multiple images.
> 
> 
> 
> 

Btw, I was bit confused 😀 that’s the reason I created this draft version to check if we are on same page

caarlos0:
AFAIK you can use `docker login` to be logged in to several registries... 

also, that reminds me, maybe this should be done on the docker image instead? 

check https://github.com/goreleaser/goreleaser/blob/master/scripts/entrypoint.sh

C123R:
> AFAIK you can use `docker login` to be logged in to several registries...

Sure, I know but I thought of using `DOCKER_CONFIG` env var instead of using `docker login` command, also used it in Github [docker-login](https://github.com/Azure/docker-login
) action.

internal/pipe/docker/docker.go
```diff
@@ -260,6 +269,70 @@ func dockerBuild(ctx *context.Context, root string, images, flags []string) erro
...
+
+	dockerConfigFile, err := os.Create(dockerConfigPath)
+	if err != nil {
+		return errors.Wrap(err, "failed to create docker registry config")
+	}
+	defer dockerConfigFile.Close()
+
+	switch dockerRegistry {
+	case "docker.pkg.github.com":
+		authToken = encodeToken("docker",  os.Getenv("GITHUB_TOKEN"))
```
radeksimko, line 302: In theory the user may also decide to push images to a registry which isn't associated with the repository where the GitHub Action runs from. So while it's reasonable to expect `GITHUB_TOKEN` to be valid in such case, it will only have access to the origin registry - so some further filtering based on image path (not just the registry hostname) may be necessary here, to avoid logging into registries that the token doesn't have access to.

C123R: @radeksimko Good Point. Prior to login, we could verify the image path using predefined variables `GITHUB_REPOSITORY` or `CI_REGISTRY_IMAGE`(GITLAB)

caarlos0:
Can't we do this in the entrypoint, as it is already with docker hub?

refs: https://github.com/goreleaser/goreleaser/blob/master/scripts/entrypoint.sh

caarlos0:
I rather leave this out for now, users can authenticate their own way outside of goreleaser... 

pr #1703 adds the login to our docker image, which has more limited effects, but may help users with this problem as well.

PR #1450, opened 2020-04-17
fix: correct the spelling of "snapshotting"

jaxxstorm:
Small and relatively insignificant change, but the correct spelling is
with two "t's"


caarlos0:
thanks

PR #1466, opened 2020-04-23
fix: runHook shell commands need to have backslashes escaped under Wi…

jftuga:
If applied, this commit will fix Pre /Post Hooks under Windows.  Before this patch, a post hook might look something like this (the `cmd` variable in `build.go`) and then fail:

    upx -9 C:github.comjftugagofwddisttest_windows_amd64gofwd.exe

After this patch, the `cmd` variable now looks like this and succeeds:

    upx -9 C:\github.com\jftuga\gofwd\dist\test_windows_amd64\gofwd.exe


internal/pipe/build/build.go
```diff
@@ -134,6 +135,11 @@ func runHook(ctx *context.Context, opts builders.Options, buildEnv []string, hoo
 			return err
 		}
 
+		// shell commands need to have backslashes escaped under Windows
+		if "windows" == runtime.GOOS {
```
caarlos0, line 139:
```suggestion
		if runtime.GOOS == "windows" {
```

jftuga: I swapped these in this commit: `swap l-value and r-value`

jftuga:
> maybe even better would be to move that function to a `run_unix.go` and create a `run_windows.go` with the backslash changes... or something else like that...

I wrote some code using `run_unix.go` and `run_windows.go` files by using `// +build !windows` and `// +build windows`.  Each file contains a `run()` definition. By the time `command` reaches this function, it already has the backslashes stripped out.  The original reason I placed this code where it is, is that I believe that `shellwords.Parse()` strips the first set of backslashes, but thus leaving the second set of backslashes because I have just doubled them.

caarlos0:
I don't have any windows boxes here, can you confirm if it is a problem in shellwords? If so, Its probably better we open an issue/PR there...

jftuga:
It looks like `mattn/go-shellwords` has been dormant for about 2 months.

jftuga:
This is now fixed in `shellwords`

https://github.com/mattn/go-shellwords/pull/39

radeksimko:
@jftuga The PR you linked to is still pending review - i.e. it's not fixed in `master` yet.

Aside from that it's worth noting that GoReleaser itself is currently not being tested on Windows.

I just tried to run all tests on a Windows VM to discover [test failures](https://gist.github.com/radeksimko/35c9e3f2bdea4dbc58472d1e2ef8c9be). Many of these may be just related to Windows-incompatible test data, but some may actually indicate bugs.

So just to set the expectations right - there's likely much more work to do before GoReleaser genuinely works on Windows.

caarlos0:
check #1747

PR #1467, opened 2020-04-24
feat(scoop): provide config option to change commit message

zlesnr:
Without this change, users unable to control the resulting commit message of the
scoop update.  In some environments this may present an issue with commit
linters that require a specific commit message format in order to build proper
change logs and make decisions.  Here we include a Scoop config option to use a
format string provided by the user during the commit.

resolves: https://github.com/goreleaser/goreleaser/issues/1468

### DIFF
--- a/internal/pipe/scoop/scoop.go
+++ b/internal/pipe/scoop/scoop.go
@@ -54,6 +54,10 @@ func (Pipe) Default(ctx *context.Context) error {
 		ctx.Config.Scoop.CommitAuthor.Email = "goreleaser@carlosbecker.com"
 	}
 
+	if ctx.Config.Scoop.CommitMessageTemplate == "" {
+		ctx.Config.Scoop.CommitMessageTemplate = "Scoop update for {{ .ProjectName }} version {{ .Tag }}"
+	}
+
 	return nil
 }
 
@@ -104,13 +108,20 @@ func doRun(ctx *context.Context, client client.Client) error {
 	if ctx.Config.Release.Disable {
 		return pipe.Skip("release is disabled")
 	}
+
+	commitMessage, err := tmpl.New(ctx).
+		Apply(ctx.Config.Scoop.CommitMessageTemplate)
+	if err != nil {
+		return err
+	}
+
 	return client.CreateFile(
 		ctx,
 		ctx.Config.Scoop.CommitAuthor,
 		ctx.Config.Scoop.Bucket,
 		content.Bytes(),
 		path,
-		fmt.Sprintf("Scoop update for %s version %s", ctx.Config.ProjectName, ctx.Git.CurrentTag),
+		commitMessage,
 	)
 }
 
--- a/www/content/scoop.md
+++ b/www/content/scoop.md
@@ -31,6 +31,9 @@ scoop:
     name: goreleaserbot
     email: goreleaser@carlosbecker.com
 
+  # The project name and current git tag are used in the format string.
+  commit_msg_template: "Scoop update for {{ .ProjectName }} version {{ .Tag }}"
+
   # Your app's homepage.
   # Default is empty.
   homepage: "https://example.com/"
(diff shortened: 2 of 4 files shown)


internal/pipe/scoop/scoop.go
```diff
@@ -101,13 +101,21 @@ func doRun(ctx *context.Context, client client.Client) error {
...
+	}
+
 	return client.CreateFile(
 		ctx,
 		ctx.Config.Scoop.CommitAuthor,
 		ctx.Config.Scoop.Bucket,
 		content.Bytes(),
 		path,
-		fmt.Sprintf("Scoop update for %s version %s", ctx.Config.ProjectName, ctx.Git.CurrentTag),
+		fmt.Sprintf(commitFormat, ctx.Config.ProjectName, ctx.Git.CurrentTag),
```
zlesnr, line 118: Probably want to document to the user that two `%s` should be set here.

www/content/scoop.md
```diff
@@ -31,6 +31,9 @@ scoop:
     name: goreleaserbot
     email: goreleaser@carlosbecker.com
 
+  # The project name and current git tag are used in the format string.
+  commit_format: "chore(scoop): update %s version %s"
```
jthurman42, line 35: This should the templating formatter

`commit_format: "chore(scoop): update {{ . ProjectName }} version {{ . Version }}"`

caarlos0: agreed!

internal/pipe/scoop/scoop.go
```diff
@@ -101,13 +101,21 @@ func doRun(ctx *context.Context, client client.Client) error {
 	if ctx.Config.Release.Disable {
 		return pipe.Skip("release is disabled")
 	}
+
+	var commitFormat string
+	if ctx.Config.Scoop.CommitFormat == "" {
```
caarlos0, line 106: this should be done in the `Default` method and should also use templates as @jthurman42 pointed out

zlesnr: Moved to default and updated to template stings.

www/content/scoop.md
zlesnr: Updated to template strings, but have a look, just passing the `ctx` context in.

zlesnr:
Thanks for the pointers.  I've updated it a bit to use the templating system that I missed at first.  Looks pretty neat.

internal/pipe/scoop/scoop.go
```diff
@@ -101,13 +105,20 @@ func doRun(ctx *context.Context, client client.Client) error {
 	if ctx.Config.Release.Disable {
 		return pipe.Skip("release is disabled")
 	}
+
+	commitMessage, err := tmpl.New(ctx).
+		Apply(ctx.Config.Scoop.URLTemplate)
```
jthurman42, line 110: `ctx.Config.Scoop.CommitTemplate`

www/content/scoop.md
```diff
@@ -31,6 +31,9 @@ scoop:
     name: goreleaserbot
     email: goreleaser@carlosbecker.com
 
+  # The project name and current git tag are used in the format string.
+  commit_format: "chore(scoop): update {{ .Config.ProjectName }} version {{ .Git.CurrentTag }}"
```
jthurman42, line 35: Should be `commit_template:`.  Also add a comment above with the default (or set the example to the default)

```yaml
  # Git commit message template used for Scoop
  # Defaults are shown.
  commit_template: "Scoop update for {{ .ProjectName }} version {{ .Version }}"
```

www/content/scoop.md
```diff
@@ -31,6 +31,9 @@ scoop:
     name: goreleaserbot
     email: goreleaser@carlosbecker.com
 
+  # The project name and current git tag are used in the format string.
+  commit_template: "Scoop update for {{ .ProjectName }} version {{ .Tag }}"
```
caarlos0, line 35: maybe rename it to `commit_msg_template`? 🤔

zlesnr: Renamed as requested.

zlesnr:
Not sure about that failing test.  Master seems to be failing for the same reason.

caarlos0:
yeah I'll take a look, thanks!

Issue #1436, opened 2020-04-12
refactor main.go


caarlos0:
main.go is already a pretty much untestable mess, we should probably refactor it.

Now that we have multiple commands, maybe using spf13/cobra makes more sense as well.

jftuga:
What advantages of cobra do you see vs kingpin?  Kingpin supports multiple commands.  I'm not partial to one or the other.

caarlos0:
I think the end code is better organized and also a bit easier to write tests to... 

also this influences me a bit towards that https://github.com/alecthomas/kingpin#contributions-only 

and the fact that most large/complex cmds use cobra as well...

PR #1472, opened 2020-04-26
feat: refactor main.go to spf13/cobra

caarlos0:
closes https://github.com/goreleaser/goreleaser/issues/1436


cmd/root.go
```diff
@@ -0,0 +1,96 @@
...
+		NewReleaseCmd().cmd,
+		NewCheckCmd().cmd,
+		NewInitCmd().cmd,
+	)
+
+	root.cmd = cmd
+	return root
+}
+
+func shouldPrependRelease(cmd *cobra.Command, args []string) bool {
```
caarlos0, line 72: this was end up being hacky... on kingpin we could make a command default... impossible by default on cobra it seems...

either way, this works 👍

PR #1481, opened 2020-04-30
feat: Add support for custom publishers

### LINKED ISSUE #1460: Support custom publishers
Currently there is many publishers available:

https://github.com/goreleaser/goreleaser/blob/5f2cf501e85d139c8a48ad9ec79cdc874b4c89ce/internal/pipe/publish/publish.go#L38-L42

while the HTTP Upload publisher looks fairly universal, it could be a lot of effort to publish to a custom, authenticated API endpoint, or even publish to a known endpoint (like S3) in a more opinionated/complex way (e.g. setting some extra HTTP headers for each object).

For that reason I'd like to propose creation of a custom publisher, which could offer more or less the same interface as the new build hooks, e.g.

```yml
publishers:
  -
    # Unique name of your publisher. Used for identification
    name: "custom"

    # IDs of the artifacts you want to publish
    ids:
     - foo
     - bar

    # Publish checksums (defaults to false)
    checksum: true

    # Publish signatures (defaults to false)
    signature: true

    # Working directory in which to execute the command
    dir: "/utils"

    # Command to be executed
    cmd: custom-publisher.sh -product={{ .ProjectName }} -version={{ .Version }} {{ .ArtifactName }}

    # Environment variables
    env:
      - API_TOKEN=secret-token
```

I'm willing to implement this, if there is agreement on the idea and direction in general.

radeksimko:
Closes #1460

This is functionally mostly ready and it has a decent test coverage too.

## Usage

It's worth noting that `env` has no inheritance of variables (unlike existing hooks), which has some tradeoffs:
 - some essential variables like `TMPDIR` or `PATH` may be expected and required by the executables that are launched
 - any sensitive variables remain scoped to GoReleaser and don't leak by default

We could perhaps pass through some safe variables like the ones mentioned above, to balance the negative effect of this?

```yml
publishers:
  -
    # Unique name of your publisher. Used for identification
    name: "custom"

    # IDs of the artifacts you want to publish
    ids:
     - foo
     - bar

    # Publish checksums (defaults to false)
    checksum: true

    # Publish signatures (defaults to false)
    signature: true

    # Working directory in which to execute the command
    dir: "/utils"

    # Command to be executed
    cmd: custom-publisher.sh -product={{ .ProjectName }} -version={{ .Version }} {{ .ArtifactPath }}

    # Environment variables
    env:
      - API_TOKEN=secret-token
```

### DIFF
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -1,0 +1,167 @@
+package exec
+
+import (
+	"fmt"
+	"os/exec"
+
+	"github.com/apex/log"
+	"github.com/goreleaser/goreleaser/internal/artifact"
+	"github.com/goreleaser/goreleaser/internal/logext"
+	"github.com/goreleaser/goreleaser/internal/pipe"
+	"github.com/goreleaser/goreleaser/internal/semerrgroup"
+	"github.com/goreleaser/goreleaser/internal/tmpl"
+	"github.com/goreleaser/goreleaser/pkg/config"
+	"github.com/goreleaser/goreleaser/pkg/context"
+	"github.com/mattn/go-shellwords"
+)
+
+func Execute(ctx *context.Context, publishers []config.Publisher) error {
+	if ctx.SkipPublish {
+		return pipe.ErrSkipPublishEnabled
+	}
+
+	for _, p := range publishers {
+		log.WithField("name", p.Name).Debug("executing custom publisher")
+		err := executePublisher(ctx, p)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func executePublisher(ctx *context.Context, publisher config.Publisher) error {
+	log.Debugf("filtering %d artifacts", len(ctx.Artifacts.List()))
+	artifacts := filterArtifacts(ctx.Artifacts, publisher)
+	log.Debugf("will execute custom publisher with %d artifacts", len(artifacts))
+
+	var g = semerrgroup.New(ctx.Parallelism)
+	for _, artifact := range artifacts {
+		artifact := artifact
+		g.Go(func() error {
+			c, err := resolveCommand(ctx, publisher, artifact)
+			if err != nil {
+				return err
+			}
+
+			return executeCommand(c)
+		})
+	}
+
+	return g.Wait()
+}
+
+func executeCommand(c *command) error {
+	log.WithField("args", c.Args).
+		WithField("env", c.Env).
+		Debug("executing command")
+
+	// nolint: gosec
+	var cmd = exec.CommandContext(c.Ctx, c.Args[0], c.Args[1:]...)
+	cmd.Env = c.Env
+	if c.Dir != "" {
+		cmd.Dir = c.Dir
+	}
+
+	entry := log.WithField("cmd", c.Args[0])
+	cmd.Stderr = logext.NewErrWriter(entry)
+	cmd.Stdout = logext.NewWriter(entry)
+
+	log.WithField("cmd", cmd.Args).Info("publishing")
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("publishing: %s failed: %w",
+			c.Args[0], err)
+	}
+
+	log.Debugf("command %s finished successfully", c.Args[0])
+	return nil
+}
+
+func filterArtifacts(artifacts artifact.Artifacts, publisher config.Publisher) []*artifact.Artifact {
+	filters := []artifact.Filter{
+		artifact.ByType(artifact.UploadableArchive),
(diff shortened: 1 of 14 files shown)


caarlos0:
Hi @radeksimko! Thanks for the great PR :) 

Looking at the code, everything seems good to me.

About your questions:

> * We could take the same shortcut, or resolve it once and for all.

I think we can do this on another PR... for now, maybe use the same behavior already on other things...

> * decide whether `ArtifactPath` (currently relative, e.g. `dist/project_1.0.0_darwin_amd64.zip`) should be absolute path, or whether we should introduce `abs` function

I vote for the `abs` function 🤔... changing that now will break a lot of tests I think...

radeksimko:
@caarlos0 Thank you for the early feedback. I have now resolved all TODOs as advised and added some docs.

PTAL.

internal/exec/exec.go
```diff
@@ -0,0 +1,167 @@
...
+	return nil
+}
+
+func filterArtifacts(artifacts artifact.Artifacts, publisher config.Publisher) []*artifact.Artifact {
+	var filter artifact.Filter
+	filters := append([]artifact.Filter{},
+		artifact.ByType(artifact.UploadableArchive),
+		artifact.ByType(artifact.UploadableFile),
+		artifact.ByType(artifact.LinuxPackage),
+		artifact.ByType(artifact.UploadableBinary))
```
caarlos0, line 87: could be simplified to:
```suggestion
	filters := []artifact.Filter{
		artifact.ByType(artifact.UploadableArchive),
		artifact.ByType(artifact.UploadableFile),
		artifact.ByType(artifact.LinuxPackage),
		artifact.ByType(artifact.UploadableBinary),
	}
```

internal/exec/exec.go
```diff
@@ -0,0 +1,167 @@
...
+
+	if publisher.Checksum {
+		filters = append(filters, artifact.ByType(artifact.Checksum))
+	}
+
+	if publisher.Signature {
+		filters = append(filters, artifact.ByType(artifact.Signature))
+	}
+
+	filter = artifact.Or(filters...)
```
caarlos0, line 97: filter could be declared here, so it would never be nil, and then the `if filter == nil {` above could be removed as well... 

```suggestion
	var filter = artifact.Or(filters...)
```

internal/logext/writer.go
```diff
@@ -16,3 +16,18 @@ func (t Writer) Write(p []byte) (n int, err error) {
 	t.ctx.Info(string(p))
 	return len(p), nil
 }
+
+// Writer writes with log.Error
+type ErrorWriter struct {
```
caarlos0, line 21: this does not look like its being used...

internal/tmpl/tmpl_test.go
```diff
@@ -152,6 +154,8 @@ func TestFuncMap(t *testing.T) {
 	var ctx = context.New(config.Project{
 		ProjectName: "proj",
 	})
+	wd, _ := os.Getwd()
```
caarlos0, line 157: maybe add an error check? not sure how getwd can fail, but if it does, tests bellow will also fail and it could be a bit harder to debug...

```suggestion
	wd, err := os.Getwd()
	require.NoError(t, err)
```

internal/exec/exec.go
radeksimko: Indeed - I don't know why did I do it in such complicated way! 😄

internal/logext/writer.go
radeksimko: the struct itself could be un-exported if that's what you're suggesting, but otherwise it's used by `NewErrWriter` below, which is in turn called in `internal/exec/exec.go` to report stderr that way.

I'm open to ideas on how/whether to report stderr that way though - I just thought it should be distinguishable from stdout and reasonably formatted (which can be difficult if it's just part of an error).

internal/tmpl/tmpl_test.go
radeksimko: Indeed - good catch!

internal/exec/exec.go
radeksimko: Good point, and I'm happy to remove the `filter == nil` check, I just thought unfiltered list is better outcome in such situation than a crash regardless of whether `nil` is preventable? But I don't have strong opinion about this.

internal/logext/writer.go
caarlos0: oh, sorry, didn't notice it...

internal/exec/exec.go
caarlos0: yeah no problem at all :)

PR #1482, opened 2020-05-01
refactor(pipe/brew): Improve error reporting

radeksimko:
I ran into this generic error in the CI job which ran as part of https://github.com/goreleaser/goreleaser/pull/1481 - I'm unsure what's the root cause (it seems to be happening on `master` too), but I thought this patch could make the error at least a bit more specific.


radeksimko:
Actually this kind of complicates a potential bug fix for the root cause, so I'm going to withdraw this one and raise a PR with the bugfix instead.

PR #1483, opened 2020-05-01
fix(pipe/brew): Default to GitHub

radeksimko:
Previously `goreleaser --snapshot` (also used here in Travis CI) would fail if no token was provided, because Homebrew requires a token type for figuring out how to construct the download URLs.

This patch makes GitHub the default token type, which is in line with how releases work now.

### Before

```
      • homebrew tap formula
         • token type                type=
   ⨯ release failed after 42.51s error=homebrew tap formula: failed to publish artifacts: token type not implemented for brew pipe
```

### After

```
      • homebrew tap formula
         • token type                type=
         • writing                   formula=dist/goreleaser.rb
         • pipe skipped              error=publishing is disabled
```

--- 

I'm assuming this fix still retains the spirit of https://github.com/goreleaser/goreleaser/pull/1425 and allows inspection of formula before upload, or when the upload fails. The only downside is that the URLs in the generated formula may not match the ones that would then be used for publishing. That hopefully isn't a big deal though, because GoReleaser doesn't allow publishing of previously generated artifacts - i.e. the formula will be generated again with correct URLs before publishing.

### DIFF
--- a/internal/pipe/brew/brew.go
+++ b/internal/pipe/brew/brew.go
@@ -26,8 +26,20 @@ var ErrNoArchivesFound = errors.New("no linux/macos archives found")
 // for linux or windows.
 var ErrMultipleArchivesSameOS = errors.New("one tap can handle only archive of an OS/Arch combination. Consider using ids in the brew section")
 
+// ErrEmptyTokenType indicates unknown token type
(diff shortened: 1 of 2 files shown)


caarlos0:
awesome, thanks!

PR #1489, opened 2020-05-08
feat: Add Windows ARM64 support

paultyng:
This will add Windows ARM64 to the valid target list.

Laptops exist running [Windows on ARM64](https://www.microsoft.com/en-us/surface/business/surface-pro-x/processor).


caarlos0:
it seems go doesn't compile go Windows ARM64: https://golang.org/doc/install/source#environment

refs https://github.com/golang/go/issues/36439

paultyng:
Sorry about that, will track the upstream!

PR #1519, opened 2020-05-14
fix: improve docker logs and errors

caarlos0:
improved error output, as it may be confusing when there's more than one image being built...


gianarb:
Thank you for that!!

PR #1520, opened 2020-05-15
feat: Add build command

### LINKED ISSUE #1469: Allow skipping post build hooks

I would like to use GoReleaser also in a CI pipeline, so I can test release-ability. Since GoReleaser has all the build flags and build targets it looks very tempting to use the skip flags and use it to just test whether everything compiles.

```
goreleaser release --rm-dist --snapshot --skip-publish --skip-sign
```

Post build hooks can however contain logic which is either unnecessary (upx compression) or even undesirable (code signing) to run in such context and there's currently no way to avoid running them.

Introduce `--skip-post-build-hooks` flag which will skip execution of post build hooks.

I appreciate there's already many `--skip-...` flags, so I'm open to alternative ideas that could avoid polluting the endless list of flags.

Passing an ENV variable to goreleaser in a CI build and adding `if` conditional to `.goreleaser.yml`.

radeksimko:
This PR implements new `build` command per discussion in #1469 and aims to prevent the sprawl of `skip-*` flags in the main `release` command.

This also potentially paves way for a future similar command, like `archive/package` as discussed in #1446 but that should be IMO discussed separately.

Closes #1469

### DIFF
--- a/internal/pipeline/pipeline.go
+++ b/internal/pipeline/pipeline.go
@@ -34,9 +34,9 @@ type Piper interface {
 	Run(ctx *context.Context) error
 }
 
-// Pipeline contains all pipe implementations in order
-// nolint: gochecknoglobals
-var Pipeline = []Piper{
+// BuildPipeline contains all build-related pipe implementations in order
+// nolint:gochecknoglobals
+var BuildPipeline = []Piper{
 	before.Pipe{},          // run global hooks before build
 	env.Pipe{},             // load and validate environment variables
 	git.Pipe{},             // get and validate git repo state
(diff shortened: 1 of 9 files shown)


internal/pipeline/pipeline.go
```diff
@@ -47,12 +47,17 @@ var Pipeline = []Piper{
...
+var Pipeline = append(BuildPipeline, []Piper{
+	archive.Pipe{},       // archive in tar.gz, zip or binary (which does no archiving at all)
+	sourcearchive.Pipe{}, // archive the source code using git-archive
+	nfpm.Pipe{},          // archive via fpm (deb, rpm) using "native" go impl
+	snapcraft.Pipe{},     // archive via snapcraft (snap)
+	checksums.Pipe{},     // checksums of the files
+	sign.Pipe{},          // sign artifacts
+	docker.Pipe{},        // create and push docker images
+	publish.Pipe{},       // publishes artifacts
+}...)
```
caarlos0, line 63: I think it can be simplified to something like:

```suggestion
var Pipeline = append(
	BuildPipeline,
	archive.Pipe{},       // archive in tar.gz, zip or binary (which does no archiving at all)
	sourcearchive.Pipe{}, // archive the source code using git-archive
	nfpm.Pipe{},          // archive via fpm (deb, rpm) using "native" go impl
	snapcraft.Pipe{},     // archive via snapcraft (snap)
	checksums.Pipe{},     // checksums of the files
	sign.Pipe{},          // sign artifacts
	docker.Pipe{},        // create and push docker images
	publish.Pipe{},       // publishes artifacts
)
```

radeksimko: Good call!

caarlos0:
Awesome, thanks for the PR 🚀

PR #1522, opened 2020-05-15
Added custom Header and Footer to releases

wesleimp:
This PR makes prossible to add custom Header and Footer to your releases

### DIFF
--- a/internal/pipe/release/body.go
+++ b/internal/pipe/release/body.go
@@ -21,9 +21,9 @@ const bodyTemplateText = `{{if .Header }}{{ .Header }}
 - ` + "`docker pull {{ . -}}`" + `
 {{- end -}}
 {{- end }}
-
-{{if .Footer }}{{ .Footer }}{{end}}
-`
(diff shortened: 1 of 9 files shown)


caarlos0:
Hey, this is already allowed, check

```console
$ goreleaser release --help
```

wesleimp:
> Hey, thanks for the PR!
> 
> This is already allowed though, check:
> 
> ```
> $ goreleaser release --help
> ```

Sure, but I cant set the header or footer by `yml` file and I need to create `header.md` or `footer.md` to use that feature, right?

wesleimp:
@caarlos0 should I close this PR?

caarlos0:
I thought about this, I think the flags are more versatile, as we can easily use them for generated headers/footers as well...

I get why you are doing it on config as well, I suggest you fix them in your CI config instead...  even something like `--release-footer <(echo "my simple footer")` should work if you don't want to create the files.

Will close this for now, thanks for the contribution anyway :)

PR #1523, opened 2020-05-15
feat: run hooks as last step in the release pipeline

gianarb:
Hello, my goreleaser friends!

!!! I would like to make this PR better (will explain what I mean
later), it is here to gather feedback about the feature itself.

I asked about running some custom code at the end of the release cycle
on Gopher Slack today.

I would like to manipulate the manifest for a bunch of Docker
images I create during the release life cycle.

My build generates docker images for multiple architectures
and in order to teach the registry about that, you are supposed to
generate or updated a manifest.

You have to do it when the images are pushed to the registry.

It looks like a custom publisher won't work because it runs before
the docker one.

This PR proposes the same approach we use for the `before` feature, but
it runs at the end of the release workflow, as the last step.

---

As I said, I just copy-pasted and hooked a bunch of code. If this has
sense, I think I would like to write a pipe called `HookRunner` in order to reuse
it for Before and After.

---

- [ ] Add documentation
- [ ] Add tests

### DIFF
--- a/internal/pipe/after/after.go
+++ b/internal/pipe/after/after.go
@@ -1,0 +1,44 @@
+package after
+
+import (
+	"fmt"
+	"os/exec"
+
+	"github.com/goreleaser/goreleaser/pkg/context"
+
+	"github.com/apex/log"
+	"github.com/fatih/color"
+	"github.com/goreleaser/goreleaser/internal/tmpl"
+	"github.com/mattn/go-shellwords"
+)
+
+type Pipe struct{}
+
+func (p Pipe) String() string {
+	return "running after hooks"
+}
+
+// Run executes the hooks
+func (Pipe) Run(ctx *context.Context) error {
+	var tmpl = tmpl.New(ctx)
+	/* #nosec */
+	for _, step := range ctx.Config.After.Hooks {
+		s, err := tmpl.Apply(step)
+		if err != nil {
+			return err
+		}
+		args, err := shellwords.Parse(s)
+		if err != nil {
+			return err
+		}
+		log.Infof("running %s", color.CyanString(step))
+		cmd := exec.Command(args[0], args[1:]...)
+		cmd.Env = ctx.Env.Strings()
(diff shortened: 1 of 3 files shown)


gianarb:
I tried to get around this feature with custom publishers but the fact that they run in parallel per artifact does not really apply to the multi-arch manifest tool. I can do what I am trying to do with `makefiles` but the idea to have all the release cycle in one place sounds appealing

caarlos0:
I think this makes sense, yes. 🙏

gianarb:
@caarlos0 should I create a pipe `HookRunner` and use it for both? pre and post?

caarlos0:
> @caarlos0 should I create a pipe `HookRunner` and use it for both? pre and post?

Maybe we can use the hook structure of builds for both before and pre as well? 🤔
I think the build hooks are more versatile and should suit more cases in the end 🤔

gianarb:
Yes as I wrote in the PR description I copy-pasted a bunch of code and I am not sure what's your idea here. Personally I think it is simple to reuse the same code. But it is your call as a maintainer to tell me what you think is easier to maintain

caarlos0:
My idea was to maybe make [this](https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/build/build.go#L108-L151) a `pkg/hookrunner` or something, and use it on all hooks 🤔

It would be a bit harder to keep compact with the current before hooks, but it should be doable with some custom yaml parser magic 🤔

wdyt?

also cc/ @radeksimko

radeksimko:
@caarlos0 sounds good. I think there is potential for sharing some logic between the hooks, although we need to keep in mind they will execute with very different context, so e.g. artifacts just won't be available in `before` hooks, for logical (ordering) reasons. But I suppose that templating logic probably accounts for that, so it's fine.

caarlos0:
PR seems abandoned, closing for now. Feel free to reopen (or open a new one) if needed.

Thanks!

PR #1535, opened 2020-05-20
feat: Added extra files to blob

### LINKED ISSUE #1512: feature: blobs.extra_files
make it possible to upload extra files on the blobs pipe.

_Originally posted by @taglme in https://github.com/goreleaser/goreleaser/discussions/1511_

wesleimp:
Closes #1512


pkg/config/config.go
```diff
@@ -374,14 +374,22 @@ type Before struct {
...
+	KMSKey     string         `yaml:",omitempty"`
+	IDs        []string       `yaml:"ids,omitempty"`
+	Endpoint   string         `yaml:",omitempty"` // used for minio for example
+	Extra      BlobExtraFiles `yaml:",omitempty"`
+}
+
+// BlobExtraFiles contains config for extrafiles
+type BlobExtraFiles struct {
+	Folder string   `yaml:",omitempty"`
+	Path   string   `yaml:",omitempty"`
```
caarlos0, line 391: why do we have both folder and path?

I would also suggest to use the same structure (or something similar) as we have on the release:

```go
ExtraFiles   []ExtraFile `yaml:"extra_files,omitempty"`
```

The folder inside the bucket is already provided in the `blob` section, and if the user needs extra files on different folders, they can just add more blob sections with the desired behavior...

how does that sound? 🤔

caarlos0:
Thanks for the PR, did a comment above, lmk what you think :)

pkg/config/config.go
wesleimp: @caarlos0 you are right, `Folder` property doesn't make sense. The `Path` property means where the files are inside the project, for example. `src/extra-fil s/*`, in this case the `yml` will looks like:

```yml
...
extra_files:
  path: src/extra-files
  files:
    - file1
```

Does this make sense to you?

caarlos0: I was thinking we can use the same `ExtraFile` struct that already exists... what do you think?

wesleimp: Nevermind, now I got your thought. I made the changes according to the existing ExtraFIles

wesleimp:
> Thanks for the PR, did a comment above, lmk what you think :)

@caarlos0 now I think all the changes are ok.  I considered all of your comments

.golangci.yml
```diff
@@ -14,3 +14,7 @@ linters:
     - godox
     - gomnd
     - gocognit
+
+issues:
+  exclude:
+    - Using the variable on range scope .* in function literal
```
caarlos0, line 20: let's please fix the issue instead of exclude it...

internal/pipe/blob/upload.go
```diff
@@ -95,21 +96,45 @@ func doUpload(ctx *context.Context, conf config.Blob) error {
 	for _, artifact := range ctx.Artifacts.Filter(filter).List() {
 		artifact := artifact
 		g.Go(func() error {
-			// TODO: replace this with ?prefix=folder on the bucket url
```
caarlos0, line 98: the TODO should probably stay...

internal/commom/extra_files.go
```diff
@@ -0,0 +1,36 @@
+package commom
```
caarlos0, line 1: I think we can rename the pkg to `extrafiles` and the main function to `Find` only... 

recommented reading: https://blog.golang.org/package-names

.golangci.yml
wesleimp: I have never seen that error before and navigating through the Github, I found most of people "solving" their problems with that.

Sorry, my bad

caarlos0: no problem :)

caarlos0:
Thanks for the PR :)

PR #1543, opened 2020-05-22
fix: include golang as a dependency of deb/rpms

caarlos0:
having in mind #1542, maybe we should add `golang` as a dependency of the rpm and deb packages as well?

tested on both ubuntu and fedora (latests) and it works, although by default ubuntu install golang 1.10 and fedora golang 1.13 (but I think if any other package provides `golang` it will work as well).

what are your thoughts on this?

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -108,6 +108,7 @@ nfpms:
     - rpm
     dependencies:
     - git
+    - golang
 snapcrafts:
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -108,6 +108,8 @@ nfpms:
     - rpm
     dependencies:
     - git
+    recommend:
```
caarlos0, line 111:
```suggestion
    recommends:
```

PR #1546, opened 2020-05-22
chore(deps): bump code.gitea.io/sdk/gitea from v0.11.3 to v0.12.0

6543:
Migrate gitea client sdk

for more info look at https://gitea.com/gitea/go-sdk/src/branch/master/docs/migrate-v0.11-to-v0.12.md
and for the changelog at https://gitea.com/gitea/go-sdk/releases

PR #1547, opened 2020-05-22
feat:  Gitea client add create file function

6543:
implement for brew and scoop support for Gitea-hosted repos

### DIFF
--- a/internal/client/client_test.go
+++ b/internal/client/client_test.go
@@ -19,8 +19,9 @@ func TestClientNewGitea(t *testing.T) {
 	ctx := &context.Context{
 		Config: config.Project{
 			GiteaURLs: config.GiteaURLs{
-				API:      "https://git.dtluna.net/api/v1",
-				Download: "https://git.dtluna.net",
+				// TODO: use a mocked http server to cover version api
+				API:      "https://gitea.com/api/v1",
+				Download: "https://gitea.com",
 			},
 		},
 		TokenType: context.TokenTypeGitea,
@@ -36,7 +37,7 @@ func TestClientNewGiteaInvalidURL(t *testing.T) {
 	ctx := &context.Context{
 		Config: config.Project{
 			GiteaURLs: config.GiteaURLs{
-				API: "://git.dtluna.net/api/v1",
+				API: "://gitea.com/api/v1",
 			},
 		},
 		TokenType: context.TokenTypeGitea,
--- a/internal/client/client_test.go
+++ b/internal/client/client_test.go
@@ -19,7 +19,8 @@ func TestClientNewGitea(t *testing.T) {
 	ctx := &context.Context{
 		Config: config.Project{
 			GiteaURLs: config.GiteaURLs{
-				API: "https://git.dtluna.net/api/v1",
+				API:      "https://git.dtluna.net/api/v1",
+				Download: "https://git.dtluna.net",
 			},
 		},
 		TokenType: context.TokenTypeGitea,
--- a/internal/client/gitea.go
+++ b/internal/client/gitea.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"crypto/tls"
+	"encoding/base64"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -85,8 +86,42 @@ func (c *giteaClient) CreateFile(
 	path,
 	message string,
 ) error {
-	//TODO: implement for brew and scoop support for Gitea-hosted repos
-	return nil
+	// use default branch
+	branchName := ""
+
+	fileOptions := gitea.FileOptions{
+		Message:    message,
+		BranchName: branchName,
+		Author: gitea.Identity{
+			Name:  commitAuthor.Name,
+			Email: commitAuthor.Email,
+		},
+		Committer: gitea.Identity{
+			Name:  commitAuthor.Name,
+			Email: commitAuthor.Email,
+		},
+	}
+
+	currentFile, resp, err := c.client.GetContents(repo.Owner, repo.Name, branchName, path)
+	// file not exist, create it
+	if err != nil {
+		if resp == nil || resp.StatusCode != http.StatusNotFound {
+			return err
+		}
+		_, _, err = c.client.CreateFile(repo.Owner, repo.Name, path, gitea.CreateFileOptions{
+			FileOptions: fileOptions,
+			Content:     base64.StdEncoding.EncodeToString(content),
+		})
+		return err
+	}
+
+	// update file
+	_, _, err = c.client.UpdateFile(repo.Owner, repo.Name, path, gitea.UpdateFileOptions{
+		FileOptions: fileOptions,
+		SHA:         currentFile.SHA,
+		Content:     base64.StdEncoding.EncodeToString(content),
+	})
+	return err
 }
 
 func (c *giteaClient) createRelease(ctx *context.Context, title, body string) (*gitea.Release, error) {
--- a/internal/client/gitea.go
+++ b/internal/client/gitea.go
@@ -228,7 +228,12 @@ func (c *giteaClient) CreateRelease(ctx *context.Context, body string) (string, 
 }
 
 func (c *giteaClient) ReleaseURLTemplate(ctx *context.Context) (string, error) {
-	return "", NotImplementedError{TokenType: context.TokenTypeGitea}
+	return fmt.Sprintf(
+		"%s/%s/%s/releases/download/{{ .Tag }}/{{ .ArtifactName }}",
+		ctx.Config.GiteaURLs.Download,
+		ctx.Config.Release.Gitea.Owner,
+		ctx.Config.Release.Gitea.Name,
+	), nil
 }
 
 // Upload uploads a file into a release repository.
--- a/internal/pipe/defaults/defaults.go
+++ b/internal/pipe/defaults/defaults.go
@@ -30,7 +30,7 @@ func (Pipe) Run(ctx *context.Context) error {
 		ctx.Config.GitLabURLs.Download = client.DefaultGitLabDownloadURL
 	}
 	if ctx.Config.GiteaURLs.Download == "" {
-		ctx.Config.GitLabURLs.Download = strings.ReplaceAll(ctx.Config.GitLabURLs.API, "/api/v1", "")
+		ctx.Config.GiteaURLs.Download = strings.ReplaceAll(ctx.Config.GiteaURLs.API, "/api/v1", "")
 	}
 	for _, defaulter := range defaults.Defaulters {
 		if err := middleware.Logging(
--- a/internal/pipe/defaults/defaults.go
+++ b/internal/pipe/defaults/defaults.go
@@ -30,7 +30,7 @@ func (Pipe) Run(ctx *context.Context) error {
 		ctx.Config.GitLabURLs.Download = client.DefaultGitLabDownloadURL
 	}
 	if ctx.Config.GiteaURLs.Download == "" {
-		ctx.Config.GitLabURLs.Download = strings.TrimLeft(ctx.Config.GitLabURLs.API, "/api/v1")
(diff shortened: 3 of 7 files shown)


internal/client/gitea.go
```diff
@@ -61,8 +62,44 @@ func (c *giteaClient) CreateFile(
...
+		Committer: gitea.Identity{
+			Name:  commitAuthor.Name,
+			Email: commitAuthor.Email,
+		},
+	}
+
+	currentFile, err := c.client.GetContents(owner, repoName, branchName, path)
+	// file not exist, create it
+	if err != nil {
+		if err.Error() != "404 Not Found" {
```
6543, line 93: gitea/go-sdk#303 will impruve it .. but we have to live with it for now

6543:
@caarlos0 I dont realy cnow what this will function will enable goreleaser - just noticed it while doing #1546

would be nice if you can test it throu :)

6543:
and feel free to change the pull title ...

6543:
@caarlos0 I removed dummy test and add a real one now CI works

only problem is codecov now thinks this pull is bad :/

internal/client/gitea.go
```diff
@@ -61,8 +62,50 @@ func (c *giteaClient) CreateFile(
 	path,
 	message string,
 ) error {
-	//TODO: implement for brew and scoop support for Gitea-hosted repos
-	return nil
+	// Gitea >= 1.12.0 Use DefaultBranch on ""
+	branchName := ""
+	if err := c.client.CheckServerVersionConstraint(">=1.12.0"); err != nil {
```
radeksimko, line 67: Is this because gitea versions older than 1.12 didn't have the concept of default branches at all, or was it just not exposed in the API? What is the meaning of the empty string in these older versions?

Either way I assume it's correct, but it would be great to explain the motivation behind this a bit more in the comment.

6543: this wont be needed for v0.12.1
-> https://gitea.com/gitea/go-sdk/pulls/352

I'll add a ToDo comment to remove this when upgrading to gitea sdk v0.12.1

internal/client/gitea.go
```diff
@@ -61,8 +62,52 @@ func (c *giteaClient) CreateFile(
 	path,
 	message string,
 ) error {
-	//TODO: implement for brew and scoop support for Gitea-hosted repos
-	return nil
+	// ToDo: remove this when upgrading to v0.12.0
```
6543, line 65:
```suggestion
	// ToDo: remove this when upgrading to v0.12.1
```

6543:
@radeksimko avter you have looked at the commits, If I should squash them just tell

caarlos0:
> @caarlos0 What are your thoughts on the deprecation?

I think it should be fine, yes...

6543:
@radeksimko @caarlos0 current state?

caarlos0:
as @radeksimko mentioned, this pr alone does nothing... need the changes on brew/scoop pipe as well...

6543:
hmm ok feel free to cherry-pick this :)

radeksimko:
@6543 I have done some refactoring which should make it easier to enable gitea for both Homebrew and Scoop.

I assume it would be better to wait [until `v0.12.1` lands](https://gitea.com/gitea/go-sdk/releases) according to your comment above though?

6543:
@radeksimko to wait for v0.12.1 is a good idear :)

6543:
@radeksimko with your refactor and v0.12.1 release - what has to be done to finish this?

radeksimko:
Good question @6543 

I think the main thing is checking the remaining docs which still claim gitea isn't supported, e.g. https://github.com/goreleaser/goreleaser/blob/b05c95a00f03fedc570a04465f3185c1b7546439/www/docs/customization/homebrew.md

>    NOTE: make sure the url_template, the token and given repo (github or gitlab) owner and name are from the
>    same kind. We will probably unify this in the next major version like it is done with scoop.
> 
>    GitHub/GitLab repository to push the formula to
>    Gitea is not supported yet, but the support coming

https://github.com/goreleaser/goreleaser/blob/b05c95a00f03fedc570a04465f3185c1b7546439/www/docs/customization/scoop.md

> Template for the url which is determined by the given Token (github or gitlab)
> Default for github is "https://github.com/<repo_owner>/<repo_name>/releases/download/{{ .Tag }}/{{ .ArtifactName }}"
> Default for gitlab is "https://gitlab.com/<repo_owner>/<repo_name>/uploads/{{ .ArtifactUploadHash }}/{{ .ArtifactName }}"
> Gitea is not supported yet, but the support coming

and double check the relevant `brew` and `scoop` packages for any similar comments in the code.

Finally - have you tried this out with a real Homebrew tap or Scoop bucket hosted in Gitea? That's probably the last thing remaining I'd also like to check myself before merging this.

radeksimko:
and this will need to be implemented too:

https://github.com/goreleaser/goreleaser/blob/01fd3e8c7b3b16e77996b0434fc49f62a033f775/internal/client/gitea.go#L197-L199

6543:
give me some time :) and thanks for the hint

caarlos0:
@6543 is this still valid?

6543:
I'll rebase ... :)

6543:
@caarlos0 I think it's valid again :)

NOTE: didn't test it out in the wild again ...

mavogel:
@6543 I could help you out testing _out in the wild_. Could you rebase for me upfront :)

6543:
@mavogel would be **great** ... sure :)

6543:
@mavogel -> 98a3d3c

6543:
looks like I need to change some things at this point ... WIP

mavogel:
ok let me know when you're done :)

6543:
@mavogel it's ready - it was just some changes missing in the test

mavogel:
@6543 I was testing it locally due to my config for gitlab/gitea you can find [here](https://gitlab.com/mavogel/release-testing#testing-against-gitea) and also set it up locally. Unfortunately, there are some manual steps to do.

When I hover over one of the releases (e.g `release-testing_0.3.46_Linux_i386.tar.gz`) to get the download URL, which will later be used within the brew formula, I get `http://localhost:3000/attachments/e26a9054-85bc-4971-898d-f7cf17c56d21`. More investigation/debugging will be necessary to find out where we can get the hash during the upload process and then pass it through as we did already for GitLab with the `ArtifactUploadHash`. 

Screenshot 2020-11-15 at 23 38 03

I'll get back to it tomorrow, but feel free to set it up locally as well. 

@caarlos0 would it make sense to add those local configs somewhere here in this repo, so people can set up their own dev env to verify those changes?

6543:
@mavogel you dont need the uuid you can use the name too as it's done on github. the UI just use the uuid since you can upload artifacts with the same name ... this should be fixed, but I had no time to do so

6543:
@mavogel  `http://localhost:3000/user/repo/releases/download/0.3.46/release-testing_0.3.46_Linux_i386.tar.gz` is also valide

mavogel:
ok, thanks for the heads up. I'll restest later

mavogel:
@6543 on publishing the homebrew taps I get an `json: cannot unmarshal array into Go value of type gitea.ContentsResponse`

I quick investigation seems like this method is causing it internally
```go
currentFile, resp, err := c.client.GetContents(repo.Owner, repo.Name, branchName, path)
// -> getParsedResponse -> json.Unmarshal(data, obj)
```
at https://gitea.com/gitea/go-sdk/src/branch/master/gitea/client.go#L242

AFAICS you're also the maintainer of the sdk :)

mavogel:
Update: 
- I got it fixed by creating an empty `Formula` folder in the `homebrew-tap` repository. 
- then it  was possible to install the binary via
```sh
brew tap mavogel/gitea-tools ssh://git@localhost:222/mavogel/homebrew-tap.git
brew install mavogel/gitea-tools/release-testing
```

So happy to have it successfully tested in the wild 🎉 
Should we open an issue anyway for the SDK function when trying to create a file in a non-existing path, should return another error than the one `json: cannot unmarshal`?

6543:
@mavogel pleace do :) - I'm happy to follow up on that specific issue on a new created issue in the sdk repository :)

6543:
> and this will need to be implemented too:
> 
> https://github.com/goreleaser/goreleaser/blob/01fd3e8c7b3b16e77996b0434fc49f62a033f775/internal/client/gitea.go#L197-L199

what is it doing?

6543:
@radeksimko added
@caarlos0 think it is ready & tested now :)

6543:
@radeksimko added
@caarlos0 think it is ready & tested now :)

6543:
@caarlos0 do you squash-merge or should i swash it?

internal/client/client_test.go
```diff
@@ -19,7 +19,8 @@ func TestClientNewGitea(t *testing.T) {
 	ctx := &context.Context{
 		Config: config.Project{
 			GiteaURLs: config.GiteaURLs{
-				API: "https://git.dtluna.net/api/v1",
+				API:      "https://git.dtluna.net/api/v1",
+				Download: "https://git.dtluna.net",
```
caarlos0, line 23: does this mean our tests actually hit this gitea server? 

if so, I would say we should change them, if not, I would say to set a fake address...

internal/pipe/defaults/defaults.go
```diff
@@ -27,6 +29,9 @@ func (Pipe) Run(ctx *context.Context) error {
 	if ctx.Config.GitLabURLs.Download == "" {
 		ctx.Config.GitLabURLs.Download = client.DefaultGitLabDownloadURL
 	}
+	if ctx.Config.GiteaURLs.Download == "" {
+		ctx.Config.GitLabURLs.Download = strings.ReplaceAll(ctx.Config.GitLabURLs.API, "/api/v1", "")
```
caarlos0, line 33: shouldn't it be gitea here instead of gitlab?

caarlos0: also this needs to be tested as well.

pkg/config/config.go
```diff
@@ -30,6 +30,7 @@ type GitLabURLs struct {
 // GiteaURLs holds the URLs to be used when using gitea.
 type GiteaURLs struct {
 	API           string `yaml:"api,omitempty"`
+	Download      string `yaml:"download,omitempty"`
```
caarlos0, line 33: need to document this option...

internal/client/client_test.go
6543: @caarlos0 this was not added by me - tested and found out that `TestClientNewGitea` rely on it

should I change it to gitea.com ?

caarlos0: I think its better yes, and maybe add a TODO to change this to a mocked http server...

6543:
@caarlos0 thanks for the advice - done

mavogel:
@6543 : there you go: https://gitea.com/gitea/go-sdk/issues/459

@caarlos0 : for the testing setup there was a lot of manual work to do until I discovered this recent [setup](https://gitea.com/gitea/go-sdk/src/branch/master/Makefile#L81) of gitea. IMHO it would make sense to add this and some additional steps (like homebrew-tap repo creation) to the docs. For `gitlab` as well, so Acc tests can be set-up quickly. What do you think? I could contribute this

caarlos0:
> @caarlos0 : for the testing setup there was a lot of manual work to do until I discovered this recent [setup](https://gitea.com/gitea/go-sdk/src/branch/master/Makefile#L81) of gitea. IMHO it would make sense to add this and some additional steps (like homebrew-tap repo creation) to the docs. For `gitlab` as well, so Acc tests can be set-up quickly. What do you think? I could contribute this

you mean as documentation? 

I think it makes sense to put somewhere, or at least link to it somewhere, yes 🤔

6543:
something missing?

caarlos0:
Thanks everyone

PR #1552, opened 2020-05-23
feat: allow signing only the source

greut:
Allow to sign the source artifact only.

I wanted to sign both checksum and source without having to sign all the artifacts

```yaml
source:
  enabled: true
checksum:
signs:
- id: default
  artifacts: checksum
- id: source
  artifacts: source
```


caarlos0:
Thanks for the PR 🚀

Issue #1452, opened 2020-04-17
Expose more snap configuration in SnapraftAppMetadata


neoaggelos:
**Is your feature request related to a problem? Please describe.**

Initial context: https://github.com/TheThingsNetwork/lorawan-stack/pull/2361#issuecomment-614973823

```
env MYVAR=$OTHER_ENV_VAR executable
```

I would assume being able to override the command is useful in other scenarios as well. This is something well supported by Snapcraft, so `goreleaser` should not block using it.

**Describe the solution you'd like**

Add `Command`, `Environ` and more fields in `SnapcraftAppMetadata`. Most can simply then be passed to the `snap.yaml` file.

It should be simple enough to have that support, and possibly falling back to defaults where needed. 

**Describe alternatives you've considered**

**Additional context**

caarlos0:
our current snap implementation is a bit hacky IMHO... I'm not sure how to add it.

If you can work on a PR it would be awesome though 🙏

neoaggelos:
I can try to take on some of these. I am mostly interested in setting the environment, the command, and possibly adding a few more files (e.g. wrapper scripts) on the resulting snap.

eberkund:
Personally I find it easier not to use goreleaser for producing snap packages. One thing I have encountered though is getting the correct version string inside my snapcrafy.yaml file.

It would be nice if goreleaser had a command which would just return the version string (as retrieved from Git), such as 0.3.5-next.

caarlos0:
@eberkund maybe this helps: https://github.com/caarlos0/svu

eberkund:
> @eberkund maybe this helps: https://github.com/caarlos0/svu

Thanks for sharing this looks similar to what I need. But it doesn't seem to work exactly the same way as goreleaser which adds a "-next" suffix to the version number of  "dirty" releases.

caarlos0:
> Thanks for sharing this looks similar to what I need. But it doesn't seem to work exactly the same way as goreleaser which adds a "-next" suffix to the version number of "dirty" releases.

yeah, this is done in the `snapshot` part of goreleaser probably (assuming your snapshot template is something like `{{.Version}}-next`.

You could use `echo "$(svu current)-next"` to get something similar...

neoaggelos:
Closing this one since #1553 has been merged.

Issue #1544, opened 2020-05-22
do not link files on docker pipe


caarlos0:
some docker pipe tests are failing with things like:

```
link testdata/Dockerfile /tmp/dockertest011168116/dist/goreleaserdocker212266563/Dockerfile: invalid cross-device link
```

seems to be related to some change in github actions, as goreleaser's code didn't change anything related to this lately.

PS: the docker pipe do some hard links indeed, maybe we should change that to copy instead as it already caused other problems in the past?

PPS: golang does not have a stdlib copy function, maybe use something like, more details https://github.com/goreleaser/goreleaser/pull/355 and https://github.com/golang/go/issues/8868

caarlos0:
refs https://github.com/goreleaser/goreleaser/pull/1545

till:
I just ran into issues with the docker build, and then learned about `extra_files` (which worked! :tada:).

Do you think it's even necessary to setup a build directory and copy/link files into a temporary directory? I can see advantages for everything, but it seems in order to avoid that the image contains unnecessary/unwanted files, `.dockerignore` is more suitable.

caarlos0:
the problem is not dockerignore itself, we do the temp dir thing because, otherwise, the dockerfile would have to `COPY dist/foo_linux_amd64/binary /etc`... and today its just `COPY binary /etc`, which works outside goreleaser as well (`go build . && docker build .`)... 

the linking thing is hacky, ideally we should copy the needed files...

till:
@caarlos0 thanks, I see the reasoning! 👍 

And yeah, maybe a copy is better or less weird on different systems.

till:
Do you have a preference as to what to use:

 - `io.Copy()`,
 - `ioutil.ReadFile()` and `ioutil.WriteFile()`
 - `os.Read()` and `os.Write()`

I am hoping all of them are cross platform? 😂

caarlos0:
`io.Copy` copies from a reader to a writer, we still need to create the target file, and there is where the problems happen... 

check https://github.com/golang/go/issues/8868 and specially Rob Pike's comments on https://go-review.googlesource.com/c/go/+/1591

caarlos0:
(but IMHO we don't need a perfect 100% correct on all platforms in existence copy file, my plan was to copy the one from that CL 🤣)

caarlos0:
Will ignore this for now, as the problem was on github actions and it is working as is for quite some time...

caarlos0:
[...]

PR #1553, opened 2020-05-23
feat: Support custom command and extra files in snaps

neoaggelos:
- Support adding extra files on snap packages (useful for static assets or wrapper scripts)
- Update documentation
- Support setting a custom command for snap apps

Also added unit tests for all new changes.

Add support for wrapper scripts on snap packages.

References #1452. If this is merged, then #1452 can close as far as we are concerned.


internal/pipe/snapcraft/snapcraft.go
```diff
@@ -314,3 +334,29 @@ func push(ctx *context.Context, snap *artifact.Artifact) error {
 	ctx.Artifacts.Add(snap)
 	return nil
 }
+
+// walks the src, recreating dirs and hard-linking files
+func link(src, dest string, mode os.FileMode) error {
```
caarlos0, line 339: this is copied from the `docker` pipe isn't it?

its ok, just wanted to note that linking thing may break in some cases (check #1544), eventually we'll give it a shot at copying the files instead... (which will then impact here as well)

www/docs/customization/snapcraft.md
```diff
@@ -114,6 +123,9 @@ snapcrafts:
         # https://docs.snapcraft.io/tab-completion-for-snaps.
         completer: drumroll-completion.bash
 
+        # You can override the command name. Defaults to the app name.
+        # command: bin/drumroll.wrapper
```
caarlos0, line 127: nitpick but please uncomment this line so it looks like the others...

```suggestion
        command: bin/drumroll.wrapper
```

www/docs/customization/snapcraft.md
```diff
@@ -114,6 +123,9 @@ snapcrafts:
         # https://docs.snapcraft.io/tab-completion-for-snaps.
         completer: drumroll-completion.bash
 
+        # You can override the command name. Defaults to the app name.
```
caarlos0, line 126: unspoken default, but we usually leave the default on its own line:

```suggestion
        # You can override the command name.
        # Defaults to the app name.
```

www/docs/customization/snapcraft.md
```diff
@@ -88,6 +88,15 @@ snapcrafts:
     # Default is empty.
     base: core18
 
+    # Add extra files on the resulting snap. Useful for including wrapper
+    # scripts or other useful static files. Source filenames are relative to the
+    # project directory. Destination filenames are relative to the snap prime
+    # directory.
```
caarlos0, line 94: Missing the default:

```suggestion
    # directory.
    # Default is empty.
```

internal/pipe/snapcraft/snapcraft.go
neoaggelos: Indeed, this is the same code. I thought about taking it into a util package but I decided against it, just to keep things simple.

Do you think there is something that needs to be improved here for this PR?

neoaggelos:
Also rebased and fixed CI errors.

caarlos0:
> CI seems to be failing with an unrelated (?) error:

yeah, just opened a PR about it #1587

internal/pipe/snapcraft/snapcraft.go
```diff
@@ -231,11 +246,16 @@ func create(ctx *context.Context, snap config.Snapcraft, arch string, binaries [
 				WithField("name", name).
 				Debug("passed binary to snapcraft")
 
+			command := name
+			if config.Command != "" {
```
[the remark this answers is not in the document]
neoaggelos: Looks better, avoided it initially so as to not mutate the config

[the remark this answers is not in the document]
neoaggelos: So that change actually breaks the tests, and it feels like patchwork to reset the config in the tests for them to work properly.

Reverted, let me know if that is OK or how to proceed.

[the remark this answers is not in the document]
neoaggelos: https://github.com/goreleaser/goreleaser/runs/745188125?check_suite_focus=true

caarlos0:
Thanks!

Issue #1532, opened 2020-05-19
Consistent versioning for snapshots


radeksimko:
**Is your feature request related to a problem? Please describe.**

I'm unsure if this is bug or a feature, but I noticed that for `--snapshot` archives the `.Version` format is different to the standard non-snapshot releases.

```
v0.2.0-SNAPSHOT-906c368
```
vs
```
0.2.0
```

I admit version formatting may not be a huge deal for snapshot builds as they're far less likely to be distributed widely, but it seems inconsistent nonetheless. Relatedly [`.Version` description in docs](https://goreleaser.com/customization/templates/) claims the `v` prefix is stripped, which doesn't apply to the snapshot mode.

This can be an issue if such snapshot versions need to be compared to the "clean" versions in semver or any other sense.

**Describe the solution you'd like**

I think the least disruptive change here could be introducing the following new variables:

 - `.IsSnapshot` (bool)
 - `.RawVersion` (string) - always returning the same format as `.Version` in non-snapshot mode

Then default `ldflags` can be retained for compatibility, but allow more flexible configuration, e.g.

```yml
ldflags:
      - '-s -w -X "version={{ .RawVersion }}" -X "prerelease={{ if .IsSnapshot }}dev{{ end }}"'
```

**Describe alternatives you've considered**

I found `RawVersion` can be done with a just a little verbose construct, but I couldn't find a way to identify snapshot builds.

```yml
ldflags:
      - '-s -w -X "version={{printf "%s.%s.%s" . Major .Minor .Patch }}" -X "prerelease="'
```

caarlos0:
> I admit version formatting may not be a huge deal for snapshot builds as they're far less likely to be distributed widely, but it seems inconsistent nonetheless. Relatedly [`.Version` description in docs](https://goreleaser.com/customization/templates/) claims the `v` prefix is stripped, which doesn't apply to the snapshot mode.

I think this is a bug 🤔



> I think the least disruptive change here could be introducing the following new variables:
> 
>     * `.IsSnapshot` (bool)
> 
>     * `.RawVersion` (string) - always returning the same format as `.Version` in non-snapshot mode
> 
> 
> Then default `ldflags` can be retained for compatibility, but allow more flexible configuration, e.g.

This sounds good to me.

--- 

On a side note, I also started some work adding `inc` functions which I though may be useful, but closed the PR due to lack of feedback... maybe it would help in your case too?  https://github.com/goreleaser/goreleaser/pull/1457

PR #1557, opened 2020-05-24
feat: Expose semver for snapshots

### LINKED ISSUE #1549: Real world example for authenticated custom publisher with curl or wget
I’m struggling to get goreleaser to push .deb files to packagecloud.io. I don’t believe the HTTP upload will work because package cloud requires POST to an endpoint with the distribution ID which is their own lookup table. However, I tried to do this with a raw curl command but couldn’t get the authentication to work properly.  

It would be very helpful if the documentation provided a real world example of using custom publisher with curl or wget or even the packagecloud CLI if someone has experience with this.

radeksimko:
Closes #1549 

--- 

I formatted the table via https://www.tablesgenerator.com/markdown_tables but I wish it was a bit easier to modify these tables manually without having to use any external tools/services.

We could make it easier by using the default left-side text-aligning, which I think would be also easier to read when rendered. That's something to discuss in a separate PR, where we could sweep all the tables at once and discuss whether it's a good idea.

### DIFF
--- a/internal/tmpl/tmpl.go
+++ b/internal/tmpl/tmpl.go
@@ -3,6 +3,7 @@ package tmpl
 
 import (
 	"bytes"
+	"fmt"
 	"path/filepath"
 	"strings"
 	"text/template"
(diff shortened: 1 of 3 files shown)


caarlos0:
LGTM, thanks!

caarlos0:
I think it was suppose to close https://github.com/goreleaser/goreleaser/issues/1532

radeksimko:
> I think it was suppose to close #1532

Indeed - I must have made a typo! Thanks for catching it.

Issue #1537, opened 2020-05-21
Brew versioned releases


edulop91:
**Is your feature request related to a problem? Please describe.**
I was trying to publish versioned releases using brew. It seems that the recommended way of doing this is `foo@1` or `foo@2`. 

For example, there is a formula for [awscli](https://github.com/Homebrew/homebrew-core/blob/master/Formula/awscli.rb) and a formula for [awscli@1](https://github.com/Homebrew/homebrew-core/blob/master/Formula/awscli%401.rb).

When I try doing [something similar](https://github.com/chanzuckerberg/blessclient/blob/v0/.goreleaser.yml#L24) with goreleaser, I get the [following formula](https://github.com/chanzuckerberg/homebrew-tap/blob/master/blessclient%400.rb#L2). This then causes the following error: 
>Invalid formula: /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/chanzuckerberg/homebrew-tap/blessclient@0.rb
blessclient@0: /home/linuxbrew/.linuxbrew/Homebrew/Library/Taps/chanzuckerberg/homebrew-tap/blessclient@0.rb:2: `@0' is not allowed as an instance variable name

**Describe the solution you'd like**

It seems that, by convention, the `@` is [replaced](https://github.com/Homebrew/homebrew-core/blob/master/Formula/awscli%401.rb#L1) by `At` in the fomula's class name. Would it make sense to automatically add this character replacement in the brew template/pipeline?

**Describe alternatives you've considered**
- automatically replace `@` for `At` [here](https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/brew/template.go#L32)
- Allow us to specify the name of the formula name + the formula's class independently

**Additional context**
I'm happy to contribute code for this!

caarlos0:
>     * automatically replace `@` for `At` [here](https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/brew/template.go#L32)

this seems good enough.

PR #1568, opened 2020-05-26
fix: Brew versioned formulae now supported

edulop91:
Will replace `@` for `AT` in brew formulae name.

Allow for Brew versioned releases (eg: foo@1). See how [awscli](https://github.com/Homebrew/homebrew-core/blob/master/Formula/awscli.rb#L1) and [awscli@1](https://github.com/Homebrew/homebrew-core/blob/master/Formula/awscli%401.rb#L1) do it.

- Fixes https://github.com/goreleaser/goreleaser/issues/1537
- https://github.com/Homebrew/homebrew-core/blob/master/Formula/awscli.rb#L1
- https://github.com/Homebrew/homebrew-core/blob/master/Formula/awscli%401.rb#L1


caarlos0:
Awesome, thanks!

PR #1569, opened 2020-05-26
docs: Format markdown tables

radeksimko:
This patch formats all existing markdown tables in docs in a way that aligns cell content to left, as opposed to the centre (in most cases). This in turn makes it easier for anyone without any markdown generators to edit the source and getting the spacing right.

### DIFF
--- a/www/docs/ci/actions.md
+++ b/www/docs/ci/actions.md
@@ -115,11 +115,11 @@ signs:
 
 Following inputs can be used as `step.with` keys
 
-| Name          | Type    | Default   | Description                               |
-|---------------|---------|-----------|-------------------------------------------|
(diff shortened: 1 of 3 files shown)


radeksimko:
@caarlos0 I'm not opposed to plugging checks into CI, but personally I think it's good to maintain some balance between bots telling contributors off for (subjectively) small things vs designing the codebase so that bots don't need to tell anyone off because they just get it right for the first time. 😄

caarlos0:
build is failing, but seems unrelated... will merge anyway.

---



> @caarlos0 I'm not opposed to plugging checks into CI, but personally I think it's good to maintain some balance between bots telling contributors off for (subjectively) small things vs designing the codebase so that bots don't need to tell anyone off because they just get it right for the first time. 😄

totally, we could run it with `--write` on an action though... :)

PR #1596, opened 2020-06-08
docs: Documentation to include simple multiple binary example

### LINKED ISSUE #1370: Documentation to include simple multiple binary example
My project has separate command-line tools in `cmd/tool1/tool1.go` and `cmd/tool2/tool2/go` etc, which I think is pretty common.

It took me a while, and a lot of searching  to find that each separate `builds` clause needs its own distinct "id". The default seems to be picked up from the main folder name (or github repo?).

It would be great to have a "multiple binaries" example  in the sections listed on https://goreleaser.com/.

Great tool, by the way!

wesleimp:
Closes #1370

### DIFF
--- a/www/docs/customization/build.md
+++ b/www/docs/customization/build.md
@@ -118,6 +118,36 @@ builds:
 !!! tip
     Learn more about the [name template engine](/customization/templates).
 
+Here is an example with multiple binaries:
+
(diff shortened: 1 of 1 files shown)


caarlos0:
thanks!

Issue #1585, opened 2020-06-03
Docker image with cross compilation for OSX, windows, linux


doodlesbykumbi:
**Is your feature request related to a problem? Please describe.**

I usually have to create a special cross-compilation image with Goreleaser for automated builds. 
Here's an example https://github.com/cyberark/secretless-broker/blob/master/bin/Dockerfile.releaser. As part of a release I include builds for a variety of targets (e.g. linux amd64, darwin amd64 etc.). 

Without a cross compilation image, I have to have my builds running in separate executors or potentially run some builds locally. This is all a lot more cumbersome than leveraging a linux image that has cross-compilation capabilities (e.g. https://hub.docker.com/r/dockercore/golang-cross/)

**Describe the solution you'd like**

goreleaser currently maintains image tags for the normal goreleaser and gorelease with cgo `v0.137` and `v0.137-cgo`. It would be amazing if the same would be done for an image with cross compilation capabilities. I've provided an example above. That would mean an additional tag for `v0.137` perhaps called `v0.137-crossgo`.

**Describe alternatives you've considered**

I considered maintaining the images myself. This likely means no-one else gets to benefit from this, in part because of discoverability since I'm a random. It also means I have to keep track of goreleaser releases to keep things up to date, which I suspect would be better done here.

**Additional context**

N/A

caarlos0:
this was added to [docs](https://goreleaser.com/cookbooks/cgo-and-crosscompiling), so I'll close this.

Thanks!

Issue #708, opened 2018-07-02
Crossbuilding with xgo


endorama:
Hello, first of all a **big** thank you for this project. Is awesome and works like a charm.

**Is your feature request related to a problem? Please describe.**
I'm having some issues cross-compiling a go package ( [github.com/99designs/keyring](https://github.com/99designs/keyring) )

Trying to cross-build from linux to osx I needed to use [`xgo`](https://github.com/karalabe/xgo). This is due to the use of the `keychain` osx feature ( no idea why it doesn't compile with CGO, I'm still a golang newbie ).

I tried looking at the docs and the repo ( issues and code ) but wasn't able to find a solution.

Building with CGO does not seems to work ( I'm waiting some more informations on how to cross build that library https://github.com/99designs/keyring/issues/23 )

**Describe the solution you'd like**
I'd like to use `xgo` to build the package.

The command I'm using is:
```
$ xgo \
    --targets="darwin/amd64,linux/amd64" \
    --dest=dist \
    github.com/99designs/keyring
```

**Describe alternatives you've considered**
Doing it manually, reproducing the goreleaser steps ( build, sign, release ).

Thank you!

caarlos0:
you don't need xgo, you just need to disable cgo.

Take a look at goreleaser's `goreleaser.yml` file: https://github.com/goreleaser/goreleaser/blob/master/.goreleaser.yml#L5-L6

Let me know if it doesn't work!

Cheers

endorama:
Hi @caarlos0 thanks for the quick reply. Unfortunately does not seem to work.

<details>
  <summary><code>.goreleaser.yml</code></summary>
  <p>

  ```yaml
# .goreleaser.yml
# Build customization
builds:
  - binary: two-factor-authenticator
    env:
      - CGO_ENABLED=0
    goos:
      - darwin
      - linux
    goarch:
      - amd64
  ```
  </p>
</details>

<details>
  <summary>output</summary>
  <p>

```
$ goreleaser release --skip-publish --rm-dist --snapshot

   • releasing using goreleaser 0.79.0...
   • loading config file       file=.goreleaser.yml
   • SETTING DEFAULTS FOR:
      • loading environment variables
      • snapshoting              
      • releasing to GitHub      
      • project name             
      • creating archives        
      • building binaries        
      • creating Linux packages with fpm
      • creating Linux packages with nfpm
      • creating Linux packages with snapcraft
      • calculating checksums    
      • signing artifacts        
[...]

PR #1597, opened 2020-06-08
feat: Add docker image for cross compilation

doodlesbykumbi:
Adds a cross compilation image that allows goreleaser to build artefacts for linux, windows and darwin. This image builds on top of a base image used to cross compile golang. It includes the MinGW compiler for Windows, and an OSX SDK. It makes life easier when seeking to cross compilation.

Without a cross compilation image users have to find bespoke solutions e.g. run builds on separate executors or potentially run some builds locally. This is all a lot more cumbersome than leveraging an image that has cross-compilation capabilities (i.e. https://hub.docker.com/r/dockercore/golang-cross/)

Associated with https://github.com/goreleaser/goreleaser/issues/1585

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -31,6 +31,21 @@ changelog:
     - Merge branch
 dockers:
 - image_templates:
+  - 'goreleaser/goreleaser:{{ .Tag }}-crossgo'
+  - 'goreleaser/goreleaser:v{{ .Major }}.{{ .Minor }}-crossgo'
+  - 'goreleaser/goreleaser:latest-crossgo'
+  dockerfile: Dockerfile.crossgo
+  binaries:
+  - goreleaser
+  build_flag_templates:
+  - "--pull"
+  - "--label=org.opencontainers.image.created={{.Date}}"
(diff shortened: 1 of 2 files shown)


caarlos0:
We have discussed this in the past, and I think the main issue here is the [macOS SDK license](https://www.apple.com/legal/sla/docs/xcode.pdf), more specifically:

> You agree not to rent, lease, lend, upload to or host on any website or server, sell, redistribute, or sublicense the Apple Software and Apple Services, in whole or in part, or to enable others to do so.

At least, this is my understanding...

Another issue is that the cross go image is still Go 1.13, but that I guess we could live with...

https://github.com/goreleaser/goreleaser/issues/708

doodlesbykumbi:
@caarlos0 Thanks. I'm going to get some legal guidance on this and get back to you. 

I wonder. The image I'm building on top of (which is published on Dockerhub) is from Docker Inc. Surely, they must be aware of the licensing considerations.

caarlos0:
this was added to [docs](https://goreleaser.com/cookbooks/cgo-and-crosscompiling), so I'll close this.

Thanks!

PR #1610, opened 2020-06-15
docs: unshallow action

caarlos0:
closes https://github.com/goreleaser/goreleaser/pull/1608

refs https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,9 +15,8 @@ jobs:
       -
         name: Checkout
         uses: actions/checkout@v2
-      -
-        name: Unshallow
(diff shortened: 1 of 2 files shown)


crazy-max:
@caarlos0 Maybe add a note about `version` input like [this](https://github.com/goreleaser/goreleaser-action#inputs)?

caarlos0:
> @caarlos0 Maybe add a note about `version` input like [this](https://github.com/goreleaser/goreleaser-action#inputs)?

seems like another PR... this one is about the cloning thing.

but yeah, we should add it :)

PR #1617, opened 2020-06-18
fix: docker login using password stdin

nboukeffa:
Instead of passing the password using the `-p` arg which is deprecated, we now use `--password-stdin` which is recommended.

Solves the following warning:
```
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
```


caarlos0:
Thanks!

PR #1622, opened 2020-06-19
chore(ci): publish to github docker registry as well

caarlos0:
DOCKER_USERNAME is the same for dockerhub and github, so it should be OK.


caarlos0:
> [multi-arch Docker image are not currently supported on GitHub Package Registry](https://github.community/t5/GitHub-API-Development-and/Handle-multi-arch-Docker-images-on-GitHub-Package-Registry/m-p/31650#M2981) in case you want to bring support for it in the future.

this is a bummer... I think I'll hold on this for now then.

Thanks for pointing that out!

PR #1625, opened 2020-06-24
fix: use correct filename for GCS downloads

abeMedia:
Currently if uploading a file to Google Cloud Storage the filename contains the path when downloading e.g. `bucket/path/to/file.ext` will download as `path_to_file.ext`.

Setting the `Content-Disposition` metadata on upload ensures the filename we have locally is what it's called when downloading from GCS.


caarlos0:
Thanks!

PR #1632, opened 2020-06-30
fix: Don't try and get release information if releases are disabled

JakeMakesStuff:
If applied, this commit will not try and get the version from Git/handle Git dirty states if releases are set to false.

Releases being disabled signifies that the user does not want to handle getting a GitHub version, so that should be handled accordingly. This is useful for local development.

### DIFF
--- a/internal/pipe/git/git.go
+++ b/internal/pipe/git/git.go
@@ -30,9 +30,14 @@ func (Pipe) Run(ctx *context.Context) error {
 		return err
 	}
 	ctx.Git = info
-	log.Infof("releasing %s, commit %s", info.CurrentTag, info.Commit)
-	ctx.Version = strings.TrimPrefix(ctx.Git.CurrentTag, "v")
-	return validate(ctx)
+	if !ctx.Config.Release.Disable {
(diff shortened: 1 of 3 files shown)


JakeMakesStuff:
Essentially it's relating to trying to test goreleaser locally. In its current state it will try and grab the version information from Git. This is annoying when you are trying to work on something locally since:
A) It is expected that the latest version is blank. If there are assets in the latest version, goreleaser will error.
B) It is expected that Git is not in a dirty state.

I tested with the following configuration which successfully worked with this PR:
```yaml
before:
  hooks:
    - go mod download
    - cd config/ && npm i && npm run build && cd ..
    - cd assets && go run . && cd ..
    - packr2

archives:
- replacements:
    darwin: Darwin
    linux: Linux
    windows: Windows
    386: i386
    amd64: x86_64

checksum:

release:
  disable: true

builds:
  -
    goos:
      - darwin
    goarch:
      - amd64
```

caarlos0:
have you tried `goreleaser release --snapshot`? I think it will solve your problem 🤔

JakeMakesStuff:
Ah that worked, I will close this - apologies!

caarlos0:
no problem :)

PR #1647, opened 2020-07-04
docs: Fix typo

kevgo:
If applied, this commit will fix a typo in the documentation.

This change is being made because "below" is mistyped with two l.


mgrachev:
@kevgo Thanks! 🙏

Issue #1643, opened 2020-07-02
Support custom token for Homebrew & Scoop pushes


radeksimko:
**Is your feature request related to a problem? Please describe.**

Currently Homebrew and Scoop publishing relies on the single token given to GoReleaser.

Usually Homebrew/Scoop formulas/manifests don't live in the same repo and in such cases publishing to Homebrew or Scoop requires passing a custom token with elevated privileges (push to both Homebrew and Scoop repositories).

This makes it inflexible for organisations with good security hygiene where Homebrew tap and Scoop bucket may be each managed centrally and tokens issued separately with just access to each individual repository.

**Describe the solution you'd like**

Allow specifying custom token for Homebrew and for Scoop, e.g.

```yaml
brews:
  -
    github:
      owner: github-user
      name: homebrew-tap
      token: {{ .Env.HOMEBREW_TOKEN }}
```

```yaml
scoop:
  bucket:
    owner: github-user
    name: scoop-bucket
    token: {{ .Env.SCOOP_TOKEN }}
```

**Describe alternatives you've considered**

 - Providing a single token with access to (1) current repository, (2) Homebrew tap, and (3) Scoop bucket
 - Use `skip_upload: true` for Homebrew and handle commit & push outside of GoReleaser

PR #1649, opened 2020-07-06
refactor: align Homebrew repo handling with Scoop

radeksimko:
This change should be practically no-op from _functional_ perspective, however **it does introduce 2 deprecations** in the `brews` section:

The following individual GitHub/GitLab sections:
```yaml
brews:
  -
    github:
      owner: github-owner
      name: homebrew-tap
```
```yaml
brews:
  -
    gitlab:
      owner: gitlab-owner
      name: homebrew-tap
```

are proposed to be deprecated in favour of a single section

```yaml
brews:
  -
    tap:
      owner: repo-owner
      name: homebrew-tap
```

which aligns Homebrew with Scoop:

```yaml
scoop:
  bucket:
    owner: user
    name: scoop-bucket
```

This better communicates to the end-user what will happen anyway. e.g. previously one could define `brews[].gitlab` and while providing GitHub token - in which case we'd generate GitHub URLs _and_ still attempt to upload the formula to GitHub (presumably failing to do so).

This makes it more obvious that one cannot publish "cross provider", e.g. release from GitHub repo and publish formula to GitLab tap. I don't think that would be a very common scenario anyway and the user can still set `skip_upload: true` and upload it themselves.

--- 

This was mainly motivated by https://github.com/goreleaser/goreleaser/pull/1547 and https://github.com/goreleaser/goreleaser/issues/1643


caarlos0:
awesome, thanks! 🙏

PR #1654, opened 2020-07-07
docs: add missing quotes around templating syntax

radeksimko:
Quotes are apparently necessary, otherwise the example leads to the following error:

```
   • releasing...
   • loading config file       file=.goreleaser.yml
   ⨯ release failed after 0.00s error=yaml: unmarshal errors:
  line 59: cannot unmarshal !!map into string
```


caarlos0:
sorry, didn't catch that on code review :P

PR #1657, opened 2020-07-08
feat: support closing milestones

### LINKED ISSUE #1415: Support Closing GitHub Milestone

It would be great if we could automatically close an associated [GitHub milestone](https://help.github.com/en/github/managing-your-work-on-github/about-milestones) when releasing. Currently this is done manually. Our milestones are named matching the target release tag name, e.g. `v1.2.3`.

New configuration to enable the handling of GitHub milestones, e.g.

```yaml
# .goreleaser.yml
milestones:
  # You can have multiple milestone configs
  -
    # Repo for the milestone.
    # Default is extracted from the origin remote URL.
    github:
      owner: user
      name: repo

    # Whether to close the milestone
    # Disable to skip performing the milestone action
    close: true

    # Handling for missing milestone when closing the milestone
    #   error: Prevent release if missing
    #   skip: Skip if missing
    #
    # Default: skip
    missing_on_close: error|skip

    # Name of milestone
    #
    # Default: {{ .Tag }}
    name: Current Release

    # Other potential settings after initial implementation

    # Handling for open issues when closing the milestone
    #   error: Prevent release if open issues in milestone
    #   leave: Leave open in milestone
    #   move: Move issues from this milestone to next
    #
    # Default: none
    open_issues_on_close: error|leave|move

    # Whether to open the milestone
    # Disable to skip the milestone action
    # Note: it can only be one: either close or open
    #
    # Default: false
    open: true

    # Name of next milestone
    next_name: Next Release
```

When added with `close: true`, this configuration triggers logic to find the GitHub milestone by name to fetch its ID, then close the milestone.

Externally running a script to wrap the goreleaser process or separate build step with its own script to handle this.

**Additional context**

The closing of the milestone triggers other automated behaviors in our repositories, such as commenting on all issues and pull requests associated with that milestone to let folks know the functionality or fix has been released and how to update. That additional functionality seems out of scope for this project. 😄 

Thank you for the consideration! I am willing to contribute if this would be approved.

bflad:
Closes #1415

---

Enables configuration to automatically close milestones matching the tag name (or any configurable name template) at the end of the publishing pipeline with GitHub, GitLab, and Gitea. In its simplest form:

```yaml
milestones:
  - close: true
```

Will close the milestone title matching the release tag after all publishing has occurred.

### DIFF
--- a/internal/client/gitea.go
+++ b/internal/client/gitea.go
@@ -60,7 +60,7 @@ func (c *giteaClient) CloseMilestone(ctx *context.Context, repo Repo, title stri
 	}
 
 	if milestone == nil {
-		return fmt.Errorf("no open milestone matching: %s", title)
+		return ErrNoMilestoneFound{Title: title}
 	}
 
 	closedState := string(gitea.StateClosed)
@@ -74,11 +74,7 @@ func (c *giteaClient) CloseMilestone(ctx *context.Context, repo Repo, title stri
 
 	_, err = c.client.EditMilestone(repo.Owner, repo.Name, milestone.ID, opts)
 
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return err
 }
 
 // CreateFile creates a file in the repository at a given path
--- a/internal/client/gitea.go
+++ b/internal/client/gitea.go
@@ -51,6 +51,36 @@ func NewGitea(ctx *context.Context, token string) (Client, error) {
 	return &giteaClient{client: client}, nil
 }
 
+// CloseMilestone closes a given milestone.
+func (c *giteaClient) CloseMilestone(ctx *context.Context, repo Repo, title string) error {
+	milestone, err := c.getMilestoneByTitle(repo, title)
+
+	if err != nil {
+		return err
+	}
+
+	if milestone == nil {
+		return fmt.Errorf("no open milestone matching: %s", title)
+	}
+
+	closedState := string(gitea.StateClosed)
+
+	opts := gitea.EditMilestoneOption{
+		Deadline:    milestone.Deadline,
+		Description: &milestone.Description,
+		State:       &closedState,
+		Title:       milestone.Title,
+	}
+
+	_, err = c.client.EditMilestone(repo.Owner, repo.Name, milestone.ID, opts)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
 // CreateFile creates a file in the repository at a given path
 // or updates the file if it exists.
 func (c *giteaClient) CreateFile(
@@ -192,4 +222,22 @@ func (c *giteaClient) Upload(
 		return RetriableError{err}
 	}
 	return nil
+}
+
+// getMilestoneByTitle returns a milestone by title.
+func (c *giteaClient) getMilestoneByTitle(repo Repo, title string) (*gitea.Milestone, error) {
+	// The Gitea API/SDK does not provide lookup by title functionality currently.
+	milestones, err := c.client.ListRepoMilestones(repo.Owner, repo.Name, gitea.ListMilestoneOption{})
+
+	if err != nil {
+		return nil, err
+	}
+
+	for _, milestone := range milestones {
+		if milestone.Title == title {
+			return milestone, nil
+		}
+	}
+
+	return nil, nil
 }
--- a/internal/client/github.go
+++ b/internal/client/github.go
@@ -65,7 +65,7 @@ func (c *githubClient) CloseMilestone(ctx *context.Context, repo Repo, title str
 	}
 
 	if milestone == nil {
-		return fmt.Errorf("no open milestone matching: %s", title)
+		return ErrNoMilestoneFound{Title: title}
 	}
 
 	closedState := "closed"
@@ -79,11 +79,7 @@ func (c *githubClient) CloseMilestone(ctx *context.Context, repo Repo, title str
 		milestone,
 	)
 
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return err
 }
 
 func (c *githubClient) CreateFile(
--- a/internal/client/github.go
+++ b/internal/client/github.go
@@ -56,6 +56,36 @@ func NewGitHub(ctx *context.Context, token string) (Client, error) {
 	return &githubClient{client: client}, nil
 }
 
+// CloseMilestone closes a given milestone.
+func (c *githubClient) CloseMilestone(ctx *context.Context, repo Repo, title string) error {
+	milestone, err := c.getMilestoneByTitle(ctx, repo, title)
+
+	if err != nil {
+		return err
+	}
+
+	if milestone == nil {
+		return fmt.Errorf("no open milestone matching: %s", title)
+	}
+
+	closedState := "closed"
+	milestone.State = &closedState
+
+	_, _, err = c.client.Issues.EditMilestone(
+		ctx,
+		repo.Owner,
+		repo.Name,
+		*milestone.Number,
+		milestone,
+	)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
 func (c *githubClient) CreateFile(
 	ctx *context.Context,
 	commitAuthor config.CommitAuthor,
@@ -186,4 +216,39 @@ func (c *githubClient) Upload(
 		return err
 	}
 	return RetriableError{err}
+}
+
+// getMilestoneByTitle returns a milestone by title.
+func (c *githubClient) getMilestoneByTitle(ctx *context.Context, repo Repo, title string) (*github.Milestone, error) {
+	// The GitHub API/SDK does not provide lookup by title functionality currently.
+	opts := &github.MilestoneListOptions{
+		ListOptions: github.ListOptions{PerPage: 100},
+	}
+
+	for {
+		milestones, resp, err := c.client.Issues.ListMilestones(
+			ctx,
+			repo.Owner,
+			repo.Name,
+			opts,
+		)
+
+		if err != nil {
+			return nil, err
+		}
+
+		for _, m := range milestones {
+			if m != nil && m.Title != nil && *m.Title == title {
+				return m, nil
+			}
+		}
+
+		if resp.NextPage == 0 {
+			break
+		}
+
+		opts.Page = resp.NextPage
+	}
+
+	return nil, nil
 }
--- a/internal/client/gitlab.go
+++ b/internal/client/gitlab.go
@@ -48,6 +48,41 @@ func NewGitLab(ctx *context.Context, token string) (Client, error) {
 	return &gitlabClient{client: client}, nil
 }
 
+// CloseMilestone closes a given milestone.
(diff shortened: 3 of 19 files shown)


internal/client/github.go
```diff
@@ -56,6 +56,36 @@ func NewGitHub(ctx *context.Context, token string) (Client, error) {
...
+
+	_, _, err = c.client.Issues.EditMilestone(
+		ctx,
+		repo.Owner,
+		repo.Name,
+		*milestone.Number,
+		milestone,
+	)
+
+	if err != nil {
```
caarlos0, line 82: could be just `return err`

internal/client/gitea.go
```diff
@@ -51,6 +51,36 @@ func NewGitea(ctx *context.Context, token string) (Client, error) {
...
+	opts := gitea.EditMilestoneOption{
+		Deadline:    milestone.Deadline,
+		Description: &milestone.Description,
+		State:       &closedState,
+		Title:       milestone.Title,
+	}
+
+	_, err = c.client.EditMilestone(repo.Owner, repo.Name, milestone.ID, opts)
+
+	if err != nil {
```
caarlos0, line 77: can just `return err`

internal/client/gitea.go
```diff
@@ -51,6 +51,36 @@ func NewGitea(ctx *context.Context, token string) (Client, error) {
...
+// CloseMilestone closes a given milestone.
+func (c *giteaClient) CloseMilestone(ctx *context.Context, repo Repo, title string) error {
+	milestone, err := c.getMilestoneByTitle(repo, title)
+
+	if err != nil {
+		return err
+	}
+
+	if milestone == nil {
+		return fmt.Errorf("no open milestone matching: %s", title)
```
caarlos0, line 63: maybe create an error type `ErrNoMilestoneFound` or something like that? 🤔

internal/client/gitlab.go
```diff
@@ -48,6 +48,41 @@ func NewGitLab(ctx *context.Context, token string) (Client, error) {
...
+		Title:       &milestone.Title,
+	}
+
+	_, _, err = c.client.Milestones.UpdateMilestone(
+		repo.String(),
+		milestone.ID,
+		opts,
+	)
+
+	if err != nil {
```
caarlos0, line 79: `return err` :)

www/docs/customization/milestone.md
```diff
@@ -0,0 +1,43 @@
...
+      name: repo
+
+    # Gitea repository for the milestone
+    # Default is extracted from the origin remote URL
+    gitea:
+      owner: user
+      name: repo
+
+    # Whether to close the milestone
+    # Default is true
```
caarlos0, line 34: Maybe the default should be false? Otherwise people might just upgrade goreleaser and get their milestones closed without understanding why 🤔

internal/pipe/milestone/milestone.go
```diff
@@ -0,0 +1,132 @@
...
+
+		// nolint: exhaustive
+		switch ctx.TokenType {
+		case context.TokenTypeGitLab:
+			{
+				if milestone.GitLab.Name == "" {
+					repo, err := git.ExtractRepoFromConfig()
+					if err != nil {
+						return err
+					}
```
caarlos0, line 43: repeated a couple of times, maybe move to outside the switch?

www/docs/customization/milestone.md
bflad: Oh ha sorry, the default was false since the configuration is boolean and it wasn't defaulted. Fixing the documentation page. 👍

internal/pipe/milestone/milestone.go
bflad: Sure thing -- I'm guessing `internal/pipe/release/release.go` will need the same treatment, but I'll leave it alone in this changeset.

bflad: Hmm it might not be so clean since it will still need the conditionals for error handling (e.g. only error when the context token matches the VCS type and its not configured). Willing to adjust as necessary though if you have ideas!

internal/client/github.go
bflad: Updated in https://github.com/goreleaser/goreleaser/pull/1657/commits/5aea6dd9fed3c81f9878bae771456b73722ba4a7 👍

internal/client/gitea.go
bflad: Updated in https://github.com/goreleaser/goreleaser/pull/1657/commits/5aea6dd9fed3c81f9878bae771456b73722ba4a7 👍

internal/client/gitea.go
bflad: Updated in https://github.com/goreleaser/goreleaser/pull/1657/commits/5aea6dd9fed3c81f9878bae771456b73722ba4a7 👍

internal/client/gitlab.go
bflad: Updated in https://github.com/goreleaser/goreleaser/pull/1657/commits/5aea6dd9fed3c81f9878bae771456b73722ba4a7 👍

www/docs/customization/milestone.md
bflad: Updated in https://github.com/goreleaser/goreleaser/pull/1657/commits/5aea6dd9fed3c81f9878bae771456b73722ba4a7 👍

pkg/config/config.go
```diff
@@ -282,6 +282,15 @@ type Release struct {
 	ExtraFiles   []ExtraFile `yaml:"extra_files,omitempty"`
 }
 
+// Milestone config used for VCS milestone.
+type Milestone struct {
+	GitHub       Repo   `yaml:",omitempty"`
+	GitLab       Repo   `yaml:",omitempty"`
+	Gitea        Repo   `yaml:",omitempty"`
```
radeksimko, line 289: Generally the way authentication is done currently we don't support more than 1 provider at a time, so you could not e.g. close a milestone in GitHub and GitLab at the same time. Only a single token is expected at a time.

Relatedly I recently refactored similar granular logic in Homebrew https://github.com/goreleaser/goreleaser/pull/1649

So I think we should just reduce these to `Repo Repo `yaml:",omitempty"`

internal/pipe/milestone/milestone.go
radeksimko: Would it help to maybe leave the GitLab/GitHub/Gitea logic entirely outside of the `pipe/milestone` and deal with the differences in `client.CloseMilestone`?

Also if the config is unified, as I suggested in my other comment, so that may help too?

pkg/config/config.go
caarlos0: True that... I think it makes sense to allow a single repo indeed.

bflad: Updated to single `repo` configuration in e95effc3db84a45d4a067a428490945baf7057df 👍

internal/pipe/milestone/milestone.go
bflad: Updated to single `repo` configuration in e95effc3db84a45d4a067a428490945baf7057df 👍

bflad:
After that last commit with the missing `Defaulter` addition (not sure if there's a good way to add testing for that sort of thing), verified end-to-end with a real repository and milestone. 🚀 

Screen Shot 2020-07-09 at 8 25 19 AM

internal/pipe/milestone/milestone.go
```diff
@@ -0,0 +1,92 @@
...
+		repo := client.Repo{
+			Name:  milestone.Repo.Name,
+			Owner: milestone.Repo.Owner,
+		}
+
+		log.WithField("milestone", name).
+			WithField("repo", repo.String()).
+			Info("closing milestone")
+
+		err = vcsClient.CloseMilestone(ctx, repo, name)
```
caarlos0, line 84: I wonder if wether should fail the release if a milestone fails to close 🤔

bflad: In my feature request, I did have this as a potential configuration option:

```yaml
    # Handling for missing milestone when closing the milestone
    #   error: Prevent release if missing
    #   skip: Skip if missing
    #
    # Default: skip
    missing_on_close: error|skip
```

Please let me know if you would prefer that (or something like it, naming could probably be better) should be implemented and I can sneak that in. 😄

caarlos0: I think its a good idea, or just skip by default... wdyt?

bflad: Settled on this in https://github.com/goreleaser/goreleaser/pull/1657/commits/409fd946f5c785f0f1822d2e30a5411dc2e6c3cf:

```yaml
    # Fail release on errors, such as missing milestone on close
    # Default is false
    fail_on_error: true
```

Which will skip errors by default. Please let me know what you think!

internal/client/gitea.go
```diff
@@ -51,6 +51,32 @@ func NewGitea(ctx *context.Context, token string) (Client, error) {
...
+	}
+
+	if milestone == nil {
+		return ErrNoMilestoneFound{Title: title}
+	}
+
+	closedState := string(gitea.StateClosed)
+
+	opts := gitea.EditMilestoneOption{
+		Deadline:    milestone.Deadline,
```
radeksimko, line 69: Probably not a big deal when closing the milestone, but I think this is prone to a race condition where the milestone's data gets modified out of bound before the modified version is sent back.

I suppose we are limited by the SDK/API here anyway?

internal/git/config_test.go
```diff
@@ -1,8 +1,9 @@
-package release
+package git_test
```
radeksimko, line 1: Any particular reason why this package should have the `_test` suffix?

bflad: Ah yes! Sorry should've left a PR comment, it was getting an import cycle otherwise:

```
# github.com/goreleaser/goreleaser/internal/git
import cycle not allowed in test
package github.com/goreleaser/goreleaser/internal/git (test)
	imports github.com/goreleaser/goreleaser/internal/testlib
	imports github.com/goreleaser/goreleaser/internal/git
FAIL	github.com/goreleaser/goreleaser/internal/git [setup failed]
```

I can undo the change to move the git config handling into this package so its re-usable, but figured the code made sense in this one spot.

radeksimko: Gotcha, I guess it's fine - we can do the refactoring in a separate PR.

internal/client/gitea.go
bflad: I was operating under the presumption that leaving pointer fields in the update options empty could blank them out in the remote object -- I do not know for sure if that's the actual case though. Happy to change this any which way.

radeksimko: Aye, sorry I meant racing another request/human outside of GoReleaser with GoReleaser, rather than the code itself being racy within GoReleaser. In other words I wondered if there was a way of sending a single "partial update" request as opposed to "get everything" -> "update everything"?

bflad: Not that I'm aware of 😅

radeksimko: 👌LGTM then

PR #1667, opened 2020-07-12
docs: Update quick-start.md

venkyvb:
Improve the quick-start documentation to reflect that `.goreleaser.yml` needs to be checked into the source control. See - https://github.com/goreleaser/goreleaser/discussions/1666

### DIFF
--- a/www/docs/quick-start.md
+++ b/www/docs/quick-start.md
@@ -23,7 +23,7 @@ $ goreleaser init
 ```
 
 You can [customize](/customization) the generated `.goreleaser.yml` or leave
-it as-is, it's up to you. Note that the `.goreleaser.yml` needs to be checked into the source control.
+it as-is, it's up to you. It is best practice to check `.goreleaser.yml` into the source control.
 
 You can test the configuration at any time by running GoReleaser with a few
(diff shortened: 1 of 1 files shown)


www/docs/quick-start.md
```diff
@@ -23,7 +23,7 @@ $ goreleaser init
 ```
 
 You can [customize](/customization) the generated `.goreleaser.yml` or leave
-it as-is, it's up to you.
+it as-is, it's up to you. Note that the `.goreleaser.yml` needs to be checked into the source control.
```
radeksimko, line 26: Strictly speaking it could be in `.gitignore` which also avoids the problem with dirty git state, but I'd say that's in most cases bad idea and there should be almost no reason not to check it in. Sensitive data can be avoided via ENV variables, which is something users should use either way, whether it's OSS or a private project.
```suggestion
it as-is, it's up to you. It is best practice to check `.goreleaser.yml` into the source control.
```

radeksimko:
Not sure why the CI isn't kicking off - I'm going to try to close and re-open the PR.

radeksimko:
Looks like there's an [ongoing incident](https://www.githubstatus.com/) affecting GitHub Actions and the API too so that may be the reason.

PR #1677, opened 2020-07-17
docs: fix OCI annotation example

jgustie:
Signed-off-by: Jeremy Gustie <jeremy@gustie.com>

If applied, this commit will change the example labels in the Docker documentation to match the OCI recommendations.

The `org.label-schema.name` label became `org.opencontainers.image.title` ([Annotations](https://github.com/opencontainers/image-spec/blob/master/annotations.md)).


caarlos0:
Thanks!

PR #1679, opened 2020-07-17
fix: Completer script link errors for snapcraft

### LINKED ISSUE #1678: Snapcraft build failures when using completers
**Describe the bug**

When creating a snap package with multiple builds and multiple apps, setting a completer on the application fails.

**To Reproduce**
```yaml
snap:
  builds:
  - build1
  - build2
  apps:
    app1:
      completer: app1.completer
    app2:
      completer: app2.completer
```

> Make sure your config file is valid by running
> `goreleaser check -f path-to-config-file`.

**Expected behavior**

The snap to build successfully

- OS: all
- OS version: all
- GoReleaser 0.140.0
- GitLab version n/a

**Additional context**

This occurs because https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/snapcraft/snapcraft.go#L273 is executer per application per binary, so `os.Link` fails the second time around with a `File Exists` error.

neoaggelos:
## Changes

Closes #1678

## Description

Do not fail on `os.Link` errors if `os.IsExist(err)`

## Considered alternatives

- Loop through the completer scripts separately and only once.
- Keep track of files that have been linked already and do not try to link again.

### DIFF
--- a/internal/pipe/snapcraft/snapcraft.go
+++ b/internal/pipe/snapcraft/snapcraft.go
@@ -270,7 +270,7 @@ func create(ctx *context.Context, snap config.Snapcraft, arch string, binaries [
 				log.WithField("src", config.Completer).
 					WithField("dst", destCompleterPath).
 					Debug("linking")
-				if err := os.Link(config.Completer, destCompleterPath); err != nil {
+				if err := os.Link(config.Completer, destCompleterPath); err != nil && !os.IsExist(err) {
 					return errors.Wrap(err, "failed to link completer")
 				}
 				if err := os.Chmod(destCompleterPath, 0644); err != nil {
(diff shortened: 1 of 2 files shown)


caarlos0:
thanks for the PR :)

can you add a test case for it as well?

thanks!

neoaggelos:
Changed approach to moving on if error is that file exists because of [CI failures for code complexity](https://github.com/goreleaser/goreleaser/runs/882656007?check_suite_focus=true#step:7:70).

Also, updated `TestCompleter` to use two different builds, effectively testing for this change.

neoaggelos:
By the way, running tests locally fail (Ubuntu 20.04 -- go version go1.14.5 linux/amd64). This should probably be a separate issue if I can replicate it.

caarlos0:
nice, thanks for the PR :) 



> By the way, running tests locally fail (Ubuntu 20.04 -- go version go1.14.5 linux/amd64). This should probably be a separate issue if I can replicate it.

awesome, let me know what you find out :)

caarlos0:
Thinking about this now, why does this happens in the first place? (having the same completer multiple times)

PR #1680, opened 2020-07-17
refactor: avoid copying the completer multiple times

caarlos0:
different approach to #1678 (and its PR #1679): we don't need to copy the completer for each binary, so we can just move that loop to outside of the binaries loop..


neoaggelos:
Just tested and it works great.

caarlos0:
Thanks @neoaggelos :)

PR #1703, opened 2020-07-23
feat: log into Github package registry with token

till:
Related: #1445, #1431

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -14,4 +14,9 @@ if [ -n "$GORELEASER_GITHUB_TOKEN" ] ; then
   export GITHUB_TOKEN=$GORELEASER_GITHUB_TOKEN
 fi
 
+if [ -n "$GITHUB_TOKEN" ]; then
+  # Log into GitHub package registry
+  echo $GITHUB_TOKEN | docker login docker.pkg.github.com -u docker --password-stdin
+fi
+
 goreleaser $@


till:
It could be as simple as that. I created a personal access token for myself, and I seem to be able to log in with `-u docker` always. Not sure how Gitlab works. :)

till:
Is gitlab as simple as, `echo $GITLAB_TOKEN | docker login registry.gitlab.com -u docker --password-stdin`?

till:
I guess, LBNL — should I use env vars for the hosts? So people can override them? E.g., default to the "cloud" names and then in case people run on-premise they can set something else?

caarlos0:
@crazy-max correct me if I'm wrong, but this will not affect the action, right?

I think it doesn't hurt doing this here, but we could probably add it to the action as well... wdyt?

crazy-max:
@caarlos0 

> correct me if I'm wrong, but this will not affect the action, right?

As I can see this only affects the GoReleaser Docker image so nothing related with our action.

> I think it doesn't hurt doing this here, but we could probably add it to the action as well... wdyt?

I think that authentication on a Docker registry (as for snapcraft packages as well) should be decoupled from our action. Some users like to use the [official docker env vars (`DOCKER_CONFIG`)](https://docs.docker.com/engine/reference/commandline/cli/#environment-variables) to authenticate against a registry and also the registry domain could vary depending on your subscription (`docker.pkg.github.com` or `containers.pkg.github.com`).

caarlos0:
> I think that authentication on a Docker registry (as for snapcraft packages as well) should be decoupled from our action. Some users like to use the [official docker env vars (`DOCKER_CONFIG`)](https://docs.docker.com/engine/reference/commandline/cli/#environment-variables) to authenticate against a registry and also the registry domain could vary depending on your subscription (`docker.pkg.github.com` or `containers.pkg.github.com`).

hmmm didn't know about that, thanks for sharing!

that being said, not sure if its worth merging this PR or not... thoughts?

till:
I'm good with not merging. Where would you do this instead?

caarlos0:
> I'm good with not merging. Where would you do this instead?

good question... I think the solution would be to each user override the entrypoint and so on...

I think we can merge this one...

PR #1707, opened 2020-07-26
remove --pull option from docker run

devmarkpro:
docker run command has not the option `--pull`
here's the `docker run` [documentation](https://docs.docker.com/engine/reference/run/)


caarlos0:
thanks

Issue #1710, opened 2020-07-29
Run before pipe after semver pipe


ryancurrah:
**Is your feature request related to a problem? Please describe.**
When using the global hook/before pipe I need to be able to echo the version to a file but that information is not available yet because the semver pipe runs after the before pipe.

**Describe the solution you'd like**
Change the ordering of the pipes so that git and semver template variables are available to the before pipe.

**Additional context**
https://github.com/goreleaser/goreleaser/blob/master/internal/pipeline/pipeline.go#L40-L43

caarlos0:
fixed in #1711

ryancurrah:
Thank you!

PR #1711, opened 2020-07-29
feat(pipeline): change the pipeline order to support additional template variables

ryancurrah:
This change allows us to use all the template variables in the named templates in the before hook.

Issue: GH-1710


caarlos0:
Makes sense, thanks!

PR #1721, opened 2020-08-04
fix(ci): htmltest failing with 416 on some sites

caarlos0:
it defaults headers to `Range: bytes=0-0`, which some providers may not support.

changing to `0-10` seems to yield a 206 on the providers I tested.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -25,4 +25,4 @@ jobs:
         run: |
           cd www
           curl https://htmltest.wjdp.uk | bash
-          ./bin/htmltest -c htmltest.yml .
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -25,4 +25,4 @@ jobs:
         run: |
           cd www
           curl https://htmltest.wjdp.uk | bash
-          ./bin/htmltest -c htmltest.yml site
+          ./bin/htmltest -c htmltest.yml .
```
caarlos0, line 28: not sure why it complains about this now ¯\_(ツ)_/¯

Issue #1734, opened 2020-08-09
Getting invalid object name 'tags/warning' when attempting to release 


AndreasBieber:
**Describe the bug**

When attempting to release via Azure DevOps Pipeline I get the following during the changelog generation which breaks the release in the end:
```
• generating changelog     
      • running git               args=[-c log.showSignature=false describe --tags --abbrev=0 tags/v0.3.18^]
      • git result                output=warning: refname 'tags/v0.3.18' is ambiguous.
v0.3.17

      • running git               args=[-c log.showSignature=false log --pretty=oneline --abbrev-commit --no-decorate --no-color tags/warning: refname tags/v0.3.18 is ambiguous...tags/v0.3.18]
      • git result                output=fatal: invalid object name 'tags/warning'.

   ⨯ release failed after 1.59s error=fatal: invalid object name 'tags/warning'.
```
When disabling changelog generation everything is working as expected.

**To Reproduce**

I'm not able to reproduce this issue because it is working locally as expected.

`.goreleaser.yaml`:
```yaml
dist: ./artifacts/dist
before:
  hooks:
    - go mod download
    - go generate ./...
builds:
- main: ./cmd/enpxio/enpxio.go
  env:
  - CGO_ENABLED=0
  goos:
    - linux
    - windows
    - darwin
  ldflags:
    - -s -w -X dev.azure.com/myorg/myproject/_git/enpxio.git/internal/meta.Version={{.Version}} -X dev.azure.com/myorg/myproject/_git/enpxio.git/internal/meta.Commit={{.Commit}} -X dev.azure.com/myorg/myproject/_git/enpxio.git/internal/meta.BuiltDate={{.Date}}
archives:
- replacements:
    darwin: Darwin
    linux: Linux
    windows: Windows
    386: i386
    amd64: x86_64
  format_overrides:
  - goos: windows
    format: zip
  name_template: "{{ .ProjectName }}_{{ .Os }}_{{ .Arch }}"
checksum:
  name_template: 'checksums.txt'
release:
  disable: true
changelog:
  skip: false
  sort: asc
  filters:
    exclude:
    - '^docs:'
    - '^test:'
    - '^chore:'
    - '^style:'
    - '^refactor:'
```

**Expected behavior**

The release should create a changelog without errors.

**Environment:**

- OS: `linux`
[...]

PR #1740, opened 2020-08-11
fix: ignore git warnings

caarlos0:
should avoid issues like https://github.com/goreleaser/goreleaser/issues/1734, although I'm not sure how to reproduce it (and therefore write a test for it).


caarlos0:
found a reproducible, added a test.

PR #1758, opened 2020-08-16
feat: update to Go 1.15 image

sj14:
**If applied, this commit will...**

Update the docker base image to use Go 1.15

**Why is this change being made?**

Some users may require Go 1.15 features.


caarlos0:
Thanks 🙏

PR #1770, opened 2020-08-25
sign: extra option allowing stdin for commands

greut:
Adding extra fields to the sign pipeline allowing it to pass either data or a file to stdin.

I've been playing with minisign (or signify) as alternative to gpg for signing and it wants a password on stdin. The `sh` workaround is okay but feels wrong.

```yaml
- id: minisign default
  artifacts: checksum
  signature: "${artifact}.minisig"
  cmd: sh
  args:
    - '-c'
    - 'echo "" | minisign -x ${signature} -Sm ${artifact} -t "{{ .ProjectName }} {{ .Tag }}"'
```

vs

```yaml
  cmd: minisign
  stdin_file: /dev/null
  args:
    - -x
    - ${signature}
    - -Sm
    - ${artifact}
    - -t
    - "{{ .ProjectName }}{{ .Tag }}"'
```

As the tests are already using gpg, I've added a key named `password` with the password set to `password`.


caarlos0:
looks good, thanks!

caarlos0:
and sorry for the late response :|

PR #1773, opened 2020-08-28
feat: add support for homebrew HEAD installations

### LINKED ISSUE #1772: Add support for homebrew HEAD installations
Add support for `brew install --HEAD $formula`. This allows users to install the version from master.

**Additional context**
https://docs.brew.sh/Formula-Cookbook#unstable-versions-head

mgrachev:
Close #1772

### DIFF
--- a/internal/pipe/brew/brew.go
+++ b/internal/pipe/brew/brew.go
@@ -265,6 +265,7 @@ func dataFor(ctx *context.Context, cfg config.Homebrew, cl client.Client, artifa
 		DownloadStrategy: cfg.DownloadStrategy,
 		CustomRequire:    cfg.CustomRequire,
 		CustomBlock:      split(cfg.CustomBlock),
+		Head:             cfg.Head,
 	}
 
 	for _, artifact := range artifacts {
--- a/internal/pipe/brew/brew_test.go
+++ b/internal/pipe/brew/brew_test.go
@@ -80,6 +80,8 @@ func TestFullFormulae(t *testing.T) {
 	data.CustomBlock = []string{"devel do", `  url "https://github.com/caarlos0/test/releases/download/v0.1.3/test_Darwin_x86_64.tar.gz"`, `  sha256 "1633f61598ab0791e213135923624eb342196b3494909c91899bcd0560f84c68"`, "end"}
 	data.Install = []string{"custom install script", "another install script"}
 	data.Tests = []string{`system "#{bin}/{{.ProjectName}} -version"`}
+	data.Head = config.HomebrewHead{URL: "https://github.com/caarlos0/test.git", Branch: "develop", Revision: "090930930295adslfknsdfsdaffnasd13"}
+
 	formulae, err := doBuildFormula(context.New(config.Project{
 		ProjectName: "foo",
 	}), data)
--- a/internal/pipe/brew/template.go
+++ b/internal/pipe/brew/template.go
@@ -20,6 +20,7 @@ type templateData struct {
 	Linux            downloadable
 	Arm              downloadable
 	Arm64            downloadable
+	Head             config.HomebrewHead
 }
 
 type downloadable struct {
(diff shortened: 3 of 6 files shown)


caarlos0:
Thinking more about it... adding only the head doesn't seem enough, as the formula doesn't have the instructions to compile anything... 

If we would need to add how to compile too, I would say this is out of scope...

mgrachev:
I  think we can generate something like this and this will work:
```ruby
class Test < Formula
  desc "Some desc"
  homepage "https://google.com"
  version "0.1.3"
  bottle :unneeded

  head do
    url "https://github.com/caarlos0/test.git", :branch => "develop", :revision => "090930930295adslfknsdfsdaffnasd13"
    depends_on "go" => :build 
  end

  if OS.mac?
    url "https://github.com/caarlos0/test/releases/download/v0.1.3/test_Darwin_x86_64.tar.gz"
    sha256 "1633f61598ab0791e213135923624eb342196b3494909c91899bcd0560f84c68"
  elsif OS.linux?
    if Hardware::CPU.intel?
      url "https://github.com/caarlos0/test/releases/download/v0.1.3/test_Linux_x86_64.tar.gz"
      sha256 "1633f61598ab0791e213135923624eb342196b3494909c91899bcd0560f84c67"
    end
    if Hardware::CPU.arm?
      if Hardware::CPU.is_64_bit?
        url "https://github.com/caarlos0/test/releases/download/v0.1.3/test_Arm64.tar.gz"
        sha256 "1633f61598ab0791e213135923624eb342196b3494909c91899bcd0560f84c67"
      else
        url "https://github.com/caarlos0/test/releases/download/v0.1.3/test_Arm6.tar.gz"
        sha256 "1633f61598ab0791e213135923624eb342196b3494909c91899bcd0560f84c67"
      end
    end
  end
  
  devel do
    url "https://github.com/caarlos0/test/releases/download/v0.1.3/test_Darwin_x86_64.tar.gz"
    sha256 "1633f61598ab0791e213135923624eb342196b3494909c91899bcd0560f84c68"
  end
  
  depends_on "gtk+"
  
  conflicts_with "svn"

  def install
    if build.head?
      system "go", "build", *std_go_args
    else
      custom install script
      another install script
    end
  end

  def caveats; <<~EOS
    Here are some caveats
  EOS
  end

  plist_options :startup => false

  def plist; <<~EOS
    it works
  EOS
  end

  test do
    system "#{bin}/foo -version"
  end
end
```

@caarlos0 What do you think about it?

caarlos0:
The problem with this is that some projects may require different things on build (flags, dependencies, etc). So it will be another build config... 

Maybe its easier to just add an example to the docs on how to create a standalone `myproject-head` formula? wdyt?

mgrachev:
Maybe we will do something like this?
```ruby
class Test < Formula
  desc "Some desc"
  homepage "https://google.com"
  version "0.1.3"
  bottle :unneeded

  head do
    url "https://github.com/caarlos0/test.git", :branch => "develop"
    depends_on "goreleaser" => :build 
  end

  if OS.mac?
  # ...
  end
 
  def install
    if build.head?
      system "goreleaser", "build", *std_go_args
    else
      custom install script
      another install script
    end
  end

  test do
    system "#{bin}/foo -version"
  end
end
```

caarlos0:
Still weird for whoever installs it... I would rather keep this on docs only. 🤔

mgrachev:
Okay 👌 
How should it look in the docs?
Could you give an example?

caarlos0:
My idea would be to to put an example (like [this](https://github.com/goreleaser/goreleaser/pull/1773#issuecomment-682677059)) in the end of the `brew.md` file... something like:

```md
## Head formulas

GoReleaser does not generate those for you, but you can create a `my-app-head.rb` file similar to this one:

-- example code here

And your users can then install them with:

brew cask install user/tap/my-app-head

Check the homebrew documentation for more info.
```

(just out of my head, feel free to change as you see fit)

Issue #1235, opened 2019-11-08
Customize go binary used to build 


natefinch:
I want to be able to build with a specific version of go, not just whatever version exists as "go" on the current system. 

**Describe the solution you'd like**
Add a GoBinaryPath (or some such) configuration option in the yaml.  This would allow you to choose one of the downloadable go binaries (like go1.13.4) and then you always know you're building with the version of Go you support (and fail if it doesn't exist).  This is a way to "pin" the go version you're building with, even if developers have different versions of go on their system.

**Describe alternatives you've considered**
A shell alias can work if you're working in a shell context, but I want to be able to do this within a magefile, so that's not really a solution for me.

This is mostly because we currently do releases from dev machines, clearly if this were being run in CI/CD, we'd just make sure the right version of Go is installed there.

caarlos0:
This is interesting, depending on how we do this, it would allow to build and ship the same thing with different go versions as well... 

is that something people usually do? I can't think of an usecase for that...

BTW, you can add custom envs to the build, maybe add a custom PATH there would solve the issue?

natefinch:
I'm not sure why anyone would release multiple binaries built with different versions of go. This is mostly an insurance policy to make sure your code is built with the version of go you expect it to be and not just whatever version happens to exist on the machine labelled "go".  

I think this is mostly only useful for people running goreleaser on their personal machine and not through CI where they should be able to specifically set what version of Go exists.

craigfurman:
This would also help goreleaser docker image users. At the moment, goreleaser will use whatever go version is built into the image: https://github.com/goreleaser/goreleaser/blob/master/Dockerfile#L1

caarlos0:
I'm not sure how the api should look like, but I agree we can add something related to this.

open to ideas.

natefinch:
I think just a configuration value that has the name of the go binary you want to use.  It'll then use that name in place of "go" when running commands that would otherwise use the "go" command. Probably most of the time it would be something in your PATH, but there's nothing saying someone couldn't choose a relative path if they wanted.   That's what I do in Mage.

GoBinary: "go1.12.3" // defaults to "go"

caarlos0:
SGTM @natefinch 

---

sorry for the late response, lost the notification somehow

arlaneenalra:
[...]

PR #1775, opened 2020-08-29
feat: Add `gobinary` option to build configs.

arlaneenalra:
If applied, this commit will add a `gobinary` option to the build config allowing specification of specific go binaries to use during a build.

The change is being made because it was listed in #1235 as requested.

See: #1235 for further details.


caarlos0:
thanks!

PR #1786, opened 2020-09-02
fix: artifact string method switch statement

abursavich:
Existing implementation seemed to rely on implicit case fallthrough, which doesn't exist in Go. Moving default return to default case helps catch this error in the future, as it will cause a "missing return" compile error.


caarlos0:
Thanks!

PR #1796, opened 2020-09-07
refactor: use git tag --points-at

### LINKED ISSUE #1793: Use git tag --points-at to detect the legitimate tag
@caarlos0 

Refs. goreleaser/goreleaser-action#238 goreleaser/goreleaser-action#239

As explained on Slack, I think `git describe` is (maybe) not a good way to detect the "proper" tag. We should first use [git tag --points-at](https://git-scm.com/docs/git-tag#Documentation/git-tag.txt---points-atltobjectgt) to detect the legitimate latest tag if a commit has multi tags (sort by `-version:creatordate`). If none is found/valid then we can use `git describe --tags --abbrev=0`.

You can find an implementation in [our GitHub Action](https://github.com/goreleaser/goreleaser-action/blob/097bb3c0c5b5249fdc72ccdff0f874aebdc93839/src/git.ts#L12-L31).

caarlos0:
closes #1793

is this what you meant @crazy-max?

### DIFF
--- a/internal/pipe/git/git.go
+++ b/internal/pipe/git/git.go
@@ -148,11 +148,26 @@ func getFullCommit() (string, error) {
 }
 
 func getTag() (string, error) {
-	if tag := os.Getenv("GORELEASER_CURRENT_TAG"); tag != "" {
-		return tag, nil
+	var tag string
+	var err error
+	for _, fn := range []func() (string, error){
+		func() (string, error) {
+			return os.Getenv("GORELEASER_CURRENT_TAG"), nil
+		},
+		func() (string, error) {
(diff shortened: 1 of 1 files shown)


crazy-max:
@caarlos0 Yes pretty much but `git tag --points-at` can return a [list of tags](https://github.com/goreleaser/goreleaser-action/blob/master/src/git.ts#L25).

caarlos0:
yeah, `git.Clean` takes the first line of STDOUT only... so.. it should be fine I think 🤔

crazy-max (approved):
Oh ok I missed that. LGTM then!

caarlos0:
thanks @crazy-max 🙏

PR #1800, opened 2020-09-08
fix: allow help command

melbahja:
Allow help commands like:
```bash
goreleaser help
goreleaser help check
```

When you run: `goreleaser help` it should output help and not run `releaser`.

Thanks.


caarlos0:
Thanks @melbahja !

PR #1805, opened 2020-09-15
feat: IDs filter for Checksums pipe

jpcoenen:
This allows specific artifacts to be excluded from the checksums file.

The reason for introducing this, is the requirement of [Terraform Registry](https://www.terraform.io/docs/registry/providers/index.html) releases to only contain the prescribed archives in the checksum file. If the file contains more, the release is not accepted by the Terraform Registry. For our specific release, we want to include some other archives for backwards compatibility, which breaks the release to the registry.

An example of how it would work:
```yaml
checksum:
  ids: 
    - foo
    - bar
```
Will only include all artifacts with the `ID` `foo` or `bar`. If `ids` is not set, the current behaviour is unchanged.


jpcoenen:
@caarlos0 is there any chance of this making it into `master`? I could really use it for my project, but if it's unwanted, I know to look for an alternative solution.

internal/pipe/checksums/checksums.go
```diff
@@ -40,14 +40,19 @@ func (Pipe) Run(ctx *context.Context) (err error) {
...
 			artifact.ByType(artifact.UploadableArchive),
 			artifact.ByType(artifact.UploadableBinary),
 			artifact.ByType(artifact.UploadableSourceArchive),
 			artifact.ByType(artifact.LinuxPackage),
-		),
-	).List()
+		)
+	} else {
+		filter = artifact.ByIDs(ctx.Config.Checksum.IDs...)
+	}
```
caarlos0, line 53: I would say this should be an `and` (ids && filters by type)

caarlos0 (changes requested):

caarlos0:
> @caarlos0 is there any chance of this making it into `master`? I could really use it for my project, but if it's unwanted, I know to look for an alternative solution.

Sorry @jpcoenen , totally lost this in the notifications... yeah, we can totally include some version of this. Just make a note.

Thanks for the PR and sorry for the delay :)

internal/pipe/checksums/checksums.go
jpcoenen: You're right. I was under the assumption that the proposed behaviour was in line with how other ID filters work, but looking at it again, that does not seem to be the case. I changed it.

jpcoenen:
> > @caarlos0 is there any chance of this making it into `master`? I could really use it for my project, but if it's unwanted, I know to look for an alternative solution.
> 
> Sorry @jpcoenen , totally lost this in the notifications... yeah, we can totally include some version of this. Just make a note.
> 
> Thanks for the PR and sorry for the delay :)

Don't worry about it. Thank you very much for getting back to and mostly for building and maintaining this awesome tool!

What do you mean by "just make a note"?

caarlos0 (approved):

caarlos0:
> What do you mean by "just make a note"?

Oh, just that I left a comment on the code 😬

caarlos0:
> Thank you very much for getting back to and mostly for building and maintaining this awesome tool!

Thanks, and thanks again for the pr 🙏

jpcoenen:
Awesome!

PR #1812, opened 2020-09-20
refactor: replace pkg/errors.Wrap with fmt.Errorf

caarlos0:
using the "new" way of wrapping errors.


radeksimko (approved):
LGTM, good cleanup!

radeksimko:
I think there's a bug somewhere though, based on the failing test.

caarlos0:
Seems related to gitea, will take a look... but I think it was introduced by #1807

PR #1823, opened 2020-09-24
fix #1821: use name template for bare binary files

### LINKED ISSUE #1821: `archives.name_template` not respected when `format: binary`
**Describe the bug**

When I set the `archives.format: binary` it no longer respects `name_template`.

**To Reproduce**

`.goreleaser.yml`:

```yml
# This is an example goreleaser.yaml file with some sane defaults.
# Make sure to check the documentation at http://goreleaser.com
project_name: arc
before:
  hooks:
    # You may remove this if you don't use go modules.
    - go mod download
    # you may remove this if you don't need go generate
    - go generate ./...
builds:
  -
    env:
      - CGO_ENABLED=0
    main: ./cmd/foo
    goos:
      - linux
      - windows
      - darwin
    goarch:
      - 386
      - amd64
      - arm
      - arm64
    goarm:
      - 6
      - 7
archives:
  -
    name_template: "{{ .ProjectName }}_{{ .Os }}_{{ .Arch }}{{ .Arm }}"
    format: binary
    replacements:
      darwin: mac
checksum:
snapshot:
  name_template: "{{ .Tag }}-next"
changelog:
  sort: asc
  filters:
    exclude:
      - '^docs:'
      - '^test:'
```

```bash
goreleaser --snapshot --skip-publish --rm-dist
```

```txt
foo_darwin_amd64
foo_linux_386
foo_linux_amd64
foo_linux_arm64
foo_linux_arm_6
foo_linux_arm_7
foo_windows_386
foo_windows_amd64
```
This is the `target` format, not the `archive` format:
https://github.com/goreleaser/goreleaser/blob/master/internal/builders/golang/targets.go#L17

**Expected behavior**

I expect that the release files would still output in the specified template format:

```txt
foo_mac_amd64
foo_linux_386
foo_linux_amd64
foo_linux_arm64
foo_linux_arm6
foo_linux_arm7
foo_windows_386
foo_windows_amd64
```

Oddly, it looks like the default would actually do what I want:

- https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/archive/archive.go#L70

But it doesn't...

- OS: mac
- GoReleaser Version 

goreleaser version 0.143.0
commit: a92774f602331c53cd6b2c3c3ec8ec9058a26a7f
built at: 2020-09-06T14:53:58Z
built by: gorelease

coolaj86:
Update `skip` to rename binary files according to `archives.name_template` in the same way that `tar`, `zip`, and others are updated.

This is to fix https://github.com/goreleaser/goreleaser/issues/1821

## Remaining Problem

The directory name and the binary name are the same by default, so I added a `bins` directory.

I realize this is probably not a desirable solution, so I'd like some direction on how to better solve for that.

### DIFF
--- a/internal/pipe/archive/archive.go
+++ b/internal/pipe/archive/archive.go
@@ -196,10 +196,17 @@ func skip(ctx *context.Context, archive config.Archive, binaries []*artifact.Art
 		if err != nil {
 			return err
 		}
+		name += binary.ExtraOr("Ext", "").(string)
+		renamePath := filepath.Join(ctx.Config.Dist, name)
+		if err := os.Rename(binary.Path, renamePath); err != nil {
+			return err
+		}
+		_ = os.Remove(filepath.Dir(binary.Path))
 		ctx.Artifacts.Add(&artifact.Artifact{
-			Type:   artifact.UploadableBinary,
-			Name:   name + binary.ExtraOr("Ext", "").(string),
-			Path:   binary.Path,
+			Type: artifact.UploadableBinary,
+			// this is the name it shows
+			Name:   name,
+			Path:   renamePath,
 			Goos:   binary.Goos,
 			Goarch: binary.Goarch,
 			Goarm:  binary.Goarm,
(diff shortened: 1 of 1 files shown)


caarlos0:
see https://github.com/goreleaser/goreleaser/issues/1821#issuecomment-702139155

Issue #1538, opened 2020-05-21
Ability to run uploads on snapshot


jaxxstorm:
**Is your feature request related to a problem? Please describe.**
When I'm testing my goreleaser configuration for a project, I have to do the following:

- generate the config
- test some config locally
- then test it in CI

In order to actually do the test in CI, i _have_ to create a tag which then might publish and artifact or upload that artifact to an s3 bucket or whichever other release configs you have.

This is pretty painful during the initial iterations and means lots of mistakes are made

**Describe the solution you'd like**

It would be great if we could have a flag that goes with the `--snapshot` flag, something like `--publish` which allows you to test this configuration and uploads the artifacts to wherever, without adding the latest release tag in Github

**Describe alternatives you've considered**
We could optionally use the new `goreleaser build` mechanism and do the uploads manually, but it defeats the object of using goreleaser :)

**Additional context**
These builds could be used by `homebrew` for its `devel` channel as well

caarlos0:
similar to https://github.com/goreleaser/goreleaser/discussions/1534 maybe?

jaxxstorm:
Yes, similar. Feel free to fold this issue into that one

caarlos0:
this is now available with the `--nightly` flag on goreleaser-pro.

closing

sriio:
@caarlos0 Any chance to get it on oss version ?

caarlos0:
not for now, no.

PR #1825, opened 2020-09-27
feat: add flag for publishing snapshot artifacts

sriio:

Add a flag for release command to force pushing  snapshot artifacts

## Why is this change being made ?

Like maven can do, during a project life, you maybe need to provide pre built snapshot binary of your application for differents reason (internal test, CD, etc...). The fact that allow snapshot publishing will help a lot.

Relate to issue #1538


sriio:
@caarlos0 Any chance to accept this PR ?

caarlos0:
> @caarlos0 Any chance to accept this PR ?

don't know yet, need to think more about it, test it and etc...

caarlos0:
the idea of snapshots is for local build only. Maybe thats not clear enough in docs, but to do what you want you should probably use tags with pre-release and/or build identifiers.

closing for now, thanks for the pr though :)

PR #1826, opened 2020-09-29
docs: how to cross-compile Go with CGO

troian:
Updating docs with CGO and cross-compiling how to.
Follow up to [conversation](https://gophers.slack.com/archives/C3T1JU24B/p1600905690004300)


caarlos0:
awesome, thanks!

PR #1832, opened 2020-10-02
feat: allows to specify custom files for a specific GOOS

ermos:
## Allows to specify custom files for a specific GOOS

Hi,

I haven't found parameters for able to push custom files in archive for a specific GOOS, so I have make this one for my personal use and I think that can be useful for other people.

### How it's work ?

It is similar format than "format_overrides", that look like : 

```yml
archives:
  - file_custom:
      - goos: linux
        files:
          - linux.sh
      - goos: darwin
        files:
          - mac.sh
```
We specify goos output and we add files in an array.


caarlos0:
You can do the same with multiple builds/archive sections.

["Yes is forever"](https://twitter.com/solomonstre/status/715277134978113536), so, I rather merge this - or at least not right now.

Thanks for the PR either way.

PR #1839, opened 2020-10-06
refactor: use require on all tests

caarlos0:
makes tests fail earlier and the output easier to read... if a test has more failures, I assume people will run them manually to debug/fix the issue anyway, therefore catching further assert/require failures.

does that make sense?

---

PS: automated change with editor replace all, maybe there are things wrong 🤣

### DIFF
--- a/internal/artifact/artifact_test.go
+++ b/internal/artifact/artifact_test.go
@@ -6,7 +6,6 @@ import (
 	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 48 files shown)


radeksimko (approved):
LGTM.

I admit I didn't read all 2k+ lines 😅  but it looks sensible from a quick glance and the CI gives us some extra confidence.

Issue #1820, opened 2020-09-24
resolve templates in env section


troian:
In some cases like [cross-compiling](https://gophers.slack.com/archives/C3T1JU24B/p1600905690004300) it is quite handy to substitute long strings especially when there are repetitive.
In [example](https://github.com/troian/golang-cross/blob/master/.goreleaser.yaml) instead of writing
```yaml
builds:
  - id: darwin-amd64
    env:
      - PKG_CONFIG_SYSROOT_DIR=/go/src/github.com/troian/golang-cross/sysroot/macos/amd64
      - PKG_CONFIG_PATH=/go/src/github.com/troian/golang-cross/sysroot/macos/amd64/usr/local/lib/pkgconfig
  - id: linux-amd64
    env:
      - CGO_ENABLED=1
      - PKG_CONFIG_SYSROOT_DIR=/go/src/github.com/troian/golang-cross/sysroot/linux
      - PKG_CONFIG_PATH=/go/src/github.com/troian/golang-cross/sysroot/linux/usr/lib/x86_64-linux-gnu/pkgconfig
```
it would be good to use path prefixes from env variable
```yaml
builds:
  - id: darwin-amd64
    env:
      - PKG_CONFIG_SYSROOT_DIR={{ .Env.SYSROOT_DIR }}/macos/amd64
      - PKG_CONFIG_PATH={{ .Env.SYSROOT_DIR }}/macos/amd64/usr/local/lib/pkgconfig
  - id: linux-amd64
    env:
      - CGO_ENABLED=1
      - PKG_CONFIG_SYSROOT_DIR={{ .Env.SYSROOT_DIR }}/linux
      - PKG_CONFIG_PATH={{ .Env.SYSROOT_DIR }}/linux/usr/lib/x86_64-linux-gnu/pkgconfig
```

m4ver1k:
@caarlos0 can I give this a try ?

caarlos0:
hey @m4ver1k , yeah go ahead! Thanks 🙏

m4ver1k:
Hi @caarlos0,
have made the changes, but the snapcraft test keeps failing, I'm on mac and installed snapcraft using brew.

```--- FAIL: TestRunPipeWithName (0.58s)
    snapcraft_test.go:141:
        	Error Trace:	snapcraft_test.go:141
        	Error:      	Received unexpected error:
        	            	failed to generate snap package: Snapping...
        	            	Sorry, an error occurred in Snapcraft:
[...]

PR #1845, opened 2020-10-06
feat: resolve templates in env section

m4ver1k:
This commit will support resolving templates in Env section

https://github.com/goreleaser/goreleaser/issues/1820

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -3,10 +3,12 @@
 package config
 
 import (
+	"bytes"
 	"io"
 	"io/ioutil"
 	"os"
 	"strings"
+	"text/template"
 
 	"github.com/apex/log"
 	yaml "gopkg.in/yaml.v2"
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -522,6 +524,45 @@ type Project struct {
 	GiteaURLs GiteaURLs `yaml:"gitea_urls,omitempty"`
 }
 
+func (p *Project) executeEnvTemplates() error {
```
caarlos0, line 527: we have our own template package (`tmpl`) with some defaults et al, please use it instead.

example usages can be found in the codebase: https://github.com/goreleaser/goreleaser/search?q=%22tmpl.New%28%22

caarlos0: also, the template should be evaluated in the build pipe, not here.

caarlos0:
Thanks for the PR, made a couple of comments above.

pkg/config/config.go
m4ver1k: Sure will make these changes,
thanks

caarlos0:
superseeded by #2246

thanks for the pr though! :)

Issue #1780, opened 2020-09-01
Prevent archive creation if platform isn't supported


travisgroth:
**Is your feature request related to a problem? Please describe.**

We currently have two builds (`app` and `app-cli`), and one of them(`app`) supports only a subset of platforms.  Currently the `app` archive is configured to contain both builds, and the `app-cli` archive contains just the cli build.  The problem we're seeing is that the `app` archive is generated containing only the `app-cli` binary for platforms where `app` is not built.  This leads to some user confusion as there are archives which appear to be missing a binary.

**Describe the solution you'd like**

There's two ways that jump out at me to help with this:

1. Add a goos/goarch parameter to the archives pipe so that it is only attempted for certain targets.  This would work just like the build options.
2. Have a way to tell the archives pipe to skip if binaries for all builds aren't found

There's probably additional permutations of those that would work.  Something like (1) seems more natural to me.

**Describe alternatives you've considered**
We could redo our approach to packaging, but I thought maybe this feature would be useful for others.

**Additional context**
N/A

caarlos0:
I think its probably better to fail if there are different binary counts for each target, forcing the user to specify the archives in those cases... 🤔

dlemel8:
Hey @caarlos0 , can you guide me a bit how to resolve this? 
I want to do this.
Thanks.

caarlos0:
On code itself I don't know yet... I'll need to take a deeper look...

caarlos0:
was easier to do than to explain, done on #1841

PR #1856, opened 2020-10-13
feat: allow to disable archive binary count check

### LINKED ISSUE #1855: Support option to allow archive with different binary count for a given platform
**Describe the bug**

A change in behavior was introduced recently (in #1841, proposed as a solution to #1780), that causes goreleaser to error when the count of binaries does not match for each platform within an archive.  In situations where this is expressly desired, there is no way to circumvent this check.

**To Reproduce**

Steps to reproduce the behavior can be found in the following goreleaser config:

https://github.com/criticalstack/crit/blob/a8ec49adbc86e953b6e89634a01cd9873c29d102/.goreleaser.yml

where two binaries are defined within a build and the `goarch` is different between the two binaries.  Here the second binary does not specify additional architectures, causing the count of binaries per platform to be different.

**Expected behavior**

As a user of goreleaser I want the option to disable this check when it is expressly not wanted/needed.  In situations like with this project:  https://github.com/criticalstack/crit there are two binaries in a given archive, one a Kubernetes bootstrapping tool, the other binary a [kind-like](https://kind.sigs.k8s.io/) container-in-container tool only for testing/local development.  The bootstrapping tool can run on multiple architectures like arm64, while the testing tool does not at this time.  It is nice with installer scripts like [this](https://github.com/criticalstack/crit/blob/main/scripts/install.sh) to simply install the contents of a given archive if it is supported (which also omits them when they are not).

caarlos0:
closes #1855

refs 229c8f54a51a8b25313a752016983c44e47e0214 #1841  #1780

### DIFF
--- a/internal/pipe/archive/archive.go
+++ b/internal/pipe/archive/archive.go
@@ -97,7 +97,7 @@ func (Pipe) Run(ctx *context.Context) error {
 				artifact.ByIDs(archive.Builds...),
 			),
 		).GroupByPlatform()
-		if err := checkArtifacts(artifacts); err != nil {
+		if err := checkArtifacts(artifacts); err != nil && !archive.AllowDifferentBinaryCount {
 			return fmt.Errorf("invalid archive: %d: %w", i, ErrArchiveDifferentBinaryCount)
 		}
 		for group, artifacts := range artifacts {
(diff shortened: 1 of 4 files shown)


caarlos0:
releasing v0.145.0...

PR #1859, opened 2020-10-15
feat: Add Shell Autocompletion to goreleaser for bash and zsh

danielhelfand:
If applied, this commit will add shell autocompletion to `goreleaser` for bash/zsh. Can also add some documentation if there is a suggested place.

**Why is this change being made?**

See documented issue #1858 

**Provide links to any relevant tickets, URLs or other resources**

Closes #1858

Issue #1858, opened 2020-10-14
Add Shell Autocompletion


danielhelfand:
**Is your feature request related to a problem? Please describe.**
When working with `goreleaser` outside of CI (e.g., working locally to develop a process that can be used in CI), it would be nice to have shell autocompletion to allow for a user to tab to autocomplete commands. This would be similar to what is outlined for [`kubectl`](https://kubernetes.io/docs/tasks/tools/install-kubectl/#enabling-shell-autocompletion). 

**Describe the solution you'd like**

When a user tabs after having `goreleaser`, all possible commands should be displayed:

```
goreleaser \t

build check help init release
```

When typing a command that has been started, autocomplete should complete the command:

```
goreleaser b \t -> goreleaser build
```

**Describe alternatives you've considered**

I could generate the autocompletion scripts myself and maintain outside the project, but I feel like it's a pretty common use case that should be simple to support.

**Additional context**
[Cobra docs](https://github.com/spf13/cobra/blob/142dfb15a8b9122dd6543adc769cbf59b976d793/shell_completions.md)

PR #1860, opened 2020-10-15
feat: add fish completion

caarlos0:
refs #1858 #1859

- added fish as well
- allow `__complete` as well (used for fish completions at least)


radeksimko (approved):
I'm not a fish user myself, so I can't test it, but the integration seems simple enough, so I suppose most bugs would likely be upstream (cobra) if anywhere.

Issue #1867, opened 2020-10-20
Bad documentation for ArtifactPath


lukasmalkmus:
**Describe the bug**

Bad documentation for name template `.ArtifactPath `: https://goreleaser.com/customization/templates

The documentation says `Relative path to artifact` but it is actually the absolute path.

**To Reproduce**

```yaml
archives:
    files:
      - "{{ dir .ArtifactPath }}/man/{{ .ProjectName }}*.1"
```

**Expected behavior**

Get the man pages included in the final archive.

**Environment (please complete the following information):**

- OS: macOS
- OS version: 10.15 (Catalina)
- GoReleaser Version: 0.145.0

**Additional context**

I have configured a different `dist` folder. I don't want to hardcode it everywhere else to access it. When releasing, a pre hook creates man pages in the dist folder. I attempted to dynamically get my relative dist folder using `{{ dir .ArtifactPath }}` but that returns the absolute path and doesn't include the files in the archive.

PR #1870, opened 2020-10-23
docs: .ArtifactPath is the absolute path, not the relative one.

flyhard:
Change the documentation to reflect reality. Currently the documentation says .ArtifactPath is relative, not absolute.

The issue was raised in #1867 and this should fix the documentation.


caarlos0:
closes #1867

PR #1875, opened 2020-10-27
docs: added apk package

hellt:
Hi,
a tiny improvement that adds apk package to the list of the supported packages in the docs


caarlos0:
Thanks!

PR #1876, opened 2020-10-28
fix: ensure uniqueness of brew's bin.install

aeneasr:
This patch ensures that there are no duplicate `bin.install` statements added to the homebrew formula. It resolves several upstream issues including https://github.com/ory/homebrew-kratos/issues/1 and others.

Closes https://github.com/ory/homebrew-kratos/issues/1


caarlos0:
Thanks!

PR #1889, opened 2020-11-05
feat: replace zglob with fileglob

caarlos0:
BREAKING CHANGE: may have unintended side-effects.

so we can be consistent and have a single glob dependency, as per https://github.com/goreleaser/nfpm/pull/244

---

one case that came out of this is, imagine a tree like:

```
.
- foo
  - bar.txt
- zap.txt
```

and a glob `./**/*.txt`.

On the previous glob library it would return both txt files, now it will only return `bar.txt`, which I think is correct, since the `./**/` implies that it should be in a subfolder.


djgilcrease:
LGTM, but quick question on why not use https://golang.org/pkg/path/filepath/#Glob

erikgeiser:
Previously zglob was used which allows for zsh-style globs. If users use these additional features their packages won't build with filepath/glob.

erikgeiser:
I can review this PR later today.

radeksimko (approved):
Looks like @djgilcrease already asked the question I would also asked and @erikgeiser answered it, so I don't have much to add here.

As for the breaking change - I would certainly call it out in the changelog and will defer to your decision about versioning or any other impact.

FWIW I would usually judge the severity (who would be impacted) based on how reasonable the previous behaviour/expectation was. As you mentioned the `./**/*.txt` behaviour currently just seems wrong anyway, so not a _huge_ deal in my opinion. 🤷

caarlos0:
depends on https://github.com/goreleaser/fileglob/pull/2

internal/pipe/archive/archive_test.go
```diff
@@ -157,8 +157,6 @@ func TestRunPipe(t *testing.T) {
 						t,
 						[]string{
 							fmt.Sprintf("README.%s.md", os),
-							"foo/bar",
```
caarlos0, line 160: not sure if this was wrong before or if it is wrong now...

internal/pipe/release/release_test.go
```diff
@@ -163,8 +163,6 @@ func TestRunPipeWithIDsThenFilters(t *testing.T) {
 	require.True(t, client.UploadedFile)
 	require.Contains(t, client.UploadedFileNames, "bin.deb")
 	require.Contains(t, client.UploadedFileNames, "bin.tar.gz")
-	require.Contains(t, client.UploadedFileNames, "release1.golden")
```
caarlos0, line 166: this was wrong before...

internal/pipe/archive/archive_test.go
```diff
@@ -426,7 +424,7 @@ func TestRunPipeInvalidGlob(t *testing.T) {
 			"ID":     "default",
 		},
 	})
-	require.EqualError(t, Pipe{}.Run(ctx), `failed to find files to archive: globbing failed for pattern [x-]: file does not exist`)
+	require.EqualError(t, Pipe{}.Run(ctx), `failed to find files to archive: globbing failed for pattern [x-]: unexpected end of input`)
```
caarlos0, line 427: better errors 🙏

internal/extrafiles/extra_files.go
```diff
@@ -6,31 +6,35 @@ import (
...
-				return result, fmt.Errorf("globbing failed for pattern %s: %w", extra.Glob, err)
+		if extra.Glob == "" {
+			continue
+		}
+		files, err := fileglob.Glob(extra.Glob)
+		if err != nil {
+			return result, fmt.Errorf("globbing failed for pattern %s: %w", extra.Glob, err)
+		}
+		if len(files) == 0 {
+			return result, fmt.Errorf("globbing failed for pattern %s: file does not exist", extra.Glob)
```
caarlos0, line 25: zglob handled it differently, actually erroring if file not found...

internal/pipe/archive/archive_test.go
caarlos0: as far as I tested, archives still work ok even if missing the "dirs"...

internal/extrafiles/extra_files.go
erikgeiser: The `file does not exist` seems a bit weird. The actual message should depend on the syntax tree of the glob expression:
* Glob only statically matches a single file which does not exist -> `file does not exist`
* Glob contains wildcard (or similar matchers) -> `no matching files`

We could do this in `fileglob.Glob` by checking if the static prefix is identical to the glob expression.

erikgeiser (approved):

internal/extrafiles/extra_files.go
caarlos0: hmm, yes, that makes sense, will add it there.

tiger5226:
Yeah same here, we had a low frequency deployment fail today. 

`failed to find files to archive: globbing failed for pattern licence*`

So apparently with the new fileglob it is now searching folders it previously did not. In our case it was a restricted folder so it was stopped and that failed the deployment. Seems like a workaround is to set the license to something other than default so it does not search for this pattern which is the default setting.

Issue #1804, opened 2020-09-14
Not having an ARM files makes brew formula un-tappable on ARM machines


gssbzn:
**Describe the bug**

Currently we got a report where not having an ARM url for one of our formulas makes the whole tap error

The generated formula can be seen [here](https://github.com/mongodb/homebrew-brew/blob/master/Formula/mongocli.rb)

**To Reproduce**

Users are doing `brew tap mongodb/brew` and getting:
```
Invalid formula <formula path>
Formulae require at least a URL

Cannot tap mongodb/brew
```

```yaml
before:
  hooks:
    - go mod tidy

builds:
  - <<: &build_defaults
      main: ./cmd/mongocli/mongocli.go
      ldflags:
        - -s -w -X github.com/mongodb/mongocli/internal/version.Version={{.Version}}
    id: macos
    goos: [darwin]
    goarch: [amd64]
    hooks:
      post: gon -log-level=error gon.json
  - <<: *build_defaults
    id: linux
    goos: [linux]
    goarch: [amd64]
  - <<: *build_defaults
    id: windows
    goos: [windows]
    goarch: [amd64]
archives:
- id: linux
  builds: [linux]
  <<: &archive_defaults
    files:
      - README.md
      - LICENSE
      - third_party_notices/**/*
    replacements:
      darwin: macos
      amd64: x86_64
  wrap_in_directory: true
  format: tar.gz
- id: macos
  builds: [macos]
  <<: *archive_defaults
  format: zip
  wrap_in_directory: false
- id: windows
  builds: [windows]
  <<: *archive_defaults
  wrap_in_directory: false
  format: zip
brews:
- name: mongocli
  ids: [macos,linux]
  tap:
    owner: mongodb
    name: homebrew-brew
  folder: Formula
  skip_upload: auto
  description: The MongoDB Command Line Interface (mongocli) is a tool for managing your MongoDB cloud services, like MongoDB Atlas, MongoDB Cloud Manager, and MongoDB Ops Manager.
  homepage: https://github.com/mongodb/mongocli
  install: |
    bin.install "mongocli"
    (bash_completion/"mongocli.sh").write `#{bin}/mongocli completion bash`
    (zsh_completion/"_mongocli").write `#{bin}/mongocli completion zsh`
    (fish_completion/"mongocli.fish").write `#{bin}/mongocli completion fish`
  test: |
    system "#{bin}/mongocli --version"
nfpms:
- package_name: mongocli
[...]

PR #1900, opened 2020-11-13
fix: improving URLs on linuxbrew

### LINKED ISSUE #1893: Releasing to a Linux-only Homebrew formula
**Describe the bug**

```
$ brew readall
Error: Invalid formula: /usr/local/Homebrew/Library/Taps/muesli/homebrew-tap/deckmaster.rb
```

The formula is generated with GoReleaser and does not target macOS. In such case, at least `depends_on :linux` should be added, but, regardless, it does not work due to the fact that macOS requires a `url` property, which is not defined on macOS.

**To Reproduce**

https://github.com/muesli/deckmaster/blob/master/.goreleaser.yml

**Expected behavior**

The generated formula should have `depends_on :linux` and handle gracefully when the tap, which may have formulas available on macOS, too, errors out due to unset `url` property.

- OS: mac
- OS version: macOS 11.0.1 Big Sur
- GoReleaser Version 0.146.0

caarlos0:
- rename some template fields 
- changing the ifs to avoid empty if statements
- adding `depends_on :linux` if only linux URLs

closes #1893
maybe helps with #1804

### DIFF
--- a/internal/pipe/brew/brew.go
+++ b/internal/pipe/brew/brew.go
@@ -310,21 +310,21 @@ func dataFor(ctx *context.Context, cfg config.Homebrew, cl client.Client, artifa
 			result.MacOS = down
 		} else if artifact.Goos == "linux" {
 			switch artifact.Goarch {
-			case "386", "amd64":
-				if result.Linux.DownloadURL != "" {
+			case "amd64":
+				if result.LinuxAmd64.DownloadURL != "" {
 					return result, ErrMultipleArchivesSameOS
 				}
-				result.Linux = down
+				result.LinuxAmd64 = down
 			case "arm":
-				if result.Arm.DownloadURL != "" {
+				if result.LinuxArm.DownloadURL != "" {
 					return result, ErrMultipleArchivesSameOS
 				}
-				result.Arm = down
+				result.LinuxArm = down
 			case "arm64":
-				if result.Arm64.DownloadURL != "" {
+				if result.LinuxArm64.DownloadURL != "" {
 					return result, ErrMultipleArchivesSameOS
 				}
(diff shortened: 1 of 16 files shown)


internal/pipe/brew/brew.go
```diff
@@ -310,21 +310,21 @@ func dataFor(ctx *context.Context, cfg config.Homebrew, cl client.Client, artifa
 			result.MacOS = down
 		} else if artifact.Goos == "linux" {
 			switch artifact.Goarch {
-			case "386", "amd64":
-				if result.Linux.DownloadURL != "" {
+			case "amd64":
```
caarlos0, line 313: we never include i386 builds, so this part of the switch was useless

internal/pipe/brew/testdata/test.rb.golden
```diff
@@ -4,24 +4,21 @@ class Test < Formula
   homepage "https://google.com"
   version "0.1.3"
   bottle :unneeded
-
```
radeksimko, line 7: I think that extra space was actually useful from grouping perspective.

radeksimko (approved):
I wonder how soon are we going to need to make a similar change to account for `darwin/arm64` (the Apple Silicon) 😄

caarlos0:
Once go 1.16 is out I believe 😂

mgrachev (approved):

mavogel (approved):
LGTM. Fixed the `golden` files to make the tests green and docs

caarlos0:
> LGTM. Fixed the `golden` files to make the tests green and docs

the tests were right to fail, bad sed on my side... reverted your commit...

PR #1906, opened 2020-11-17
feat: add support for git branch in templates

SuperQ:
Add a `.Branch` field to the text templates populated by the current git
branch.

Signed-off-by: Ben Kochie <superq@gmail.com>

### DIFF
--- a/internal/pipe/git/git.go
+++ b/internal/pipe/git/git.go
@@ -39,6 +39,7 @@ func (Pipe) Run(ctx *context.Context) error {
 
 // nolint: gochecknoglobals
 var fakeInfo = context.GitInfo{
+	Branch:      "none",
 	CurrentTag:  "v0.0.0",
 	Commit:      "none",
 	ShortCommit: "none",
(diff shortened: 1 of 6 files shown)


internal/pipe/git/git.go
```diff
@@ -65,6 +66,10 @@ func getInfo(ctx *context.Context) (context.GitInfo, error) {
 }
 
 func getGitInfo() (context.GitInfo, error) {
+	branch, err := getBranch()
+	if err != nil {
+		return context.GitInfo{}, errors.Wrap(err, "couldn't get current branch")
```
caarlos0, line 71:
```suggestion
		return context.GitInfo{}, fmt.Errorf("couldn't get current branch: %w", err)
```

SuperQ: Thanks, I started building this on a branch before this new pattern was added.

internal/pipe/git/git.go
```diff
@@ -123,6 +130,10 @@ func validate(ctx *context.Context) error {
 	return nil
 }
 
+func getBranch() (string, error) {
+	return git.Clean(git.Run("rev-parse", "--abbrev-ref", "HEAD", "--quiet"))
```
caarlos0, line 134: I think there are no tests checking that this gets the correct output...

SuperQ: This is the typical command I use to get the current branch. But you're right, I don't know how to mock a test for this. Some hints would be welcome. But I know it works. :smile: 

```
[git/goreleaser] $ git rev-parse --abbrev-ref HEAD --quiet
branch_template
```

caarlos0: Yeah, it works, but as-is if we remove the method, or change it to something that doesn't work, no tests will break..

here's an example: https://github.com/goreleaser/goreleaser/blob/master/internal/pipe/git/git_test.go#L30-L42

SuperQ: That's great, thanks.

caarlos0 (approved):

PR #1909, opened 2020-11-19
docs: fixing internal links that result in a broken navigation sidebar

pdxfixit:
I suspect this is something about the mkdocs template, but I'm not familiar enough, and this is easy.

Explanation:
Following any of these links, the resulting page reads just fine, but the navigation sidebar is lacking part of the expected path, resulting in links that are invalid.

If applied, this commit will fix some internal links, which appear to work, but result in a navigation sidebar with incorrect links.

This change is being made because I couldn't (fairly quickly) figure out where else an appropriate change might be made -- in the mkdocs config or in an upstream project.

### DIFF
--- a/www/docs/customization/archive.md
+++ b/www/docs/customization/archive.md
@@ -76,7 +76,7 @@ archives:
 ```
 
 !!! tip
-    Learn more about the [name template engine](/customization/templates).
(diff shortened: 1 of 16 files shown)


caarlos0:
Thanks!

Issue #1908, opened 2020-11-18
Make generated brew formulas `brew audit` compatible


lukasmalkmus:
**Is your feature request related to a problem? Please describe.**

I'm using `goreleaser` to release a Homebrew formula for my binary release. I'm using the `custom_block` config key to set a `license` field in the resulting formula as well. But when running `brew audit` on a formula generated by `goreleaser`, it yields some issues that could easily be omitted:

```
Trailing whitespace detected.
`license` (line 21) should be put before `bottle` (line 6)
```

**Describe the solution you'd like**

I propose two changes:

1. Just trim the trailing whitespaces. In theory, this should be a no brainer.
2. Instead of having the license set in the `custom_block`, make `license` a first class key in the `brews` config.

**Describe alternatives you've considered**

None

**Additional context**

None

---

I would love to contribute and make a PR if my issue sounds reasonable.

caarlos0:
I remember trying fixing those issues, and then more issues would arise from the fact its not a valid brew formula (its a tap)... and the I just kind of gave up.

But I agree, license should be first class and removing the trailing whitespace should be easy enough to do too.

lukasmalkmus:
Awesome, will figure out a PR on the weekend 👍

lukasmalkmus:
Closed by #1911.

caarlos0:
Many thanks @lukasmalkmus 🤘

PR #1911, opened 2020-11-22
feat: make generated brew formulas `brew audit` compatible (#1908)

lukasmalkmus:
This PR resolves the issues described in #1908. I want to split this into two commits:

1. Add the license field (optional). If no SPDX license identifier is set, the field won't be present in the generated Formula.
2. Get rid of the whitespaces in otherwise empty lines.

Figuring the first one out was easy but I kinda struggle with the whitespaces. Really never thought that was the hard part... but gotemplates 🤦

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -101,6 +101,7 @@ brews:
     folder: Formula
     homepage:  https://goreleaser.com
     description: Deliver Go binaries as fast and easily as possible
+    license: MIT
     test: |
       system "#{bin}/goreleaser -v"
     dependencies:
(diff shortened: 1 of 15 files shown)


caarlos0 (commented):
LGTM so far :)

caarlos0:
> Figuring the first one out was easy but I kinda struggle with the whitespaces. Really never thought that was the hard part... but gotemplates 🤦

yeah that's always the worst, maybe we can remove them after compiling the template though 🤔

lukasmalkmus:
> yeah that's always the worst, maybe we can remove them after compiling the template though

Yes, I added that. Might be the quickest way to resolve this instead of dealing with templating (even if that would be the proper way I guess).

lukasmalkmus:
Ready for review.

caarlos0 (approved):

caarlos0:
Thanks!

PR #1936, opened 2020-12-07
feat: add Os and Arch template options to the Binary name field

sylviamoss:
If applied, this PR will add the capability to use `{{ .Os }}` and `{{.Arch}}` in the binary name. 

I'm implementing this change because I was expecting these to work at this point of the build process. I'm using this specific configuration for the binary name, to make it easier for [Packer](https://github.com/hashicorp/packer) when managing plugin's binaries for different os/arch without needing to add the information to the binary name itself afterward.

Here a failing release example for a repository I'm using for testing:
https://github.com/sylviamoss/packer-provisioner-comment/runs/1511229354?check_suite_focus=true#step:6:58

I tested the changes locally and here's the output line with the binary name: 
```
   • building binaries
      • building                  binary=/Users/sylviamoss/Packer/plugins-example/packer-provisioner-comment/dist/packer-provisioner-comment_freebsd_amd64/packer-provisioner-comment_v0.2.2-SNAPSHOT-a1d03f2_freebsd_amd64
```


caarlos0:
I think you would want to do that by using `archive.format: binary` and setting the templates there... does that make sense?

sylviamoss:
I'm creating zip archives. The way our tool works is by keeping the binaries in a folder to use them at the runtime. One can have binaries for different os/arch for the same plugin and that's why we differentiate them by their names. We are parsing the binary name and adding these information to it, it would be nice if they were created already with them.

caarlos0 (commented):
LGTM overall, can you update the docs too? (`www/docs/customization/build.md`)

sylviamoss:
@caarlos0 I updated the documentation but rebased with master and messed everything up. 🤦🏼

radeksimko (approved):
LGTM.

I don't have access to Vercel to verify how the markdown is rendered, but it looks good.

www/docs/customization/build.md
```diff
@@ -159,6 +159,17 @@ builds:
       - windows
 ```
 
+The binary name field supports [templating](/customization/templates/). The following build details are exposed:
+
+| Key     | Description                      |
+|---------|----------------------------------|
+| .Os     | `GOOS`                           |
+| .Arch   | `GOARCH`                         |
+| .Target | `GOARM`                          |
```
caarlos0, line 168: this doesn't seem correct...

sylviamoss: that's true, I did some confusion here 🤦🏼 
fixing it now

caarlos0 (approved):

caarlos0:
LGTM, thanks!

PR #1944, opened 2020-12-11
fix: nfpm signing error with `--skip-sign` when gpg key is not found

zbindenren:
Currently the following command will fail:

```
goreleaser release --snapshot --rm-dist --skip-sign
⨯ release failed after 4.74s error=nfpm failed: signing error: armored detach sign: reading PGP key file: open /tmp/key.gpg: no such file or directory
```

If I do not want to sign the packages with the `--skip-sign` flag (just want to build the rpm,debs etc...)  the command should not fail when there is no signing key available.

I am not sure if the fix is in the correct place :man_shrugging: Maybe there is a better way to do it.

### DIFF
--- a/internal/pipe/nfpm/nfpm.go
+++ b/internal/pipe/nfpm/nfpm.go
@@ -218,6 +218,12 @@ func create(ctx *context.Context, fpm config.NFPM, format, arch string, binaries
 		},
 	}
 
+	if ctx.SkipSign {
+		info.APK.Signature = nfpm.APKSignature{}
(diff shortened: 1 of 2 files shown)


caarlos0:
hey, thanks for the PR! Can you add a test for it as well?

PS: the build issue is unrelated to your change.

zbindenren:
@caarlos0 added a test. let me know if it is ok.

caarlos0:
Looks great, thanks!

PR #1947, opened 2020-12-12
feat(ci): test on windows

caarlos0:
refs https://github.com/goreleaser/nfpm/issues/260

this is probably going to be big...


internal/testlib/string.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// EqualIgnoreCRLF requires equality ignoring line feed differences.
+func EqualIgnoreCRLF(t *testing.T, expected, actual string) {
```
djgilcrease, line 11: Instead of this I would add a `.gitattributes` like https://github.com/goreleaser/nfpm/blob/master/.gitattributes

caarlos0:
One problem I found is: https://github.com/golang/go/issues/41809 

So, when archiving on windows, the archives might be wrong, since go won't set the executable bit of the files...

I have no idea how to fix this yet 🤔

caarlos0:
who am I kidding

Issue #1952, opened 2020-12-18
Darwin ARM64 Support


coolaj86:
I noticed that builds for Apple Silicon / M1 / ARM64 are not being built with goreleaser, even though I have go1.16 installed, which supports them.

```bash
go version
```

```
go1.16beta1 darwin/amd64
```

```bash
GOOS=darwin GOARCH=arm64 go build .
```

This is relevant to me in particular because I work on [webinstall.dev](https://webinstall.dev), which already supports M1 / Apple Silicon builds as soon as they are published, but many of the supported Go projects are using Goreleaser.

**Describe the solution you'd like**

I imagine that there's a simple line in a file somewhere to turn on darwin arm64 builds for Go versions that support it.

**Describe alternatives you've considered**

I can build by hand for the time being, or maybe fork goreleaser and hunt for the darwin options.

**Additional context**

N/A

caarlos0:
Currently goreleaser allows only stable goos/goarches, but I guess it would be nice to have a beta feature for this... 

also need to handle brew arm64, for example.

coolaj86:
Why not introspect what Go can build and use that as the list rather than hard-coding it?

coolaj86:
Using `go tool dist list`:

```txt
aix/ppc64
android/386
android/amd64
android/arm
android/arm64
darwin/amd64
darwin/arm64
dragonfly/amd64
freebsd/386
freebsd/amd64
freebsd/arm
freebsd/arm64
illumos/amd64
ios/amd64
ios/arm64
js/wasm
linux/386
linux/amd64
linux/arm
linux/arm64
linux/mips
linux/mips64
linux/mips64le
linux/mipsle
linux/ppc64
linux/ppc64le
linux/riscv64
linux/s390x
netbsd/386
netbsd/amd64
netbsd/arm
netbsd/arm64
openbsd/386
openbsd/amd64
openbsd/arm
openbsd/arm64
openbsd/mips64
plan9/386
plan9/amd64
plan9/arm
solaris/amd64
windows/386
windows/amd64
windows/arm
```

stack72:
I think limiting to Go 1.16.x is already causing a lot of us pain in supporting our users who are on the new apple macbooks...

caarlos0:
Sorry @stack72 , didn't get what you meant... can you explain?

stack72:
My apologies - it sent before my second sentence

[...]

PR #1956, opened 2020-12-19
feat: support go 1.16 and apple silicon

caarlos0:
closes https://github.com/goreleaser/goreleaser/issues/1952

will only be merged once go 1.16 is out.


radeksimko (approved):
LGTM, I tried using it to generate a compatible `darwin/arm64` build of `terraform-ls`:
https://github.com/hashicorp/terraform-ls/issues/349

Admittedly I don't have the new MacBook to test that it actually works, but compilation was successful.

The only question I would have is: What happens when `goreleaser release` runs under Go 1.15 or lower? I'm assuming this would gracefully fail, right?

caarlos0:
> The only question I would have is: What happens when `goreleaser release` runs under Go 1.15 or lower? I'm assuming this would gracefully fail, right?

yes, go build will fail on the build step...

bouk:
It would be good at add arm64 as a default here: https://github.com/goreleaser/goreleaser/blob/264dedc12a7ecab5a7ce5592e9e4aac1ca0494f6/internal/builders/golang/build.go#L52

caarlos0:
> It would be good to add arm64 as a default here

done in #1999

.github/workflows/build.yml
```diff
@@ -34,7 +34,8 @@ jobs:
         name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.15
+          stable: false
+          go-version: '1.16.0-beta1'
```
radeksimko, line 38:
```suggestion
          go-version: '1.16'
```

radeksimko (approved):
1.16 was just released:
https://golang.org/doc/go1.16

I think the only last missing piece of the puzzle is 1.16 CircleCI Docker image (which AFAIK is used for generating docs):
https://hub.docker.com/r/circleci/golang/tags?page=1&ordering=last_updated&name=1.16

crazy-max (approved):
LGTM minus upgrading GHA workflow to use Go 1.16.

PR #1958, opened 2020-12-21
feat: allow automatic snapshot on dirty state

### LINKED ISSUE #1324: Option to skip arbitrary stages
I have a use case where I sometimes want to run `goreleaser` locally for dev purposes. For that case, I need to skip as many cases as I can to increase local iteration speed.

Right now we have

        --skip-publish            Skips publishing artifacts
        --skip-sign               Skips signing the artifacts
        --skip-validate           Skips several sanity checks

but stuff such as `--skip-archive`, `--skip-checksum` are missing.

Maybe goreleaser should have those? Or maybe a generic flag like:

     --skip=PHASE[,PHASE[,PHASE...]]
     --skip=archive,checksum,sign,release,before

Maybe goreleaser isn't created for local iteration, but given the amount of configuration I'm supplying (-ldflags and build options etc) I'd love to use goreleaser locally as well to build binaries to run.

**Additional context**
N/A

crazy-max:
Introduce a new flag for build/release commands called `--snapshot-auto`. This new flag will detect if Git is in dirty state and create a snapshot if so. Footprint in the project is pretty minimal to avoid unattended regressions for the current implementation of snapshot.

The purpose of this new flag is for CI and automation to avoid specific conditions like:

https://github.com/goreleaser/goreleaser/blob/9423da1b607b26916664cb26016811b54359e46c/.github/workflows/build.yml#L85-L89

I am open to any suggestion or alternative proposal.

Could also partially fix #1324 (see the comment https://github.com/goreleaser/goreleaser/issues/1324#issuecomment-634189064)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -78,15 +78,11 @@ jobs:
           snapcraft login --with <(echo "$SNAPCRAFT_LOGIN")
       -
         name: GoReleaser
-        if: success()
+        if: success() && (startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/master')
         env:
           GITHUB_TOKEN: ${{ secrets.GH_PAT }}
         run: |
-          if [[ $GITHUB_REF == refs/tags/v* ]]; then
-            ./goreleaser
-          elif [[ $GITHUB_REF == refs/heads/master ]]; then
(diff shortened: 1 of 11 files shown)


caarlos0:
Awesome, thanks @crazy-max 🚀

PR #1961, opened 2020-12-22
feat: allow to template build targets

### LINKED ISSUE #1852: Add support for only building a subset of defined operating systems (GOOS) and architectures (GOARCH)

I would like to be able to specify a subset of defined operating systems and/or architectures when doing a build, e.g. via `--os`/`--goos` or `--arch`/`--goarch` flags.

A possible alternative would be to allow specifying `goos` and/or` goarch` via environment variables, e.g. something like:

```yaml
builds:
  - ... trimmed ...
    goos:
      - "{{.Env.GOOS}}"
    goarch:
      - "{{.Env.GOARCH}}"
```

**Additional context**

The same thing was asked in https://github.com/goreleaser/goreleaser/discussions/1810.

A related idea also hinted at implementing `--goos` and `--goarch` flags: https://github.com/goreleaser/goreleaser/issues/1324#issuecomment-634257437.

crazy-max:
Fixes #1852

### DIFF
--- a/internal/builders/golang/build.go
+++ b/internal/builders/golang/build.go
@@ -79,7 +79,11 @@ func (*Builder) Build(ctx *context.Context, build config.Build, options api.Opti
 	var cmd = []string{build.GoBinary, "build"}
 
 	var env = append(ctx.Env.Strings(), build.Env...)
-	env = append(env, target.Env()...)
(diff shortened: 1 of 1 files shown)


caarlos0:
discussed on slack, will go with a different approach.

PR #1970, opened 2020-12-28
docs: add changie to changelog generators

miniscruff:
Adds a link to changie an alternative changelog generator.
Changie uses goreleaser for releases so it is, at least a little, tested.

Not sure if anything else was needed for doc changes.


caarlos0:
Thanks!

PR #1979, opened 2020-12-29
feat: simplify docker images

caarlos0:
- we don't need the `latest-$arch` imgs as it handled by the manifest
- the v0.X imgs are never used anyway, removing

We'll now only have `latest` and `v0.x.y` images.


crazy-max:
@caarlos0 If you want some hints about image versioning, you can take a look at https://github.com/crazy-max/ghaction-docker-meta#semver

PR #1987, opened 2021-01-04
Add `--goos` and `--goarch` flags.

tylerhou:
This PR refactors flag code and adds the `--goos` and `--goarch` flags. The flags let the user specify a set of OSes and architectures to build from the command line. See https://github.com/goreleaser/goreleaser/issues/1852.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -17,15 +17,30 @@ type buildCmd struct {
 	opts buildOpts
 }
 
+type sharedBuildOpts struct {
+	config       string
+	snapshot     bool
+	skipValidate bool
+	rmDist       bool
+	deprecated   bool
+	parallelism  int
+	timeout      time.Duration
+}
+
 type buildOpts struct {
-	config        string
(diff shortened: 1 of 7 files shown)


edigaryev:
This has been open for a while and would be pretty nice to have this merged. Any chance anyone could have a look at it?

caarlos0:
I'm not sure how we should implement this... but in any case, I would keep things simpler by removing the negated stuff, as of which, if provided, only the given targets should be built.

But I'm not sure about the imple yet... this "leaks" too much from the cmd into the api... I'll need to give it some more thought.

In the meantime, thanks for the PR and sorry for the delay :)

tylerhou:
No problem; I know open source is hard so don't feel pressured to merge. Let me know if you'd like me to make changes as well, or come up with something that doesn't leak as much.

tylerhou:
One way to not leak into context.go might be to directly consume the flags in shouldBuildTarget.

PR #2000, opened 2021-01-07
refactor: use t instead of tt

caarlos0:
very error-prone, as we can see in some tests, we were using `t` instead of `tt` later on, making the `go test` output be wrong, basically.


mavogel (commented):
LGTM and you were faster :)

PR #2002, opened 2021-01-07
feat(http): support custom headers in http upload

sirlatrom:
This commit will add an additional field to elements under `uploads:`, allowing the user to specify custom headers for the `PUT` (or other mode) request.

GitLab has a Generic Packages Repository which allows uploads using various auth methods, one being the CI job token. However, it must be specified not as basic authentication but rather as a custom header, namely `JOB-TOKEN: $CI_JOB_TOKEN`. This change supports specifying that header as well as any other headers required by various upload destinations.

See also: https://docs.gitlab.com/ee/user/packages/generic_packages/index.html#publish-a-generic-package-by-using-cicd

### DIFF
--- a/internal/http/http.go
+++ b/internal/http/http.go
@@ -240,6 +240,11 @@ func uploadAsset(ctx *context.Context, upload *config.Upload, artifact *artifact
 	log.Debugf("generated target url: %s", targetURL)
 
 	var headers = map[string]string{}
+	if upload.CustomHeaders != nil {
+		for name, value := range upload.CustomHeaders {
+			headers[name] = value
+		}
+	}
 	if upload.ChecksumHeader != "" {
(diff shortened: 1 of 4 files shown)


caarlos0 (commented):
LGTM, can you add a test case as well?

sirlatrom:
> 
> 
> LGTM, can you add a test case as well?

Yes, I've added it now, along with an example in the docs: https://github.com/goreleaser/goreleaser/pull/2002/files#diff-8be32e869556dd405c589f4df19e68aef2c4cab8ab8a1867224460cc4bd6cc2eR472-R486

sirlatrom:
I failed to check the new test locally, turns out I can't write valid Go code 😅 Fixed now.

caarlos0 (approved):
nice, thanks!

PR #2003, opened 2021-01-07
feat: allow to use nfpm packages in the docker pipe

### LINKED ISSUE #1491: Allow Docker builds to have access to NFPM outputs

I'm working on a project that uses GoReleaser to produce both Debian packages and Docker images. I would like to be able to construct the Docker images by running `apt-get install <Debian package that GoReleaser just built>`. (Of course, other commands are usually necessary, but being able to use the package for the main installation would be helpful.) That avoids the need to write out the extra files that the binary needs three times (in the NFPM config, the Docker config, and the Dockerfile) and dependencies on other packages twice. It also ensures that there is minimal difference in environment between a normal install and a Docker image.

I would like the package built in the NFPM pipe to be made available during the Docker build. The configuration might look something like this, analogous to the `builds` key for defining which raw binaries are available:

```yaml
nfpms:
  - id: <nfpm-id>
    ...

dockers:
  - nfpms:
      - <nfpm-id>
```

Then it would be possible to do
```dockerfile
COPY <nfpm-id>.deb /tmp
RUN apt-get install /tmp/<nfpm-id>.deb
```

or something similar.

Maybe the format (deb/RPM) could be specified somehow, or maybe this would just make all the formats for that build available.

Perhaps this could be part of something more general for sharing files between pipes, but I don't know GoReleaser well enough to say what that would look like.

One can list out all the dependencies and extra files repeatedly, but that duplication is annoying.

One could imagine including the package file in `extra_files`, but in fact that field disallows both templates (the name of the file will generally include the version, at least, so templates would be necessary) and [files inside the dist directory](https://github.com/goreleaser/goreleaser/pull/1175). I also don't know if there's any guarantee that the pipes are run in the necessary order.

caarlos0:
The idea is to simplify things a bit.

Since we already filter by ID on the docker pipe, there is no need to filter by binary name as well.

So, we can just drop that filter, which allows us to easily add NFPM packages to the filter.

So, instead of having only the binaries, we'll have all binaries and nfpm packages that match goos/goarch/goarm and the IDs filter (if provided).... which allows us to install the packages directly instead of doing the same thing via binaries + extra_files.

closes  #1503
closes #1491

---

TODO:

- [x] tests
- [x] review docs


djgilcrease (approved):
LGTM

PR #2007, opened 2021-01-08
feat: install completions on linux packages

caarlos0:
install fish, bash and zsh completions on the linux packages :)


djgilcrease (approved):
LGTM

Issue #2029, opened 2021-01-20
map has no entry for key "Os"


azr:
**Describe the bug**

The Os ( and maybe Arch ? ) fields can't be accessed 

related: #2021, https://github.com/goreleaser/goreleaser/pull/1936

**To Reproduce**

Tag a release

```yaml
# This is an example goreleaser.yaml file with some sane defaults.
# Make sure to check the documentation at http://goreleaser.com
env:
  - CGO_ENABLED=0
before:
  hooks:
    - go test ./...

    # You may remove this if you generate these on your own
    - go install github.com/hashicorp/packer/cmd/mapstructure-to-hcl2
    - go generate ./...

    # Install packer-plugins-check to check binary compatibility with Packer
    - go install github.com/hashicorp/packer/cmd/packer-plugins-check

builds:
  # A separated build to run the packer-plugins-check only once for a linux_amd64 binary
  -
    id: plugin-check
    mod_timestamp: '{{ .CommitTimestamp }}'
    hooks:
      post:
        # This will check plugin compatibility by making sure packer can load the binary
        - cmd: packer-plugins-check -load={{ .Name }}
          dir: "{{ dir .Path}}"
    flags:
      - -trimpath #removes all file system paths from the compiled executable
    ldflags:
      - '-s -w -X main.Version={{.Version}} -X main.VersionPrerelease= '
    goos:
      - linux
    goarch:
      - amd64
    binary: '{{ .ProjectName }}_v{{ .Version }}_{{ .Os }}_{{ .Arch }}_x5'
  - 
    mod_timestamp: '{{ .CommitTimestamp }}'
    flags:
      - -trimpath #removes all file system paths from the compiled executable
    ldflags:
      - '-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}'  ## can be useful to set variables 
    goos:
      - freebsd
      - windows
      - linux
      - darwin
    goarch:
      - amd64
      - '386'
      - arm
      - arm64
    ignore:
      - goos: darwin
        goarch: '386'
      - goos: linux
        goarch: amd64
    binary: '{{ .ProjectName }}_v{{ .Version }}_{{ .Os }}_{{ .Arch }}_x5'
archives:
- format: zip
  name_template: '{{ .ProjectName }}_v{{ .Version }}_{{ .Os }}_{{ .Arch }}_x5'
checksum:
[...]

Issue #2021, opened 2021-01-15
Documentation indicates that {{.Os}} and {{.Arch}} can be used in binary field


jaqx0r:
In https://goreleaser.com/customization/build/ the documentation indicates that the `binary` field of the `build` list can be templated:

```
The binary name field supports templating. The following build details are exposed:

Key	Description
.Os	GOOS
.Arch	GOARCH
.Arm	GOARM
.Ext	Extension, e.g. .exe
.Target	Build target, e.g. darwin_amd64
```

However in #1151 the issue was raised that these do not work, and the bug was closed as wontfix, suggesting that it is not possible for `.Os` to be templated in this context.

The documentation never gives an example of how to use the `.Os` or `.Arch` template variables in any case, so one fumbles around with attempts like:

```
builds:
  - id: mtail
    main: ./cmd/mtail/main.go
    binary: mtail_{{.Version}}_{{.Os}}_{{.Arch}}
```
in a `.goreleaser.yml` end up with goreleaser reporting errors like
```
  ⨯ release failed after 2.74s error=template: tmpl:1:21: executing "tmpl" at <.Os>: map has no entry for key "Os"
```

Please update the documentation to either:
 * demonstrate correct usage of these template fields with a working example, OR
 * remove the section suggesting that it is possible to template the binary name.

Thanks in advance!

caarlos0:
that feature is only on master... our docs are not versioned, unfortunately..

jaqx0r:
Sorry I don't understand, do you mean templates in 'binary' will be supported after a release is cut?

caarlos0:
releasing v0.155.0, should fix this.

jaqx0r:
Thanks!

azr:
Hello there, when will this version be available in the github-action ? Is there a way for me to use it already ?

caarlos0:
> Hello there, when will this version be available in the github-action ? Is there a way for me to use it already ?

the action uses latest by default, so its available already.

azr:
Hey @caarlos0, thanks for answering, that's weird because I still have the same `executing "tmpl" at <.Os>: map has no entry for key "Os"` issue, ( yes I'm using latest ).

caarlos0:
> Do you have any tip ?

Will need more details... can you open a new issue and follow the instructions in the template? Thx

PR #2030, opened 2021-01-21
fix: brew default install when using templates in the binary name

caarlos0:
This was a long time TODO that should probably be removed anyway.

In the specific case of using templates in the `builds.binary` field, even if we evaluate it at defaults time, the formula wouldn't work when using multiple platforms, so it would have to be customized anyway.

This simplifies things by removing one complicated, probably wrong default (hence the warning) in favor of a simple probably wrong default (which will still have the warning).

All things considered I think this should be ok.

closes https://github.com/goreleaser/goreleaser/issues/2029
see #2021
see #1936

### DIFF
--- a/internal/pipe/brew/brew.go
+++ b/internal/pipe/brew/brew.go
@@ -83,28 +83,8 @@ func (Pipe) Default(ctx *context.Context) error {
 		var brew = &ctx.Config.Brews[i]
 
 		if brew.Install == "" {
-			// TODO: maybe replace this with a simpler also optimistic
-			// approach of just doing `bin.install "project_name"`?
-			var installs []string
-			for _, build := range ctx.Config.Builds {
-				if !isBrewBuild(build) {
-					continue
-				}
-				install := fmt.Sprintf(`bin.install "%s"`, build.Binary)
-				// Do not add duplicate "bin.install" statements when binary names overlap.
-				var found bool
-				for _, instruction := range installs {
-					if instruction == install {
-						found = true
-						break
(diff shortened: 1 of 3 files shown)


radeksimko (approved):
Am I right that this reduces the number of binaries which would otherwise appear in the Homebrew formula to 1? Not that I think >1 binaries for macOS is common - just checking.

I think distributing non-cgo and cgo binary of the same program as shown in the (now removed) test is unlikely to be practical, and other cases should probably be just covered by multiple `brew` statements. If each binary does a different thing, then it probably should be installed separately anyway?

Either way if someone comes up with a valid use case, then I reckon we can assess and maybe turn `brew.Install` into `brew.Installs` or something like that.

caarlos0:
> Am I right that this reduces the number of binaries which would otherwise appear in the Homebrew formula to 1? Not that I think >1 binaries for macOS is common - just checking.

yes.

> I think distributing non-cgo and cgo binary of the same program as shown in the (now removed) test is unlikely to be practical, and other cases should probably be just covered by multiple `brew` statements. If each binary does a different thing, then it probably should be installed separately anyway?

The cgo stuff on the test wasn't doing anything, was just to assert that both builds were added to the formula.

---

a different approach to this would be build the install list in the `Run` part instead of `Default`, but I think this should be OK as the warning was there since forever 🤔

caarlos0:
Also notice it was only adding builds that have `darwin amd64` as target, even if a linuxbrew only formula should work too 🤣

Issue #2066, opened 2021-02-18
bug: failed to build for darwin_arm64


mr-karan:
**Describe the bug**

While using `goreleaser` (config file below), I started to get this error since yesterday:

```
   • building binaries
      • building                  binary=/home/runner/work/doggo/doggo/dist/doggo_linux_arm64/doggo
      • building                  binary=/home/runner/work/doggo/doggo/dist/doggo_darwin_amd64/doggo
      • building                  binary=/home/runner/work/doggo/doggo/dist/doggo_windows_amd64/doggo.exe
      • building                  binary=/home/runner/work/doggo/doggo/dist/doggo_linux_amd64/doggo
      • building                  binary=/home/runner/work/doggo/doggo/dist/doggo_darwin_arm64/doggo
   ⨯ release failed after 50.68s error=failed to build for darwin_arm64: # github.com/mr-karan/doggo/cmd/doggo
/opt/hostedtoolcache/go/1.15.8/x64/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/tmp/go-link-445048011/go.o: file not recognized: File format not recognized
collect2: error: ld returned 1 exit status


Error: The process '/opt/hostedtoolcache/goreleaser-action/0.156.2/x64/goreleaser' failed with exit code 1
```
**To Reproduce**

```yaml
env:
  - GO111MODULE=on
  - CGO_ENABLED=0

builds:
  - binary: doggo
    goos:
      - windows
      - darwin
      - linux
    goarch:
      - amd64
      - arm64
    goarm:
      - 6
      - 7
    ldflags:
      - -s -w -X "main.buildVersion={{ .Tag }} ({{ .ShortCommit }} {{ .Date }})"
    dir: ./cmd/doggo/
```

**Expected behavior**

Last I tried this was on `goreleaser 0.155.2` which worked without issues (it automatically skipped the invalid platform I think):

```
   • building binaries
      • building                  binary=/home/runner/work/doggo/doggo/dist/doggo_linux_arm64/doggo
      • building                  binary=/home/runner/work/doggo/doggo/dist/doggo_darwin_amd64/doggo
      • building                  binary=/home/runner/work/doggo/doggo/dist/doggo_linux_amd64/doggo
      • building                  binary=/home/runner/work/doggo/doggo/dist/doggo_windows_amd64/doggo.exe
```

**Environment (please complete the following information):**

- OS: linux
- OS version: ubuntu:20.04
[...]

Issue #1653, opened 2020-07-07
Builds have default GOOS & GOARCH even for explicit targets


radeksimko:
**Describe the bug**

When building for explicit targets, the internal `config.Build` receives default GoOS and GoArch
https://github.com/goreleaser/goreleaser/blob/752d4e5435468a9c4b9e911840101d2af2f5acaa/internal/builders/golang/build.go#L44-L49

which then trips up this logic in Homebrew pipe:
https://github.com/goreleaser/goreleaser/blob/752d4e5435468a9c4b9e911840101d2af2f5acaa/internal/pipe/brew/brew.go#L116-L123

as it falsely identifies builds as Homebrew when they're not.

The combination (multiple builds + use of explicit targets + homebrew) is rare enough that it doesn't affect as many people I suppose, but it does result in duplicate `install` stanzas in the published formula:

```ruby
  def install
    bin.install "binary-name"
    bin.install "binary-name"
  end
```

**To Reproduce**

```yaml
builds:
  -
    targets:
      - darwin_amd64
      - windows_386
      - windows_amd64
  -
    targets:
      - linux_386
      - linux_amd64
brews:
  -
    tap:
      owner: myself
      name: homebrew-tap
```

**Expected behavior**

 - GoOS & GoArch remains empty if targets are explicitly defined
 - Homebrew doesn't publish duplicate `install` stanza

**Environment (please complete the following information):**

- OS: macOS
- OS version: 10.15.5
- GoReleaser Version: v0.139.0

**Workaround:**

This can be probably worked around by specifying the `install` config option which overrides the detection:

```yaml
install: |
      bin.install "program"
```

caarlos0:
Q: why do you use the `targets` config instead of `goos` and `goarch`?  Asking because, although it is allowed, it was not documented... The instance I remember people using it is as a hacky way to get goreleaser to release rust binaries for example 🤔

radeksimko:
Good question.

There is a list of os/arch combinations which we intend to support/build, for various reasons and we only want to support these and no others. Adding new combination is always a very deliberate decision and we don't want that to happen by accident.

1. I value explicitness and readability above most things
2. writing down all possible combinations and then listing the ones we don't intend to support in ignore list seems quite error-prone, also because the list of supported GOOS and GOARCH is a moving target

caarlos0:
That makes sense. I think there are probably more bug hidden in this (the lack of docs being one issue), but I agree we should tackle then.

PR #2069, opened 2021-02-18
fix: try not to break if run on go1.15

caarlos0:
ignore darwin/arm64 if its go 1.15...

I don't really like this... but maybe its better leaving this here for a month or so than dealing with support...

refs #2066


radeksimko (approved):
LGTM.

btw. I think this highlights the potential value of more explicit build targets as we discussed in https://github.com/goreleaser/goreleaser/issues/1653#issuecomment-654845644

Issue #2095, opened 2021-03-08
nfpm `package_name` isn't used in deb file name


eriksw:
**Describe the bug**

Setting the `package_name` option doesn't affect the file name of the produced`.deb` file.

The package has the correct `Package: ...` in its control tarball, but it's a nontrivial violation of end-user expectations for a `.deb` to have a filename that has no relation to the actual package name.

**To Reproduce**

(For context, I encountered this while trying to prepare a PR to make [apt-gcs](https://github.com/dhaivat/apt-gcs) have built packages.)

```yaml
before:
  hooks:
    - go mod tidy
builds:
  - id: gs
    binary: gs
    env:
      - CGO_ENABLED=0
    flags:
      - -trimpath
    goos:
      - linux
    ldflags:
      - -s -w
    main: ./cmd/main.go
nfpms:
  - package_name: apt-transport-gs
    builds:
      - gs
    maintainer: https://github.com/dhaivat/apt-gcs
    homepage: https://github.com/dhaivat/apt-gcs
    license: MIT
    formats:
      - deb
    bindir: /usr/lib/apt/methods
archives:
  - format: binary
checksum:
  disable: true
changelog:
  sort: asc
  filters:
    exclude:
      - '^docs:'
      - '^test:'
```

**Expected behavior**

Produces deb packages named like `apt-transport-gcs_[...].deb`

**Actual behavior**

```
➜  dist git:(master) ✗ ls *.deb | sort
apt-gcs_v0.0.0-SNAPSHOT-6e5a477_linux_386.deb
apt-gcs_v0.0.0-SNAPSHOT-6e5a477_linux_amd64.deb
apt-gcs_v0.0.0-SNAPSHOT-6e5a477_linux_arm64.deb
```

Note that the `Package:` is correct. Again, the issue is that the package does not match the filename:
```
➜  dist git:(master) ✗ ar x apt-gcs_v0.0.0-SNAPSHOT-6e5a477_linux_amd64.deb
➜  dist git:(master) ✗ tar xzvf control.tar.gz
x ./md5sums
x ./conffiles
x ./control
➜  dist git:(master) ✗ cat control
Package: apt-transport-gs
Version: 0.0.0~SNAPSHOT-6e5a477
Section:
Priority:
Architecture: amd64
Maintainer: https://github.com/dhaivat/apt-gcs
Installed-Size: 9144
Homepage: https://github.com/dhaivat/apt-gcs
Description: no description given
```

[...]

PR #2097, opened 2021-03-08
feat: package name on nfpm file name template

caarlos0:
Although we are changing a default here, I think almost all cases fall under:

1. a single package with the same name as the project name
2. multiple packages with different names, in which case the user is manually setting `file_name_template`

For 2, this changes nothing.

For 1, the default `PackageName` is `ProjectName`, so it also changes nothing.

In any case, this new default is more sensible and is what I as a user would expect.

Thoughts?

closes https://github.com/goreleaser/goreleaser/issues/2095

### DIFF
--- a/internal/pipe/nfpm/nfpm.go
+++ b/internal/pipe/nfpm/nfpm.go
@@ -26,7 +26,7 @@ import (
 	"github.com/goreleaser/goreleaser/pkg/context"
 )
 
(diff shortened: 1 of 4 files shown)


caarlos0:
(big diff due to formatting changes by gofumpt, sorry about that)

radeksimko (approved):
Agreed with the risk assessment. 👍🏻

PR #2098, opened 2021-03-09
Fix nfpm windows build

GGP1:


GGP1:
This PR should be ommited as it contains errors. I will open a new one once I fix them.

internal/pipe/nfpm/nfpm.go
```diff
@@ -285,6 +285,11 @@ func create(ctx *context.Context, fpm config.NFPM, format, arch string, binaries
 		info.Deb.Signature = nfpm.DebSignature{}
 	}
 
+	for _, content := range info.Contents {
+		content.Source = filepath.ToSlash(content.Source)
```
caarlos0, line 289: probably could be on line 199

GGP1: Yes, that's exactly what I changed

PR #2110, opened 2021-03-15
docs: Add info on skip_upload to scoop 

cugu:
Scoop supports `skip_upload`, just like Brew, but it was missing in the docs.


caarlos0 (approved):
Thanks!

PR #2122, opened 2021-03-18
docs: expand GitLab CI examples

tedtramonte:
Docker-in-Docker is overkill for simple binary and archive release purposes and the more verbose configuration for it is intimidating for it to be the only example of using GoReleaser in GitLab CI.

This commit adds an example `.gitlab-ci.yml` and some additional explanation on the general GitLab CI process, as I stumbled onto some quirks that were hard to find documentation on while implementing GoReleaser into my own CI pipeline.

### DIFF
--- a/www/docs/ci/gitlab.md
+++ b/www/docs/ci/gitlab.md
@@ -1,7 +1,52 @@
 # GitLab CI
 
-To create GitLab releases and push images to a Docker registry, add a file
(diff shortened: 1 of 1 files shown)


caarlos0 (commented):
Seems simpler indeed!

Mind opening a PR to https://gitlab.com/goreleaser/example as well?

PR #2124, opened 2021-03-19
fix: default gomips

caarlos0:
default to hardfloat, as in `go build`.

an empty gomips was causing goreleaser to fail validation as well.

closes https://github.com/goreleaser/goreleaser/discussions/2105


radeksimko (commented):
LGTM

Issue #1688, opened 2020-07-20
Provide a name template for the current build module/package


dionysius:
**Is your feature request related to a problem? Please describe.**
I have version informations in an separate package within the same repo, since I provide functions to use, handle and display version related informations differently. That means I can't use `-X main.xyz=...` and am forced to give the full go package name to that package in ldflag's `-X` argument. There seems to be no module related name in https://goreleaser.com/customization/templates/

**Describe the solution you'd like**
- I'd like to have a name for the current _root_ package, which is a module in my case, which would be `<"current-dir">/<builds.dir>`. Suggestion `.Module`
- And/Or a name for the current binary package, which would be `<"current-dir">/<builds[].dir>/<builds[].main>`. Suggestion `.Package` (read: package of the binary)

- If I had a `.Module` I could do `-X "{{.Module}}/cli/version.Version={{.Version}}"`
- If I had a `.Package` I could do `-X "{{replace .Package "/cmd/lxe" ""}}/cli/version.Version={{.Version}}"`

Reference:
- `go list -m` prints the module
- `go list -f {{ .ImportPath }} <dir>` prints the package

```bash
$ go list -m
github.com/automaticserver/lxe
$ go list -f '{{ .ImportPath }}' ./cmd/lxe/
github.com/automaticserver/lxe/cmd/lxe
$ go list -f '{{ .ImportPath }}' ./cli/version/
github.com/automaticserver/lxe/cli/version
$ cd cmd/lxe/; go list -m # module is printed in every subpackage
github.com/automaticserver/lxe
```

**Describe alternatives you've considered**
A working alternative is providing that path hardcoded in the environment:
```yaml
builds:
- binary: lxe
  main: cmd/lxe/main.go
  env:
  - GO111MODULE=on
  - GO_MODULENAME=github.com/automaticserver/lxe
  ldflags: 
    - -s -w
    - -X "{{.Env.GO_MODULENAME}}/cli/version.Version={{replace .Version (print "-" .ShortCommit) ""}}"
    - -X "{{.Env.GO_MODULENAME}}/cli/version.GitCommit={{.FullCommit}}"
    - -X "{{.Env.GO_MODULENAME}}/cli/version.BuildDate={{time "Mon, 02 Jan 2006 15:04:05 -0700"}}"
    - -X "{{.Env.GO_MODULENAME}}/cli/version.PackageName={{.ProjectName}}"
```
[...]

PR #2128, opened 2021-03-21
feat: allow to use ModulePath on templates

caarlos0:
closes https://github.com/goreleaser/goreleaser/issues/1688

### DIFF
--- a/internal/pipeline/pipeline.go
+++ b/internal/pipeline/pipeline.go
@@ -4,6 +4,7 @@ package pipeline
 import (
 	"fmt"
 
+	"github.com/goreleaser/goreleaser/internal/pipe/gomod"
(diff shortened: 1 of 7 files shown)


internal/pipeline/pipeline.go
```diff
@@ -37,6 +38,7 @@ type Piper interface {
 // BuildPipeline contains all build-related pipe implementations in order.
 // nolint:gochecknoglobals
 var BuildPipeline = []Piper{
+	gomod.Pipe{},           // setup gomod-related stuff
```
radeksimko, line 41: Wouldn't some `GO*` environment variables potentially affect the output of this, i.e. are you sure the order is correct here? 🤔

caarlos0: afaik no 🤔 

if something appears we can change though

PR #2145, opened 2021-03-29
fix: update go-gitlab to fix upload file issue

favadi:
Gitlab.com has some changed (bug?) recently that returns 500 internal server error if the filename contains "/".
Update go-gitlab dependency to get the workaround from upstream: https://github.com/xanzy/go-gitlab/pull/1113.

> If applied, this commit will...

Fix the uploading error for Gitlab.com.

> Why is this change being made?

To update go-gitlab dependency to get the workaround from upstream.

> Provide links to any relevant tickets, URLs or other resources

* https://github.com/xanzy/go-gitlab/pull/1113


caarlos0:
Thanks

PR #2169, opened 2021-04-13
Added ppc64le support

mgiessing:
If applied, this commit will...add IBM Power architecture for build commands and the according docker image

Why is this change being made?
It is made to directly include ppc64le during build/release process of new versions. Without this change it is required to clone the repository on a Power system and manually `make build` the goreleaser binary.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -18,6 +18,7 @@ builds:
     - amd64
     - arm
     - arm64
+    - ppc64le
   mod_timestamp: '{{ .CommitTimestamp }}'
   flags:
     - -trimpath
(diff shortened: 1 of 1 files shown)


caarlos0:
is it common to use ppc64le as development machines? 

asking because I never actually seen it, and this is the first time someone asks for this arch on our releases... 

also saw that your bio says you work for ibm, so I wonder how biased you are towards this.

IMHO, we should support only arches that most people actually use for development and on CI systems, as it doesn't impact goreleaser's ability to generate builds for the other arches any way.

mgiessing:
Well, I'd say it depends. For sure most of the users are working on x86 as this is by far the most common architecture. 
Power systems are usually seen in mid-range to large enterprises, but people indeed develop on these machines.

In the end it is not that much effort to clone this repository and build it on ppc64le, but to answer your question with another question (and I know that's usually not what people want to hear):

Would it be problematic for users or the maintenance of the releases itself to have ppc64le builds here? As Go provides multiarch, it is usually just a different flag value and I'd assume the maintenance effort wouldn't be too big.

In the end it is of course just a little bit of comfortability to directly have the power binary / docker image instead of cloning & building.

caarlos0:
> Would it be problematic for users or the maintenance of the releases itself to have ppc64le builds here? As Go provides multiarch, it is usually just a different flag value and I'd assume the maintenance effort wouldn't be too big.

just a bit longer build times... and the fact that I won't actually test it...

mgiessing:
Then I'd say if you don't feel comfortable with respect to your mentioned points you can close it without merging, but if I can assist or help you in any Power related topic when merging, then let me know.

caarlos0:
since no one else pinged about this, and its a rather unusual architecture, I'll not be merging this one - or at least not now, in order to focus on more common architectures.

In any case, thanks for the PR 🙏

PR #2032, opened 2021-01-22
feat: add `--build-id` flag for build command.

joselitofilho:
## If applied, this commit will...
This proposal of this PR is to add the `--build-id` flag for build command to build only the passed IDs.

## Why is this change being made?
This change is being made to choose which of the builds in the `YAML` file will be executed.

For example...
```yaml
builds:
  - main: ./cmd/main.go
    id: main
    env:
      - CGO_ENABLED=0
  - main: ./configs/main.go
    id: conf
```
When you run the command `goreleaser build --build-id=main ...` only the build with ID main will be executed.

PR #2179, opened 2021-04-21
feat: build single target and/or build id

caarlos0:
a rather simple way to implement a "single target build" approach.

- `goreleaser build --single-target`: builds only `runtime`'s `GOOS`/`GOARCH`
- `GOOS=linux GOARCH=darwin goreleaser build --single-target`: builds for specified `GOOS`/`GOARCH` environment variables
- `goreleaser build --single-target --id foo`: builds only `runtime`'s `GOOS`/`GOARCH` and build id `foo`

Cool things about this:
- no need to mess with anything but the command line
- known interface (GOOS/GOARCH envs)

Notes:
- if `--id` is passed but we have either 0 (default config) or 1 build configs, it is ignored (but warns)

What do you think?

--

closes https://github.com/goreleaser/goreleaser/pull/1987
closes https://github.com/goreleaser/goreleaser/issues/1852
closes https://github.com/goreleaser/goreleaser/pull/2032

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"os"
 	"runtime"
 	"time"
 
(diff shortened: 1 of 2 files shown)


radeksimko (approved):
LGTM! :shipit:

till:
This looks great!

edigaryev:
@caarlos0 thanks a lot for this! Would you mind cutting a new release?

PR #2180, opened 2021-04-21
docs: document required GitHub Actions token privileges

radeksimko:
This is to encourage use of reduced scope of the token by taking advantage of recently announced changes: https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/

### DIFF
--- a/www/docs/ci/actions.md
+++ b/www/docs/ci/actions.md
@@ -21,6 +21,9 @@ on:
   pull_request:
   push:
 
+permissions:
+  contents: write
+
 jobs:
   goreleaser:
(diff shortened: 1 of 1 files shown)


radeksimko:
As far as I can tell https://github.com/goreleaser/goreleaser/pull/2177 should not require any more permissions, but please correct me if I'm wrong @caarlos0

caarlos0:
> As far as I can tell #2177 should not require any more permissions, but please correct me if I'm wrong @caarlos0

I think you're right.

caarlos0 (approved):

PR #2205, opened 2021-05-06
fix: allow underscores in formula name

kreempuff:
I was trying to create two formulas per release. One signifying the latest and
another with the version explicitly: formula.rb and formula@v0.0.0.rb

The filenames are fine but the formula name for the explicit version
was created as FormulaATv0.0.0 which isn't valid Ruby.

Looking at the function history I couldn't find why underscores were being replaced
since underscores are valid for Ruby identifiers. I opted to replace dots in a formula with
underscores

- https://github.com/goreleaser/goreleaser/commit/1399a396036ea246522a4cbae7aef376d6dbbda2
- https://ruby-doc.org/docs/ruby-doc-bundle/Manual/man-1.4/syntax.html#ident

### DIFF
--- a/internal/pipe/brew/brew.go
+++ b/internal/pipe/brew/brew.go
@@ -323,8 +323,13 @@ func split(s string) []string {
 	return strings
 }
 
+// formulaNameFor transforms the formula name into a form
+// that more resembles a valid Ruby class name
+// e.g. foo_bar@v6.0.0-rc is turned into FooBarATv6_0_0RC
+// The order of these replacements is important
 func formulaNameFor(name string) string {
 	name = strings.ReplaceAll(name, "-", " ")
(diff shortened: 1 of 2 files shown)


caarlos0:
build will fail, as I'm pretty sure there are tests for this...

caarlos0:
> Looking at the function history I couldn't find why underscores were being replaced
> since underscores are valid for Ruby identifiers. I opted to replace dots in a formula with
> underscores

don't recall as well...

kreempuff:
> build will fail, as I'm pretty sure there are tests for this...

Makes sense, I'll update failing ones

kreempuff:
I think underscores are removed to make the Formula name cleaner

I reverted the change, added what I wanted, and added a comment on the function

caarlos0 (approved):

PR #2210, opened 2021-05-11
fix(ci): use official login action and switch to github secret for ghcr

crazy-max:
Use the official [docker/login-action](https://github.com/docker/login-action) to login against [Docker Hub](https://github.com/docker/login-action#docker-hub) and [GHCR](https://github.com/docker/login-action#github-container-registry).

[Container registry now supports GITHUB_TOKEN](https://github.blog/changelog/2021-03-24-packages-container-registry-now-supports-github_token/) so no need to use a PAT anymore :)

You may need to manage write and read access of GitHub Actions for repositories in the container settings:

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,25 +68,27 @@ jobs:
         name: Diff
         run: git diff
       -
-        name: Docker Login
-        if: success() && startsWith(github.ref, 'refs/tags/v')
-        env:
-          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
-          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -95,8 +97,3 @@ jobs:
           elif [[ $GITHUB_REF == refs/heads/master ]]; then
             ./goreleaser --snapshot
           fi
-      -
-        name: Clear
-        if: always() && startsWith(github.ref, 'refs/tags/v')
-        run: |
-          rm -f ${HOME}/.docker/config.json
```
crazy-max, line 102: This is [handled by docker login action](https://github.com/docker/login-action#inputs).

caarlos0:
Thanks @crazy-max 🎉

PR #2219, opened 2021-05-13
feat: Use GitLab Direct Asset Links

### LINKED ISSUE #2218: Support GitLab Direct Links
When producing releases to GitLab the assets uploaded (e.g build output) are not in a predicable format. The current format for uploaded artifacts, as linked form the release is: `{{gitlab-server}}/{{project-path}}/uploads/{{random-id}}/{{filename}}`.

This makes it difficult when writing scripts that download updates, as the `random-id` must be updated, rather than simply bumping the version number in the URL.

Use GitLabs direct asset linking ([assets/links](https://docs.gitlab.com/13.10/ee/api/releases/links.html#create-a-link) docs) to produce URLs in the form: `{{gitlab-server}}/{{project-path}}/-/releases/{{version}}/downloads/{{file-path}}/{{filename}}`.

This would allow simple updating of download URLs just by changing the `version` value in the URL, rather than having to identify the `random-id` provided.

None

**Additional context**
Has a requirement on go-gitlab supporting this functionality, which I have implemented here: https://github.com/xanzy/go-gitlab/pull/1135. 

I would be happy to contribute this functionality.

56KBs:
Implement the use of Direct Asset Links when uploading artifacts to a GitLab release.

This makes the URLs for artifacts uploaded to GitLab upon a release have a deterministic format for each release.

Closes #2218


caarlos0:
does this mean we can remove the `releaseLink.ID` stuff? 🤔

mavogel:
... and the `ArtifactUploadHash` stuff as well? :)

56KBs:
> does this mean we can remove the `releaseLink.ID` stuff? 🤔

@caarlos0 could you point out what you're referring to? I can only see one reference which just prints the URL.

> ... and the `ArtifactUploadHash` stuff as well? 🎉

Thanks for that @mavogel - I've removed `ArtifactUploadHash` functionality as you're right, it is no longer required. However I am yet to test the scoop/brew functionality as I've never used either of them so will need to take some time to lookup how they function, etc

mavogel:
Nice @56KBs! You can us https://gitlab.com/mavogel/release-testing as a  baseline for the brew stuff or use https://github.com/goreleaser/acceptance-tests:

TLDR; is
- create a repository in GitLab called, e.g. `homebrew-tap`
- release with `goreleaser` and the configured brew (as described in the release-testing for example)
- after success run the following commands
```
# tap it locally
brew tap mavogel/tools git@gitlab.com:mavogel/homebrew-tap.git
# install the binary
brew install mavogel/tools/release-testing
# run it to verify
release-testing --help
```

let me know if you need testing support

56KBs:
Thanks @mavogel - That made my life really easy (Except for me trying to use a private GitLab instance first which went very poorly).

Running it did highlight an issue with the template/URL name, which I've fixed now too.

If you'd like to see the end result for the URLs & homebrew generated repo, feel free to look here:

- https://gitlab.com/56KBs/goreleaser-testing/-/releases
- https://gitlab.com/56KBs/goreleaser-testing-homebrew

caarlos0:
Amazing @56KBs! Thanks a ton!

Q: does this need an specific version of gitlab? If so, we should probably point it out in the docs.

Other than that, lgtm, although it might be a breaking change for people using gitlab and using custom templates with `ArtifactUploadHash` (which I think chances are none-to-rare and they'll probably be glad to not have to use it anymore anyway)

56KBs:
> Amazing @56KBs! Thanks a ton!

Glad to contribute 😃 

> Q: does this need an specific version of gitlab? If so, we should probably point it out in the docs.

It was first documented in GitLab version 13.6, however, it looks to have been implemented but undocumented since at least 12.10. Happy to add a note to the relevant section of the docs with this requirement.

> Other than that, lgtm, although it might be a breaking change for people using gitlab and using custom templates with `ArtifactUploadHash` (which I think chances are none-to-rare and they'll probably be glad to not have to use it anymore anyway)

I'd agree it technically is breaking, but I also assumed the chance someone was using that template value would be really unlikely.

caarlos0:
> It was first documented in GitLab version 13.6, however, it looks to have been implemented but undocumented since at least 12.10. Happy to add a note to the relevant section of the docs with this requirement.

yes I think it would be better to mention that 🙏

Thanks!

56KBs:
> yes I think it would be better to mention that 🙏


Updated the docs in the two places that mention the GitLab version supported. If there is anything else needed for this PR, let me know :)

caarlos0 (approved):

Issue #2150, opened 2021-03-29
ability to announce to twitter


caarlos0:
Would be nice to be able to automatically announce releases to twitter, usually do it manually... having it automated would be neat.

leonwright:
Hmm.. Sounds like something I'd love to work on! I'll get on it.

leonwright:
@caarlos0 do you have a recommendation on where to put the logic for this feature in keeping with your current structure?

caarlos0:
I would probably create an announcer interface (like the current publisher one for example) and a Twitter implementation 🤔

We could then call all the announcers in the end of the release...

leonwright:
Alright, thanks for the help!

leonwright:
@caarlos0 Hi, I have completed a basic implementation of this. Below is an example of the config that will need to be added for this to work. Could you please test this and let me know if this functionality is what you expected before I go ahead and clean up and submit a final pull request?

PR #2227, opened 2021-05-18
Twitter announcer

leonwright:
Allow users the ability to announce via Twitter once they have completed a release.

This change was requested by owner in issue #2150 

Issue #2150

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,8 @@ require (
 	github.com/caarlos0/ctrlc v1.0.0
 	github.com/caarlos0/go-shellwords v1.0.12
 	github.com/campoy/unique v0.0.0-20180121183637-88950e537e7e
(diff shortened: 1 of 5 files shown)


caarlos0:
Looks good overall but IMHO secrets should not be in the config file... maybe allow them from env only?

caarlos0:
about the config, maybe something like this would be better (as we could add other announcers laters):

```yaml
announce:
  twitter:
    enabled: true
    message_template: 'New {{ .ProjectName }} version is out! Get it while it's fresh: https://github.com/caarlos0/{{ .ProjectName }}'
```

Then new announcers could be another section inside announce.

Also wonder if it would make sense to allow multiple twitter instances... but I'm assuming no 🤔

leonwright:
Thanks, I will update and revert.

caarlos0:
completed in #2248

thanks for the work here @leonwright 🤘

PR #2228, opened 2021-05-19
docs: fix config field

sixg0000d:
field `name_template` not found in type config.NFPMOverridables, but it's `file_name_template` found.


caarlos0:
thanks!

Issue #1802, opened 2020-09-10
Problems with signing artifacts in-place


jkowalski:
I'm trying to create a signing pipeline that will use `rpm --addsign` to sign RPM packages in-place.

```
signs:
  - id: sign-rpm-packages
    cmd: rpm
    artifacts: all
    args: ["--addsign", "${artifact}"]
    ids:
    - rpm-packages
  - id: sign-checksum
    artifacts: checksum
```

There are 2 problems:

1. Goreleaser appears to invoke the `rpm --addsign` command for all RPMs (which is correct), but also for the `checksums` file (which is unexpected and RPM fails to sign). If I remove 'ids' it will attempt to invoke the script on RPM, DEB, ZIP, TAR.GZ and checksums file (which I think is the correct behavior in terms of file selection).

Q: why does selecting `rpm-packages` also include checksums?

2. The second pipeline `sign-checksum` runs before/in parallel to the first pipeline, but because files are being modified in-place, the checksums file it signs is no longer valid.

Q: Is there a way to have checksums (re)generated after signing other artifacts and just before signing the actual checksums file?

I was able to workaround those issues by using a somewhat ugly script: https://github.com/kopia/kopia/blob/master/tools/sign.sh which ignores some files passed to it and regenerates checksums just before signing them. I don't know if this is intentional or by accident, but the checksums file is always passed last to the signing script, otherwise it would not work.

Please advise, I really hope there's a cleaner way to achieve signing in place without jumping through so many hoops.

(the full YAML file (with the workaround described) is in https://github.com/kopia/kopia/blob/master/.goreleaser.yml)

caarlos0:
hmm, those all seem valid bugs, yes, thanks for reporting.

The root cause, though, will be eventually fixed in https://github.com/goreleaser/nfpm/issues/212

caarlos0:
So the issues here are:

1. need to be able to skip checksums from the sign pipe
1. the problem with the checksums not being re-generated after signing

problem 2 is a little bit trickier to solve, and I'm not sure we should.
problem 1 is a bug indeed.

in any case, we need to port the new nfpm options to goreleaser (signing), and it will solve both problem in this particular case at least. #1829

NorseGaud:
2 is also an issue for cmd: gon. If it is possible to skip the checksum file uploading, that would be great.

caarlos0:
problem 1 should be fixed on v0.166.0, can you try?

PR #2235, opened 2021-05-21
feat: fine grained sign

caarlos0:
refs #1802

allow more options on `artifacts`.

### DIFF
--- a/internal/pipe/sign/sign.go
+++ b/internal/pipe/sign/sign.go
@@ -93,7 +93,7 @@ func (Pipe) Run(ctx *context.Context) error {
 				return fmt.Errorf("invalid list of artifacts to sign: %s", cfg.Artifacts)
 			}
 
-			if cfg.Artifacts != "checksum" && len(cfg.IDs) > 0 {
+			if len(cfg.IDs) > 0 {
 				filters = append(filters, artifact.ByIDs(cfg.IDs...))
 			}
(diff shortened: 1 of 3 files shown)


internal/pipe/sign/sign.go
```diff
@@ -81,14 +81,21 @@ func (Pipe) Run(ctx *context.Context) error {
...
+				filters = append(filters, artifact.ByType(artifact.LinuxPackage))
 			case "none":
 				return pipe.ErrSkipSignEnabled
 			default:
 				return fmt.Errorf("invalid list of artifacts to sign: %s", cfg.Artifacts)
 			}
+
+			if cfg.Artifacts != "checksum" && len(cfg.IDs) > 0 {
+				filters = append(filters, artifact.ByIDs(cfg.IDs...))
+			}
```
radeksimko, line 98: Just to clarify - do we not do the filtering by both IDs and checksum because there will always be only one checksum for the release? If so, then I'm assuming this then also applies to `source` - which is not part of the condition here?

caarlos0: the id filter skips itself when the type is checksum

https://github.com/goreleaser/goreleaser/blob/master/internal/artifact/artifact.go#L227-L230

I think this if there was from before this... or just to be extra sure... not really remember 🤔

caarlos0: I don't even need the if there actually, forgot it there 🤣

radeksimko (approved):

PR #2236, opened 2021-05-21
feat: allow to template signs[].signature 

### LINKED ISSUE #2195: Allow variable interpolation in sings[].signature

Sometimes it can be useful to identify what key was used to generate a particular sig file (for example when keys are being rotated). This is currently only possibly by _hard coding_ the key identifier, such as

```yaml
signs:
  -
    id: with_key_id
    signature: "${artifact}.72D7468F.sig"
    args: ["--batch", "--no-tty", "-u", "{{ .Env.GPG_KEY_ID }}", "--output", "${artifact}.{{ .Env.GPG_KEY_ID }}.sig", "--detach-sign", "${artifact}"]
    artifacts: checksum
```

I would like to be able to use variable such as `{{ .Env.GPG_KEY_ID }}` inside `signature` field.

Besides hard-coding the key ID I attempted to omit `signature` from configuration and just specify the sig filename in `args`:

```yaml
signs:
  -
    id: with_key_id
    args: ["--batch", "--no-tty", "-u", "{{ .Env.GPG_KEY_ID }}", "--output", "${artifact}.{{ .Env.GPG_KEY_ID }}.sig", "--detach-sign", "${artifact}"]
    artifacts: checksum
```

which does make `gpg` generate the file into the right location, but then following pipes can't find it, because they only look for the filename defined as part of `signature` - which makes sense as it's hard for GoReleaser to guess what `gpg` did and where did I write the sig file (if it even did).

caarlos0:
closes  #2195

### DIFF
--- a/internal/pipe/sign/testdata/README.md
+++ b/internal/pipe/sign/testdata/README.md
@@ -8,73 +8,84 @@ key works.
 This uses the `--homedir .` option to create the test keys so that we do not touch
 the local keyring file.
 
-1.  Create signing keys
+## 1.  Create signing keys
 
-    cd $GOPATH/src/github.com/goreleaser/goreleaser/internal/pipeline/sign/testdata/gnupg
-    gpg --homedir . --quick-generate-key --batch --passphrase '' nopass default default 10y
+```sh
+cd $GOPATH/src/github.com/goreleaser/goreleaser/internal/pipeline/sign/testdata/gnupg
+gpg --homedir . --quick-generate-key --batch --passphrase '' nopass default default 10y
+```
 
-1.  Check that the key exists
+## 2. Check that the key exists
 
-    ## $ gpg --homedir . --list-keys
+```sh
+gpg --homedir . --list-keys
+pub rsa2048 2017-12-13 [SC][expires: 2027-12-11]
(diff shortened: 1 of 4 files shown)


internal/pipe/sign/testdata/README.md
```diff
@@ -8,73 +8,84 @@ key works.
...
+rm foo foo.sig
+```
 
-    rm foo foo.sig
+## 4. Make sure you have keyrings for both gpg1 and gpg2
 
-1)  Make sure you have keyrings for both gpg1 and gpg2
+travis-ci.org runs on an old Ubuntu installation which
+has gpg 1.4 installed. We need to provide keyrings that
+have the same keys and users for both formats.
```
radeksimko, line 57: Probably not related to this PR, but this looks a bit out-dated as GoReleaser no longer uses Travis from what I can tell?

radeksimko (approved):

internal/pipe/sign/testdata/README.md
caarlos0: yeah, still useful to test with both gpg1 and gpg2 I think... 🤔

just fixed formatting, in another pr I'll take a proper look at it...

PR #2241, opened 2021-05-21
feat: Scoop pipe - Write the manifest before skips, like Brew pipe.

eolder:
This change makes the scoop pipe write out the manifest before the skip conditions.

By writing the manifest before the skip section of the pipe, you can do dry runs of the scoop pipeline. This is how the brew pipe does it.

Example behavior before change:
```
    ...
      • homebrew tap formula
         • writing                   formula=dist/app.rb
         • pipe skipped              error=brew.skip_upload is set
      • scoop manifests  
         • pipe skipped              error=scoop.skip_upload is true
    ...
```
After
``` 
    ...
      • homebrew tap formula
         • writing                   formula=dist/app.rb
         • pipe skipped              error=brew.skip_upload is set
      • scoop manifests  
         • writing                   manifest=app.json
         • pipe skipped              error=scoop.skip_upload is true
    ...
```

### DIFF
--- a/internal/pipe/scoop/scoop.go
+++ b/internal/pipe/scoop/scoop.go
@@ -5,6 +5,8 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


caarlos0:
nice! thanks for the pr!

can you add a test for it as well?

eolder:
In `e5456fa` I added a test for the modification as well as a small change - the manifest is now written to dist instead of the base dir.

caarlos0 (approved):

Issue #1638, opened 2020-07-01
enable file mapping in archives


rbonthond:
I'm building multiple artifacts with archives, and each has it's on documentation/settings/etc. I would like to be able to map files to a different location using "files" field.

Something like:

```
archives:
  - <<: &archive_defaults
      name_template: "{{ .ProjectName }}-{{ .Binary }}-{{ .Version }}-{{ .Os }}-{{ .Arch }}"
      format: tar.gz
      wrap_in_directory: true
      files:
        - pkg/{{ .Binary }}/etc/**:etc/**
        - pkg/{{ .Binary }}/doc/**:doc/**
    id: cmdA
    builds: [cmdA]
  - <<: *archive_defaults
    id: cmdB
    builds: [cmdB]
  - <<: *archive_defaults
    id: cmdC
    builds: [cmdC]
```

jaxxstorm:
I also need this. I have `files` which are in subfolders, and I'd like to be able to but them in the root of the build directory, but I currently can't.

jaxxstorm:
It looks like this was attempted here

https://github.com/goreleaser/goreleaser/pull/1351

and discussed here: https://github.com/goreleaser/goreleaser/issues/679

@caarlos0 is there a workaround for this?

jaxxstorm:
for anyone who has a similar issue, I worked around it with a pre-hook.

Add the files you want to add to your `.gitignore` so that goreleaser doesn't yell at you for having a dirty working tree:

```
custom-file
```
 (note, don't use globs)

Then add in your pre-hook, simply copy the file to the root of your directory:

```
before:
  hooks:
    - cp path/to/my/file/custom-file .
```

Then you can specify the file in the archive as normal.

This unfortunately means you'll have a file in your local working copy that is erroneous, because we don't have hooks on the archive or in the global section

ainsleyclark:
Im having problems with this also. 

I have multiple folders that reside under /build. That have to be copied to the main directory when releasing otherwise the app will fail.

The solution described above would work for files, but not for when you are copying whole directories and if they have the same name as the ones in your root folder.

Is there a quick solution for this? Perhaps integration with CI?

Thanks @caarlos0

ainsleyclark:
Can I just confirm that the feature was implemented but never pulled in? 

https://github.com/goreleaser/goreleaser/pull/1351

marcauberer:
[...]

PR #2253, opened 2021-05-25
feat: builds.output_path

tmeckel:
This Pull Request will add the option `output_path` to a build definition in the `config.yaml` file.

This will add the ability to fully control the path of the created binary by `goreleaser`.  This comes in handy, when you create for example a (ZIP) archive, for distributing Terraform providers inside a CI/CD pipeline and you use `goreleaser` as a matrix build tool.

### DIFF
--- a/internal/pipe/build/build.go
+++ b/internal/pipe/build/build.go
@@ -186,10 +186,17 @@ func buildOptionsForTarget(ctx *context.Context, build config.Build, target stri
 
 	build.Binary = binary
 	name := build.Binary + ext
+	outputPath := fmt.Sprintf("%s_%s", build.ID, target)
+	if build.OutputPath != "" {
+		outputPath, err = tmpl.New(ctx).WithBuildOptions(buildOpts).Apply(build.OutputPath)
+		if err != nil {
+			return nil, err
+		}
+	}
 	path, err := filepath.Abs(
 		filepath.Join(
 			ctx.Config.Dist,
-			fmt.Sprintf("%s_%s", build.ID, target),
+			outputPath,
 			name,
 		),
 	)
--- a/internal/pipe/build/build_test.go
+++ b/internal/pipe/build/build_test.go
@@ -747,6 +747,24 @@ func TestBuildOptionsForTarget(t *testing.T) {
 				Arch:   "amd64",
 			},
 		},
+		{
+			name: "binary name with Os and Arch template variables and output directory",
+			build: config.Build{
+				ID:     "testid",
+				Binary: "testbinary_{{.Os}}_{{.Arch}}",
+				Targets: []string{
+					"linux_amd64",
+				},
+				OutputPath: "outputpath/{{.Os}}/{{.Arch}}",
+			},
+			expectedOpts: &api.Options{
+				Name:   "testbinary_linux_amd64",
+				Path:   filepath.Join(tmpDir, "outputpath", "linux", "amd64", "testbinary_linux_amd64"),
+				Target: "linux_amd64",
+				Os:     "linux",
+				Arch:   "amd64",
+			},
+		},
 	}
 
 	for _, tc := range testCases {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -203,6 +203,7 @@ type Build struct {
 	ModTimestamp string         `yaml:"mod_timestamp,omitempty"`
 	Skip         bool           `yaml:",omitempty"`
 	GoBinary     string         `yaml:",omitempty"`
+	OutputPath   string         `yaml:"output_path,omitempty"`
 }
 
 type HookConfig struct {
--- a/www/docs/customization/build.md
+++ b/www/docs/customization/build.md
@@ -126,6 +126,12 @@ builds:
     # Useful for library projects.
     # Default is false
     skip: false
+
+    # If defined, the created binary of a (matrix) build is 
+    # saved to the specified directory. The directory is always a sub-directory 
+    # of the dist folder. If the binary attribute itself is a path,
+    # the binary is saved in the path underneath the path specified here.
+    output_path: "my-path/{{.Os}}/{{.Arch}}"
 ```
 
 !!! tip


caarlos0:
hey @tmeckel, thanks for the pr!

just want to understand: you need this to change the path of the binary inside the zip file, correct?

if so, if we had file mappings in the archive pipe, this won't be needed, correct?

tmeckel:
@caarlos0 yes one of the benefits of having the ability  to fully control the output path can be the fact you can simply archive the created artifacts without having to fiddle with some sort of "file mappings" and stuff. Either with the "archive pipe" you mentioned or with another (external) script. But aside that, further processing of the goreleaser artifacts is often much more easier when you can control where they are created, than arduously rearranging them afterwards. 

Even if `goreleaser` is mainly focussing on creating "releases", it's a really great matrix build tool compared to tools like gox or somethings. And even better, if you start using `goreleaser` simply for creating artifacts for multiple target platforms, you don't have to switch your tool set later, if you're required to create fully fledged releases.

Apropos archive pipeline and file mappings: according to the documentation of the archive pipeline (https://goreleaser.com/customization/archive/) I don't have any clue how I could rearrange (manipulate) the files paths inside the created archive. Any guidance (help)?

caarlos0:
> @caarlos0 yes one of the benefits of having the ability to fully control the output path can be the fact you can simply archive the created artifacts without having to fiddle with some sort of "file mappings" and stuff. Either with the "archive pipe" you mentioned or with another (external) script. But aside that, further processing of the goreleaser artifacts is often much more easier when you can control where they are created, than arduously rearranging them afterwards.
> 
> Even if `goreleaser` is mainly focussing on creating "releases", it's a really great matrix build tool compared to tools like gox or somethings. And even better, if you start using `goreleaser` simply for creating artifacts for multiple target platforms, you don't have to switch your tool set later, if you're required to create fully fledged releases.

hmm, yeah, assuming the usage of `goreleaser build` only, it makes sense indeed.



> Apropos archive pipeline and file mappings: according to the documentation of the archive pipeline ([goreleaser.com/customization/archive](https://goreleaser.com/customization/archive/)) I don't have any clue how I could rearrange (manipulate) the files paths inside the created archive. Any guidance (help)?

its not yet possible, see #1638

caarlos0:
One possible issue I see with this: the user could define the "same" path for all binaries, which means builds would override each other 🤔 

not sure if we need to fix/prevent it, but maybe document? 

Also, maybe we could instead make `binary` templateable? 🤔

tmeckel:
> 
> 
> One possible issue I see with this: the user could define the "same" path for all binaries, which means builds would override each other 🤔
> 
> not sure if we need to fix/prevent it, but maybe document?
> 
when this could be the case, then this should definitely be documented. This could cause headaches 👎🏼 

> Also, maybe we could instead make `binary` templateable? 🤔

Well, if this could be done without creating a breaking change it would be a much more general solution than having a new configuration property :+1

caarlos0:
> Well, if this could be done without creating a breaking change it would be a much more general solution than having a new configuration property :+1

afaik it could yeah... once upon a time it couldn't because docker builds would need to specify which binaries to copy, but that's no longer the case...

tmeckel:
> 
> 
> > Well, if this could be done without creating a breaking change it would be a much more general solution than having a new configuration property :+1
> 
> afaik it could yeah... once upon a time it couldn't because docker builds would need to specify which binaries to copy, but that's no longer the case...

So what would you proposal here. Should I cancel my PR? Or should I add the functionality for templating the `binary` setting to this PR. If so what would you propose how this should be implemented.

caarlos0:
I was digging into it... so, binary is templateable already... the issue is that it still puts the binaries in a subfolder in dist, e.g.:

```yaml
binary: "{{.Os}}/{{.Arch}}/{{.ProjectName}}_{{.Os}}_{{.Arch}}"
```

becomes

```sh
./dist/goreleaser_linux_arm64/linux/arm64/goreleaser_linux_arm64
```

so, I'm wondering if this is just a matter of lack of documentation or if you need the binaries to not be wrapped in `{{ .ProjectName }}_{{ .Os }}_{{ .Arch }}`?

caarlos0:
FWIW, just tried a config like this here:

```yaml
builds:
  - env:
      - CGO_ENABLED=0
    goos:
      - linux
      - darwin
    goarch:
      - arm64
      - amd64
    binary: "{{.Os}}/{{.Arch}}/{{.ProjectName}}_{{.Os}}_{{.Arch}}"

```

and archives also look as what I think you would expect:

```sh
❯ tar tvf dist/foobar_v0.0.0-SNAPSHOT-7efc141_linux_amd64.tar.gz 
-rwxrwxr-x carlos/carlos 1355776 2021-06-02 02:08 linux/amd64/foobar_linux_amd64
```

tmeckel:
> 
> 
> I was digging into it... so, binary is templateable already... the issue is that it still puts the binaries in a subfolder in dist, e.g.:
> 
> ```yaml
> binary: "{{.Os}}/{{.Arch}}/{{.ProjectName}}_{{.Os}}_{{.Arch}}"
> ```
> 
> becomes
> 
> ```shell
> ./dist/goreleaser_linux_arm64/linux/arm64/goreleaser_linux_arm64
> ```
> 
> so, I'm wondering if this is just a matter of lack of documentation or if you need the binaries to not be wrapped in `{{ .ProjectName }}_{{ .Os }}_{{ .Arch }}`?

Exactly that's why I introduced the `output_path` property, because the current code in `build.go` uses a static pattern to build to output path:

image

and because I'm fan of not introducing breaking changes (especially for a tool like `goreleaser` that is heavily used!) I changed the generation of  the output path as the following, where the old strategy is the default and is only overwritten when the `output_path` is set:

image

but I've to admit that if there would be a way to fully control the output path **without** introducing a new config setting this would be great! 👍🏼 

And about your test with the `binary` settings using a pattern `"{{.Os}}/{{.Arch}}/{{.ProjectName}}_{{.Os}}_{{.Arch}}"`:

This will still create those "default" folders inside the `dist` directory. The pattern is only effective for this sub directory.

I used the following pattern:

`binary: "microsoft/azuredevops/{{ .Version }}/{{ .Os }}_{{ .Arch }}/terraform-provider-azuredevops" `

which creates the following directories inside dist:

image

But my approach creates the following structure, when the `output_path` is set like the following and the `binary` settings is set to a static string:

* `binary: "terraform-provider-azuredevops"`
* `output_path: "microsoft/azuredevops/{{ .Version }}/{{ .Os }}_{{ .Arch }}"`

image

With this directory structure I'm able create a ZIP file, which contains *all* binary versions for the different target platforms in one single shot:

```shell
$ cd dist
$ zip -9rv microsoft-azuredevops-1.1.0.zip microsoft
```

This ZIP file then has the required directory structure that is expected the `terraform-bundle` command.

image

I hope I could clarify my use-case and motivation for the PR 😀

caarlos0:
hmm got it...

will need to think for a while on a good way to introduce this without breaking changes...

I think we could assume most people only use the `binary` property to change the binary name instead of its full path, and for the people that do change it, I think they would expect it to change the full path (like you do)

so, far, 2 ideas:

## 1

- on Defaults, check if binary has a `/` or `\`, if not, prefix with `{{.ProjectName}}_{{.Os}}_{{.Arch}}{{.Arm}}{{.Mips}}` (or something like that)
- remove that sprintf

## 2

- add a new `wrap_dir` (or something with a better name) that defaults to true on build
- only if its true prepend the sprintf thing

2nd option seems to be more backwards-compatible... just need to find a good name for it :P

tmeckel:
# 3rd

Accept my PR and go with `output_path` 👍🏼  😀 


But either way I'm happy that I got my point across and that you grasped the use case. I'm fine with any solution here 😸

caarlos0:
having two options in which you can specify the path looks a bit confusing IMHO...

tmeckel:
> 
> 
> having two options in which you can specify the path looks a bit confusing IMHO...

That's why I said it's kinda crazy... but if it comes to me I set the `binary` setting to a constant value (obviously because I wasn't aware of the fact that's possible to use a template string 😁 ) but you can consider `output_path` as a "base path" or something as I wrote in the "documentation".

But I agree _ease of use_ and _predictability_ are two very valid factors one should consider designing a tool or API.

tmeckel:
> 
> 2nd option seems to be more backwards-compatible... just need to find a good name for it :P

A good name for the option might be `architecture_dir` or `create_architecture_dir` because in the current implementation it's the path for the combination of  `.Os` and `.Arch` in a matrix build.

caarlos0:
check https://github.com/goreleaser/goreleaser/pull/2280 plz :) @tmeckel

Issue #2159, opened 2021-04-06
brew: Add explicit OS/arch dependencies to communicate compatibility gracefully


radeksimko:
**Describe the bug**

Users attempting to install a program via Homebrew formula generated by GoReleaser on unsupported OS/arch may receive confusing errors and be unable to `brew tap` the whole repo just because it contains invalid formulas with effectively missing `url` (after none of [these conditions](https://github.com/goreleaser/goreleaser/blob/41356ea84d734928493a7342d3c44e4b163237bc/internal/pipe/brew/template.go#L52-L89) are met).

**To Reproduce**

I don't have a full repro case (yet), but we maintain a generator that generates formulas in pretty much the same way as GoReleaser and people have reported failures on Apple Silicon https://github.com/hashicorp/homebrew-tap/issues/134 so I'm working with the assumption that this must affect GoReleaser too, it just probably isn't a common enough problem people run into (yet).

**Expected behavior**

User is presented with a clear error message that their OS/arch is not supported and (most importantly) they're able to still `brew tap` the tap even if it contains formulas which are not supported on their platform.

**Additional context**

I believe we could just set some default dependencies in formulas automatically as we know what's supported (OS/arch) based on the builds anyway.

For example
```ruby
depends_on arch: [:x86_64] if OS.mac?
```
or alternatively
```ruby
on_macos do
  depends_on arch: [:x86_64]
end
```

--- 

Apologies for the slightly vague description and lack of repro case - I'm just currently swamped with other unrelated work and wanted to create this issue to track the problem until I get to troubleshoot it properly and ideally send a patch.

caarlos0:
sounds good, I researched a while ago how to do this and didn't found docs for it... glad to see its possible! 🙏

caarlos0:
> Maybe it has been fixed?

no idea, need to test 🤔

caarlos0:
check https://github.com/goreleaser/goreleaser/pull/2261

seems to be a bit better yes, per my tests at least...

if someone can help double-checking it would be amazing!

caarlos0:
#2261 was partially reverted due to #2270

see Homebrew/discussions#1595

caarlos0:
yeah, makes sense... I think the custom requirement is a good option... will look into how that can be made...

ryancurrah:
Our company is starting to ship M1's to developers and when they are installing our internal tools through brew formula created by go releaser they are running into the same problem.

ryancurrah:
[...]

PR #2261, opened 2021-05-27
fix: improve brew formula

caarlos0:
closes https://github.com/goreleaser/goreleaser/issues/2159

would be amazing if someone can help testing this 🙏

### DIFF
--- a/internal/pipe/brew/brew.go
+++ b/internal/pipe/brew/brew.go
@@ -198,8 +198,39 @@ func buildFormula(ctx *context.Context, brew config.Homebrew, client client.Clie
 	return doBuildFormula(ctx, data)
 }
 
+func fixDataDownloads(data templateData) templateData {
+	if data.MacOSAmd64.DownloadURL != "" {
+		data.HasMacOSDownloads = true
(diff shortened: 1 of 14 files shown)


mgrachev (approved):
LGTM 👍

radeksimko (commented):
This is better than it was, but I reckon that a part of this PR was reverted due to https://github.com/goreleaser/goreleaser/issues/2270 - which is reasonable solution for unblocking people 👍🏻  but it still leaves the main case described as part of the original issue unresolved.

I think that `depends_on :arch` was the right approach, but from what I can tell it evaluates multiple architectures there as "AND", rather than "OR", which then makes it an unsatisfiable condition for any multi-arch builds.

https://github.com/Homebrew/brew/blob/3476ca7b8ff889ef61e65783d35d1b2073bd4012/Library/Homebrew/requirements/arch_requirement.rb#L21-L27

I'm not sure this is actually intended on Homebrew's side, so I'm going to open an issue 🤔

radeksimko:
See https://github.com/Homebrew/discussions/discussions/1595

caarlos0:
yes, I forgot to reopen the issue, thanks for pinging 🙏🏻

PR #2267, opened 2021-05-30
feat: release header/footer

caarlos0:
Adds `release.header` and `release.footer`.

The `--release-notes` et al flags sets the contents/header/footer of the changelog itself, not the release. This is a different feature - hopefully useful :)

Also created a `golden` package with helper functions to compare golden files :)


radeksimko (commented):
Looks useful, although I don't have a use case for it personally 👍🏻

zhouhaibing089:
This is pretty useful! I'd hope this can be templated, though.

Issue #2201, opened 2021-05-04
custom publishers should preserve env vars


meier-christoph:
**Is your feature request related to a problem? Please describe.**

Custom publishers is a nice feature and I totally get why you didn't copy env vars but it should at least have an option to so.

As it is now, in my case at least, I need to set PATH, TMP, LOCALAPPDATA and some more and to make it worse .Env.PATH != .Env.Path and since I need to run this on both windows and linux ... well its a mess.

**Describe the solution you'd like**

A simple flag which, if is set, preserves existing env vars for custom publishers. Leave it off by default for backwards compatibility.

caarlos0:
seems easy enough, but maybe would be even better to keep the "expected" envs by default (PATH, TMP, HOME, etc?)

cc/ @radeksimko thoughts?

radeksimko:
Yeah, that sounds reasonable. We could pass through the following:

- `HOME`
- `USER`
- `USERPROFILE`
- `TMPDIR`
- `TMP`
- `TEMP`
- `PATH`

which at AFAIK corresponds to some basic common operations within Go stdlib; across Windows and Unix - at least that's what we did some time ago in another project: https://github.com/hashicorp/terraform-ls/blob/bc69f994136d741deb627f68b57479bb70c6eb9e/internal/terraform/exec/exec.go#L24-L32

I wouldn't oppose _strongly_ an introduction of some sort of opt-in inheritance, but if the above pass-through list can solve most of the problems then I'd be much happier.

caarlos0:
> I wouldn't oppose _strongly_ an introduction of some sort of opt-in inheritance, but if the above pass-through list can solve most of the problems then I'd be much happier.

yeah, that sounds like the best option for me as well

PR #2277, opened 2021-06-02
fix(publishers): inherit some system variables

caarlos0:
closes https://github.com/goreleaser/goreleaser/issues/2201

### DIFF
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -1,7 +1,9 @@
+// Package exec can execute commands on the OS.
 package exec
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 
 	"github.com/apex/log"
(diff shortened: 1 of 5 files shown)


internal/exec/exec.go
```diff
@@ -59,7 +65,14 @@ func executeCommand(c *command) error {
...
 		// nolint: gosec
 	cmd := exec.CommandContext(c.Ctx, c.Args[0], c.Args[1:]...)
-	cmd.Env = c.Env
+	cmd.Env = []string{}
+	for _, key := range passthroughEnvVars {
+		if value := os.Getenv(key); value != "" {
+			cmd.Env = append(cmd.Env, key+"="+value)
+		}
+	}
+	cmd.Env = append(cmd.Env, c.Env...)
```
radeksimko, line 74: How would this work with duplicates - e.g. when `PATH` is declared explicitly in config and also passed through? I'm assuming that last one wins? Is it worth deduplicating it?

radeksimko (approved):
LGTM aside from the misspells and the inline question about duplicates.

internal/exec/exec.go
caarlos0: afaik last one wins, so there is little to no use in deduplicating it 🤔

radeksimko (approved):

PR #2278, opened 2021-06-02
fix: brew require arch

### LINKED ISSUE #2159: brew: Add explicit OS/arch dependencies to communicate compatibility gracefully

caarlos0:
closes #2159
refs #2261
see https://github.com/Homebrew/discussions/discussions/1595


internal/pipe/brew/template.go
```diff
@@ -39,6 +41,36 @@ const formulaTemplate = `# typed: false
 {{ if .CustomRequire -}}
 require_relative "{{ .CustomRequire }}"
 {{ end -}}
+
+class SupportedArchRequirement < Requirement
```
radeksimko, line 45: Am I right (per your comments in the linked discussion) that for this to work (and Ruby to be satisfied) this extra class would have to be in a separate file?

If so then we'd be making the assumption that the whole tap/repo is basically owned by GoReleaser, which seems like a flawed assumption?

radeksimko (commented):
Given the relative silence in the linked discussion and relatively suboptimal available "easy/quick" solution (based on my understanding) I'm wondering if it's better to instead treat this as "upstream bug" and have it fixed upstream, in Homebrew as initially discussed?

https://github.com/Homebrew/discussions/discussions/1595#discussioncomment-816103

caarlos0:
Yep, probably would be cleaner than this workaround...

caarlos0:
since no news on this, will close.

internal/pipe/brew/template.go
ryancurrah: For us our tap is managed only by goreleaser.

caarlos0: in any case, this goes to the tap itself, it does not need to be a separated file...

PR #2280, opened 2021-06-03
feat: builds. no_unique_dist_dir

caarlos0:
still not happy about the name, impl looks a bit nicer though.

closes  #2253

This might include a breaking change: before you could set a path on the binary name, e.g., `foo/mybin`, which would cause snap and nfpm packages to try to put the binary at `/usr/local/bin/foo/mybin`. EDIT: snap would actually fail, just nfpm would do this.

After this, the paths are correct.

---

it seems this is specially useful for terraform providers...


tmeckel:
@caarlos0 another good option to have the use case of a fully configurable output path implemented 👍🏼 But this does not solve your objection about having the ability to specify path templates at two locations. One here with the `dist_path` settings and then with the `binary` setting. 

My initial idea (before i changed my mind and went with the `output_path`) to get rid of the automatically created "target path" (`fmt.Sprintf("%s_%s", build.ID, target)`) was to introduce a setting `no_architecture_dir` which is a `boolean` and if set to true prevents the execution of the `fmt.Sprintf` function call.

image

caarlos0:
yeah, I thought it would make sense, but having a bool that disables the unique dir is nicer IMHO.

Just pushed, can you test this revision in your pipeline just to be sure it works as we expect?

caarlos0:
also cc/ @radeksimko since it involves terraform providers...

internal/pipe/build/build.go
```diff
@@ -198,7 +202,7 @@ func buildOptionsForTarget(ctx *context.Context, build config.Build, target stri
 	}
 
 	log.WithField("binary", path).Info("building")
-	buildOpts.Name = name
+	buildOpts.Name = filepath.Base(name)
```
caarlos0, line 205: this might be a breaking change... if you have very peculiar config...

tmeckel:
@caarlos0 I can say this works as good as my PR #2253 So from my point of view I could retract (close) my PR.

https://github.com/goreleaser/goreleaser/pull/2253#issuecomment-853124533

image

image

caarlos0:
nice, I built from yours so your name will still be in the commit, basically changed names/docs a bit and also add that fix for nfpm/snap packages...

Thanks again for the pr/efforts testing et all, really appreciate 🙏

tmeckel:
@caarlos0 

> 
> 
> nice, I built from yours so your name will still be in the commit, basically changed names/docs a bit and also add that fix for nfpm/snap packages...

Awesome 👍🏼 

> 
> Thanks again for the pr/efforts testing et all, really appreciate 🙏

I think it's an example how open source development should work: finding together the best solution. It was really nice to have worked with you on this 👍🏼

caarlos0:
> I think it's an example how open source development should work: finding together the best solution. It was really nice to have worked with you on this 👍🏼

thanks, likewise 🖤

PR #2287, opened 2021-06-05
feat: remove :latest imgs from release notes

caarlos0:
I was thinking... does it makes sense to add the `:latest` docker images to the release notes? After all, **it will be wrong** eventually anyway...

I'm thinking... maybe its better if we just don't print those imgs there?

wdyt?

### DIFF
--- a/internal/pipe/release/body.go
+++ b/internal/pipe/release/body.go
@@ -2,6 +2,7 @@ package release
 
 import (
 	"bytes"
+	"strings"
 	"text/template"
 
 	"github.com/goreleaser/goreleaser/internal/artifact"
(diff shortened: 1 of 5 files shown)


marcauberer:
@caarlos0 Because in older releases the image for `latest` is also included, but will reference a newer image version? Yes, that is a justified thought ...

marcauberer (approved):

caarlos0:
> Because in older releases the image for `latest` is also included, but will reference a newer image version?

exactly...

radeksimko (approved):
I agree, `:latest` does have its place e.g. on the website, but it doesn't seem that relevant in release notes for a particular version.

PR #2326, opened 2021-07-01
pipe/release: template header and footer

zhouhaibing089:
A continuation of #2267 - the header and footer in release configuration
can be templated as well to support more use cases like attaching
published docker image information(without using docker pipe).


caarlos0:
Thanks!

Issue #2332, opened 2021-07-04
Goreleaser Docker Image does not support Multi-Arch builds


msvechla:
**Describe the bug**

The current goreleaser Docker image does not support multi-arch builds, because the Docker buildx plugin is not installed in the image.

This currently results in the following error in CI:

```
   • building docker image     buildx=true image=msvechla/vaultbot:1.12.0-rc1-amd64
   ⨯ release failed after 162.95s error=failed to build docker image: msvechla/vaultbot:1.12.0-rc1-amd64: 
unknown flag: --load
See 'docker --help'.
```

**To Reproduce**

```yaml
project_name: vaultbot
builds:
  - env: [CGO_ENABLED=0]
    binary: vaultbot
    goos:
      - linux
      - windows
      - darwin
    goarch:
      - amd64
      - arm64
dockers:
- image_templates: 
  - "msvechla/vaultbot:{{ .Version }}-amd64"
  dockerfile: Dockerfile
  use_buildx: true
  build_flag_templates:
  - "--platform=linux/amd64"
- image_templates: 
  - "msvechla/vaultbot:{{ .Version }}-arm64"
  dockerfile: Dockerfile
  use_buildx: true
  build_flag_templates:
  - "--platform=linux/arm64"
docker_manifests:
- name_template: msvechla/vaultbot:{{ .Version }}
  image_templates:
  - "msvechla/vaultbot:{{ .Version }}-amd64"
  - "msvechla/vaultbot:{{ .Version }}-arm64"
```

**Expected behavior**

Working multi-arch builds with buildx.

**Environment (please complete the following information):**

- OS: linux
- OS version: -
- GoReleaser Version: 0.173.1

I will open a PR to add `buildx` to the goreleaser Docker image. Let me know what you think.
Thanks a lot for this awesome project!

caarlos0:
fixed on #2333 assuming the parent host supports multi-arch builds.

PR #2333, opened 2021-07-04
feat: add docker-buildx to Docker image

msvechla:
This commit will add the `buildx` docker plugin, to enable multi-arch Docker builds.
For more information see #2332.

Let me know what you think, whether this makes sense, or if some modifications are required.
Thanks a lot!


Dockerfile
```diff
@@ -16,3 +16,7 @@ RUN chmod +x /entrypoint.sh
 
 COPY goreleaser_*.apk /tmp/
 RUN apk add --allow-untrusted /tmp/goreleaser_*.apk
+
+RUN mkdir -p $HOME/.docker/cli-plugins/ && \
```
caarlos0, line 20: I think we can just add `docker-cli-buildx` to the list of packages to install in line 3...

msvechla: even better, I will give it a try and update the PR if it works. Thanks for the hint!

msvechla: works great, I updated the PR

crazy-max:
I don't think it will work as intended because you need a buildkit container to allow building against other platforms (through [`docker-container`](https://github.com/docker/buildx/blob/master/docs/reference/buildx_create.md#driver) driver).

Dockerfile
crazy-max: You can also use our [buildx-bin](https://github.com/docker/buildx#from-dockerfile) image to add buildx to your image: https://github.com/docker/buildx#from-dockerfile

msvechla:
Yes, this is not the only configuration required, but this is the only configuration required from the `goreleaser` docker image. You can see my successful pipeline run here: https://gitlab.com/msvechla/vaultbot/-/jobs/1399876389 

There I use my custom image of `goreleaser` with these changes applied.
I also had to do some other things to prepare the linked service DinD environment, but it is working.

This is the `.gitlab-ci.yml` file I am using:

```release:
  stage: release
  image: docker:stable
  before_script:
    - mkdir -p $HOME/.docker/cli-plugins/
    - wget -O $HOME/.docker/cli-plugins/docker-buildx $BUILDX_URL
    - chmod a+x $HOME/.docker/cli-plugins/docker-buildx
    - "echo -e '{\n  \"experimental\": \"enabled\"\n}' | tee $HOME/.docker/config.json"
    - docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
    - docker buildx create --use --driver docker-container --name multiarch --platform=${BUILDX_PLATFORM}
  services:
    - docker:dind

  variables:
    # Disable shallow cloning so that goreleaser can diff between tags to
    # generate a changelog.
    GIT_DEPTH: 0

  script: |
    rm -f gl-code-quality-report.json
    docker run --rm --privileged \
      -v $PWD:/go/src/gitlab.com/msvechla/vaultbot \
      -w /go/src/gitlab.com/msvechla/vaultbot \
      -v /var/run/docker.sock:/var/run/docker.sock \
      -e DOCKER_USERNAME -e DOCKER_PASSWORD -e DOCKER_REGISTRY  \
      -e GITLAB_TOKEN \
      msvechla/goreleaser:latest release --rm-dist
  only:
    - tags
```

caarlos0:
I think you are probably using the host as the DOCKER_HOST, so probably it has the builder stuff ready.

I believe this should be fine.

Thanks for the PR! 🙏

caarlos0 (approved):

Issue #2334, opened 2021-07-04
Gitlab Releases broken


msvechla:
**Describe the bug**

Gitlab releases are broken for gitlab.com:

```
 • github/gitlab/gitea releases
         • token type                type=gitlab
         • creating or updating release repo= tag=v1.12.0-rc1
         • projectID                 name=5174485 owner= projectID=5174485
   ⨯ release failed after 4.14s error=github/gitlab/gitea releases: failed to publish artifacts: GET https://gitlab.com/api/v4/projects/5174485/releases/v1.12.0-rc1: 404 {message: 404 Not Found}
```

The release is not there, as it is expected, because goreleaser should create it.

**To Reproduce**

```yaml
project_name: vaultbot
release:
  gitlab:
    owner: ""
    name: "{{.Env.CI_PROJECT_ID}}"
```

**Expected behavior**

A new Gitlab release is created successfully.

**Environment (please complete the following information):**

- OS: mac
- OS version: Darwin C02Z1054LVDQ 20.5.0 Darwin Kernel Version 20.5.0: Sat May  8 05:10:33 PDT 2021; root:xnu-7195.121.3~9/RELEASE_X86_64 x86_64
- GoReleaser Version: 0.173.1
- GitLab version: gitlab.com

When performing the same request against the URL we can see that the Gitlab API returns `404`:

```
❯ curl --header "PRIVATE-TOKEN: $GITLAB_COM_TOKEN"  https://gitlab.com/api/v4/projects/5174485/releases/v1.12.0-rc1
{"message":"404 Not Found"}
```

It looks like the code expects a `403` for non-existent releases: https://github.com/goreleaser/goreleaser/blame/master/internal/client/gitlab.go#L211

Maybe there was an API change here, not sure. I can create a PR and add the `404` to the condition as well.

PR #2335, opened 2021-07-04
fix: new gitlab releases throwing 404 due to API change #2334

msvechla:
This commit fixes releases on gitlab.com.
Currently, new releases are not working, because the API returns `404` for non-existing releases, which causes goreleaser to throw an error and fail.

See also #2334 for more info.

Let me know what you think, thanks a lot!

### DIFF
--- a/internal/client/gitlab.go
+++ b/internal/client/gitlab.go
@@ -208,11 +208,11 @@ func (c *gitlabClient) CreateRelease(ctx *context.Context, body string) (release
 	name := title
 	tagName := ctx.Git.CurrentTag
 	release, resp, err := c.client.Releases.GetRelease(projectID, tagName)
-	if err != nil && (resp == nil || resp.StatusCode != 403) {
+	if err != nil && (resp != nil && resp.StatusCode != 403 && resp.StatusCode != 404) {
 		return "", err
 	}
 
-	if resp.StatusCode == 403 {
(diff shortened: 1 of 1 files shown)


internal/client/gitlab.go
```diff
@@ -208,11 +208,11 @@ func (c *gitlabClient) CreateRelease(ctx *context.Context, body string) (release
 	name := title
 	tagName := ctx.Git.CurrentTag
 	release, resp, err := c.client.Releases.GetRelease(projectID, tagName)
-	if err != nil && (resp == nil || resp.StatusCode != 403) {
+	if err != nil && (resp.StatusCode != 403 && resp.StatusCode != 404) {
```
caarlos0, line 211: why remove the `resp == nil` check?

msvechla: I removed it, because the `resp` is now `nil` when no release is present. So the error condition always triggered. If  we want to keep the `nil` check, we will have to change the conditional logic a bit more

msvechla: Never mind, my bad. I added back the `nil` check, definitely makes sense.

caarlos0:
LGTM, lets just see if @mavogel has any objections

mavogel (approved):
Thanks for catching this one!

PR #2347, opened 2021-07-18
feat: file mappings in archives

### LINKED ISSUE #1638: enable file mapping in archives

caarlos0:
at last, closes #1638

I think I got to something that both does what everyone wants (set destination, mode, owner, etc) and keeps compatibility with the previous behavior.

Added a lot of tests, so it should be "easy" to review what happens in which case... let me know if you think I missed any possible relevant case.

Also, sorry, the PR is a bit bigger than I though, but most of it is tests so I guess thats good.

### Usage

**Adds `README.md` at the root of the archive:**

```yaml
 - README.md
```

**Adds all `md` files to the root of the archive:**

```yaml
 - '*.md'
```

**Adds all `md` files to the root of the archive:**

```yaml
 - src: '*.md'
```

**Adds all `md` files in the current folder to a `docs` folder in the archive:**
```yaml
 - src: '*.md'
   dst: docs
```

**Recursively adds all `go` files to a `source` folder in the archive:**

```yaml
 # in this case, cmd/myapp/main.go will be added as source/cmd/myapp/main.go
 - src: '**/*.go'
   dst: source
```

**Recursively adds all `go` files to a `source` folder in the archive, stripping their parent folder:**

```yaml
 # in this case, cmd/myapp/main.go will be added as source/main.go
 - src: '**/*.go'
   dst: source
   strip_parent: true
 ```

**Set file info:**

```yaml
 - src: '**/*.go'
   dst: source
   info:
     owner: root
     group: root
     mode: 0755
     mtime: 2008-01-02T15:04:05Z  
 ```


ReddicoAinsley:
Amazing, thank you @caarlos0

caarlos0:
probably later today, worst case scenario: this weekend

PR #2350, opened 2021-07-19
fix: gitignore patching needs leading newline (#2349)

### LINKED ISSUE #2349: [Bug]:  goreleaser init breaks .gitignores without trailing newline
### What happened?

My repo's `.gitignore` did not have a trailing newline (which is a valid `.gitignore`)
Ran `goreleaser init` on the repo, which tried do add `dist/` to my `.gitignore` but without a prefixed newline `\n`
This broke my current `.gitignore'  and didn't achieve `goreleaser`'s goal of ignoring `dist/`

```
$ mkdir test
$ cd test
$ git init
$ echo -n "foo" > .gitignore
$ goreleaser init
$ cat .gitignoreecho -n "foo" 
foodist/
```

### goreleaser version

```bash
goreleaser version 0.173.2
commit: 5bcfa9042d0efd419b660a3b63c815a13d919e30
built at: 2021-07-06T08:42:56Z
built by: goreleaser
module version: v0.173.2, checksum: h1:LhajqMn6G6kaXCdAWGn5cHwL1mCbhvX4QlHDL9O33Dc=
```

### GoReleaser Check

### Search

### Code of Conduct

### Additional context

Fix is simply adding a leading newline here:
https://github.com/goreleaser/goreleaser/blob/df1ea4851ad1ea5afb323c1213ca6648ef2637f5/cmd/init.go#L43

Making a PR for it.

neomantra:
If applied, this commit will inject an extra leading newline when patching `.gitignore`, so that it works properly on existing `.gitignore` files that do not have trailing newlines (#2349)

**Why is this change being made?**

Per #2349, the current way breaks `.gitignore` files without a trailing newline.

closes #2349

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -40,7 +40,7 @@ func newInitCmd() *initCmd {
 				return err
 			}
 			defer gitignore.Close()
-			if _, err := gitignore.WriteString("dist/\n"); err != nil {
+			if _, err := gitignore.WriteString("\ndist/\n"); err != nil {
 				return err
 			}
 
--- a/cmd/init_test.go
+++ b/cmd/init_test.go
@@ -32,7 +32,7 @@ func TestInitGitIgnoreExists(t *testing.T) {
 
 	bts, err := os.ReadFile(".gitignore")
 	require.NoError(t, err)
-	require.Equal(t, "mybinary\ndist/\n", string(bts))
+	require.Equal(t, "mybinary\n\ndist/\n", string(bts))
 }
 
 func TestInitFileExists(t *testing.T) {


cmd/init.go
```diff
@@ -40,7 +40,7 @@ func newInitCmd() *initCmd {
 				return err
 			}
 			defer gitignore.Close()
-			if _, err := gitignore.WriteString("dist/\n"); err != nil {
+			if _, err := gitignore.WriteString("\ndist/\n"); err != nil {
```
caarlos0, line 43: I don't think we need the `\n` in the beginning 🤔

caarlos0:
this looks fine, but can you explain how not having a `\n` in the end of `.gitignore` breaks it?

neomantra:
By break, I mean that if my original `.gitignore` had:
```
foo<no-line-break>
```
After `goreleaser init`, it would say:
```
foodist/
```
Which would no longer ignore the `foo` directory, breaking the expected `.gitignore` setup.  And it also wouldn't ignore `dist/`.

caarlos0:
oh, gotcha... yah, that makes sense.

Thanks for the pr!

caarlos0 (approved):

Issue #2355, opened 2021-07-20
goreleaser test


caarlos0:
Would be good to have a `goreleaser test` command that does the following:

- check config file for deprecations
- skip publish
- if using gomod proxy feature, skip the proxying
- skips tag validation

---

The idea came while working with @myitcv  on https://github.com/cue-lang/cue/pull/1105

myitcv:
Reflecting on my experience with `goreleaser` to date (which is very limited!), I have one thought/idea that might be considered for this feature.

I would very much like to see exactly how assets etc will be named. It is a crucial aspect of the release assets that get posted to GitHub that they are named correctly. Consider release `v0.4.0` of CUE for example:

https://github.com/cue-lang/cue/releases/tag/v0.4.0

The assets that are created also have `v0.4.0` in the name, e.g. `cue_v0.4.0_darwin_amd64.tar.gz`.

I can't see via the existing `goreleaser release --rm-dist --snapshot` testing mechanism how to "fake" the version that will actually get released. Instead, assets get created with what appears to be some variation of local tags associated with the commit. 

So being able to supply a the "version" in this test flow would be very useful.

(apologies if there is a way of doing this that I have missed!)

caarlos0:
maybe we could revive https://github.com/goreleaser/goreleaser/pull/1457 and then change the default `snapshot.name_template` from `{{ .Tag }}-SNAPSHOT-{{.ShortCommit}}` to `{{ incpatch .Tag }}`. 🤔

myitcv:
Another drive-by thought, that builds on the idea of being able to "inject" the version (https://github.com/goreleaser/goreleaser/issues/2355#issuecomment-884107609): have the output of `goreleaser test` be stable, so that it can be compared against a golden file. That way, we can add `goreleaser test` to our CI, and assert that its output will be a known thing.

caarlos0:
Another possible idea of implementation, which I think lives here as well: IMHO, the `test` command should be able to check the environment variables as well.

E.g., I use an `{{ .Env.FOO_KEY }}` somewhere, if I run `goreleaser test` without setting it, it should exit 1 telling me I'm missing that. 

At least for me, most of my mistakes consist of forgetting to wire some environment variable in some workflow somewhere, so having that would be amazing :)

myitcv:
> At least for me, most of my mistakes consist of forgetting to wire some environment variable in some workflow somewhere, so having that would be amazing :)

[...]

PR #1457, opened 2020-04-20
feat: add incpatch/minor/major to templates

caarlos0:
- added inc template functions for semantic versions

refs #1455

PR #2358, opened 2021-07-22
feat: inc parts of semver, change snapshot default name

caarlos0:
revives https://github.com/goreleaser/goreleaser/pull/1457 adding some new template functions:

- incpatch
- incminor
- incmajor

and change the `goreleaser init`  generated config snapshot name to `{{ incpatch .Tag }}-next`

refs https://github.com/goreleaser/goreleaser/issues/2355

### DIFF
--- a/internal/pipe/git/git.go
+++ b/internal/pipe/git/git.go
@@ -32,9 +32,7 @@ func (Pipe) Run(ctx *context.Context) error {
 		return err
 	}
 	ctx.Git = info
-	if !ctx.Snapshot {
-		log.Infof("releasing %s, commit %s", info.CurrentTag, info.Commit)
-	}
+	log.Infof("building commit=%s, latest tag=", info.Commit, info.CurrentTag)
 	ctx.Version = strings.TrimPrefix(ctx.Git.CurrentTag, "v")
 	return validate(ctx)
 }
--- a/internal/pipe/git/git.go
+++ b/internal/pipe/git/git.go
@@ -32,7 +32,7 @@ func (Pipe) Run(ctx *context.Context) error {
 		return err
 	}
 	ctx.Git = info
-	log.Infof("building commit=%s, latest tag=", info.Commit, info.CurrentTag)
+	log.WithField("commit", info.Commit).WithField("latest tag", info.CurrentTag).Info("building...")
 	ctx.Version = strings.TrimPrefix(ctx.Git.CurrentTag, "v")
 	return validate(ctx)
 }
--- a/internal/pipe/git/git.go
+++ b/internal/pipe/git/git.go
@@ -32,6 +32,9 @@ func (Pipe) Run(ctx *context.Context) error {
 		return err
 	}
 	ctx.Git = info
+	if !ctx.Snapshot {
+		log.Infof("releasing %s, commit %s", info.CurrentTag, info.Commit)
+	}
 	ctx.Version = strings.TrimPrefix(ctx.Git.CurrentTag, "v")
 	return validate(ctx)
 }
--- a/internal/pipe/git/git.go
+++ b/internal/pipe/git/git.go
@@ -32,7 +32,6 @@ func (Pipe) Run(ctx *context.Context) error {
 		return err
 	}
 	ctx.Git = info
-	log.Infof("releasing %s, commit %s", info.CurrentTag, info.Commit)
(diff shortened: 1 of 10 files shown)


internal/pipe/git/git.go
```diff
@@ -32,7 +32,9 @@ func (Pipe) Run(ctx *context.Context) error {
 		return err
 	}
 	ctx.Git = info
-	log.Infof("releasing %s, commit %s", info.CurrentTag, info.Commit)
+	if !ctx.Snapshot {
+		log.Infof("releasing %s, commit %s", info.CurrentTag, info.Commit)
```
radeksimko, line 36: I agree that _releasing_ is not relevant here, but wouldn't it be still useful to output the commit SHA as a confirmation of what exact SHA did GoReleaser pick up?

radeksimko (commented):
I relate to @myitcv in #2355 in the sense that the current name with `major.minor.patch-SNAPSHOT-commit` is not very conventional but I still believe that there's value in it being different from a regular "stable" release. This PR would make the two indistinguishable which I would be unfortunate IMO.

How about changing it something like `major.minor.patch-prerelease.commit` or ``major.minor.patch-prerelease` or anything similar that follows semver more closely?

https://semver.org/#spec-item-9

> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version.

radeksimko:
Apologies for bit of a trigger happiness with the last comment, after re-reading the discussion in https://github.com/goreleaser/goreleaser/issues/2355 I suppose that the problem you're all trying to solve is predictability of the name of the resulting artifact.

I obviously can't speak for any majority of users, but from my experience this has not been a common concern for us at HashiCorp. I understand the CI-testability use case, but a real concern I would have if snapshots were named the same as the upcoming version is that someone will accidentally publish that as a "stable" version and then override with a real stable release later.

Having the commit SHA there is from my experience valuable to make releasing more error-prone and to easily set expectations for end-users _by default_.

I suppose that one group of users will have to end up specifying this manually. I would personally be more inclined to keep the _default_ options _safer_.

https://goreleaser.com/customization/snapshots/

myitcv:
@radeksimko I'm very inexperienced with `goreleaser`, so my comments should be taken with a pinch of salt!

> but a real concern I would have if snapshots were named the same as the upcoming version is that someone will accidentally publish that as a "stable" version and then override with a real stable release later.

Please can you explain this point a bit further? Because it's entirely possible I'm not entirely following, or I have a different mental model. I expect `goreleaser test` to, by default, leave no trace of what it did locally. No files, no docker images, nothing. It goes without saying this mode would also not upload/push anything, ever. For those people who do want to test a docker image, perhaps a flag `--nocleanup` could be provided... but at this point I'm getting out of my depth!

My desire to have `goreleaser test` accept a version string is that it allows us to simulate the entire release process without actually mutating any remote state (GitHub release assets, docker images etc). The stable output of `goreleaser test` would allow us to affirm what _would_ happen if we were to run things for real. Because without that, the only time we can truly test what will be published (in terms of asset names) is when it comes to cutting a release. That just feels fragile, especially if you consider `goreleaser` updates along the way (which will hopefully be stable!)

But again, I really defer to those with greater experience of `goreleaser`. Please consider my experience reports as just that: I'm very happy being guided to doing things a different way, just so long as it gives us comfort with our `goreleaser` setup.

radeksimko:
> I expect goreleaser test to, by default, leave no trace of what it did locally. No files, no docker images, nothing. It goes without saying this mode would also not upload/push anything, ever.

That more or less aligns with my thinking but my this PR would change behaviour of the `goreleaser build --snapshot` and `goreleaser release --snapshot` commands. I would be more than happy for a `test` command to behave as you described.

I do believe the `test` command should perhaps use different (default?) naming convention for the snapshot as it has different use cases.

caarlos0:
Yeah, I think that changing the default here is indeed no good.

Maybe we could instead have a different field for the actual `test` command later, which will actually be something @myitcv is interested in as well, e.g. `test.tag_template: v99.9.9` (which by default could be `{{ inpatch .Tag }}`).

wdyt?

caarlos0:
should be better now.

I did change the default on the config generated by `goreleaser init`, which shiould not a breaking change, but will be a bit better than the previous value `TAG-next`...

radeksimko (approved):
LGTM 🚢

PR #2361, opened 2021-07-23
feat(snapcraft): manage channel upload

guilhem:
By default, snap store manage "edge", "beta", "candidate" and "stable".
Default to all this channels when package is in grade stable. But only on
"edge" and "beta" for grade "devel".

This fix this TODO: https://github.com/goreleaser/goreleaser/blob/bf19dc1079887d31e58065771bb069953b1704d6/internal/pipe/snapcraft/snapcraft.go#L342

### DIFF
--- a/internal/pipe/snapcraft/snapcraft.go
+++ b/internal/pipe/snapcraft/snapcraft.go
@@ -4,6 +4,7 @@ package snapcraft
 import (
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -21,6 +22,8 @@ import (
 	"github.com/goreleaser/goreleaser/pkg/config"
 	"github.com/goreleaser/goreleaser/pkg/context"
 )
+
+const releasesExtra = "releases"
 
 // ErrNoSnapcraft is shown when snapcraft cannot be found in $PATH.
 var ErrNoSnapcraft = errors.New("snapcraft not present in $PATH")
@@ -79,6 +82,14 @@ func (Pipe) Default(ctx *context.Context) error {
 		snap := &ctx.Config.Snapcrafts[i]
 		if snap.NameTemplate == "" {
 			snap.NameTemplate = defaultNameTemplate
+		}
+		if len(snap.ChannelTemplates) == 0 {
+			switch snap.Grade {
+			case "devel":
+				snap.ChannelTemplates = []string{"edge", "beta"}
+			default:
+				snap.ChannelTemplates = []string{"edge", "beta", "candidate", "stable"}
+			}
 		}
 		if len(snap.Builds) == 0 {
 			for _, b := range ctx.Config.Builds {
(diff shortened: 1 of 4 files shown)


internal/pipe/snapcraft/snapcraft.go
```diff
@@ -373,9 +384,31 @@ func link(src, dest string, mode os.FileMode) error {
...
+func copyFile(src, dst string, mode os.FileMode) error {
+	// Open original file
+	original, err := os.Open(src)
+	if err != nil {
+		return fmt.Errorf("fail to open '%s': %w", src, err)
+	}
+	defer original.Close()
+
+	// Create new file
+	new, err := os.OpenFile(dst, os.O_CREATE|os.O_WRONLY, mode)
```
caarlos0, line 403: should probably truncate as well?

internal/pipe/snapcraft/snapcraft.go
```diff
@@ -373,9 +384,31 @@ func link(src, dest string, mode os.FileMode) error {
...
-			return err
+		if err := copyFile(path, dst, mode); err != nil {
+			return fmt.Errorf("fail copy file '%s': %w", path, err)
 		}
-		return os.Chmod(dst, mode)
+		return nil
 	})
 }
+
+func copyFile(src, dst string, mode os.FileMode) error {
```
caarlos0, line 394: curious why this change? did you find any issues with the hard linking one?

www/docs/customization/snapcraft.md
```diff
@@ -60,12 +60,21 @@ snapcrafts:
     # store.
     description: This is the best drum roll application out there. Install it and awe!
 
+    # Channels in store where snap will be pushed.
+    # Default depends on grade.
```
caarlos0, line 64: probably worth saying here which channels are default for which grades?

caarlos0:
Thanks for the PR! Just a couple of comments :)

internal/pipe/snapcraft/snapcraft.go
guilhem: Yes, hard link work only if both files are on the same FS.
My `/tmp` is managed outside, so tests are failing with:
```
failed to link completer: link testdata/foo-completer.bash /tmp/TestCompleter1448278326/001/dist/foo_arm6/prime/testdata/foo-completer.bash: invalid cross-device link
```
Generally, hardlinks are a bad idea.
I tried to use Symlink but `chmod` is inherited and failed.

www/docs/customization/snapcraft.md
guilhem: Grade say something about it, but we may be more explicit indeed :)

internal/pipe/snapcraft/snapcraft.go
guilhem: nice shot ;)
Updating it

guilhem:
@caarlos0 I updated to use a template for channels

caarlos0 (approved):
lgtm, thanks for the PR! 🤘

PR #2362, opened 2021-07-24
fix: copying file on docker pkg instead of hard linking

caarlos0:
since we are now copying instead of hard-linking the files on the snapcraft pipe, let's do the same for the docker pipe.

This also means we can deduplicate some code!

refs #2358

### DIFF
--- a/internal/pipe/snapcraft/snapcraft.go
+++ b/internal/pipe/snapcraft/snapcraft.go
@@ -4,7 +4,6 @@ package snapcraft
 import (
 	"errors"
 	"fmt"
-	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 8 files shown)


internal/pipe/snapcraft/snapcraft.go
```diff
@@ -433,9 +385,5 @@ func processChannelsTemplates(ctx *context.Context, snap config.Snapcraft) ([]st
 		channels = append(channels, channel)
 	}
 
-	if len(channels) == 0 {
```
caarlos0, line 436: this will never be true because of the defaults.

caarlos0:
@goreleaser/goreleaser let me know if you see any problems with this, tests apparently are passing 🙏

PR #2372, opened 2021-07-27
feat: skip push docker prerelease or stable

vandot:
Replace `skip_push` `auto` flag  in `docker` and `docker_manifest` with `prerelease` and `stable` flags.

I have a need to skip push when either it is `prerelease` or not, this is the case
```
- name_template: repo/image:{{ .Major }}.{{ .Minor }}.{{ .Patch }}
  image_templates:
  - repo/image:{{ .Version }}-amd64
  - repo/image:{{ .Version }}-armv7
  - repo/image:{{ .Version }}-arm64v8
  skip_push: prerelease
- name_template: repo/image:{{ .Major }}.{{ .Minor }}.{{ .Patch }}-{{ .Prerelease }}
  image_templates:
  - repo/image:{{ .Version }}-amd64
  - repo/image:{{ .Version }}-armv7
  - repo/image:{{ .Version }}-arm64v8
  skip_push: stable
  ```
First one should be pushed if its stable release and skipped if its a prerelease, other one vice versa.

Maybe I missed something and this can be accomplished in some other way?

### DIFF
--- a/internal/pipe/docker/docker.go
+++ b/internal/pipe/docker/docker.go
@@ -185,8 +185,11 @@ func process(ctx *context.Context, docker config.Docker, artifacts []*artifact.A
 	if ctx.SkipPublish {
 		return pipe.ErrSkipPublishEnabled
 	}
-	if strings.TrimSpace(docker.SkipPush) == "auto" && ctx.Semver.Prerelease != "" {
-		return pipe.Skip("prerelease detected with 'auto' push, skipping docker publish")
+	if strings.TrimSpace(docker.SkipPush) == "prerelease" && ctx.Semver.Prerelease != "" {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by caarlos0** [comment]
assuming the dockerfile is the same, you could instead:

```
name_template: repo/image:{{ .Major }}.{{ .Minor }}.{{ .Patch }}{{ with  .Prerelease }}-{{ . }}{{ end }}
```

and that would have the same effect.

does that solves your case?

vandot:
There is no :exploding_head: for a reaction :smile: 
Thanks, that solves our case perfectly.
I will close the PR

caarlos0:
Glad to hear it 🤘

PR #2373, opened 2021-07-30
fix: upload gcs blobs to bucket root

abeMedia:
Currently, when trying to upload files to the root of a gcs bucket using the config below it actually creates a directory called `/` with all the files inside it.

```yml
blobs:
  - provider: gs
    bucket: my-bucket-name
    ids:
      - myapp
    folder: '/'
```      

This fix allows uploading to the root of a gcs bucket.

### DIFF
--- a/internal/pipe/blob/upload.go
+++ b/internal/pipe/blob/upload.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"strings"
 
 	"github.com/apex/log"
 	"github.com/goreleaser/goreleaser/internal/artifact"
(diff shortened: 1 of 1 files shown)


caarlos0:
I think you could just not inform anything on `folder`? Its the same behavior on other places AFAIK 🤔

abeMedia:
If `folder` is an empty string it goes to `{{.PackageName}}/{{.Version}}`.

caarlos0:
ah, makes sense... can you add a test for it as well?

thanks!

abeMedia:
Sure thing!

PR #2404, opened 2021-08-16
fix: Enable darwin/arm64 builds for go1.17 and later

jorng:
This refactors the check for disabling `darwin/arm64` builds on Go versions less than `go1.16`. Previously, it would only match on the substring `go1.16`, which limits this to only work with `go1.16.*` (not anything later).

This is needed because `arm64` builds will start breaking with go1.17.

The new `isGo116orLater` function now verifies the go version matches based on a regex (this uses a regex due to the _slightly_ unconventional style of go version strings). The regex captures the version digits, and verifies that the major version is >=`1` and the minor version is >=`16` (if major version == `1`).

Additionally, the new `isGo116orLater` function also now runs the GoBinary command from the `build.Dir`, matching how the rest of the build process works. This resolves issues with relative `gobinary` paths (i.e. wrapper scripts that may be checked into the current repository).

### DIFF
--- a/internal/builders/golang/targets.go
+++ b/internal/builders/golang/targets.go
@@ -1,9 +1,10 @@
 package golang
 
 import (
-	"bytes"
 	"fmt"
 	"os/exec"
+	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/apex/log"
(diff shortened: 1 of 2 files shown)


caarlos0:
this check can actually be removed and we can expire the deprecation notice: https://goreleaser.com/deprecations/#builds-for-darwinarm64

jorng:
Ok. Interesting way to go.

Unfortunately that doesn't fix the other problem that this fixed (running `GoBinary` from `build.Dir`).

caarlos0:
hmm, correct, that may be a different issue, can you open a PR just for it?

jorng:
#2411

PR #2411, opened 2021-08-17
fix: Run 'go version' command in build.Dir

jorng:
Currently, the `goVersion` function is run without any directory set. This is problematic when a build uses the `dir` config in combination with the `gobinary` config. When `gobinary` is a relative path (like a script in the current repository), goVersion silently fails, returning an empty string.

This commit refactors `goVersion` to execute the command with `Dir` set to the `build.Dir`, in addition to now returning an error, so that issues may be bubbled up in the build log, rather than silently failing.

It also adds a new helper function to facilitate running `goVersion` by creating a temporary executable that outputs a given `version` string. This function is only currently used by `TestWithDefaults`.


jorng:
Pushed a fix for the failing test.

caarlos0:
thanks @jorng !

Issue #2421, opened 2021-08-23
Allow overriding of `version` that goreleaser grabs from the git environment


aaomidi:
I would like to specify a `version` for my goreleaser builds. Unfortunately not all git packages follow semver versioning, which means the packages generated by `goreleaser` might not even be installable with apt/yum/etc due to the version name.

For example:
```
$ sudo dpkg -i dist/contact-auditor_release-2021-08-16_linux_amd64.deb 
dpkg: error processing archive dist/contact-auditor_release-2021-08-16_linux_amd64.deb (--install):
 parsing file '/var/lib/dpkg/tmp.ci/control' near line 2 package 'contact-auditor':
 'Version' field value 'release-2021-08-16': version number does not start with digit
Errors were encountered while processing:
 dist/contact-auditor_release-2021-08-16_linux_amd64.deb
 ```
 
 So what I'm asking for would look like this in the configuration file:

```
version: {{ .ENV.SOMEVARIABLE }}
builds:    
  - id: contact-auditor
    main: ./cmd/contact-auditor
    binary: contact-auditor
    goarch:
        - amd64
    env:
        - CGO_ENABLED=0
    goos:
        - linux

nfpms:
  - id: contact-auditor-package
    package_name: contact-auditor
    builds:
      - contact-auditor
    formats:
      - rpm
      - deb
    bindir: /opt/boulder/bin

```

caarlos0:
You can override the `ldflags` in the build section and set the version to anything there https://goreleaser.com/customization/build/

On the nfpms side,unfortunatelly the `version` field is not exposed.

But, in any case, goreleaser is meant to work only with semantic versions, not sure how you even got this far - https://goreleaser.com/limitations/semver/

aaomidi:
I've made the necessary changes on a cloned version of `goreleaser`, I assume there would be no interest in taking this in as a PR?

caarlos0:
as pointed out on #2422, rather not do this right now. Closing, thanks for the issue though 🙏

Issue #413, opened 2017-11-07
Version override without relying on GIT tag


tj:
Hey!

My use-case is a little strange, but I need to release Up's Pro binaries to my own release storage on S3, so I'm doing:

```
$ goreleaser -p 1 --rm-dist -config .goreleaser.yml --skip-publish
```

But the since I may have `v0.1.0` for the OSS version, and `v0.1.0` in the "pro" branch, they obviously conflict so I can't version. I'm not sure what the best solution would be, maybe to support tag prefixes like `pro-v1.0.0`, or `--version v1.0.0` to specify this at build time.

Any suggestions?

thanks man!

caarlos0:
Hi @tj 

Yeah, I haven't predicted this kind of use case... let me think about this for a while :P

tj:
I'll think on it too hahah. Supporting a prefix miiiight be the simplest, `--skip-publish` is fine for my case, I have a little tool for uploading to my system already so that part isn't a big deal.

Happy to figure out a PR if that sounds reasonable

caarlos0:
@tj yeah, if I understood it right, this seems like the simplest way...

would be awesome if you could provide a patch (you can also "test it in production" in your codebase hehe) 

let me know if you need help with anything :P

tj:
ah weird I think this is a non-issue actually. I think I had an older build of goreleaser which sorted the versions so 0.3.0 would be > 0.1.0-pro and it would error thinking that commit was not v0.3.0. Seems ok now!

chnrxn:
I would also like to have more flexibility on the version number without relying only on a git tag.

Either a tag prefix or a flag would also solve it for me. Any chance of this happening?

PR #2422, opened 2021-08-23
Version override

aaomidi:
#2421, #413

Remaining:
- [ ] documentation updates


caarlos0:
I'm not sure of the side effects of this, but my main concern is that you are releasing one thing (the tag with the date on it) and users will see other thing unrelated to that (the semver)...

maybe its best to use semver? 🤔

caarlos0:
Rather not merge this in right now. Thanks for the PR though.

PR #2435, opened 2021-08-27
fix: build dir that dont exist yet

### LINKED ISSUE #2434: [Bug]: premature `go version` validation breaks goreleaser flows
### What happened?

The commit https://github.com/goreleaser/goreleaser/commit/013bd69126459125694d7cb2c434dd9ba63e5a5b breaks https://github.com/caddyserver/caddy goreleaser's flow. One of the steps in the `before.hooks` is to create the directory to be used as the build dir later on. The referenced commit validates the `go version` before running any of the `before.hooks`, but it also tries to switch into the build dir before it's created because it doesn't get created until the `before.hook` runs.

Have the `builds.dir` specified to a directory which gets created as part of the `before.hooks`, then run `goreleaser check` or `build` or `release`. The run will fail before running any of the before hooks.

Caddy's `.goreleaser.yml` file:
https://github.com/caddyserver/caddy/blob/84b906a248a7cf7299f39f2f997822cb38387759/.goreleaser.yml#L1-L25

Failing run:
https://github.com/caddyserver/caddy/runs/3424377795?check_suite_focus=true#step:3:24

### goreleaser version

```bash
goreleaser version 0.176.0
commit: dd5ccf7170741e0303e90e81faf9073d03ac8f43
built at: 2021-08-24T23:49:11Z
built by: goreleaser
module version: v0.176.0, checksum: h1:z3hvWAsJCTWS2fmfpbhWYEtIZuxWiwDcq22R2KUZcV8=
```

### GoReleaser Check

### Search

### Code of Conduct

### Additional context

The workaround (for us) is to create the directory before invoking goreleaser. This isn't nice because the task ends up being duplicated across 2 flows: `goreleaser check` in our lint run, and the plain `goreleaser` release flow.

About the `goreleaser check shows no errors`, although I checked that box, the issue is with the command itself failing prematurely.

caarlos0:
if the `build.dir` doesn't exist, ignore it while checking the go version.

IMHO, its unlikely that the before hook of a build will create a folder and set a go wrapper in there... 

For now I think we can just ignore the dir if we fail to stat it.

cc/ @jorng  

closes #2434

### DIFF
--- a/internal/builders/golang/targets.go
+++ b/internal/builders/golang/targets.go
@@ -2,6 +2,7 @@ package golang
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 	"regexp"
 	"strings"
@@ -133,7 +134,11 @@ var (
 
 func goVersion(build config.Build) ([]byte, error) {
 	cmd := exec.Command(build.GoBinary, "version")
-	cmd.Dir = build.Dir // Set Dir to build directory in case of reletive path to GoBinary
+	// If the build.Dir is acessible, set the cmd dir to it in case
+	// of reletive path to GoBinary
+	if _, err := os.Stat(build.Dir); err == nil {
+		cmd.Dir = build.Dir
+	}
 	bts, err := cmd.CombinedOutput()
 	if err != nil {
 		return nil, fmt.Errorf("unable to determine version of go binary (%s): %w", build.GoBinary, err)
--- a/internal/builders/golang/build_test.go
+++ b/internal/builders/golang/build_test.go
@@ -88,6 +88,36 @@ func TestWithDefaults(t *testing.T) {
 			},
 			goBinary: "go",
 		},
+		"empty with custom dir": {
+			build: config.Build{
+				ID:     "foo2",
+				Binary: "foo",
+				Dir:    "./testdata",
+			},
+			targets: []string{
+				"linux_amd64",
+				"linux_386",
+				"linux_arm64",
+				"darwin_amd64",
+				"darwin_arm64",
+			},
+			goBinary: "go",
+		},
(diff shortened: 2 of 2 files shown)


jorng:
Makes sense. Setting Dir during the version check more specifically for the instance that both Dir and GoBinary are set, in case GoBinary is a relative path. 

That's how I hit the bug originally. I wrote a wrapper script to "natively" compile for `darwin/amd64` when running on `darwin/arm64` using Rosetta. The builds all failed when I had Dir+GoBinary set.

internal/builders/golang/targets.go
```diff
@@ -133,7 +134,11 @@ var (
 
 func goVersion(build config.Build) ([]byte, error) {
 	cmd := exec.Command(build.GoBinary, "version")
-	cmd.Dir = build.Dir // Set Dir to build directory in case of reletive path to GoBinary
+	// If the build.Dir is acessible, set the cmd dir to it in case
+	// of reletive path to GoBinary
+	if _, err := os.Stat(build.Dir); err == nil {
```
jorng, line 139: A potentially faster way might be to check if `Dir` and `GoBinary` are filled out, and if `GoBinary` is relative. Since that is the only instance that we need to actually change directory when running GoBinary. Not 100% sure if the string operations would for sure faster than the syscall, it would need some testing.

Maybe something like:

```suggestion
	if build.Dir != "" && build.GoBinary != "go" && (!strings.HasPrefix(build.GoBinary, "/") && strings.Contains(build.GoBinary, "/")){
```

caarlos0: I think its ok as is...

PR #2465, opened 2021-09-07
feat: support templates for scm urls

steveazz:
Thank you for this really well polished project and the hard work put into it 🙌 

Background
---
When a git repository is hosted in multiple GitLab instances the
`.goreleaser.yml` needs to take in consideration both APIs endpoints. At
the moment it defaults to GitLab.com and you can override it with
`gitlab_urls` however this forces you to only support 1 GitLab instance.

We need this for
https://gitlab.com/gitlab-com/gl-infra/infrastructure/-/issues/14122
where we have a tool that is developed on GitLab.com but then mirrored
to an internal GitLab instance since we need it to operate GitLab.com
even when it's down.

Solution
---
Support templates like `{{ .Env.CI_SERVER_URL }}` for the
`gitlab_urls`, `github_urls`  and `gitea_urls` so it can use environment
variables and the same `.goreleaser` file can be used in multiple SCM
instances.

### DIFF
--- a/www/docs/scm/gitlab.md
+++ b/www/docs/scm/gitlab.md
@@ -18,8 +18,8 @@ env_files:
 
 ## GitLab Enterprise or private hosted
 
-You can use GoReleaser with GitLab Enterprise by providing its URLs in
-the `.goreleaser.yml` configuration file:
+You can use GoReleaser with GitLab Enterprise by providing its URLs in the
+`.goreleaser.yml` configuration file. This takes a normal string or a template value.
 
 ```yaml
 # .goreleaser.yml
--- a/internal/client/gitea.go
+++ b/internal/client/gitea.go
@@ -21,7 +21,12 @@ type giteaClient struct {
 	client *gitea.Client
 }
 
-func getInstanceURL(apiURL string) (string, error) {
+func getInstanceURL(ctx *context.Context) (string, error) {
+	apiURL, err := tmpl.New(ctx).Apply(ctx.Config.GiteaURLs.API)
+	if err != nil {
+		return "", fmt.Errorf("templating Gitea API URL: %w", err)
+	}
+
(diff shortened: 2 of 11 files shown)


caarlos0:
Hey! 

Thanks for the kind word and for the PR. 

Can you do the same for github and gitea? (just so we have feature parity)

Would also be neat to add a note the docs saying its templateable, but I can do that later if you like

Thanks!

www/docs/scm/gitlab.md
```diff
@@ -18,8 +18,8 @@ env_files:
 
 ## GitLab Enterprise or private hosted
 
-You can use GoReleaser with GitLab Enterprise by providing its URLs in
-the `.goreleaser.yml` configuration file:
+You can use GoReleaser with GitLab Enterprise by providing its URLs in the
+`.goreleaser.yml` configuration file. This takes a normal string or a template value.
 
 ```yaml
```
steveazz, line 24: **thought:** Not sure if we want to give a commented example here that it can support templates as well like below

```
# .goreleaser.yml
gitlab_urls:
  # api: "{{ .ENV. CI_API_V4_URL }}
  api: https://gitlab.mycompany.com/api/v4/
  download: https://gitlab.company.com
  # set to true if you use a self-signed certificate
  skip_tls_verify: false
```

steveazz:
@caarlos0 this should be ready for another round of review! Let me know what you think and I'm sorry it turned out to be a large PR 🙇

www/docs/scm/gitlab.md
caarlos0: no need IMHO, docs are clear enough 🙏

caarlos0 (approved):
LGTM

caarlos0 (approved):

Issue #2467, opened 2021-09-08
[Bug]: nil pointer reference in Gitlab client


ppx17:
### What happened?

While trying to run Goreleaser against a private Gitlab instance I get this panic when it attempts to create the release.

I think this is caused by 5bcfa9042d0efd419b660a3b63c815a13d919e30 where the resp == nil check is changed into a resp != nil check, causing the code to continue to line 215 when resp == nil, where it accesses a property on resp.

Stacktrace:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x947f75]
goroutine 9 [running]:
github.com/goreleaser/goreleaser/internal/client.(*gitlabClient).CreateRelease(0xc0005b03d0, 0xc0002a0000, {0xc000166070, 0x64})
	github.com/goreleaser/goreleaser@v0.177.0/internal/client/gitlab.go:215 +0x3b5
github.com/goreleaser/goreleaser/internal/pipe/release.doPublish(0xc0002a0000, {0x18836b8, 0xc0005b03d0})
	github.com/goreleaser/goreleaser@v0.177.0/internal/pipe/release/release.go:125 +0x271
github.com/goreleaser/goreleaser/internal/pipe/release.Pipe.Publish({}, 0xc0002a0000)
	github.com/goreleaser/goreleaser@v0.177.0/internal/pipe/release/release.go:111 +0x4c
github.com/goreleaser/goreleaser/internal/middleware.ErrHandler.func1(0x15f5e61)
	github.com/goreleaser/goreleaser@v0.177.0/internal/middleware/error.go:13 +0x22
github.com/goreleaser/goreleaser/internal/middleware.Logging.func1(0x15f5e61)
	github.com/goreleaser/goreleaser@v0.177.0/internal/middleware/logging.go:33 +0x142
github.com/goreleaser/goreleaser/internal/pipe/publish.Pipe.Run({}, 0x2)
	github.com/goreleaser/goreleaser@v0.177.0/internal/pipe/publish/publish.go:60 +0x109
github.com/goreleaser/goreleaser/internal/middleware.ErrHandler.func1(0x15d8fcd)
	github.com/goreleaser/goreleaser@v0.177.0/internal/middleware/error.go:13 +0x22
github.com/goreleaser/goreleaser/internal/middleware.Logging.func1(0x15d8fcd)
[...]

PR #2473, opened 2021-09-10
fix: panic on gitlab create release

steveazz:
Background
---
In 5bcfa9042d0efd419b660a3b63c815a13d919e30 the condition was changed to
that it can support the new status code returned by GitLab. However, the
condition is now checked `resp != nil` and not `resp == nil` which causes
the condition not to trigger when the resp is nil and only returns an
error. We can reproduce this issue by running
`TestGitLabCreateReleaseUknownHost` test added in this commit.

Solution
---
If we look at other error handling
https://github.com/goreleaser/goreleaser/blob/91a8abb2e6c5915d93866cb19be6b67db4a193f4/internal/client/gitlab.go#L106
we can see that we check if the `resp` is nil as well before continuing.

Added tests for the following use cases:
- When there is an error and no response was returned.
- When there is a 403 or 404 error so we expect to create a release first.
- When there is an unknown error (not 404, 403) we just bail.

### DIFF
--- a/internal/client/gitlab.go
+++ b/internal/client/gitlab.go
@@ -213,7 +213,7 @@ func (c *gitlabClient) CreateRelease(ctx *context.Context, body string) (release
 	name := title
 	tagName := ctx.Git.CurrentTag
 	release, resp, err := c.client.Releases.GetRelease(projectID, tagName)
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):

caarlos0:
Thanks!

caarlos0:
closes #2467

PR #2474, opened 2021-09-10
feat: support subgroups in GitLab

steveazz:
Background
---
GitLab supports
[subgroups](https://docs.gitlab.com/ee/user/group/subgroups/) so a
the repository can look something like
`gitlab.com/top-level-group/sub-group/repository` for example https://gitlab.com/gitlab-com/gl-infra/jsonnet-tool.

At the moment this detection doesn't work out of the box and requires
the user to set the
[`owner`](https://gitlab.com/gitlab-com/gl-infra/jsonnet-tool/-/blob/8b0781e5769d42d43d2cff5583d89d1fe4af10c5/.goreleaser.yml#L32-35)

---

Feel free to tell me it's out of scope to support such behavior since other SCMs don't support this feature!


caarlos0 (approved):
Thanks for the PR, and sorry for the delay reviewing it 🙏

PR #2480, opened 2021-09-12
feat: improve output and pipe skipping

caarlos0:
- pipes may optionally implement the new `Skipper` interface, if they do and the `Skip` method returns `true`, the entire pipe will be skipped and not even its name will show up on the output
  - this is also true for defaulters, publishers and announcers
- less log output also means its more likely that the user will see the deprecation notices and other warnings, as well as making debugging a bit easier
- pipes that implement skipper may also avoid repetitive checks like if its skip publish/announce is set and so on
- hopefully improve test coverage in some places
- left a dozen of todos for the future me
- the `dist/config.yaml` file should now be shorter by avoiding running defaults for features that will be skipped anyway - this should help when debugging user issues, hopefully
- improved brew and scoop run vs publish issue
  - on run, it only creates the formula/manifest in the `dist` folder, and add it as an artifact
  - later, if publishing is enabled, it uploads the files
- fixed some `String` impls here and there, and some godocs as well 

---

Its a very big PR, so sorry, but I got carried away... let me know if you have any questions.

---

If you can help testing it, it would be very much appreciated! I did test everything and test coverage is pretty good, but giving it a run on a project you have shouldn't hurt :) 

---

### Example

Before, with a config file like this:

```yaml
builds:
  - env:
      - CGO_ENABLED=0
    goos: [darwin]
    goarch: [amd64]
checksum:
snapshot:
  name_template: "{{ incpatch .Version }}-next"

```

You'll get an output like this:

image

Whereas after this patch, you'll get something like:

image


internal/middleware/skip/skip.go
```diff
@@ -0,0 +1,27 @@
...
+	Skip(ctx *context.Context) bool
+}
+
+// Maybe returns an action that skips immediately if the given p is a Skipper
+// and its Skip method returns true.
+func Maybe(skipper interface{}, next middleware.Action) middleware.Action {
+	if skipper, ok := skipper.(Skipper); ok {
+		return func(ctx *context.Context) error {
+			if skipper.Skip(ctx) {
+				return nil
```
caarlos0, line 21: TODO: add a debug log in here

PR #2489, opened 2021-09-15
docs: fix environment variable to be COSIGN_PWD

cpanato:
this is a small doc fix to update the documentation and use the  `COSIGN_PWD` instead of `GPG_PASSWORD` since it defaults to use `cosign`


caarlos0:
Thanks 🙏

PR #2501, opened 2021-09-22
feat(docker): install tini to handle Ctrl+C

troian:
sometimes releasing may take a while. when running it in docker `Ctrl-C` is not passed properly to the goreleaser process.

[why tini](https://github.com/krallin/tini/issues/8)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,9 +7,10 @@ RUN apk add --no-cache bash \
 	git \
 	mercurial \
 	make \
-	build-base
+	build-base \
+	tini
 
-ENTRYPOINT ["/entrypoint.sh"]
+ENTRYPOINT ["/sbin/tini", "--", "/entrypoint.sh"]
 CMD [ "-h" ]
 
 COPY scripts/entrypoint.sh /entrypoint.sh


caarlos0:
awesome, thanks!

caarlos0:
GoReleaser itself handles it fine... 

but I was just testing this patch, and still ctrl+c doesn't work... investigating why

troian:
hm, thats interesting. i just checked, works on mine setup.

@mislav check out description from tini author https://github.com/krallin/tini/issues/8#issuecomment-146135930
in short docker sends sigterm to PID 1 only which in this case is `/entrypoint.sh` script.

caarlos0:
oh, it works, I was forgetting `-ti` on my `docker run`

but it also works without tini... 

so.. seems like this is not needed 🤔

troian:
oh, i know where issue is. i forgot to add one more thing to entrypoint.sh. making pr

ps. it should work without `-ti`

troian:
folks, i'm really sorry for confusion i have created. let me put some more clarification into the issue.

> One problem is, if you run Bash as PID 1, then all signals you send to your Docker container (e.g. using docker stop or docker kill) end up sent to Bash, which does not forward them anywhere (unless you code it yourself). In other words, if you use Bash to run Goreleaser, and then run docker stop, then Goreleaser will never see the stop command!

in this case as goreleaser is running from within `entrypoint.sh` it does not receive signals

caarlos0:
oh, ok, got it... thanks @troian

PR #2510, opened 2021-09-26
feat: add SOURCE_DATE_EPOCH support for templates

shibumi:
This commit is going to add support for the SOURCE_DATE_EPOCH variable. This is important for reproducible builds.
During reproducible builds the build environment has to respect the standardized SOURCE_DATE_EPOCH variable for injecting a fake date

This change is being made for allowing goreleaser to build artifacts reproducible.

Relevant resources for this can be found here: https://reproducible-builds.org/docs/source-date-epoch/

EDIT: I am very confident that this is the wrong position for the code right now. It might make more sense to check the env variable when wrapping the additional context, right?

### DIFF
--- a/internal/tmpl/tmpl.go
+++ b/internal/tmpl/tmpl.go
@@ -4,8 +4,10 @@ package tmpl
 import (
 	"bytes"
 	"fmt"
+	"os"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 	"text/template"
 	"time"
@@ -67,6 +69,8 @@ const (
 func New(ctx *context.Context) *Template {
 	sv := ctx.Semver
 	rawVersionV := fmt.Sprintf("%d.%d.%d", sv.Major, sv.Minor, sv.Patch)
+	// call applySourceDateEpoch for respecting the SOURCE_DATE_EPOCH env variable
+	buildDate := applySourceDateEpoch()
 
 	return &Template{
 		fields: Fields{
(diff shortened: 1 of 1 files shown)


caarlos0:
hmm, that's not really needed I think... 

you can set `builds.mod_timestamp` to `{{ .Env.SOURCE_DATE_EPOCH }}` and that would be it 🤔 (also could set the custom ldflags)

for example, in this config I use the commit timestamp as date the date for everything: https://github.com/caarlos0/goreleaserfiles/blob/main/build.yml

shibumi:
Right now I set the following in my goreleaser config:
```yaml
project_name: in-toto
builds:
  - ldflags:
      - "-s -w"
      - "-extldflags=-zrelro"
      - "-extldflags=-znow"
      - "-X main.tag={{.Version}}"
      - "-X main.commit={{.FullCommit}}"
      - "-X main.date={{.Date}}"
    env:
      - "CGO_ENABLED=0"
      - "GO111MODULE=on"
      - "GOFLAGS=-mod=readonly -trimpath"
```

Setting the commit date should work and it would still be reproducible, however what if someone wants to explicitly set Date?
Of course you could argue that it is possible to set SOURCE_DATE_EPOCH manually via an environment variable, but note that SOURCE_DATE_EPOCH does not need to be a date. For example you can set SOURCE_DATE_EPOCH=0 and the build environment is responsible to set the date string to `1970-01-01`.

How does `builds.mod_timestamp` exactly work?

In my opinion it would be easier for downstream (Linux distributions packaging their packages) if goreleaser would respect SOURCE_DATE_EPOCH without any further goreleaser config patches, because otherwise downstream would be forced to patch every goreleaser config in the wild just for getting reliable reproducibility.

I am not an expert in this area. @kpcyrd can you give us some hints?

caarlos0:
builds.mod_timestamp sets the modified timestamp on the output binary

https://goreleaser.com/customization/build/

I'm not sure about adding this feature though... how deep do we need to integrate it? Is it only for binaries or also packages, checksums, etc etc etc? 

full disclosure: I didn't know https://reproducible-builds.org until now, please forgive me if its a dumb question hehe

shibumi:
Using {{ .CommitDate }} should definitely work, but what about builds via tarballs? Will the CommitDate just be empty? If so, we can maybe close this PR and I will just use CommitDate for future builds. This should work.

caarlos0:
> what about builds via tarballs?

not sure I understand your question, but goreleaser only works against a git repository... so that seems not be an issue 🤔

shibumi:
We will just stick with .CommitDate. This is actually exactly what I am looking for. Closing this.

Issue #2388, opened 2021-08-09
homebrew/scoop/gofish: allow to customize the branch


caarlos0:
currently it always goes to the default branch... we should allow to customize that for cases in which the default branch is protected.

see https://github.com/goreleaser/goreleaser/discussions/2387

siddhant-deepsource:
Hey @caarlos0,

I would like to help out to fix this. Would need some help and guidance as to how should I approach this. We can maybe create a new field `branch` which the users can specify in the `goreleaser` config. It can default to `master` or `main` (whichever is currently followed).

If the user sets this branch, we can commit to this branch instead. We  might have to handle the case in which the set branch by user doesn't exist. What do you think? Looking forward to your views about this. Thanks 🙂

caarlos0:
my idea was:

- add a `branch` field which defaults to empty  - which clients will handle as default branch if I recall correctly
- if the branch is set, check if it exists, create it in case it doesn't
- commit the file on the given branch

that should probably be handled in the [client package](https://github.com/goreleaser/goreleaser/tree/master/internal/client)

siddhant-deepsource:
@caarlos0 Neat, should we also create pull request from the set branch to the default branch?

caarlos0:
we might, but then we should do it for both gitlab, github and gitea

caarlos0:
this is merged, will be released on next. 

Closing

thanks everyone!

PR #2539, opened 2021-09-29
Draft: feat: allow default branch to be used in Gitlab CreateFile

drewstinnett:
This commit will look up the default branch name in the Gitlab project, and if it can't find one, fall back to `master`, when using the CreateFile function

Previously, the branch used in CreateFile for Gitlab operations was hard coded to `master`. Now that Gitlab has switched over to `main` as the default branch, things like brew publishing breaks since the `master` branch doesn't exist. 

There isn't currently a test in place for any of the CreateFile stuff in `gitlab_test.rb`, so not sure what to do there...open to ideas though!

https://about.gitlab.com/blog/2021/03/10/new-git-default-branch-name/

Thanks for the consideration of this patch, and definitely open to changes/suggestions/etc!


caarlos0:
we have https://github.com/goreleaser/goreleaser/issues/2388 

I think we could allow to pass the branch, and get the default only if its empty. That's for all impls (gitlab, github and gitea)

wdyt?

drewstinnett:
Ahhh I had not realized these were the same, good point! Yes, that makes a lot of sense here. I'd be happy to adjust the PR to use the behavior described above and in #2388! If it's cool with you, I'll adjust the MR later today (or maybe later this week, depending on time haha)

caarlos0:
sure thing! thanks so much @drewstinnett 🙏

btw, let me know if you need help on anything, I'm usually on gophers slack and also on our discord server 👌

drewstinnett:
Def will, ty @caarlos0, you rock!

caarlos0:
I would say that, to simplify, we could leave out the create branch thing...

- if config branch is empty, use default branch
- otherwise, use the branch provided, regardless of wether it exists or not

that way, if the user introduces a typo (e.g. `maste` instead of `master`), the error won't be "silenced"... 


also, to get the default branch on github: https://pkg.go.dev/github.com/google/go-github/v39/github#Repository.GetDefaultBranch

gitea seems to have it as well: https://pkg.go.dev/code.gitea.io/sdk/gitea#Repository

drewstinnett:
Ahhh good call, will adjust!

drewstinnett:
Okey doke, removed the `CreateBranch` stuff, and added the `GetDefaultBranch`, along with some tests to Github and Gitea (in addition to the one that's in there already for Gitlab)

Happy to adjust as needed, and ty for the consideration!

caarlos0:
Amazing! Thanks you!

PR #2541, opened 2021-09-29
feat: improve brew.install guessing

### LINKED ISSUE #2488: [Bug]: Homebrew tap's autogenerated `install` doesn't use binary name
### What happened?

I build a binary (named `kurtosis`) that gets packaged into an archive, uploaded as a Github release to a repo (https://github.com/kurtosis-tech/kurtosis-core-release-artifacts), and is then downloadable via Homebrew formula (https://github.com/kurtosis-tech/homebrew-tap/blob/master/cli.rb). The issue is that the generated formula has:
```
  def install
    bin.install "kurtosis-core-release-artifacts"
  end
```
which is incorrect - the name of the binary as defined in the build is `kurtosis`, and `kurtosis-core-release-artifacts` is the name of the repo that contains the releases with binaries. This results in a failure upon `brew install` complaining:
```
Errno::ENOENT: No such file or directory: kurtosis-core-release-artifacts
```
which is because the archive TARs actually contain a single binary named `kurtosis` (per what's defined in the build).

This seems to be related to the following line, which shows up with `goreleaser check`:
```
      • homebrew tap formula
         • optimistically guessing `brew[0].install` to be `bin.install "kurtosis-core-release-artifacts"`
```

The relevant parts of my `.goreleaser.yml` (`CLI_BINARY_FILENAME` is set to `kurtosis`):
```yaml
builds:
  - id: cli
    main: ./cli
    binary: "{{ .Env.CLI_BINARY_FILENAME }}"
    goos:
      - linux
      - darwin
    goarch:
      - 386
      - amd64
      - arm64

# In order for releasing-to-Github to work, we have to create these archives
archives:
  - id: cli
    builds:
      - cli
    # Kurtosis is a private repo so we don't want to package any files except the binaries, but GoReleaser by default packages the README, LICENSE, and changelog
    # Worse, setting this to be emptylist won't work because GoReleaser takes emptylist to mean "package the defaults" per:
    #  https://github.com/goreleaser/goreleaser/blob/73641c71ac0d9f78e2409b0cea1bf5b42111b933/internal/pipe/archive/archive.go#L61
    # Therefore, we set this to a glob of files that definitely don't exist
    files:
      - "THIS_FILE_DOES_NOT_EXIST"
    name_template: kurtosis-cli_{{ .Version }}_{{ .Os }}_{{ .Arch }}

# In order for Homebrew to work, we have to create a repo where the CLI release artifacts will be uploaded
release:
  github:
    owner: kurtosis-tech
    name: kurtosis-core-release-artifacts

  # NOTE: these are the archive IDs, not the build/binary IDs
  ids:
    - cli

brews:
  - name: cli
    ids:
      - cli
    tap:
      owner: kurtosis-tech
      name: homebrew-tap
    commit_author:
      name: kurtosisbot
    commit_msg_template: "Automated formula update for the CLI, version {{ .Tag }}"
    homepage: "https://www.kurtosistech.com"
    description: "CLI for managing Kurtosis environments."
```

AFAICT, just upload your binary to an intermediate repo, then create a Homebrew formula based off it.

### goreleaser version

```bash
0.179.0
```

### GoReleaser Check

### Search

### Code of Conduct

### Additional context

_No response_

caarlos0:
closes #2488

this will now actually guess the brew install section based on the binaries contained in the archives being used.

I'm not sure if this is a good approach though, as it could break existing users...

### DIFF
--- a/internal/pipe/archive/archive.go
+++ b/internal/pipe/archive/archive.go
@@ -171,6 +171,7 @@ func create(ctx *context.Context, arch config.Archive, binaries []*artifact.Arti
 			return fmt.Errorf("failed to add: '%s' -> '%s': %w", f.Source, f.Destination, err)
 		}
 	}
+	bins := []string{}
 	for _, binary := range binaries {
 		if err := a.Add(config.File{
 			Source:      binary.Path,
@@ -178,6 +179,7 @@ func create(ctx *context.Context, arch config.Archive, binaries []*artifact.Arti
 		}); err != nil {
 			return fmt.Errorf("failed to add: '%s' -> '%s': %w", binary.Path, binary.Name, err)
 		}
+		bins = append(bins, binary.Name)
 	}
 	ctx.Artifacts.Add(&artifact.Artifact{
 		Type:   artifact.UploadableArchive,
@@ -192,6 +194,7 @@ func create(ctx *context.Context, arch config.Archive, binaries []*artifact.Arti
 			"ID":        arch.ID,
 			"Format":    arch.Format,
 			"WrappedIn": wrap,
+			"Binaries":  bins,
 		},
 	})
 	return nil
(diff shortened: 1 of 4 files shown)


radeksimko (commented):
This looks like a reasonable fix, can you enumerate for what users/configs would this be considered as breaking? 🤔

PR #2544, opened 2021-09-30
feat: add template support for homebrew tap owner

terbolous:
This is a minor fix to add support for templates in the Homebrew tap owner field.
Our use case for this was to get the owner field from environment variables


caarlos0:
Thanks!

PR #2545, opened 2021-09-30
docs: update CircleCI example

felicianotech:
The Go image referenced in the example is a legacy image and will be deprecated. I've used instead our (CircleCI) next-gen Go image, which is what should be used going forward.


caarlos0:
thanks!

PR #2557, opened 2021-10-05
feat: Replace Make with Task

andreynering:
As discussed in private, this PR is a proposal to replace the Makefile with Task: https://taskfile.dev/

- Added a `Taskfile.yml`
- Changed CI to use it instead of Make
- Updated CONTRIBUTING documentation

I didn't delete the Makefile yet because some contributors may be used to it. It's your choice to delete it some day or not.

```bash
$ task --list
task: Available tasks for this project:
* build: 	Build the binary
* ci: 		Run all the tests and code checks
* cover: 	Run all the tests and opens the coverage report
* dev: 		Setup pre-commit
* docs:build: 	Build documentation
* docs:serve: 	Start documentation server
* fmt: 		gofmt and goimports all go files
* imgs: 	Download and resize images
* setup: 	Install dependencies
* test: 	Run tests
* todo: 	Show to-do items per file
```

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -1,0 +1,79 @@
+# https://taskfile.dev
+
+version: '3'
+
+env:
+  GO111MODULE: on
+  GOPROXY: https://proxy.golang.org,direct
+
+vars:
+  DOCKER: '{{default "docker" .DOCKER}}'
+
+tasks:
+  dev:
+    desc: Setup pre-commit
+    cmds:
+      - cp -f scripts/pre-commit.sh .git/hooks/pre-commit
+
+  setup:
+    desc: Install dependencies
+    cmds:
+      - go mod tidy
+
+  build:
+    desc: Build the binary
+    cmds:
+      - go build
+
+  test:
+    desc: Run tests
+    env:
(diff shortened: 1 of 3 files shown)


Taskfile.yml
```diff
@@ -0,0 +1,79 @@
...
+    cmds:
+      - '{{.DOCKER}} run --rm -it -p 8000:8000 -v ${PWD}/www:/docs docker.io/squidfunk/mkdocs-material'
+
+  docs:build:
+    desc: Build documentation
+    cmds:
+      - yum install -y jq
+      - pip install mkdocs-material mkdocs-minify-plugin lunr
+      - ./scripts/get-releases.sh
+      - (cd www && mkdocs build)
```
andreynering, line 73: I realized that you could probably just use Docker in this task as well, just like in the above task.

That would certainly help contributors as less dependencies would be needed.

caarlos0: this task is actually used on vercel... will check there if we have go-task/docker and if not if its easy to install

caarlos0:
Thanks! I'll fix the vercel stuff on master! 💙

PR #2559, opened 2021-10-06
chore(ci): cache go-build results on github actions

caarlos0:
according to [this](https://github.com/mvdan/github-actions-golang), this seems like a good idea to speed up tests/builds a bit

wdyt?


radeksimko (approved):
👍🏻  This also seems to align with the official source:

https://github.com/actions/cache/blob/main/examples.md#go---modules

PR #2564, opened 2021-10-06
add ko support for building and pushing container images 

### LINKED ISSUE #2556: [Feature Request]: add google/ko as brand new builder engine for building and pushing container images

ko[^1] is a simple, fast container image builder for Go applications.

It's ideal for use cases where your image contains a single Go application without any/many dependencies on the OS base image (e.g., no cgo, no OS package dependencies).

**Additional context**
[^1]: https://github.com/google/ko

cc: @dentrax @erkanzileli @imjasonh

developer-guy:
Signed-off-by: Batuhan Apaydin batuhan.apaydin@trendyol.com
Co-authored-by: Furkan furkan.turkal@trendyol.com (@dentrax)
Co-authored-by: Erkan Zileli erkan.zileli@trendyol.com  (@erkanzileli) 

fix #2556 

to add ko support for building and pushing container images 

<details>

<summary> Sample .goreleaser.yml </summary>

```YAML
# This is an example .goreleaser.yml file with some sensible defaults.
# Make sure to check the documentation at https://goreleaser.com
before:
  hooks:
    # You may remove this if you don't use go modules.
    - go mod tidy
builds:
  - env:
      - CGO_ENABLED=0
    goos:
      - linux
      - windows
      - darwin
    ldflags:
      - -s -w -X github.com/GoTurkiye/hello-world-cli/cmd.GitCommit={{ .Commit }} -X github.com/GoTurkiye/hello-world-cli/cmd.Version={{ .Version }}
archives:
  - replacements:
      darwin: Darwin
      linux: Linux
      windows: Windows
      386: i386
      amd64: x86_64
checksum:
snapshot:
  name_template: "{{ incpatch .Version }}-next"
changelog:
  sort: asc
  filters:
    exclude:
      - '^docs:'
      - '^test:'
dockers:
  - image_templates: [ "ghcr.io/goturkiye/{{ .ProjectName }}"]
    goos: linux
    goarch: amd64
    use: ko  ## notice here


```

</details>

Screen Shot 2021-10-07 at 00 05 18
Screen Shot 2021-10-07 at 00 04 09

### DIFF
--- a/internal/pipe/defaults/defaults_test.go
+++ b/internal/pipe/defaults/defaults_test.go
@@ -99,7 +99,7 @@ func TestFillPartial(t *testing.T) {
 
 	ctx = context.New(config.Project{
 		GiteaURLs: config.GiteaURLs{
-			API: "https://gitea.com/api/v1",
+			API: "https://gitea.com/api/v1/",
 		},
 	})
 	ctx.TokenType = context.TokenTypeGitea


developer-guy:
> thinking more about this, I got more questions:
> 
> * ko has a lot of config that the docker pipe wont support... what should we do about that?
> * in the same line, we are inferring the module from `./go.mod`, but what happens if the user has many `go.mod` files?
> 
> everything would likely work for "simple" projects, but not sure about more complex ones...

@imjasonh maybe can help us here.

imjasonh:
Sorry for the delayed response.

I'm a bit worried that `ko`-in-goreleaser will be confusing since `ko` itself reuses some of goreleaser's config types in `.ko.yaml`, and it may be confusing to decide where to describe certain things between goreleaser (how the binary gets built) and .ko.yaml (how the image(s) get built)

For the question of "how to configure `ko`", could we have a special `ko:` section in the goreleaser config? This would let users describe their module without having to inspect go.mod(s), and could also help with unsupported fields in the `docker:` field, since we won't need to use it.

wdyt?

cc @jonjohnsonjr

developer-guy:
> Sorry for the delayed response.
> 
> I'm a bit worried that `ko`-in-goreleaser will be confusing since `ko` itself reuses some of goreleaser's config types in `.ko.yaml`, and it may be confusing to decide where to describe certain things between goreleaser (how the binary gets built) and .ko.yaml (how the image(s) get built)
> 
> For the question of "how to configure `ko`", could we have a special `ko:` section in the goreleaser config? This would let users describe their module without having to inspect go.mod(s), and could also help with unsupported fields in the `docker:` field, since we won't need to use it.
> 
> wdyt?
> 
> cc @jonjohnsonjr

thank you for such a great explanation, I think both methods (current solution and your suggestion) would work fine but yes I agree with your concerns about configuring ko while using it in GoReleaser because ko also builds the binary first which means that fields defined in `build:` section will be ignored in this case, or somehow we should pass build parameters to ko while building the binary.

WDYT @caarlos0?

caarlos0:
I'm thinking maybe we could instead use ko as a dependency and skip the go build part (as we already do that), and just build the images... would that be possible (haven't investigate ko's codebase yet)? if so, wdyt?

Dentrax:
> I'm thinking maybe we could instead use ko as a dependency and skip the go build part (as we already do that), and just build the images... would that be possible (haven't investigate ko's codebase yet)? if so, wdyt?

I'm not so sure whether instead of using the binary, and importing `ko` code as a library is an anti-pattern or not. Is ko intended to be used as a library? We can use it as a library ofc, but that is a side-effect of how ko code is structured, which is subject to change. To clarify aforementioned concerns and the getting feedback from "ko as a dependency", cc'ing @imjasonh 🙃

imjasonh:
`ko` can be used as a library, some folks on Skaffold are also looking into integrating `ko` using its library methods.

The main functions you'd probably care about are [`build.NewGo`](https://pkg.go.dev/github.com/google/ko/pkg/build#NewGo) to create a new builder, and [`publish.NewDefault`](https://pkg.go.dev/github.com/google/ko/pkg/publish#NewDefault) to push the image.

caarlos0:
cc/ @cpanato

caarlos0:
closing for now, as we'll discuss this in our community meeting

jonjohnsonjr:
@caarlos0 tried searching briefly for docs around a community meeting and couldn't find anything -- have a link handy?

caarlos0:
sure, here you go: https://github.com/goreleaser/community/pull/2

PS: we don't have a proper date yet, but likely later january or early february

developer-guy:
let's kick this off and rewrite it by using ko as a library instead of calling it directly through the binary. 

This repo will help us with that: _https://github.com/imjasonh/miniko_

/cc @imjasonh @caarlos0

PR #2583, opened 2021-10-15
feat: allow to template builds.env

caarlos0:
This PR allows templates in the `builds.env` section.

This should help improve configs like [this one from cosign](https://github.com/sigstore/cosign/blob/main/.goreleaser.yml#L45-L92) to something like:

```yaml
builds:
- id: cosign
  binary: cosign
  main: .
  goos:
    - linux
    - darwin
    - windows
  goarch:
    - amd64
    - arm64
  env:
    - CGO_ENABLED=0
    # option 1:
    - CC_darwin_amd64=o64-clang
    - CCX_darwin_amd64=o64-clang+
    - CC_darwin_arm64=aarch64-apple-darwin20.2-clang
    - CCX_darwin_arm64=aarch64-apple-darwin20.2-clang++
    - CC_windows_amd64=x86_64-w64-mingw32-gc
    - CCX_windows_amd64=x86_64-w64-mingw32-g++
    - 'CC={{ index .Env (print "CC_" .Os "_" .Arch) }}'
    - 'CCX={{ index .Env (print "CCX_" .Os "_" .Arch) }}'
    # option 2:
    - >-
        {{- if eq .Os "darwin" }}
          {{- if eq .Arch "amd64"}}CC=o64-clang{{- end }}
          {{- if eq .Arch "arm64"}}CC=aarch64-apple-darwin20.2-clang{{- end }}
        {{- end }}
        {{- if eq .Os "windows" }}
          {{- if eq .Arch "amd64" }}CC=x86_64-w64-mingw32-gcc{{- end }}
        {{- end }}
    - >-
        {{- if eq .Os "darwin" }}
          {{- if eq .Arch "amd64"}}CXX=o64-clang+{{- end }}
          {{- if eq .Arch "arm64"}}CXX=aarch64-apple-darwin20.2-clang++{{- end }}
        {{- end }}
        {{- if eq .Os "windows" }}
          {{- if eq .Arch "amd64" }}CXX=x86_64-w64-mingw32-g++{{- end }}
        {{- end }}
```

thoughts?


caarlos0:
> I see that the variables are only passed to the build commands, not to hooks - which seems like a good idea in principle IMO, but I'm not sure this intention/behaviour is immediately obvious.

noticed that as well, and also not sure why... I think its more likely to be a bug...

caarlos0:
FWIW @radeksimko right now hooks see the build.envs...

caarlos0:
ok, assuming the env on hooks as well, this is basically wrong. Will close

PR #2609, opened 2021-10-28
feat: release only specific build ids

ryancurrah:
In order to allow releasing specific builds I added a new argument to release called 'only-build'. This argument allows one to select which build ids to build by using a regular expression. This is handy when you only want to release a specific build or test a specific build without having to wait for everything else to build as well.

Let me know what you think. If this seems like a feasible feature I will add some tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ www/docs/static/releases*.json
 completions/
 .vscode/
 .task/
+.idea/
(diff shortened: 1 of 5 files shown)


caarlos0:
for testing locally you can use [`goreleaser build --id`](https://goreleaser.com/cmd/goreleaser_build).

if you sometimes need to release only parts of your code, you'll be better of with explicit different goreleaser config files.

so, sorry, but for now, this is a no.

thanks for the pr though.

cheers!

ryancurrah:
No worries thanks for considering it.

PR #2613, opened 2021-10-29
docs: add USERS.md to track users of goreleaser

### LINKED ISSUE #2612: docs: add USERS.md to track users of goreleaser
Idea:

It would be great to have a list of all organizations using goreleaser. 

Having USERS file to keep track of who is using goreleaser via a PR could be nice way to to do this. Maybe add them later to the docs page too?

Nothing more. 😄

dirien:

fixes #2612


caarlos0:
ah, that's a nice idea! Thanks!

PR #2615, opened 2021-10-30
Add Hugo to USERS.md

bep:


bep:
You should fix your PR workflow setup; spamming users with emails for a small change in a text file is not ... great:

image

caarlos0:
> You should fix your PR workflow setup; spamming users with emails for a small change in a text file is not ... great:

the auto-assign stuff is failing for a while now.. I removed it...

the docs failure was a 404 in some url, fixed it, but will also change it to run only when docs changed as well

thanks for the feedback, its great as always 🤟

Issue #2617, opened 2021-10-31
Add support for replacements in builds section


tricky42:
**Is your feature request related to a problem? Please describe.**
I wanted to add sbom generation to the Ory Kratos projects which uses goreleaser. The goal is to create a SBOM file with the same name as the release archive. In the Kratos `.goreleaser.yml` file several `replacements` get defined: https://github.com/ory/kratos/blob/bf80e9614b048a3d846615b2eb2ca61ac7703800/.goreleaser.yml#L168

SBOM generation gets integrated via a hook in the `builds` section. 
```
    hooks:
      post:
        - cmd: cyclonedx-gomod app -licenses -json -output "./dist/{{ .ProjectName }}_{{ .Version }}_{{ .Target }}.bom.json"
```
As `replacements` are not supported in the `builds` section there is no "good" way to define the filename in the hook definition correctly.

**Describe the solution you'd like**
Either add support for replacements in the `builds` sections or add support for the definition of global replacements.

**Describe alternatives you've considered**
1. use {{ if }} statements e.g. {{ if eq .Env.GOOS "darwin" }}macos{{end}}
2. add mapping to release section
```
extra_files:
- src: foo_1.2.3_darwin_amd64.bom.json
  dst: foo_1.2.3_macos_amd64.bom.json
```

**Additional context**

caarlos0:
we already have an issue to add BOM support to goreleaser... build-level replacements is unlikely to happen, so, closing this...

#2597

Issue #2597, opened 2021-10-21
[Proposal] SBOM generation


VinodAnandan:
The proposal is to publish SBOM (“Software Bill of Materials”) artifacts for goreleaser as part of each release.

e.g.:  https://github.com/CycloneDX/cyclonedx-gomod#goreleaser-
https://github.com/CycloneDX/cyclonedx-gomod/releases/tag/v1.0.0

caarlos0:
FWIW, `go version -m` can have a wrong output if using `go mod vendor`.

https://carlosbecker.com/posts/supply-chain-goreleaser-go-mod-proxy/

caarlos0:
please take a look at #2618

wagoodman:
_(this comment is meant to be a follow on to https://github.com/goreleaser/goreleaser/pull/2618#issuecomment-957825874 from the POC PR)_

While chatting the POC over with @spiffcs and @luhring we started to wonder if it made sense to explore the idea of raising SBOMs as a first-class artifact type. This new `sbom` artifact type would be defined as "a description of another artifact or set of artifacts". It would have it's own `sboms` config section and would act/look very similar to the `signs` config section. 

The docs for such a `sboms` section might look like this:

```yaml
# .goreleaser.yml
sboms:
  -
    # ID of the bom config, must be unique.
    #
    # Defaults to "default".
    id: foo

    # List of Names/templates of the bom files created at this step.
    #
    # Available environment variables:
    # - '${artifact}': the path to the artifact that will be cataloged
    # - '${artifactID}': the ID of the artifact that will be cataloged
    #
    # Defaults to empty list.
    sbom: 
      - "${artifact}.spdx.sbom"

    # Path to the signature command
    cmd: syft

    # Command line templateable arguments for the command
    args: ["-o", "spdx", "--file", "${bom}", "${artifact}"]

    # Which artifacts to catalog
    #
    #   source:   source archive
    #   package:  linux packages (deb, rpm, apk)
    #   archive:  archives from archive pipe
    #   binary:   binaries output from the build stage
    #   container:  reference to the unpublished container image 
    #... there maybe other artifacts that make sense here!
    #
    # Defaults to `binary`
    artifacts: container

    # IDs of the artifacts to catalog.
    #
    # If `artifacts` is source this fields has no effect.
    #
    # Defaults to empty (which implies no filtering).
    ids:
      - foo
      - bar

```

With example usage:

```yaml
sboms:
  - 
    id: source-sbom
[...]

PR #2618, opened 2021-11-01
feat: improve build hooks to make it easier to create BOMs

caarlos0:
**THIS IS JUST A PoC FOR NOW**

refs #2617 #2597

the idea consists of:

- having an `artifact` env in the hooks
- allowing to add "custom artifacts" on hooks
- with that, one can easily wire things like `cyclonedx-gomod` to the pipeline
- the archive will have the related bom within it
  - not sure what should happen when using universal binaries as well
  - right now, the archive will not have any BOM in it
- this is also passing GOOS/GOARCH/etc to the hook commands
- this is not by any means a "final implementation" (or anything near that), just want to collect feedback on the API,  concerns, etc...

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -32,6 +32,12 @@ builds:
     - -trimpath
   ldflags:
     - -s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{ .CommitDate }} -X main.builtBy=goreleaser
+  hooks:
+    post:
+      - cmd: "cyclonedx-gomod app -licenses -json -output {{.Env.artifact}}.bom.json"
+        custom_artifact:
+          path: "{{.Env.artifact}}.bom.json"
+          name: goreleaser.bom.json
 
 universal_binaries:
 - replace: true
--- a/internal/artifact/artifact.go
+++ b/internal/artifact/artifact.go
@@ -56,6 +56,8 @@ const (
 	GoFishRig
 	// ScoopManifest is an uploadable scoop manifest file.
 	ScoopManifest
+	// Custom file.
+	Custom
 )
 
 func (t Type) String() string {
@@ -86,6 +88,8 @@ func (t Type) String() string {
 		return "GoFish Rig"
 	case ScoopManifest:
 		return "Scoop Manifest"
+	case Custom:
+		return "Custom file"
 	default:
 		return "unknown"
 	}
--- a/internal/pipe/archive/archive.go
+++ b/internal/pipe/archive/archive.go
@@ -89,7 +89,7 @@ func (Pipe) Run(ctx *context.Context) error {
 	g := semerrgroup.New(ctx.Parallelism)
 	for i, archive := range ctx.Config.Archives {
 		archive := archive
-		artifacts := ctx.Artifacts.Filter(
+		binaries := ctx.Artifacts.Filter(
 			artifact.And(
 				artifact.Or(
 					artifact.ByType(artifact.Binary),
@@ -98,17 +98,23 @@ func (Pipe) Run(ctx *context.Context) error {
 				artifact.ByIDs(archive.Builds...),
 			),
 		).GroupByPlatform()
-		if err := checkArtifacts(artifacts); err != nil && !archive.AllowDifferentBinaryCount {
+		customs := ctx.Artifacts.Filter(
+			artifact.And(
+				artifact.ByType(artifact.Custom),
+				artifact.ByIDs(archive.Builds...),
+			),
+		).GroupByPlatform()
+		if err := checkArtifacts(binaries); err != nil && !archive.AllowDifferentBinaryCount {
 			return fmt.Errorf("invalid archive: %d: %w", i, ErrArchiveDifferentBinaryCount)
 		}
-		for group, artifacts := range artifacts {
+		for group, artifacts := range binaries {
 			log.Debugf("group %s has %d binaries", group, len(artifacts))
 			artifacts := artifacts
 			g.Go(func() error {
 				if packageFormat(archive, artifacts[0].Goos) == "binary" {
 					return skip(ctx, archive, artifacts)
 				}
-				return create(ctx, archive, artifacts)
+				return create(ctx, archive, artifacts, customs[group])
 			})
 		}
 	}
@@ -126,7 +132,7 @@ func checkArtifacts(artifacts map[string][]*artifact.Artifact) error {
 	return ErrArchiveDifferentBinaryCount
 }
 
-func create(ctx *context.Context, arch config.Archive, binaries []*artifact.Artifact) error {
+func create(ctx *context.Context, arch config.Archive, binaries, customs []*artifact.Artifact) error {
 	format := packageFormat(arch, binaries[0].Goos)
 	folder, err := tmpl.New(ctx).
 		WithArtifact(binaries[0], arch.Replacements).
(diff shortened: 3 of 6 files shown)


VinodAnandan:
Cc: @nscuro,@tricky42, @knqyf263, @masahiro331, @samj1912, @ShubhamPalriwala, @JimBugwadia

nscuro:
First off, **thank you** for this. It's really appreciated. 

> this is also passing GOOS/GOARCH/etc to the hook commands

Great, that's super helpful! But besides `GOARCH`, `GOOS` etc., another thing that can influence module/package selection are build tags. Thus, when generating an SBOM, it's important to capture this information as well. Currently, it's possible to define them in a structured way for each `build`, however they are not accessible for `post` hooks, so they have to be set manually again:

```yml
builds:
  - # ...
    tags:
      - foo
      - bar
    hooks:
      post:
        - cmd: cyclonedx-gomod ...
          env:
            - GOFLAGS="-tags=foo,bar"
```

It would be great if we could have them available in `post` hooks as well. Providing them as `,`-joined string like `foo,bar` should be sufficient, so we could do something like this:

```yml
builds:
  - #...
    tags:
      - foo
      - bar
    hooks:
      post:
        - cmd: cyclonedx-gomod app ...
          # Or, alternatively:
          env:
            - GOFLAGS="-tags={{ .BuildTags }}"
```

> not sure what should happen when using universal binaries as well
> right now, the archive will not have any BOM in it

IIRC, universal binaries are just `amd64` and `arm64` merged together? In that case, the correct behavior would be to either include both SBOMs, or merge them into one. The latter would require goreleaser to "know" that it's handling BOMs, and how to correctly merge them.

radeksimko (commented):
The API and POC looks reasonable to me

wagoodman:
Thanks for this POC... it has helped a lot in figuring possible directions! After trying this out I have some questions / observations:

1. The advantage of build hooks is that there is a clear way to create an SBOM for a binary that was built... and you can create a binary for one or more architectures easily. There is some repetitiveness with this approach but with yaml references and anchors these sections can be dried out.

2. If you are using a tool that can intake multiple sources and binaries to be included in a single SBOM or there is a non 1:1 mapping of input sources to output SBOMs then build hooks may be a problem. Build hooks enforce a 1:1 relationship of input artifacts to SBOMs generated.

3. The default assumption of including the SBOM in any created archives seems reasonable, though there are reasons when this isn't desirable. For large applications the SBOM can also get pretty large, and even though it's being included in a gzipped tar, if the tar is being downloaded constantly (due to install scripts) then excluding the SBOM from that tar may be useful. There is a similar reasoning when you are only interested in the SBOM and not the application. Is there a way to configure goreleaser with the changes on this branch to allow for attachments of the SBOM to a release without encapsulating it in a tar?

4. The current approach seems to support signing SBOM material by including the SBOM in the tar with the build artifact and signing the tar. Is there an easy approach to specify to goreleaser to sign the SBOM directly?

5. Some SBOM tools are beginning to add attestation capabilities. I'm still trying to figure if this affects the features on this branch or not. That is, would this be something that goreleaser is aware of (say, in the `signs` section)? Or would this be out-of-band to goreleaser entirely?

6. How would capturing SBOMs for a built container image work?

VinodAnandan:
@wagoodman:  Thanks for the detailed proposal.
Could you please help explain the use cases with a source SBOM? Do you think a source sbom be generated without "Incorrect/Extra Data"?

wagoodman:
> Could you please help explain the use cases with a source SBOM?

Righto, @VinodAnandan 👍  ...are you referring to the example from https://github.com/goreleaser/goreleaser/issues/2597#issuecomment-957842629 ? Specifically:

> - one SBOM for the source tar ...

If so, here is my [opinionated] take: a source SBOM could contain at least a catalog of all of the in-project source code files, hash of each file, and license declared for each file. The source SBOM would not be responsible for Go module references and their relationships. When making an SBOM from a binary (compiled from the code contained in the source SBOM) then one could make a single external-document reference to the source SBOM. For folks interested in both the source and the module information for a project, this separation of concerns can help as the project get's bigger. Especially if you have multiple binaries being generated (and therefor multiple SBOMs being generated, one for each binary)... it wouldn't be ideal to have duplicate source descriptions in each SBOM. 

Regarding the next question:

> Do you think a source sbom be generated without "Incorrect/Extra Data"?

I'm not certain what specifically you're asking here. Do you mean having bad/extra references to source files that the go compiler may exclude given a set of build flags? Or something else? 

For the usecase highlighted above, this doesn't really apply; there is one SBOM for the state of the repo (the "source SBOM") and one or more SBOMs that may reference the source SBOM but distinctly describe what was found in the binary artifact, so is separate from any additional build concerns, or declaring which source files "belong" to which binary artifact.

(Let me know if I misinterpreted anything!)

VinodAnandan:
If I understood correctly, it is an inventory/metadata of the source code files (.go) of an application and dependencies (vendor)? I think the challenge will be to get the information of the actual source code files used for the binary and provide an accurate SBOM? Also, I think the source SBOM without considering the compiler SBOM, won't be useful as a compromised compiler can override? I am trying to understand the business/security use cases of the source sbom. I think the git repository also provides an inventory of the source code repo.

Apologies I haven't provided more context about  "Incorrect/Extra Data". I was referring to this particular issue -https://github.com/CycloneDX/cyclonedx-gomod/issues/20 . I also believe it is important to have an accurate SBOM to get the best possible value from it."

VinodAnandan:
If I understood correctly, it is an inventory/metadata of the source code files (.go) of an application and dependencies (vendor)? I think the challenge will be to get the information of the actual source code files used for the binary and provide an accurate SBOM? Also, I think the source SBOM without considering the compiler SBOM, won't be useful as a compromised compiler can override? I am trying to understand the business/security use cases of the source sbom. I think the source code repository (git repo) can also provides an inventory information?

Apologies I haven't provided more context about  "Incorrect/Extra Data". I was referring to this particular issue -https://github.com/CycloneDX/cyclonedx-gomod/issues/20 . I also believe it is important to have an accurate SBOM to get the best possible value from it.

VinodAnandan:
If I understood correctly, it is an inventory/metadata of the source code files (.go) of an application and dependencies (vendor)? I think the challenge will be to get the information of the actual source code files used for the binary and provide an accurate SBOM? Also, I think the source SBOM without considering the compiler SBOM, won't be useful as a compromised compiler can override? I am trying to understand the business/security use cases of the source sbom. I think the source code repository (git repo) can also provide the inventory information?

Apologies I haven't provided more context about "Incorrect/Extra Data". I was referring to this particular issue -CycloneDX/cyclonedx-gomod#20 . I also believe it is important to have an accurate SBOM to get the best possible value from it.

luhring:
@wagoodman @VinodAnandan This is an interesting discussion! 💥  If I'm following correctly, I would add this too —

I think the important bit here is that SBOMs should be expressed **in reference to a specific artifact**, which is the "thing" said to contain the software described in the bill of materials.

This artifact can be a **compiled Go binary**, but it also can be **something else**. (Even for Go projects!)

It's valid to have an SBOM that describes a different (but perhaps related) artifact besides a compiled binary — for example, this may be a Docker image, a TAR, or even a source code project (like when a Go project is a library consumed by others).

This means that asking the question _"will scanning a given input help to generate my **Go binary's** SBOM?"_ makes sense in some use cases — but is less relevant in other use cases.

❤️  I really like where @wagoodman is heading with this idea for GoReleaser, because it means Go project maintainers get more capabilities unlocked, and they have the freedom to ship artifacts accompanied by SBOMs, even when shipping more than just a compiled Go binary.

caarlos0:
closing this one in favor of #2649 and #2648 

feel free to continue the discussion though 🙏

VinodAnandan:
Thanks for the explanation. I think I understand where you are coming from. I was trying to understand how valuable this source SBOM data will be and how to utilise it for a meaningful purpose. For the security use cases, I think the SBOM generated during the build/compilation has more value than the source one for compiled programming languages like go. Could you please let me know if you think otherwise?

PR #2620, opened 2021-11-01
feat(source): add prefix_template option

twpayne:
If applied, this change will add a `prefix_template` option to the source archive pipe. This allows the generated tarball to use a subdirectory for all files in the archive, as described in [`git-archive`](https://git-scm.com/docs/git-archive).

Currently, archives generated by the source archive step do not use a prefix, whereas traditionally the prefix of `{{ .ProjectName }}-{{ .Version }}/` is used. See https://github.com/twpayne/chezmoi/issues/1576.

The default prefix is none, to maintain compatibility for existing users.


caarlos0:
noice, thanks @twpayne !

PR #2623, opened 2021-11-02
fix: remove .git suffix by default from announcers

### LINKED ISSUE #2622: [Bug]: Twitter announce incorrect URL
### What happened?

I use twitter announce (which works fine), but it posted the incorrect URL. 

This is most likely not a bug in Goreleaser and probably not released to Twitter announce, but to the value of `.GitURL`.

When importing a new repsitory on Github, it suggest to add a remote like: 

    git remote add origin https://github.com/speedata/ets.git

(notice the `.git` suffix)

Now I set up Twitter announce with the following setting:

```
announce:
  twitter:
    # Whether its enabled or not.
    # Defaults to false.
    enabled: true
```

and on `goreleaser release --rm-dist` it publishes a twitter message with an incorrect link. I don't have it anymore but I strongly believe it was `https://github.com/speedata/ets.git/releases/tag/v0.0.2`, including the `.git` suffix. This link gives a 404.

Perhaps it is possible to strip the `.git` suffix from the `.GitURL` or to override this in the yaml file? The `github_urls:` are documented to be for Github enterprise only, so I didn't try them.

Sorry that I have deleted the link too quickly. 

### goreleaser version

```bash
goreleaser --version
goreleaser version 0.183.0
commit: be6199f08167cc85b0b16cefc02d8106237853b1
built by: homebrew

https://goreleaser.com
```

### GoReleaser Check

### Search

### Code of Conduct

### Additional context

_No response_

Dentrax:
Fixes #2622

Signed-off-by: Furkan <furkan.turkal@trendyol.com>
Co-authored-by: Batuhan <batuhan.apaydin@trendyol.com>
Co-authored-by: Erkan <erkan.zileli@trendyol.com>

...

...

...

### DIFF
--- a/internal/pipe/discord/discord.go
+++ b/internal/pipe/discord/discord.go
@@ -16,7 +16,7 @@ const (
 	defaultAuthor          = `GoReleaser`
 	defaultColor           = "3888754"
 	defaultIcon            = "https://goreleaser.com/static/avatar.png"
-	defaultMessageTemplate = `{{ .ProjectName }} {{ .Tag }} is out! Check it out at {{ .GitURL }}/releases/tag/{{ .Tag }}`
+	defaultMessageTemplate = `{{ .ProjectName }} {{ .Tag }} is out! Check it out at {{ trimsuffix .GitURL ".git" }}/releases/tag/{{ .Tag }}`
 )
 
 type Pipe struct{}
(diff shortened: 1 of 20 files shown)


caarlos0 (commented):
this looks good! thanks!

I think its probably worth mentioning somewhere (maybe on template.md) that the giturl may not be a http url if, for instance, your remote uses ssh for example...

so, to ensure correctness, its better to always set the actual URL on the announce templates...

that, could be done in another PR though, so I'm going to go ahead and merge this!

thanks again!

PR #2624, opened 2021-11-02
docs: Add Ministry of Justice UK to list of users

jasonBirchall:
Overview
---

This PR will add Ministry of Justice in the UK to the list of proud users of this tool. Using goreleaser dramatically  improves release time/velocity of our developers, which in turn improves the experience of the British public using digital justice tools and services.

Keep up the great work!


caarlos0:
noice! thanks

PR #2626, opened 2021-11-03
feat: add message template support for git providers

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>
Co-authored-by: Furkan Türkal <furkan.turkal@trendyol.com>
Co-authored-by: Erkan Zileli <erkan.zileli@trendyol.com>

Fixes #2625 

to support message templating while announcing for supported git providers such as GitHub, GitLab, Gitea

PR #2637, opened 2021-11-06
refactor: release url

caarlos0:
refs #2626

A possible better impl:

- defaults the release url in the releases pipe
- adds docs
- move `ReleaseURL` to the ctx root
- fixed release url docs

cc/ @developer-guy @erkanzileli @Dentrax


Dentrax:
This is a much way better impl, thanks!

PR #2639, opened 2021-11-07
feat: krew support

### LINKED ISSUE #2598: [Proposal] krew support
I wanted to preflight the idea of supporting krew in goreleaser.

So i could imagine some scenarios:

1.) Create krew file for your own index -> https://krew.sigs.k8s.io/docs/developer-guide/custom-indexes. This would be similar to our brew / gofish

2.) Create the .krew.yaml file for the krew-release-bot -> https://github.com/rajatjindal/krew-release-bot

With 2. you need to first manually create a PR in the krew-index repo. Then the krew-release-bot, when added to the github action, will pick up any further changes.

What are your thoughts?

caarlos0:
adds support to creating and publishing krew plugins to a custom repository (not the [main one](https://github.com/kubernetes-sigs/krew-index)).

please help reviewing if this makes sense :) 

closes #2598


internal/pipe/krew/krew.go
```diff
@@ -0,0 +1,325 @@
...
+		return pipe.Skip("krews.skip_upload is set")
+	}
+
+	if strings.TrimSpace(cfg.SkipUpload) == "auto" && ctx.Semver.Prerelease != "" {
+		return pipe.Skip("prerelease detected with 'auto' upload, skipping krew publish")
+	}
+
+	repo := client.RepoFromRef(cfg.Index)
+
+	gpath := buildPluginPath(pluginsFolder, plugin.Name)
```
dirien, line 269: Hi @caarlos0,

Can we add here maybe the https://github.com/kubernetes-sigs/krew/tree/master/cmd/validate-krew-manifest

> Ensure plugin manifests are valid YAML and pass Krew manifest validation (optionally, you can use the validate-krew-manifest tool for static analysis).


to validate the krew file? just in case?

internal/pipe/krew/krew.go
```diff
@@ -0,0 +1,325 @@
...
+		if err != nil {
+			return result, err
+		}
+
+		goarch := []string{art.Goarch}
+		if art.Goarch == "all" {
+			goarch = []string{"amd64", "arm64"}
+		}
+
+		for _, arch := range goarch {
```
dirien, line 205: Would this work also with eg. cosign binary? As you now they have this multibinaries with same Arch/OS combination

internal/artifact/artifact.go
```diff
@@ -54,6 +54,8 @@ const (
 	BrewTap
 	// GoFishRig is an uploadable Rigs rig food file.
 	GoFishRig
+	// KrewPlugin is a krew plugin file.
+	KrewPlugin
```
dirien, line 58: IMHO, i would call it only krew?

internal/artifact/artifact.go
```diff
@@ -84,6 +86,8 @@ func (t Type) String() string {
 		return "Brew Tap"
 	case GoFishRig:
 		return "GoFish Rig"
+	case KrewPlugin:
+		return "Krew Plugin"
```
dirien, line 90: Not sure about the name. Because actually they are kubectl plugins... Tricky

dirien (commented):
Hi @caarlos0,

great work! I left some comments.

internal/pipe/krew/krew.go
caarlos0: from what I looked into their code/index, seems like only one binary is allowed... maybe we should even hardfail on those cases

internal/pipe/krew/krew.go
caarlos0: yeah, was working on it :)

caarlos0:
right now kind of depends on https://github.com/kubernetes-sigs/krew/pull/736

internal/artifact/artifact.go
caarlos0: I have no idea... I think the correct would be `KrewManifest` 🤔  (judging by `validate-krew-manifest`)... will rename :)

internal/artifact/artifact.go
caarlos0: ended up naming it krew plugin manifest, which is the most specific name...

dirien:
> right now kind of depends on [kubernetes-sigs/krew#736](https://github.com/kubernetes-sigs/krew/pull/736)

Or we need to vendor it, if it's not getting fixed in a near term?

caarlos0:
> Or we need to vendor it, if it's not getting fixed in a near term?

or we use my fork meanwhile :P

ahmetb:
that's massive. 👀 but I'll review.

caarlos0:
> that's massive. 👀 but I'll review.

mostly golden files for krew plugin manifests generated by goreleaser :D

internal/pipe/krew/krew.go
```diff
@@ -0,0 +1,330 @@
...
+		}
+	}
+	return nil
+}
+
+func doRun(ctx *context.Context, krew config.Krew, cl client.Client) error {
+	if krew.Index.Name == "" {
+		return pipe.Skip("krew manifest name is not set")
+	}
+	if krew.ShortDescription == "" {
```
caarlos0, line 85: probably need to add more cases here (e.g. empty description)

pkg/config/config.go
```diff
@@ -147,6 +147,22 @@ type Homebrew struct {
...
 
+// Krew contains the krew section.
+type Krew struct {
+	IDs                   []string     `yaml:"ids,omitempty"`
+	Name                  string       `yaml:"name,omitempty"`
+	Index                 RepoRef      `yaml:"index,omitempty"`
+	CommitAuthor          CommitAuthor `yaml:"commit_author,omitempty"`
+	CommitMessageTemplate string       `yaml:"commit_msg_template,omitempty"`
+	Caveats               string       `yaml:"caveats,omitempty"`
+	ShortDescription      string       `yaml:"short_description,omitempty"`
```
ahmetb, line 158: FWIW: we expect users to provide something max 50 chars here (we provide this feedback during the code-review in krew-index)

pkg/config/config.go
```diff
@@ -147,6 +147,22 @@ type Homebrew struct {
...
+	Description           string       `yaml:"description,omitempty"`
```
ahmetb, line 159: FWIW: we expect the author to line-wrap this at 80 chars as a multi-line YAML string (we provide this feedback during the code-review in krew-index) but it's not a hard requirement especially when URLs are in the description.

pkg/config/config.go
```diff
@@ -147,6 +147,22 @@ type Homebrew struct {
...
 }
 
+// Krew contains the krew section.
+type Krew struct {
+	IDs                   []string     `yaml:"ids,omitempty"`
+	Name                  string       `yaml:"name,omitempty"`
+	Index                 RepoRef      `yaml:"index,omitempty"`
+	CommitAuthor          CommitAuthor `yaml:"commit_author,omitempty"`
+	CommitMessageTemplate string       `yaml:"commit_msg_template,omitempty"`
+	Caveats               string       `yaml:"caveats,omitempty"`
```
ahmetb, line 157: (ditto same comments as Description below)

I mostly say these because you might need to carry over the YAML node as-is (with indentation and which YAML multiline mode is used) to the final Krew YAML rendering. Similarly, if there are structural differences such as whitespaces between the old YAML vs the new YAML, then these PRs need to get manually reviewed (krew-release-bot won't be able to approve)

ahmetb (commented):
Thank you for this, looking forward to using this!

ahmetb (commented):
Thank you for this, looking forward to using this!

internal/pipe/krew/krew.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	return cl.CreateFile(ctx, cfg.CommitAuthor, repo, content, gpath, msg)
+}
+
+func buildManifestPath(folder, filename string) string {
+	return path.Join(folder, filename)
+}
+
+type Manifest struct {
```
ahmetb, line 296: FWIW we offer this as a type in https://pkg.go.dev/sigs.k8s.io/krew/pkg/index#Plugin

pkg/config/config.go
```diff
@@ -147,6 +147,22 @@ type Homebrew struct {
 	Goarm                 string               `yaml:"goarm,omitempty"`
 }
 
+// Krew contains the krew section.
+type Krew struct {
```
ahmetb, line 151: Alternative thought: What if the user provided a template YAML file here in each krews[], and you filled it up by rendering the template? This is what [krew-release-bot](https://github.com/rajatjindal/krew-release-bot) does. That way, you get to preserve the shape of the YAML and only edit the pieces you're interested.

I understand Goreleaser helps a lot here by automatically generating `spec.platforms` but what if user still provided a YAML file that just lacked `spec.platforms` but most everything else like description etc was in place?

I suggest templating this because (1) we expect field ordering consistency between YAMLs of different plugins (2) I am not sure if stuff like caveats/description fields belong to .goreleaser.yml from a logical standpoint.

caarlos0: on the point 1, we can order the artifacts by goos/goarch so they are always in the same order... 

on point 2, we already have those fields for other things like homebrew... so I think it should be ok 🤔

internal/pipe/krew/krew.go
caarlos0: I saw, didn't use it because it comes with quite a few dependencies we don't need... seemed easy enough to copy the relevant parts

pkg/config/config.go
caarlos0: will note all those in the docs...

PR #2648, opened 2021-11-08
feat: add sbom generation

wagoodman:
This PR adds a new `sboms` section to generate SBOMs for one or more generated artifacts. Example usage:
```
sboms:
  -
    id: binary-sbom
    artifacts: binary

```
See doc updates in this PR for more detailed configuration explanation. With this approach SBOMs generated can be checksummed and referenced under the `signs` section.

_(This is a follow up from https://github.com/goreleaser/goreleaser/issues/2597#issuecomment-957842629 .)_


internal/pipe/pipe.go
```diff
@@ -25,6 +25,9 @@ var ErrSkipSignEnabled = Skip("artifact signing is disabled")
 // It means that the part of a Piper that validates some things was not run.
 var ErrSkipValidateEnabled = Skip("validation is disabled")
 
+// ErrSkipSBOMEnabled happens if --skip-sbom is set.
+var ErrSkipSBOMEnabled = Skip("artifact cataloging is disabled")
```
caarlos0, line 29: not being used anywhere...

internal/pipe/sbom/sbom.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/goreleaser/goreleaser/internal/tmpl"
+	"github.com/goreleaser/goreleaser/pkg/config"
+	"github.com/goreleaser/goreleaser/pkg/context"
+)
+
+// Pipe that catalogs common artifacts as an SBOM.
+type Pipe struct{}
+
+func (Pipe) String() string                 { return "cataloging artifacts" }
+func (Pipe) Skip(ctx *context.Context) bool { return len(ctx.Config.SBOMs) == 0 }
```
caarlos0, line 27:
```suggestion
func (Pipe) Skip(ctx *context.Context) bool { return ctx.SkipSBOMCataloging || len(ctx.Config.SBOMs) == 0 }
```

internal/pipe/sign/sign.go
```diff
@@ -77,11 +77,12 @@ func (Pipe) Run(ctx *context.Context) error {
 					artifact.ByType(artifact.UploadableSourceArchive),
 					artifact.ByType(artifact.Checksum),
 					artifact.ByType(artifact.LinuxPackage),
+					artifact.ByType(artifact.SBOM),
 				))
 			case "archive":
 				filters = append(filters, artifact.ByType(artifact.UploadableArchive))
-			case "binary":
-				filters = append(filters, artifact.ByType(artifact.UploadableBinary))
```
caarlos0, line 84: likely removed by mistake

internal/pipeline/pipeline.go
```diff
@@ -4,6 +4,8 @@ package pipeline
 import (
 	"fmt"
 
+	"github.com/goreleaser/goreleaser/internal/pipe/sbom"
+
```
caarlos0, line 8:
```suggestion
```

pkg/defaults/defaults.go
```diff
@@ -5,6 +5,8 @@ package defaults
 import (
 	"fmt"
 
+	"github.com/goreleaser/goreleaser/internal/pipe/sbom"
+
```
caarlos0, line 9:
```suggestion
```

internal/pipe/sbom/sbom.go
```diff
@@ -0,0 +1,222 @@
...
+			switch cfg.Artifacts {
+			case "binary":
+				sbom = "{{ .ArtifactName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}.sbom"
+			default:
+				sbom = "{{ .ArtifactName }}.sbom"
+			}
+			cfg.SBOMs = []string{sbom}
+		}
+		if len(cfg.Args) == 0 {
+			cfg.Args = []string{"--file", "{{ .Dist }}/$sbom0", "--output", "spdx-json", "$artifact"}
```
caarlos0, line 49: on the sign pipe prefix the file with the dist when setting the env... I would say we follow the same path here so both usages/impls look similar...

caarlos0 (commented):
I like this!

internal/pipe/sign/sign.go
wagoodman: indeed it was! I'll restore

internal/pipe/pipe.go
wagoodman: correct! With the addition of https://github.com/goreleaser/goreleaser/pull/2648#discussion_r747133802 now it should be 👍

cmd/release.go
```diff
@@ -139,6 +141,7 @@ func setupReleaseContext(ctx *context.Context, options releaseOpts) *context.Con
 	ctx.SkipAnnounce = ctx.Snapshot || options.skipPublish || options.skipAnnounce
 	ctx.SkipValidate = ctx.Snapshot || options.skipValidate
 	ctx.SkipSign = options.skipSign
+	ctx.SkipSBOMCataloging = options.skipSBOMCataloging
```
caarlos0, line 144: should `--skip-validate` infers `--skip-sbom`?

internal/artifact/artifact.go
```diff
@@ -56,6 +56,8 @@ const (
 	GoFishRig
 	// ScoopManifest is an uploadable scoop manifest file.
 	ScoopManifest
+	// SBOM is a Software Bill of Materials file.
+	SBOM
 )
 
 func (t Type) String() string {
```
caarlos0, line 63: missing sbom bellow here and in the test

internal/pipe/pipe.go
caarlos0: seems like still unused... 

in any case, I don't think its really needed... the `Skip` method in the pipe handles it :)

internal/pipe/sbom/sbom.go
```diff
@@ -0,0 +1,228 @@
...
+	templater := tmpl.New(ctx).WithEnv(env)
+	if a != nil {
+		env["artifact"] = a.Path
+		env["artifactID"] = a.ID()
+
+		templater = templater.WithArtifact(a, nil)
+
+		var paths []string
+		for idx, sbom := range cfg.SBOMs {
+			path, err := templater.Apply(filepath.Join("{{ .Dist }}", expand(sbom, env)))
```
caarlos0, line 141: could be 
```suggestion
			path, err := templater.Apply(filepath.Join(ctx.Dist, expand(sbom, env)))
```

internal/tmpl/tmpl.go
```diff
@@ -27,6 +27,7 @@ type Fields map[string]interface{}
 const (
 	// general keys.
 	projectName     = "ProjectName"
+	dist            = "Dist"
```
caarlos0, line 30: I think this is not needed... our internal code can use `ctx.Config.Dist` and user-configs should be scoped to dist already...

caarlos0 (commented):
Its looking good!

I think we now need:
- more tests
- docs
- `task fmt`/`task lint` to make ci happy

let me know how I can help :)

wagoodman:
Righto @caarlos0 , thanks for the extra review (the comments are helpful ❤️ ). I'll adjust the current branch and start adding docs / tests (it seems like the direction has been settled)... when I'm done I'll take it out of draft (🎉 ).

caarlos0:
amazing, let me know if you need any help in the meantime (I'm on our dicord and also gophers slack as well)

cmd/release.go
wagoodman: My take is that skipping validation shouldn't imply skipping SBOM generation. That is, cataloging artifacts isn't a form of validation and indeed creates artifacts of their own.

internal/tmpl/tmpl.go
wagoodman: Circling back to this --I added `Dist` as an option to allow for tool arguments that have to deal with outputting files to allow them to specify the dist directory explicitly:
```
...
-
  Cmd: my-tool
  Args:
    - --file
    - "{{ .Dist }}/some.sbom"
...
```

wagoodman: After more thought I decided to go the route of changing the cwd of the SBOM tool to `Dist` and removing `Dist` from templating. This way all cataloging tools are oriented towards what goreleaser generates instead of another location. An additional pro is that it makes it harder to misuse the `$sbom#` variables (you don't need to remember to prefix `{{ .Dist }}/` to all paths). 

The downside of this is that if tooling depends on configuration in the root of the repo this will now need to be explicitly passed in as an option.

cmd/release.go
caarlos0: hmm, yeah, that makes sense...

internal/tmpl/tmpl.go
caarlos0: yes, that was my idea as well!

sorry it took so long to reply, crazy week :P

wagoodman:
@caarlos0 I think this one is ready for review 🎉 , when CI gets kicked off I'll come back and see what the results were 🤞

caarlos0 (commented):
just a couple of comments, but, looking good!

I also played with it locally, and it works like a charm!

I would say we can enable this for archives on our own `.goreleaser.yaml` as well, dogfooding and all. 

Probably sign the sboms as well? Would generate a lot of files, but should be ok I think 🙇‍♂️

www/mkdocs.yml
```diff
@@ -91,6 +91,8 @@ nav:
     - customization/snapcraft.md
     - customization/docker.md
     - customization/docker_manifest.md
+  - Software Bill of Materials:
+      - Cataloging artifacts: customization/sbom.md
```
caarlos0, line 95: this could probably be just:

```suggestion
  - customization/sbom.md
```

www/docs/customization/sbom.md
```diff
@@ -0,0 +1,100 @@
...
+The artifact cataloging step can analyze one or more artifacts generated by Goreleaser and output one or more
+SBOM files into the dist directory.
+
+The default is configured to create an SBOM for each binary produced with [Syft](https://github.com/anchore/syft).
+To enable artifact cataloging just add:
+
+```yaml
+# .goreleaser.yml
+sboms:
+  - artifacts: binary
```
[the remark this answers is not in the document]
wagoodman: yeah, that makes a lot of sense. I'll change the default artifacts value and update the docs example

wagoodman:
@caarlos0 thanks! I added the new `sboms` section to the local `.goreleaser.yaml` and also added a new `signs` config. You certainly won't have a shortage of files on releases!

wagoodman:
I'm going to add a few more tests under `sbom_test.go` to get coverage up and I realized I'm missing tests for `sbom` artifacts under `signs_test.go`, so I'll add those too.

www/docs/customization/sbom.md
```diff
@@ -0,0 +1,102 @@
...
+These environment variables might be available in the fields that are templateable:
+
+- `${artifact}`: the path to the artifact that will be cataloged (unless "artifacts" config item is "any")
+- `${artifactID}`: the ID of the artifact that will be cataloged (unless "artifacts" config item is "any")
+- `${document}`:  the SBOM filename generated (corresponds to `${document0}` if the "artifacts" config item is "any")
+- `${document#}`: the SBOM filenames generated, where `#` corresponds to the list index under the "documents" config item (e.g. `${document0}`)
+
+## Limitations
+
+Container images generated by Goreleaser are not available to be cataloged by the SBOM tool.
```
caarlos0, line 101: As far as I tested, it don't work with macos universal binaries as well... so its probably worth listing it here too..

wagoodman: I think I can get that added to syft and released shortly. I'll update here if it ends up being too large of a lift / too long.

caarlos0: ah, that sounds good too!

wagoodman: Syft release v0.32.0 now has mac multi-architecture support! (updated the pipeline to use the latest release too) 🎉

caarlos0: niceee

caarlos0:
Thanks @wagoodman!

wagoodman:
Righto! Twas fun to develop --Thanks for the conversations and reviews 🙌!

PR #2649, opened 2021-11-08
feat: add sbom generation pipe (after signing)

wagoodman:
**This is a PoC only not intended for merging**

This is a follow up from https://github.com/goreleaser/goreleaser/issues/2597#issuecomment-957842629 .

This is also an alternative to https://github.com/goreleaser/goreleaser/pull/2648 , where the SBOM generation process takes place after artifact signing. This is to allow for cataloging of container artifacts:

```
  -
    id: container-sbom
    artifacts: container
    ids:
      - primary-image
```

The downside to this approach is that SBOMs cannot be captured in the `checksum` and `signs` sections directly.

This additionally required changing the `docker.Pipe` object to always output an artifact even if publishing is being skipped.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ www/docs/static/releases*.json
 completions/
 .vscode/
 .task/
+.idea/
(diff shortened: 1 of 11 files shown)


wagoodman:
Closing this in favor of #2648 , see https://github.com/goreleaser/goreleaser/issues/2597#issuecomment-971976928

Issue #2647, opened 2021-11-08
[Bug]: changelog is too long


till:
### What happened?

I am re-publishing an old code base that hasn't been tagged before and currently the github release fails as it exceeds the number of allowed characters for the changelog:

```
  ⨯ release failed after 167.72s error=github/gitlab/gitea releases: failed to publish artifacts: POST https://api.github.com/repos/org/repository/releases: 422 Validation Failed [{Resource:Release Field:body Code:custom Message:body is too long (maximum is 125000 characters)}]
```

### How can we reproduce this?

Have a changelog with more than `125000` characters. Unfortunately I can't give you access to the builds.

The workflow I am using is very straight forward:

```
---
name: release

on:
  push:
    tags:
      - 'v*'

jobs:
  release:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
        with:
          fetch-depth: 0
      - uses: actions/setup-go@v2
        with:
          go-version: 1.17
      - uses: docker/login-action@v1
        with:
          registry: registry.example.org
          username: ${{ secrets.DOCKER_USERNAME }}
          password: ${{ secrets.DOCKER_PASSWORD }}
      - uses: goreleaser/goreleaser-action@v2
        with:
          version: latest
          args: release --rm-dist
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

Everything works (building and publishing docker images), just the release on Github doesn't get created and therefor no binaries are uploaded.

### goreleaser version

```bash
Via `goreleaser/goreleaser-action@v2`: v0.184.0
```

### GoReleaser Check

- [X] goreleaser check shows no errors

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

dirien:
Hi @caarlos0, i can have a look into it.

I would suggest, if this is a hard cap from the github api to truncate the text to the max char size.

caarlos0:
> I would suggest, if this is a hard cap from the github api to truncate the text to the max char size.

yes, that would be my idea as well... 

feel free to tackle this :)

till:
Thanks, both of you! :)

Easy fix is good fix.

dirien:
thanks @caarlos0. Will tackle this first thing.

dirien:
hi @till and @caarlos0, sorry took a day longer :/

till:
[...]

PR #2661, opened 2021-11-10
fix: [Bug]: changelog is too long

dirien:

fixes: #2647

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -12,6 +12,13 @@ import (
 	"github.com/goreleaser/goreleaser/pkg/context"
 )
 
+const (
+	// maxReleaseBodyLength defines the max characters size of the body
+	maxReleaseBodyLength = 125000
+	// ellipsis to be used when release notes body is too long
+	ellipsis = "..."
+)
+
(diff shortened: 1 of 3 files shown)


till (approved):
🎉

internal/pipe/release/body.go
```diff
@@ -67,5 +72,10 @@ func describeBody(ctx *context.Context) (bytes.Buffer, error) {
 		ReleaseNotes: ctx.ReleaseNotes,
 		DockerImages: dockers,
 	})
+	// Truncate the release notes if it's too long (github doesn't allow more than 125000 characters)
+	if out.Len() > maxReleaseNotesBodyLength {
+		out.Truncate(maxReleaseNotesBodyLength - len(ellipsis))
+		out.Write([]byte(ellipsis))
+	}
```
caarlos0, line 79: since this is github-specific, maybe we should put it in the [`github` impl](https://github.com/goreleaser/goreleaser/blob/master/internal/client/github.go) only?

dirien: yepp you are right, I changed the location to the github impl. Thanks for the input.

caarlos0:
noice, thanks!

PR #2663, opened 2021-11-12
feat: nfpm conventional file name

caarlos0:
adds a new template variable on nfpm config with the result of [`ConventionalFileName`](https://github.com/goreleaser/nfpm/blob/v2.8.0/nfpm.go#L100) of each packager.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -157,7 +157,7 @@ scoop:
   license: MIT
 
 nfpms:
-  - file_name_template: '{{ .ProjectName }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}'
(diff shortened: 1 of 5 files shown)


radeksimko (approved):
The change in itself looks reasonable but it's not clear to me what end-user use case are you hoping to address with it? 🤔

caarlos0:
> The change in itself looks reasonable but it's not clear to me what end-user use case are you hoping to address with it? 🤔

if people want to create the debs/rpm/etc and manage their own deb/rpm/etc repositories, usually these filenames need to be named properly... this allows to more easily do that...

radeksimko:
Oh, I see so the _convention_ here means convention of the package managers/repositories nfpm is targeting - that makes sense! I'm so out of touch with Linux package management internals these days that I forgot how this is important. 😄

PR #2670, opened 2021-11-15
feat: CHANGELOG categories

### LINKED ISSUE #2669: CHANGELOG categories 🙏
I think it would be cool to sort the commits in the CHANGELOG by type.

I've taking to using your style of `labeling` my commits with `feat:` or `fix:` etc etc and I think it'd be cool for goreleaser to auto create Markdown sections based on how I am labeling my commits for example:

- fix: bug in parser
- feat: one word... 🌈 rainbows!
- chore: updated deps
- chore: reorg of file parser folders

Would make:

---

### Features

 - `COMMIT` one word... 🌈 rainbows! (userid)
 
 ### Fixes
 
 - `COMMIT` bug in parser (userid)
 
 ### Minor Fixes
 
 - `COMMIT` updated deps (userid)
 - `COMMIT` reorg of file parser folders (userid)
 
 ---

Also I really like the Native Github API CHANGELOG, but just as a summary.  So you could have both and it would end with a:

### Summary

Full Changelog: v1.0.11...v1.0.12

### Contributors

AVATARs

dirien:
Fixes #2669 request from @blacktop

The implementation could look like this:

image

The example is in a test repo, with no commits so dont wonder in looks blant. And the `use` is set to `git` so no userids. You can change this if you want. So it looks like this:

image

I tried to follow the [guidance](https://www.conventionalcommits.org/en/v1.0.0/) from @aalmiray and sorted the commits in features, Bug fixes and Others. 

You need to follow the conventional commits rules `<type>[optional scope]: <description>` to end in the right group. `feat:` goes to `Features`, `fix:` goes to `Bug Fixes`, etc.

You can activate the grouping feature via the flag `group: true` in the `changlog` block of the goreleaser config.

Looking for your feedback!

### DIFF
--- a/www/docs/customization/changelog.md
+++ b/www/docs/customization/changelog.md
@@ -25,6 +25,20 @@ changelog:
   # Default is empty
   sort: asc
 
+  # Group commits messages by given regex and title.
+  # Order value defines the order of the groups.
+  # Proving no regex means all commits will be grouped under the default group.
+  # Default is no groups.
+  groups:
+    - title: Features
+      regexp: "^.*feat[(\\w)]*:+.*$"
+      order: 0
+    - title: 'Bug fixes'
+      regexp: "^.*fix[(\\w)]*:+.*$"
+      order: 1
+    - title: Others
+      order: 999
+
   filters:
     # Commit messages matching the regexp listed here will be removed from
     # the changelog
--- a/internal/pipe/changelog/changelog.go
+++ b/internal/pipe/changelog/changelog.go
@@ -67,8 +67,40 @@ func (Pipe) Run(ctx *context.Context) error {
 
 	changelogElements := []string{
 		"## Changelog",
-		strings.Join(entries, changelogStringJoiner),
 	}
+
+	if len(ctx.Config.Changelog.Groups) > 0 {
+		log.Debug("grouping entries")
+		groups := ctx.Config.Changelog.Groups
+
+		sort.Slice(groups, func(i, j int) bool { return groups[i].Order < groups[j].Order })
+		for _, group := range groups {
+			items := make([]string, 0)
+			if len(group.Regexp) > 0 {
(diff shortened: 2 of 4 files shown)


blacktop:
This looks great!  Is it possible to create sub-categories for `optional scope` or does that over complicate things?

dirien:
@blacktop: good point. I am think if it will be deterministic enough. I would not want that some spelling differences will lead in subgroups with only one item.

dirien:
What do you think @caarlos0? Should we start first with un-scoped groups? I am not sure, I like the high level ordering currently.

www/docs/customization/changelog.md
```diff
@@ -25,6 +25,13 @@ changelog:
   # Default is empty
   sort: asc
 
+  # Group commits by the following criteria:
+  # - 'feat': commits that add new features.
+  # - 'fix': commits that fix existing issues.
+  # - 'other': all other commits.
+  # Default is false
+  group: true
```
caarlos0, line 33: I'm thinking if it would make sense to make this "configurable", e.g.:

```yaml
groups:
- title: Features
  regexp: "^.*(feat):*.*$"
  order: 0
- title:  'Bug fixes'
  regexp: "^.*(fix):*.*$"
  orde: 1
- title: Others
  order: 999
```

then the user could setup the categories as they wish.


wdyt?

dirien: @caarlos0 done!

dirien:
@caarlos0 added the very dynamic grouping logic. Attention: Groups without a regex is greedy and will collect all issues, the other groups in the list will not be considered further

caarlos0:
This looks amazing, thanks @dirien !

caarlos0:
PS: I also added this to our goreleaser config, please review and let me know what you think :)

PR #2672, opened 2021-11-16
fix: apk add --no-cache

caarlos0:
prevents apk cache

noticed it while playing with https://contains.dev/goreleaser/goreleaser/latest/layers :D


dirien (approved):
Good catch!

PR #2673, opened 2021-11-18
fix: add env to the sign schema

cpanato:
- schema: add env to the sign schema

/assign @caarlos0


caarlos0:
oh, this should have been auto-gen... 

thanks anyway!

cpanato:
hum, where to check why was not generated?

caarlos0:
its on the pro fork only because it has extra flags, will need to move some stuff around

caarlos0:
done in f545fee

PR #2676, opened 2021-11-19
refactor: use os and io packages instead of io/ioutil

Clivern:
`io/ioutil` is deprecated so i replaced it according to https://golang.org/doc/go1.16#ioutil


caarlos0:
neat, thanks!

PR #2677, opened 2021-11-21
feat: template on extra_files

### LINKED ISSUE #2655: Ability to Use Templates With checksum and release extra_files

As part of the required release process for newer Terraform Providers, a "manifest" file that matches the versioned naming convention of other release files must be included so the Terraform Registry ingest process works correctly: https://www.terraform.io/docs/plugin/framework/publishing.html#add-a-version-manifest

It's currently possible to use the `extra_files` `glob` support to fuzzy match the correct file naming (as shown below), however this could introduce false negatives or positives. Given that other portions of the `goreleaser` configuration support templates, it would be wonderful if it could be included with `extra_files` as well for parity.

The configuration could either support templates in the existing `glob` field, e.g.

```yaml
checksum:
  # ... other configuration ...
  extra_files:
    - glob: {{ .ProjectName }}_{{ .Version }}_manifest.json
release:
  # ... other configuration ...
  extra_files:
    - glob: {{ .ProjectName }}_{{ .Version }}_manifest.json
```

Or the extra files configuration could separately include a new template-supported field, e.g. `name_template`

```yaml
checksum:
  # ... other configuration ...
  extra_files:
    - name_template: {{ .ProjectName }}_{{ .Version }}_manifest.json
release:
  # ... other configuration ...
  extra_files:
    - name_template: {{ .ProjectName }}_{{ .Version }}_manifest.json
```

A current workaround that is possible today with `goreleaser` is moving/copying the file to the correct naming convention prior to executing the release then glob-ing against the expected pattern in `extra_files`, e.g.

```yaml
checksum:
  # ... other configuration ...
  extra_files:
    - glob: 'terraform-provider-*_*_manifest.json'
release:
  # ... other configuration ...
  extra_files:
    - glob: 'terraform-provider-*_*_manifest.json'
```

Unfortunately there is room for false negatives and false positives and it is a little unclear what exactly is desired without additional configuration comments.

**Additional context**

Thank you very much for this excellent project! Please let me know if you are interested in the feature and I would be happy to contribute the changes.

Aside: I will be filing a separate feature request to see if we could also have extra files support static file renaming as part of this process (so it doesn't need to also be done outside `goreleaser`), but it would likely best be paired with this feature so it matches a "source" and "target". Filed as https://github.com/goreleaser/goreleaser/issues/2656.

caarlos0:
closes #2655

### DIFF
--- a/internal/extrafiles/extra_files_test.go
+++ b/internal/extrafiles/extra_files_test.go
@@ -38,7 +38,7 @@ func TestBadTemplate(t *testing.T) {
 func TestShouldGetSpecificFile(t *testing.T) {
 	globs := []config.ExtraFile{
 		{},                        // empty glob, will be ignored
-		{Glob: "./testdata/sub3"}, // will get a file1.golden as well, but will be overriden
+		{Glob: "./testdata/sub3"}, // will get a file1.golden as well, but will be overridden
 		{Glob: "./testdata/file1.golden"},
 	}
 
--- a/internal/extrafiles/extra_files_test.go
+++ b/internal/extrafiles/extra_files_test.go
@@ -4,15 +4,45 @@ import (
 	"testing"
 
 	"github.com/goreleaser/goreleaser/pkg/config"
+	"github.com/goreleaser/goreleaser/pkg/context"
 	"github.com/stretchr/testify/require"
 )
 
+func TestTemplate(t *testing.T) {
+	globs := []config.ExtraFile{
+		{Glob: "./testdata/file{{ .Env.ONE }}.golden"},
+	}
+
+	ctx := context.New(config.Project{})
+	ctx.Env["ONE"] = "1"
+
+	files, err := Find(ctx, globs)
+	require.NoError(t, err)
+	require.Len(t, files, 1)
+
+	require.Equal(t, "testdata/file1.golden", files["file1.golden"])
+}
+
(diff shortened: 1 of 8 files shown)


internal/extrafiles/extra_files_test.go
```diff
@@ -4,15 +4,45 @@ import (
...
+
+	files, err := Find(ctx, globs)
+	require.Empty(t, files)
+	require.EqualError(t, err, `failed to apply template to glob "./testdata/file{{ .Env.NOPE }}.golden": template: tmpl:1:22: executing "tmpl" at <.Env.NOPE>: map has no entry for key "NOPE"`)
+}
+
 func TestShouldGetSpecificFile(t *testing.T) {
 	globs := []config.ExtraFile{
+		{},                        // empty glob, will be ignored
+		{Glob: "./testdata/sub3"}, // will get a file1.golden as well, but will be overriden
```
bflad, line 41: To appease the linter:

```suggestion
		{Glob: "./testdata/sub3"}, // will get a file1.golden as well, but will be overridden
```

bflad (commented):
I think overall this satisfies the use case well. 👍  I cannot think of a reason why supporting templates on top of glob syntax might cause issues since alternative matchers (`{` `}`) should still play well.

PR #2678, opened 2021-11-21
feat: remap extra_files

### LINKED ISSUE #2656: Ability to Remap Source Names with checksum and release extra_files

As part of the required release process for newer Terraform Providers, a "manifest" file that matches the versioned naming convention of other release files must be included so the Terraform Registry ingest process works correctly: https://www.terraform.io/docs/plugin/framework/publishing.html#add-a-version-manifest

This file is expected to be fairly static in nature and included with each Terraform Provider project, similar to managing a NPM module `package.json` in that ecosystem. The below examples use `terraform-registry-manifest.json`, however there is not an expected naming standard of the source since the Terraform Registry's requirement is the final versioned naming during its ingress process. 

The extra files configuration could introduce a source file name/template field, then paired with a "static" target file name/template field (see also: https://github.com/goreleaser/goreleaser/issues/2655), automatically handle the file name remapping as part of the release process, without actually copying/moving the file:

```yaml
checksum:
  # ... other configuration ...
  extra_files:
    - name_template: {{ .ProjectName }}_{{ .Version }}_manifest.json
      source: terraform-registry-manifest.json
release:
  # ... other configuration ...
  extra_files:
    - name_template: {{ .ProjectName }}_{{ .Version }}_manifest.json
      source: terraform-registry-manifest.json
```

Developers could theoretically move the file to match the next release naming before each release. This would be very tedious!

A more automated workaround that is possible today with `goreleaser` is moving/copying the file to the correct naming convention prior to executing the release, e.g. in GitHub Actions

```yaml
      - name: Copy Terraform Registry Manifest
        run: cp terraform-registry-manifest.json ${{ github.event.repository.name }}_${GITHUB_REF/refs\/tags\/v/}_manifest.json
      - name: goreleaser release
        uses: goreleaser/goreleaser-action@v2
```

Technically this introduces a dirty VCS state unless it is also ignored, e.g. via `.gitignore`, which is additional complexity for project maintainers who may not be familiar with this ecosystem.

**Additional context**

Thank you very much for this excellent project! Please let me know if you are interested in the feature and I would be happy to contribute the changes.

caarlos0:
depends on #2677 

closes #2656

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -8,8 +8,15 @@ linters:
     - wastedassign
     - revive
     - forbidigo
+    - tagliatelle
 
 linters-settings:
   forbidigo:
     forbid:
       - 'ioutil\.*'
+  tagliatelle:
+    case:
+      use-field-name: false
+      rules:
+        yaml: snake
+        json: snake
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -125,35 +125,101 @@ tasks:
       - echo {{.NEXT}}
       - git push origin --tags
 
+  goreleaser:test:pkg:
+    desc: Test a package
+    cmds:
+      - docker run --platform linux/{{.Platform}} --rm --workdir /tmp -v $PWD/dist:/tmp {{.Image}} sh -c '{{.Cmd}} && goreleaser --version'
+
   goreleaser:test:rpm:
     desc: Tests rpm packages
+    vars:
+      rpm: 'rpm --nodeps -ivh'
     cmds:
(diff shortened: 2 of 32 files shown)


bflad (commented):
This is very exciting 🎉  Do we think there is any concern that someone may try to configure only `name_template` and not `glob`? It looks like it will silently skip that extra files section, but maybe it warrants an upfront check just so the developer understands what is happening (or at least a warning log). It may also be good to clarify in the documentation that name_template is designed to be the target artifact name. 👍

caarlos0:
> Do we think there is any concern that someone may try to configure only `name_template` and not `glob`?

this might already happen if the glob evaluates to empty or is an empty string... it just skips it.

Will add a warn though.

PR #2683, opened 2021-11-22
feat: allow to use .PreviousTag on templates

caarlos0:
Previous tag will be available as a template field. It will be empty if there are no previous tag (e.g. first ever release of a repo).

Changelog still works by using the first commit on those cases (as it already does today).

### DIFF
--- a/internal/pipe/git/git.go
+++ b/internal/pipe/git/git.go
@@ -110,8 +110,11 @@ func getGitInfo() (context.GitInfo, error) {
 		}, ErrNoTag
 	}
 
-	// if we dont get a previous tag, its fine
-	previous, _ := getPreviousTag(tag)
+	previous, err := getPreviousTag(tag)
+	if err != nil {
+		// shouldn't error, will only affect templates
+		log.Warnf("couldn't find any tags before %q", tag)
+	}
 
 	return context.GitInfo{
 		Branch:      branch,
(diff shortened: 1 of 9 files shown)


caarlos0:
one problem with this: on the first tag, this will actually be a commit... 🤔 which per-se is not that bad, its just that the name will not convey what it actually is at some point 🤔

internal/pipe/git/git.go
```diff
@@ -203,6 +211,14 @@ func getTag() (string, error) {
 	return tag, err
 }
 
+func getPreviousTag(current string) (string, error) {
+	if tag := os.Getenv("GORELEASER_PREVIOUS_TAG"); tag != "" {
```
blacktop, line 215: :+1:

blacktop:
> one problem with this: on the first tag, this will actually be a commit... 🤔 which per-se is not that bad, its just that the name will not convey what it actually is at some point 🤔

Maybe a default `v0.0.0` would make sense?  I tried `v0.0.0` and a commit hash in github's URL and both 404ed (I know that is not the only purpose of this feature, however.  I kinda feel like a zero-ed out version would make more sense than a commit hash?

blacktop:
I'm not intimately familiar with your code, but everything looked good to me. 👍

caarlos0:
> Maybe a default `v0.0.0` would make sense? I tried `v0.0.0` and a commit hash in github's URL and both 404ed (I know that is not the only purpose of this feature, however. I kinda feel like a zero-ed out version would make more sense than a commit hash?

my comment was outdated, it will now be empty as stated in the docs... which makes sense, as if there is no previous tag, that's what I would expect...

dirien (approved):
LGTM 👍🏻

PR #2689, opened 2021-11-23
chore: update strings when using a custom distribution directory

cpanato:
When using a custom distribution directory goreleaser outputs the commands/information using the default distribution directory `./dist`

This PR updates the strings to use the custom name if that is set, to avoid any confusion for the user when reading the logs


caarlos0:
Thanks!

PR #2694, opened 2021-11-23
chore: avoid task error on fmt

wesleimp:
removing the `-s` flag since it's been removed in [`gofumpt v0.2.0`](https://github.com/mvdan/gofumpt/blob/master/CHANGELOG.md#020---2021-11-10)


caarlos0:
Thanks!

Issue #2697, opened 2021-11-24
chore: Rename yml to yaml


dirien:
**Is your feature request related to a problem? Please describe.**
According to [yaml.org FAQ](https://yaml.org/faq.html):

> Please use ".yaml" when possible.

**Describe the solution you'd like**
Rename `.golangci.yml` and `.goreleaser.yml` to `.golangci.yaml` and `.goreleaser.yaml`

Unfortunatly is Taskfile not supporting yaml atm -> https://github.com/go-task/task/issues/584

PR #2698, opened 2021-11-24
chore: Rename yml to yaml

dirien:
According to yaml.org FAQ:

> Please use ".yaml" when possible.

Fixes: #2697


caarlos0:
¯\_(ツ)_/¯

PR #2700, opened 2021-11-24
feat: add git summary support (aka git describe --dirty --tags --always)

ekristen:
If applied, this commit will make `git describe --dirty --tags --always` available under the `context.GitInfo` as `Summary` and ultimately available in template replacements as `.Summary`

This change is being made because the above commands output is very useful as an alternative for build outputs.

This command when run on a tagged commit will return just the tag, when run on any other commit, it will provide `TAG-N-COMMIT` where N is the number of commits since last tag, and commit is the short commit id.

I would be happy to rename this to any other field that makes more sense, I prefer the `TAG-N-COMMIT` vs `TAG-SNAPSHOT-COMMIT` format, this just allows options. With further search and replace, you can change the `-` to `.` or `+`, depending on your semver desires.

## Examples

Simple

```yaml
snapshot:
  name_template: '{{ .Summary }}'
```

Advanced, if you want to replace `-` with `.`

```yaml
snapshot:
  name_template: '{{ replace .Summary "-" "." }}'
```

### DIFF
--- a/internal/pipe/git/git.go
+++ b/internal/pipe/git/git.go
@@ -45,6 +45,7 @@ var fakeInfo = context.GitInfo{
 	Commit:      "none",
 	ShortCommit: "none",
 	FullCommit:  "none",
+	Summary:     "none",
 }
 
 func getInfo(ctx *context.Context) (context.GitInfo, error) {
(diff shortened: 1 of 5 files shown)


wesleimp:
Noice. 😃

I'm only missing some tests.

ekristen:
@wesleimp as soon as I figure out how to test this I'll add another commit. :)

ekristen:
I added some tests for the Summary field, it's possible that I could add a more specific test set, but I need to test out this testlib a bit more if we want a more comprehensive test of a more complex example.

wesleimp:
> I added some tests for the Summary field, it's possible that I could add a more specific test set, but I need to test out this testlib a bit more if we want a more comprehensive test of a more complex example.

@ekristen I think that a good and simple test would be to check if the `.Summary` contains `.Tag` and `.Commit`. eg

```go
require.Contains(t, ctx.Git.Summary, ctx.Git.Tag)
require.Contains(t, ctx.Git.Summary, ctx.Git.Commit)
```

I think it's enough to ensure the `.Summary` field. WDYT @caarlos0?

caarlos0 (approved):
LGTM, thanks!

ekristen:
Thanks for the quick response! Amazing.

PR #2701, opened 2021-11-25
feat: allow a --output flag on goreleaser build

### LINKED ISSUE #2685: Allow output template changes for single-target mode

For `--single-target` builds where the user wants to just get a binary output, it would be nice to be able to override the `dist` and `name_template`. This way our project can have a simple `make build` that outputs a `./project_name` binary.

A configuration option like this:

```yaml
single_target:
  destination: './'
  name_template: "{{ .Binary }}"
```

We could have a post-build script that moves the `dist/binary_name` into the local dir, but I think it would be cleaner if the single-target mode moved the built binary into place.

**Additional context**

This would help with the transition away from [the Prometheus promu builder](https://github.com/prometheus/promu).

caarlos0:
closes #2685

This would add similar behavior as `go build`, a `-o` flag.

so, when you `goreleaser build --single-target -o foo`, it will save the binary as `foo` (if you have only one build config, otherwise this also needs the `--id` flag).

Hopefully this helps replacing some build tasks on makefiles :)

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"runtime"
 	"time"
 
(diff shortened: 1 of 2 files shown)


mgrachev (approved):
LGTM 👌

SuperQ:
I was thinking it would be better to use a config file option, so multiple build IDs could be done in one command.

caarlos0:
hmm, if that's the case, then yeah, it might make more sense... will do as you suggested

SuperQ:
I'm also happy to help work on this.

caarlos0:
> I'm also happy to help work on this.

feel free to do so if you want :)

caarlos0:
@SuperQ what about this: when using --single-target, it always copies the binary to the root folder (with same name already setup on `.goreleaser.yaml`), and you can still rename with `--output`... what do you think?

PR #2708, opened 2021-11-29
feat: remove docker images from changelog

### LINKED ISSUE #2707: [proposal] do not automatically list docker images on release notes
In the beginning, GoReleaser didn't had customization options regarding the release notes - like footer and header. 

That's one of the reasons why we always listed the docker images in the release notes.

Now, we have those options and it would be easy to add them as a footer.

The reason I propose this is that, if you want to add something to the footer that is related to the changelog, it will actually be after the docker images list, which is a bit weird.

Also, some people might not want to list docker images at all, or list them without the `docker pull` part, or something else.

I think not automatically adding them would be a good move long-term.

But that's just what I think, and I would love to hear your thoughts as well :) cc/ @goreleaser/goreleaser

caarlos0:
Rationale is in #2707, just opening it here fwiw

closes #2707

### DIFF
--- a/internal/pipe/release/body.go
+++ b/internal/pipe/release/body.go
@@ -2,55 +2,26 @@ package release
 
 import (
 	"bytes"
-	"strings"
 	"text/template"
 
-	"github.com/goreleaser/goreleaser/internal/artifact"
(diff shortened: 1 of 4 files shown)


caarlos0:
I think this should be ok to merge.

PR #2709, opened 2021-11-29
feat: no main error page

caarlos0:
refs https://github.com/goreleaser/goreleaser-action/issues/317

### DIFF
--- a/www/docs/errors/no-main.md
+++ b/www/docs/errors/no-main.md
@@ -2,6 +2,14 @@
 
 This usually happens if you're trying to build a library or if you didn't setup the `builds.main` section in your `.goreleaser.yml` and you `main.go` is not in the root folder.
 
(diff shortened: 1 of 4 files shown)


www/docs/errors/no-main.md
```diff
@@ -0,0 +1,25 @@
+# Build does not contain a main function
+
+This usually happens if you're trying to build a library or if you didn't setup the `builds.main` section in your `.goreleaser.yml` and you `main.go` is not in the root folder.
```
greenpau, line 3: @caarlos0 , it is worth adding the error message. May improve SEO.

```
build failed after 0.11s error=build for ovn_exporter does not contain a main function
```

caarlos0: good catch ,thanks! 🙏

PR #2734, opened 2021-12-05
feat: Use templates for repo's Name and Owner

SVilgelm:
Fixes a gap in re-using the ProjectName and other template Fields in the repo's Name and Owner

It helps a lot in re-using a goreleaser config in new projects, so only `project_name` can be replaced with the name of new project.


caarlos0:
doesn't extracting it from the repo url work in your case? if no, why?

SVilgelm:
I'm not sure I understand how the extracting can help me, maybe I was not clear enough.
Here is my template: https://github.com/sv-tools/go-repo-template/blob/main/.github/goreleaser-cli.yml

when I create a new project, I just copy the config and change two fields: `project_name` and `release.github.name`
this PR will help me in changing only the `project_name` filed for new projects

I want to have be able to use the templates in repo's name:
```yaml
project_name: go-repo-template

release:
  github:
    owner: sv-tools
    name: "{{ .ProjectName }}"
  prerelease: auto
```

SVilgelm:
or do you mean that I need to remove 
```yaml
  github:
    owner: sv-tools
    name: go-repo-template
```
and the goreleaser will pick it up automatically?

caarlos0:
you don't really need to set this bit:

```yaml
  github:
    owner: sv-tools
    name: "{{ .ProjectName }}"
```

goreleaser infers it from the repository remote URL... 

you only need to set it if the URL doesn't match the actual repo (e.g. renamed) or if its failing for some reason...

SVilgelm:
cool, thank you, I didn't know about this, but I still think it would be great if we can use the templates for repos as well. I was only one place where I was not able to use templates

SVilgelm:
I had couple of old projects where the actual url and repo were different, we did not use the goreleaser for those projects, but it can be one more reason to use it

PR #2741, opened 2021-12-06
Fix JSON schema for ReleaseNotesMode

dolmen:
In the JSON schema, the `keep-existing` value [was missing](https://github.com/goreleaser/goreleaser/blob/167d95e67fe0266fd75bc5dffc949f5db550dc73/www/docs/static/schema.json#L1716) for `ReleaseNotesMode`.
```
					"mode": {
						"enum": [
							"append",
							"prepend",
							"replace"
						],
						"type": "string",
						"default": "keep-existing"
					}
```

I fixed a typo in the struct tags and rebuilt the schema (`task schema:generate`). This adds `keep-existing` in the enum.

Note: [`schema-pro.json`](https://github.com/goreleaser/goreleaser/blob/167d95e67fe0266fd75bc5dffc949f5db550dc73/www/docs/static/schema-pro.json#L1884) is also affected, but I don't know how it is generated.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -477,7 +477,7 @@ type Release struct {
 	Header                 string      `yaml:"header,omitempty"`
 	Footer                 string      `yaml:"footer,omitempty"`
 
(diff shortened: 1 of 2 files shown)


caarlos0:
Good catch, thanks!

PR #2745, opened 2021-12-07
feat: add support for GitLab Generic Packages

### LINKED ISSUE #2006: Feature request: Switch GitLab release uploads to use Generic Packages Repository
With GitLab [Generic Packages Repository](https://docs.gitlab.com/ee/user/packages/generic_packages/#publish-a-generic-package-by-using-cicd) being available in Core/Free, it would be a better target for uploading binaries than the currently used [Upload Project File API endpoint](https://docs.gitlab.com/ce/api/projects.html#upload-a-file).

There are multiple benefits over "project files":
* The users can browse Generic Packages separately from and in addition to Releases in the GitLab UI
* Can still be used as release assets, as all you specify there is a URL
* API access for _downloading_ binaries is not supported for "project files", but is for Generic Packages

Replace the upload method for GitLab releases to use the Generic Packages API, making sure the Release Asset URLs are correspondingly replaced.

Using the HTTP uploads feature to _additionally_ upload Generic Packages, as is possible with #2002 and #2005 combined. However, it requires additional configuration by the user, and results in two uploads per binary - one for "project files" and one for Generic Packages.

**Additional context**
GitLab documentation says the following about Generic Packages:

> Version history
> 
> - Introduced in GitLab 13.5.
> - It’s deployed behind a feature flag, enabled by default.
> - It’s enabled on GitLab.com.
> - It’s able to be enabled or disabled per-project.
> - It’s recommended for production use.
> - For GitLab self-managed instances, GitLab administrators can opt to disable it. 

- https://docs.gitlab.com/ee/user/packages/generic_packages/#publish-a-generic-package-by-using-cicd

neomantra:
GitLab's Generic Package Repository allows uploading files
to GitLab, as an alternative to attachments.

Added the flag `use_package_registry` to the `gitlab_urls` config.
When false, the default, it will behave as before.
When true, it will publish the file to the Generic Package Registry.
Either way, the URL is bound to the release the same

Updated `go-gitlab` to version 0.52.2 which has the new
`GenericPackages.PublishPackageFile` function.

----
Fixes #2006.  

I tried to run local tests on it and it was not working (failures with `internal/pipe` and weird assembly output);  however, I think that is unrelated to this PR as I've tested this many times before without issue.  My apologies if this is not the case and the CI fails.

### DIFF
--- a/www/docs/scm/gitlab.md
+++ b/www/docs/scm/gitlab.md
@@ -28,6 +28,9 @@ gitlab_urls:
   download: https://gitlab.company.com
   # set to true if you use a self-signed certificate
   skip_tls_verify: false
+  # set to true if you want to upload to the Package Registry rather than attachments
+  # Only works with GitLab 13.5+
+  use_package_registry: false
 ```
 
 If none are set, they default to GitLab's public URLs.
@@ -37,6 +40,19 @@ If none are set, they default to GitLab's public URLs.
     on [release](https://docs.gitlab.com/ee/user/project/releases/index.html) functionality
     and [direct asset linking](https://docs.gitlab.com/ee/user/project/releases/index.html#permanent-links-to-release-assets).
 
+## Generic Package Registry
+
+GitLab introduced the [Generic Package Registry](https://docs.gitlab.com/ee/user/packages/package_registry/index.html) in Gitlab 13.5.
+
+Normally, `goreleaser` uploads release files as "attachments", which may have [administrative limits](https://docs.gitlab.com/ee/user/admin_area/settings/account_and_limit_settings.html).  Notably, hosted gitlab.com instances have a 10MB attachment limit which cannot be changed.
+
+Uploading to the Generic Package Registry does not have this restriction.  To use it instead, set `use_package_registry` to `true`.
+
+```yaml
+# .goreleaser.yml
+gitlab_urls:
+  use_package_registry: true
+```
 
 ## Example release
 
--- a/internal/client/gitlab.go
+++ b/internal/client/gitlab.go
@@ -377,19 +377,41 @@ func (c *gitlabClient) Upload(
 		projectID = ctx.Config.Release.GitLab.Owner + "/" + projectID
 	}
 
-	log.WithField("file", file.Name()).Debug("uploading file")
-	projectFile, _, err := c.client.Projects.UploadFile(
-		projectID,
-		file.Name(),
-		nil,
-	)
-	if err != nil {
-		return err
+	var baseLinkURL string
+	if ctx.Config.GitLabURLs.UsePackageRegistry {
+		log.WithField("file", file.Name()).Debug("uploading file as generic package")
+		_, _, err := c.client.GenericPackages.PublishPackageFile(
+			projectID,
+			ctx.Config.ProjectName,
+			ctx.Version,
+			artifact.Name,
+			file,
+			nil,
+		)
+		if err != nil {
+			return err
+		}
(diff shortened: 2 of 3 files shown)


www/docs/scm/gitlab.md
```diff
@@ -37,6 +40,14 @@ If none are set, they default to GitLab's public URLs.
     on [release](https://docs.gitlab.com/ee/user/project/releases/index.html) functionality
     and [direct asset linking](https://docs.gitlab.com/ee/user/project/releases/index.html#permanent-links-to-release-assets).
 
+## Generic Package Registry
+
+GitLab introduced the [Generic Package Registry](https://docs.gitlab.com/ee/user/packages/package_registry/index.html) in Gitlab 13.5.  Normally, `goreleaser` uploads release files as "attachments", which may have [administrative limits](https://docs.gitlab.com/ee/user/admin_area/settings/account_and_limit_settings.html).  Notably, hosted gitlab.com instances have a 10MB attachment limit which cannot be changed.  Uploading to the Generic Package Registry does not have this restriction.  To use it instead, set `use_package_registry` to `true`.
+
+```
```
caarlos0, line 47:
```suggestion
```yaml
# .goreleaser.yml
```

www/docs/scm/gitlab.md
```diff
@@ -37,6 +40,14 @@ If none are set, they default to GitLab's public URLs.
     on [release](https://docs.gitlab.com/ee/user/project/releases/index.html) functionality
     and [direct asset linking](https://docs.gitlab.com/ee/user/project/releases/index.html#permanent-links-to-release-assets).
 
+## Generic Package Registry
+
+GitLab introduced the [Generic Package Registry](https://docs.gitlab.com/ee/user/packages/package_registry/index.html) in Gitlab 13.5.  Normally, `goreleaser` uploads release files as "attachments", which may have [administrative limits](https://docs.gitlab.com/ee/user/admin_area/settings/account_and_limit_settings.html).  Notably, hosted gitlab.com instances have a 10MB attachment limit which cannot be changed.  Uploading to the Generic Package Registry does not have this restriction.  To use it instead, set `use_package_registry` to `true`.
```
caarlos0, line 45:
```suggestion
GitLab introduced the [Generic Package Registry](https://docs.gitlab.com/ee/user/packages/package_registry/index.html) in Gitlab 13.5.
Normally, `goreleaser` uploads release files as "attachments", which may have [administrative limits](https://docs.gitlab.com/ee/user/admin_area/settings/account_and_limit_settings.html).  Notably, hosted gitlab.com instances have a 10MB attachment limit which cannot be changed.
Uploading to the Generic Package Registry does not have this restriction.  To use it instead, set `use_package_registry` to `true`.
```

caarlos0 (commented):
looking good overall, added just a couple of comments

www/docs/scm/gitlab.md
neomantra: Sure, changing this now -- did you make (and want)  the `not`  in line 47 the color red somehow or is that just how GitHub is showing it?

neomantra: OK pushed that (without anything with `not` coloring).

neomantra:
I have rebased this PR to main and squashed the doc PR commit.   
There are less file changes now (e.g. `go.mod`) because `go-gitlab` was upgraded to `0.52.2` in  2a340e9125f417495eac191af03c880e61004344.

caarlos0:
LGTM! Thanks for the hard work on this!

PR #2750, opened 2021-12-08
feat: add generic webhook announcer

### LINKED ISSUE #2711: HTTP Announcer
If I had a custom service that accepted an HTTP request, I'd like to be able to send an "announcement" by sending an arbitrary HTTP request to a service I own.

Support to send an http request as part of the "announcement" step.

A custom publisher perhaps? But announcing is a step _after_ publishing.

**Additional context**
n/a

dirien:
fixes #2711 

How to test:

Head over to https://webhook.site/ and get a webhook endpoint.

Add following snippet to your `goreleaser.yaml`

```yaml
announce:
  webhook:
    enabled: true
    check_cert: true

    message_template: '{ "message": "{{ .ProjectName }} {{ .Tag }} is out! Check it out at {{.ReleaseURL }}"}'

    endpoint_url: "<endpoint>"
```

Lookig for your feedback!

### DIFF
--- a/internal/pipe/webhook/webhook.go
+++ b/internal/pipe/webhook/webhook.go
@@ -1,0 +1,119 @@
+package webhook
+
+import (
+	"crypto/tls"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/apex/log"
+	"github.com/caarlos0/env/v6"
+	"github.com/goreleaser/goreleaser/internal/tmpl"
+	"github.com/goreleaser/goreleaser/pkg/context"
+)
+
+const (
+	defaultMessageTemplate = `{ "message": "{{ .ProjectName }} {{ .Tag }} is out! Check it out at {{ .ReleaseURL }}"}`
+	ContentTypeHeaderKey   = "Content-Type"
+	UserAgentHeaderKey     = "User-Agent"
+	UserAgentHeaderValue   = "gorleaser"
+	AuthorizationHeaderKey = "Authorization"
+	DefaultContentType     = "application/json; charset=utf-8"
+)
+
+type Pipe struct{}
+
+func (Pipe) String() string                 { return "webhook" }
+func (Pipe) Skip(ctx *context.Context) bool { return !ctx.Config.Announce.Webhook.Enabled }
+
+type Config struct {
+	BasicAuthHeader   string `env:"BASIC_AUTH_HEADER_VALUE"`
+	BearerTokenHeader string `env:"BEARER_TOKEN_HEADER_VALUE"`
+}
+
+func (p Pipe) Default(ctx *context.Context) error {
+	if ctx.Config.Announce.Webhook.MessageTemplate == "" {
+		ctx.Config.Announce.Webhook.MessageTemplate = defaultMessageTemplate
+	}
+	if ctx.Config.Announce.Webhook.ContentType == "" {
+		ctx.Config.Announce.Webhook.ContentType = DefaultContentType
+	}
+	return nil
+}
+
+func (p Pipe) Announce(ctx *context.Context) error {
+	var cfg Config
+	if err := env.Parse(&cfg); err != nil {
+		return fmt.Errorf("announce: failed to announce to webhook: %w", err)
+	}
+
+	endpointURLConfig, err := tmpl.New(ctx).Apply(ctx.Config.Announce.Webhook.EndpointURL)
+	if err != nil {
+		return fmt.Errorf("announce: failed to announce to webhook: %w", err)
+	}
+	if len(endpointURLConfig) == 0 {
+		return errors.New("announce: failed to announce to webhook: no endpoint url")
+	}
+
+	if _, err := url.ParseRequestURI(endpointURLConfig); err != nil {
+		return fmt.Errorf("announce: failed to announce to webhook: %w", err)
+	}
+	endpointURL, err := url.Parse(endpointURLConfig)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


dirien:
Add Auth-support via ENV variables:

```
If your endpoints are not secure, you can use following environment variables to configure them:

- BASIC_AUTH_HEADER_VALUE like `Basic <base64(username:password)>`
- BEARER_TOKEN_HEADER_VALUE like `Bearer <token>`
```

caarlos0 (commented):
maybe we could use httptest to test more of the behavior.

also not sure if we need to start already with all those tls options... maybe we can leave them out initially? thoughts?

other than that, lgtm

dirien:
> maybe we could use httptest to test more of the behavior.
> 
> also not sure if we need to start already with all those tls options... maybe we can leave them out initially? thoughts?
> 
> other than that, lgtm

Very good, I removed the mTLS part. We can add this later if needed and pumped up the Test with a httptest Webserver.

internal/pipe/webhook/webhook_test.go
```diff
@@ -0,0 +1,258 @@
...
+		Response: "Thanks for the announcement!",
+		UUID:     uuid.New(),
+	}
+
+	srv := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		defer r.Body.Close()
+		body, err := io.ReadAll(r.Body)
+		if err != nil {
+			panic(err)
+		}
```
caarlos0, line 107: why not

```suggestion
		require.NoError(t, err)
```

?

dirien: I was not sure if its part of the test or a testsupporting failure 😄

caarlos0: it works! there are several of those in this file, I think we can replace them all :D

dirien: done

internal/pipe/webhook/webhook.go
```diff
@@ -0,0 +1,158 @@
...
+func (p Pipe) Default(ctx *context.Context) error {
+	if ctx.Config.Announce.Webhook.MessageTemplate == "" {
+		ctx.Config.Announce.Webhook.MessageTemplate = defaultMessageTemplate
+	}
+	if ctx.Config.Announce.Webhook.ContentType == "" {
+		ctx.Config.Announce.Webhook.ContentType = DefaultContentType
+	}
+	if !ctx.Config.Announce.Webhook.CheckCert {
+		ctx.Config.Announce.Webhook.CheckCert = true
+	}
```
caarlos0, line 64: since `CheckCert` is a bool, this means it will always check cert...

maybe another approach would be to use something like `skip_tls_verify` (or something along these lines), which defaults to `false`...

internal/pipe/webhook/webhook.go
```diff
@@ -0,0 +1,158 @@
...
+
+const (
+	defaultMessageTemplate = `{ "message": "{{ .ProjectName }} {{ .Tag }} is out! Check it out at {{ .ReleaseURL }}"}`
+	ContentTypeHeaderKey   = "Content-Type"
+	UserAgentHeaderKey     = "User-Agent"
+	UserAgentHeaderValue   = "gorleaser"
+	AuthorizationHeaderKey = "Authorization"
+	DefaultContentType     = "application/json; charset=utf-8"
+
+	EndpointRegex = "(http|nats)(s?)://.*"
```
caarlos0, line 27: why nats?

internal/pipe/webhook/webhook.go
```diff
@@ -0,0 +1,158 @@
...
+	case http.StatusForbidden:
+		return ErrForbidden
+	case http.StatusNotFound:
+		return ErrNotFound
+	case http.StatusUnprocessableEntity:
+		return ErrUnprocessableEntityError
+	case http.StatusTooManyRequests:
+		return ErrTooManyRequest
+	default:
+		return errors.New(resp.Status)
```
caarlos0, line 156: I would just wrap them all with something like:

```
fmt.Errorf("request failed with status %v", resp.Status)
```

and drop all these custom errors... we are not handling them anyway

caarlos0 (commented):
couple more comments

internal/pipe/webhook/webhook.go
dirien: Its a very good cloud native eventing. NATS trigger are very common and its not hurting to support it. In case people use NATS as backbone for their Serverless landscape. Many reasons.

internal/pipe/webhook/webhook.go
dirien: done! Thanks for the hint, you are right!

internal/pipe/webhook/webhook.go
dirien: Good point, i changed it! Thx for pointing out!

internal/pipe/webhook/webhook.go
dirien: NATS trigger are dope? it does not hurt to let it in and I wanted to use for an article when this feature is out :)?

caarlos0: I would say maybe its better to not even check the protocol via regex and let the user do whatever? 

if it doesn't work, it'll fail anyway... this way we don't need to at some point add other protocols 😁

dirien: Sounds good! I removed it! :)

caarlos0:
amazing, thanks @dirien 🤘

PR #2770, opened 2021-12-17
Support extra_files in publishers

### LINKED ISSUE #2768: Extra Files Support with Custom Publishers

Our `.goreleaser.yml` makes use of the `extra_files` configuration of the `release` pipe as well as a custom publisher:

```yaml
publishers:
  - name: example
    checksum: true
    signature: true
    cmd: example upload-file {{ abs .ArtifactPath }}
release:
  extra_files:
    - glob: '{{ .ProjectName }}_manifest.json'
      name_template: '{{ .ProjectName }}_{{ .Version }}_manifest.json'
```

The extra files are added to the release, but not published using the custom publisher. This appears to be because the release pipe, which adds those artifacts to the internal Context, runs after the custom publishers pipe.

https://github.com/goreleaser/goreleaser/blob/505888f41be5308eb7d5c6fb25df82a1bda4cc1a/internal/exec/exec.go#L105

https://github.com/goreleaser/goreleaser/blob/505888f41be5308eb7d5c6fb25df82a1bda4cc1a/internal/pipe/release/release.go#L134-L148

https://github.com/goreleaser/goreleaser/blob/505888f41be5308eb7d5c6fb25df82a1bda4cc1a/internal/pipe/publish/publish.go#L35-L52

I'm guessing that we wouldn't want to adjust the pipe ordering that is currently happening, so with that in mind, three mutually exclusive proposals:

1. Changing the logic in checksums to add the extra files to the Context artifacts, similar to how the release pipe currently works. I'm guessing the release pipe would then need to be careful to not add/overwrite duplicates. I'm not sure the other implications of doing this as other pipes may not be expecting these additional UploadableFile artifacts. It also might not be clear that adding that extra files configuration in checksums applies to the rest of the release pipeline.
1. Adding `extra_files` configuration to custom publishers (similar to checksums and release) and handling to the custom publishers pipe execution logic. I'm guessing it would need to mimic the current checksums pipe behavior so these extra files aren't included unexpectedly in other custom publishers and the release pipe.
1. Creating a separate top-level configuration section for "extra files" that tag along through all the pipes (potentially filterable as a new id in other pipes). A new "extra files" pipe that reads this configuration and saves it into the Context should execute before checksums since that is the first extra files usage.

Separately calling the custom publisher on the extra files after calling `goreleaser`.

**Additional context**

Please let me know if you would accept any type of contribution here and happy to submit one. Thank you for this excellent project!

bflad:
Implements `extra_files` support within individual `publishers`. This matches proposal 2 from https://github.com/goreleaser/goreleaser/issues/2768, but I will also try to submit a separate pull request for proposal 3.

Closes #2768


bflad:
See also #2771 which would supersede this pull request.

caarlos0:
Thanks, I think this is simpler than the other option, so for now, lets go with it!

Thanks for working in this anyway!

PR #2779, opened 2021-12-21
feat: do not release goreleaser for arm6

caarlos0:
has virtually no downloads, and was never really tested...

I think we can have fewer targets, only the most common ones. Should speed up the release a bit and also lower the number of files in the release

wdyt?

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -25,7 +25,6 @@ builds:
     - arm
     - arm64
   goarm:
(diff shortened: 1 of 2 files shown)


caarlos0:
problem is that both armv6 and armv7 are "armhf" on debian, so the filenames colide as well... and since its EOL soon and is not downloaded most of time, IMHO, we can remove it now...

PR #2793, opened 2021-12-28
feat(deps): update go-github to v41

caarlos0:
just because I saw its outdated :)


Dentrax:
Interesting, Dependabot did not notice this? Is it because of major version change?

caarlos0:
I think so yeah 🤔

PR #2798, opened 2021-12-30
fix: sbom binary filter

cpanato:
When using the `sbom` config with `binary` artifact it was generating a duplicate list of binaries to generate the sbom and that generate the `artifacts.json` with duplicate entries and that will fail in the release step when trying to upload the files to the GitHub, for example, it tries to push duplicate files and that fail in the GitHub side.

applying this change the list is correct.


caarlos0:
Good catch! Thanks!

cpanato:
cc @caarlos0

PR #2811, opened 2022-01-07
feat: allow multiple ids on universal binaries

### LINKED ISSUE #2759: [Feature]: Support multiple ids within same universal binary
Following this [conversation](https://gophers.slack.com/archives/C3T1JU24B/p1639154901035400)

For now `universal_binaries` assumes all binaries for different arch created within the same build id
```yaml
  - id: app
    binary: akash
    main: ./cmd/app
    goarch:
      - amd64
      - arm64
    goos:
      - darwin
universal_binaries:
  - id: app
    name_template: '{{ .ProjectName }}'
    replace: true
```

unfortunately, cross-compile cases put a limit in form of specific C/C++ compiler to be set for the particular arch
```yaml
  - id: app-darwin-amd64
    binary: akash
    main: ./cmd/app
    goarch:
      - amd64
    goos:
      - darwin
    env:
      - CC=o64-clang
      - CXX=o64-clang++
    flags:
      - -trimpath
      - -mod=readonly
      - "{{ .Env.BUILD_FLAGS }}"
    ldflags:
      - "{{ .Env.LD_FLAGS }}"
      - -linkmode=external
  - id: app-darwin-arm64
    binary: app
    main: ./cmd/app
    goarch:
      - arm64
    goos:
      - darwin
    env:
      - CC=oa64-clang
      - CXX=oa64-clang++
    flags:
      - -trimpath
      - -mod=readonly
      - "{{ .Env.BUILD_FLAGS }}"
    ldflags:
      - "{{ .Env.LD_FLAGS }}"
      - -linkmode=external
```

such a case could be resolved with following extension
```yaml
universal_binaries:
  - ids:
    - app-darwin-amd64
    - app-darwin-arm64
    name_template: '{{ .ProjectName }}'
    replace: true
```

caarlos0:
fixes #2759

* behavior still the same if no `ids` is provided
* if `ids` is provided, it will be used to filter, and the resulting universal binary will have the ID set in the `id` field

since before it was not possible to set multiple ids, and the previous behavior will remain, this is not a breaking change IMO

### DIFF
--- a/internal/pipe/universalbinary/universalbinary.go
+++ b/internal/pipe/universalbinary/universalbinary.go
@@ -34,6 +34,9 @@ func (Pipe) Default(ctx *context.Context) error {
 		if unibin.ID == "" {
 			unibin.ID = ctx.Config.ProjectName
 		}
+		if len(unibin.IDs) == 0 {
+			unibin.IDs = []string{unibin.ID}
+		}
 		if unibin.NameTemplate == "" {
 			unibin.NameTemplate = "{{ .ProjectName }}"
 		}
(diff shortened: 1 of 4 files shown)


caarlos0:
cc/ @troian

troian:
thanks! 🚀

Issue #2822, opened 2022-01-10
Errors when trying to install using go install method.


pacoorozco:
### What happened?

This bug is related to the #2788 which is closed but the provided solution I don't think is good enough.
Following the documentation, there's a way to install the `goreleaser` using go install.
```
$ go install github.com/goreleaser/goreleaser@latest
go: downloading github.com/vartanbeno/go-reddit/v2 v2.0.1
../go/pkg/mod/github.com/goreleaser/goreleaser@v1.2.5/internal/pipe/reddit/reddit.go:10:2: reading github.com/vartanbeno/go-reddit/go.mod at revision v2.0.1: unknown revision v2.0.1
```
and the `goreleaser` command **is not installed**

### How can we reproduce this?

Do
```
$ go install github.com/goreleaser/goreleaser@latest
```

### goreleaser version

```bash
1.2.5
```

### GoReleaser Check

- [X] goreleaser check shows no errors

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Additional context

The provided solution, using `@main` instead of `@latest` is not working neither

```
go install github.com/goreleaser/goreleaser@main
go: downloading github.com/goreleaser/goreleaser v1.2.6-0.20220108174805-8833bfcf4d13
go: downloading github.com/google/go-github/v41 v41.0.0
go: downloading github.com/caarlos0/env/v6 v6.9.1
go: downloading github.com/vartanbeno/go-reddit/v2 v2.0.0
../go/pkg/mod/github.com/goreleaser/goreleaser@v1.2.6-0.20220108174805-8833bfcf4d13/internal/pipe/reddit/reddit.go:10:2: github.com/vartanbeno/go-reddit/v2@v2.0.0: verifying module: checksum mismatch
        downloaded: h1:vjb+mY2GwC00uKeXdM+efhzEJoScc0oQfP5CZrE2R28=
        sum.golang.org: h1:fxYMqx5lhbmJ3yYRN1nnQC/gecRB3xpUS2BbG7GLpsk=

SECURITY ERROR
This download does NOT match the one reported by the checksum server.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.
```

caarlos0:
https://github.com/vartanbeno/go-reddit/issues/26#issuecomment-985060736 

you'll need to use the default go mod proxy...

[...]

PR #2823, opened 2022-01-10
fix: go install and go-reddit

caarlos0:
closes https://github.com/goreleaser/goreleaser/issues/2822
see https://github.com/vartanbeno/go-reddit/issues/26

not much I can do besides this


developer-guy (approved):
LGTM

PR #2836, opened 2022-01-16
chore: remove stale trigger paths

twpayne:
If applied, this commit will tidy up some stale paths in the GitHub Actions `docs` workflow.

`.github/workflows/htmltest.yml` no longer exists, and the `www/*` path catches changes to `www/htmltest.yml`.


caarlos0:
thanks!

PR #2838, opened 2022-01-19
feat: arch user repository integration

caarlos0:
### TODO

- [x] generate the pkgbuild file based on config/artifacts
- [x] clone, update and commit to a git repo using a configurable ssh key
- [x] docs
- [x] make `package` templateable
- [ ] publish goreleaser pkgbuild itself (likely on next release)

### Questions

- it seems that AUR defaults to master, although I couldn't find any references for that... is that assumption correct?
- the auth method is a bit cumbersome, is there a better way?
- are the docs clear enough?
- not sure about the `aurs` field in the config file, but seemed the most appropriate 
- should we also upload PKGBUILDs et al to the release?

---
**This feature is being sponsored by [Charm](https://charm.sh) ✨**


.goreleaser.yaml
```diff
@@ -161,6 +161,33 @@ rigs:
...
+aurs:
+  - homepage:  https://goreleaser.com
+    description: Deliver Go binaries as fast and easily as possible
+    maintainer: 'Carlos Alexandro Becker <carlos at becker dot software>'
+    contributors:
+      - 'Fernandez Ludovic <lfernandez dot dev at gmail dot com>'
+    license: MIT
+    private_key: '{{ .Env.AUR_KEY }}'
+    git_url: 'ssh://aur@aur.archlinux.org/goreleaser-bin.git'
+    package: |-
```
dirien, line 173: Hi @caarlos0, Shoukd we really put installation code onto the goreleaser.yaml?

dirien (commented):
LGTM, left a comment :)

internal/pipe/aur/tmpl.go
```diff
@@ -0,0 +1,93 @@
...
+{{ fixLines . }}
+}
+{{ end }}`
+
+const srcInfoTemplate = `pkgbase = {{ .Name }}
+	pkgdesc = {{ .Desc }}
+	pkgver = {{ .Version }}
+	pkgrel = {{ .Rel }}
+	url = {{ .Homepage }}
+	license = {{ .License }}
```
Foxboron, line 73: The template doesn't generate a valid `.SRCINFO` file. All fields should be with a newline delimiter and beyond this point it seems like this template will gobble up any newlines past `license`? The test data is incorrect for reference.

internal/pipe/aur/aur.go
```diff
@@ -0,0 +1,477 @@
...
+	"github.com/goreleaser/goreleaser/internal/pipe"
+	"github.com/goreleaser/goreleaser/internal/tmpl"
+	"github.com/goreleaser/goreleaser/pkg/config"
+	"github.com/goreleaser/goreleaser/pkg/context"
+	"golang.org/x/crypto/ssh"
+)
+
+const (
+	pkgBuildExtra     = "AURConfig"
+	defaultSSHCommand = "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i {{ .KeyPath }} -F /dev/null"
```
Foxboron, line 28: This isn't particularly good and should much rather pin the AUR host key instead of ignoring it.

internal/pipe/aur/tmpl.go
```diff
@@ -0,0 +1,93 @@
...
+	{{- range .Depends -}}
+	depends = {{ . }}
+	{{ end }}
+	{{- range .Conflicts -}}
+	conflicts = {{ . }}
+	{{ end }}
+	{{- range .Provides -}}
+	provides = {{ . }}
+	{{ end }}
+	{{- range .ReleasePackages -}}
```
Foxboron, line 86: `.SRCINFO` made by `makepkg --printsrcinfo` would structure this as
<pre>
source = 1
source = 2
source = 3
sha256sums = 1
sha256sums = 2
sha256sums = 3
</pre>

However I read through the parser in aurweb and I think this should be fine? I don't know if any other `.SRCINFO` parser would depend on the above ordering.

internal/pipe/aur/aur.go
```diff
@@ -0,0 +1,477 @@
...
+	for i := range ctx.Config.AURs {
+		pkg := &ctx.Config.AURs[i]
+
+		pkg.CommitAuthor = commitauthor.Default(pkg.CommitAuthor)
+		if pkg.CommitMessageTemplate == "" {
+			pkg.CommitMessageTemplate = defaultCommitMsg
+		}
+		if pkg.Name == "" {
+			pkg.Name = ctx.Config.ProjectName + "-bin"
+			if len(pkg.Conflicts) == 0 {
```
Foxboron, line 50: I would move this logic for  `Conflicts` and `Provides` out of the `pkg.Name` if clause and unconditionally check if the user have included `ProjectName` in the arrays. This would probably help produce more correct `PKGBUILD` files as people don't need to (and probably don't) realize that these are required.

Foxboron:
Yo, I'm one of the AUR moderators and general Go packager in Arch Linux. I looked over the code and it seems fine. I tried answering the questions at the top of the PR as well. Feel free to ask if you wonder about anything else.

>it seems that AUR defaults to master, although I couldn't find any references for that... is that assumption correct?

Yes, and this won't change as the underlying AUR repository is using git namespaces where the branches are hardcoded to `master`. The docs should probably be updated to reference that.

>the auth method is a bit cumbersome, is there a better way?

There isn't. SSH keys is the only way and I'm not aware of any changes being introduced in the Python rewrite.

> are the docs clear enough?

I think there needs to be a link to the [AUR submission guidelines](https://wiki.archlinux.org/title/AUR_submission_guidelines#Rules_of_submission). Rest of the links are a good guideline.

internal/pipe/aur/tmpl.go
```diff
@@ -0,0 +1,93 @@
+package aur
+
+type templateData struct {
+	Name            string
+	Desc            string
+	Homepage        string
+	Version         string
```
Foxboron, line 7: Version needs to be stripped of any prefix letters. Things like `v1.0.0` trips up the version checker in `pacman`.

.goreleaser.yaml
caarlos0: we would need to put it somewhere... this way is analogous to `brews.install`...

internal/pipe/aur/aur.go
caarlos0: good call ,fixed

internal/pipe/aur/tmpl.go
caarlos0: that's ok, when it gets to this point, goreleaser already stripped out the `v` prefix :)

internal/pipe/aur/aur.go
Foxboron: Sorry, I was imprecise. You need to also check if `pkg.Name != ctx.Config.ProjectName` and if that is the case do the provides and conflicts. This is implied in the if-clause but there won't be any guards against this if `pkg.Name` is defined.

internal/pipe/aur/tmpl.go
caarlos0: yup, fixed!

internal/pipe/aur/tmpl.go
caarlos0: I tested it on a docker container and it seems to be working... so I guess its fine 🤔

internal/pipe/aur/aur.go
caarlos0: we don't actually need those, I was using it for testing at some point, but it can be removed now...

Foxboron: Lovely :)

caarlos0:
Hey @Foxboron, many thanks for the review! I think I addressed all your points, feel free to take another look if you can 🙏

I also made a few more improvements:

- enforce -bin suffix on pkgname
- maintainers can be a list
- more tests 🙏

Foxboron:
It looks fine on my end. I like the fact you enforce `-bin` on `pkgname`.  The only thing that might be missing is `validpgpkeys`, but I'm not sure how often people sign their releaser artfifacts with `goreleaser`. Rest looks good to me :)

Issue #2808, opened 2022-01-05
support generating an additional SBOM file based on CycloneDX format


developer-guy:
### Is your feature request related to a problem? Please describe.

No, it is not related to a problem.

### Describe the solution you'd like

We have many tools that can allow us to generate an SBOM file based on CycloneDX format. So, let's add that support to GoReleaser. IMHO, when people specify command [cyclonedx-gomod](https://github.com/CycloneDX/cyclonedx-gomod) instead of syft, we can use that command to add arguments for it as we did for syft in [line#55](https://github.com/goreleaser/goreleaser/blob/d6f136de3c21d1254a2778a3c2e0986e89a221a8/internal/pipe/sbom/sbom.go?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L55).

There is also a GitHub Action to download that tool that can help us quickly download and use it with the support that we think to add to GoReleaser.

👉 https://github.com/CycloneDX/gh-gomod-generate-sbom

### Describe alternatives you've considered

Alternatively, Syft can generate CycloneDX format-based SBOM files, so, we can create an additional flag which we can accept a type of SBOM according to that we can override the --output argument of the Syft tool to let it generate a CycloneDX format based SBOM file.

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Additional context

Please feel free to assign it to me @caarlos0. I'm willing to work on it 🙋🏻‍♂️

dirien:
Sounds good!

jonasagx:
On alternatives:
Syft, as of v0.34.0, supports the following flags for output: json, text, table, cyclonedx (for xml), cyclonedx-json, spdx-tag-value, spdx-json.

There is also a Syft-based GitHub action to generate SBOMs https://github.com/marketplace/actions/anchore-sbom-action

VinodAnandan:
[...]

PR #2839, opened 2022-01-19
fix: dedupliate cataloging binary artifacts

wagoodman:
This PR makes the following adjustments:
- primarily restores the capability to reference binary artifacts when `sbom[].artifact == binary`
- adds the ability to have templates within `sbom[].env` values
- only prepends `dist` config value to paths that relative paths (takes no action on absolute paths)
- decomposes templating code further for the SBOM pipe and surround with more testing

Why make this change? primarily to unlock more workflows when generating SBOMs.

Not all tooling can work with encapsulations of go binaries (zip/tar) and instead need access to the binary directly. Since goreleaser manages creating the binary it is possible to expose the original binary to SBOM tooling even though it may be an encapsulation that is attached to a release (e.g. a zip or tar.gz)

Take for example using `cyclonedx-gomod app`:

```
#.goreleaser.yaml

archives:
  - format: tar.gz

sboms:
  - documents:
      - "{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}.cdx.sbom"
    artifacts: binary
    cmd: cyclonedx-gomod
    env: 
      - GOOS={{ .Os }}
      - GOARCH={{ .Arch }}
    args: ["app",  "-json", "-output", "$document", "$PWD"]

```

With `artifacts: binary` the SBOM tool gets access to the unarchived binary even though `archives[0].format = tar.gz` and there is **_no_** `archives[].format = binary` set. Additionally this uses goreleaser-sourced variables to seed the correct go-specific environment variables that the tool keys in on.

Relevant links:
- Support additional SBOM files based on CycloneDX: https://github.com/goreleaser/goreleaser/issues/2808
- Removal of binary from the SBOM pipe artifact filter: https://github.com/goreleaser/goreleaser/pull/2798

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,6 +28,7 @@ require (
 	github.com/muesli/coral v1.0.0
 	github.com/muesli/mango-coral v1.0.1
 	github.com/muesli/roff v0.1.0
+	github.com/scylladb/go-set v1.0.2
 	github.com/slack-go/slack v0.10.2
 	github.com/stretchr/testify v1.7.0
 	github.com/ulikunitz/xz v0.5.10
--- a/go.sum
+++ b/go.sum
@@ -322,6 +322,8 @@ github.com/envoyproxy/protoc-gen-validate v0.6.2/go.mod h1:2t7qjJNvHPx8IjnBOzl9E
 github.com/fatih/color v1.7.0/go.mod h1:Zm6kSWBoL9eyXnKyktHP6abPY2pDugNf5KwzbycvMj4=
 github.com/fatih/color v1.13.0 h1:8LOYc1KYPPmyKMuN8QV2DNRWNbLo6LZ0iLs8+mlH53w=
 github.com/fatih/color v1.13.0/go.mod h1:kLAiJbzzSOZDVNGyDpeOxJ47H46qBXwg5ILebYFFOfk=
+github.com/fatih/set v0.2.1 h1:nn2CaJyknWE/6txyUDGwysr3G5QC6xWB/PtVjPBbeaA=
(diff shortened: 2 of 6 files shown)


caarlos0 (approved):
LGTM, seems like its missing a go mod tidy though

caarlos0:
Thanks for the PR, its looking good to me overall! 

The only pointers I have are:

- the deduplication thing maybe is something worth moving to `artifacts.go`
- there is a test failing
- it looks like its missing a `go mod tidy`

other than that, lgtm!

PS: sorry for the delay reviewing this, was focusing on other things.
Thanks!

wagoodman:
@caarlos0 no worries -- this one fell off my radar too.

I went ahead and moved the filter function to `artifacts.go`, fixed the failing tests, rebased + go mod tidied. Should be good to go 👍

wagoodman:
@caarlos0 nudge --let me know if there is anything else you'd like changed/checked on this PR.

go.mod
```diff
@@ -28,7 +28,8 @@ require (
 	github.com/muesli/coral v1.0.0
 	github.com/muesli/mango-coral v1.0.1
 	github.com/muesli/roff v0.1.0
-	github.com/slack-go/slack v0.10.2
+	github.com/scylladb/go-set v1.0.2
+	github.com/slack-go/slack v0.10.1
```
caarlos0, line 32:
```suggestion
	github.com/slack-go/slack v0.10.2
```

caarlos0 (approved):
lgtm except the accidental dep downgrade

go.mod
wagoodman: curious! I've adjusted the commit to keep v0.10.2 of slack 👍

caarlos0:
Thanks!

PR #2846, opened 2022-01-23
chore: update grammar in markdown

khrisrichardson:
... update the `www/docs/how-it-works.md` file with different grammar in first sentence and format with `prettier www/docs/how-it-works.md --write --prose-wrap always --parser markdown`

...

...


caarlos0:
Thanks!

Issue #2829, opened 2022-01-11
Conditional Build Handling Based on OS Running goreleaser


bflad:
### Is your feature request related to a problem? Please describe.

We have a `goreleaser` configuration that is setup with a single cross-compilation `build` section. We are in the process of trying to conditionally enable CGO/netcgo if the `goreleaser` process is running on the same OS that is being built (or very particularly in our case, if it is running on `darwin`). Builds for other OS should not try to enable that additional configuration as the cross-compilation will fail.

### Describe the solution you'd like

Initial proposal:

`goreleaser` could potentially expose the `runtime.GOOS` of the `goreleaser` process as a field in the templating engine, such as `.CurrentOS`, then I think we could theoretically setup our configuration like:

```yaml
builds:
  - id: darwin
    flags:
      - -trimpath
    goos:
      - darwin
    goarch:
      - amd64
      - arm64  
    ldflags:
      - '-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}'
    mod_timestamp: '{{ .CommitTimestamp }}'
    tags:
      - {{ if eq .CurrentOS "darwin" }}netcgo{{ end }}
  - id: default
    flags:
      - -trimpath
    goos:
      - freebsd
      - linux
      - windows
    goarch:
      - '386'
      - amd64
      - arm
      - arm64
    ldflags:
      - '-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}'
    mod_timestamp: '{{ .CommitTimestamp }}'
```

Other thoughts:

Supporting special build configuration that is per os/arch/target within a build would then allow the configuration to be DRY, e.g.

```yaml
builds:
  -
    flags:
      - -trimpath
    goos:
      - darwin
      - freebsd
      - linux
      - windows
    goarch:
      - '386'
      - amd64
      - arm
      - arm64
    ldflags:
      - '-s -w -X main.version={{.Version}} -X main.commit={{.Commit}}'
    mod_timestamp: '{{ .CommitTimestamp }}'
    target_specific: # or os_specific or arch_specific or something allowing os/arch/target configuration
      darwin_amd64:
        tags:
          - {{ if eq .CurrentOS "darwin" }}netcgo{{ end }}
      darwin_amd64:
        tags:
          - {{ if eq .CurrentOS "darwin" }}netcgo{{ end }}
```

[...]

PR #2859, opened 2022-01-30
feat: expose runtime goos and goarch on templates and metadata

caarlos0:
Expose `{{ .Runtime.Goos }}` and `{{ .Runtime.Goarch }}`

refs  #2829


bflad:
This appears to be working as expected, at least with my testing of build `tags`. 👍

PR #2860, opened 2022-01-30
feat: allow to override build flags by goos/goarch/etc

caarlos0:
refs  #2829

allow to override ldflags et al based on the build target

## TODO

- [x] allow to template goos/goarch/etc to use it together with #2859

### DIFF
--- a/internal/builders/golang/build.go
+++ b/internal/builders/golang/build.go
@@ -129,21 +129,35 @@ func (*Builder) Build(ctx *context.Context, build config.Build, options api.Opti
 	return nil
 }
 
+func withOverrides(build config.Build, options api.Options) config.BuildDetails {
+	for _, o := range build.BuildDetailsOverrides {
+		if o.Goos == options.Goos &&
+			o.Goarch == options.Goarch &&
+			o.Gomips == options.Gomips &&
+			o.Goarm == options.Goarm {
+			return o.BuildDetails
+		}
+	}
+	return build.BuildDetails
+}
+
 func buildGoBuildLine(ctx *context.Context, build config.Build, options api.Options, artifact *artifact.Artifact, env []string) ([]string, error) {
+	details := withOverrides(build, options)
+
 	cmd := []string{build.GoBinary, "build"}
-	flags, err := processFlags(ctx, artifact, env, build.Flags, "")
+	flags, err := processFlags(ctx, artifact, env, details.Flags, "")
 	if err != nil {
 		return cmd, err
 	}
 	cmd = append(cmd, flags...)
 
-	asmflags, err := processFlags(ctx, artifact, env, build.Asmflags, "-asmflags=")
+	asmflags, err := processFlags(ctx, artifact, env, details.Asmflags, "-asmflags=")
 	if err != nil {
 		return cmd, err
 	}
 	cmd = append(cmd, asmflags...)
 
-	gcflags, err := processFlags(ctx, artifact, env, build.Gcflags, "-gcflags=")
+	gcflags, err := processFlags(ctx, artifact, env, details.Gcflags, "-gcflags=")
 	if err != nil {
 		return cmd, err
 	}
@@ -151,7 +165,7 @@ func buildGoBuildLine(ctx *context.Context, build config.Build, options api.Opti
 
 	// tags is not a repeatable flag
 	if len(build.Tags) > 0 {
-		tags, err := processFlags(ctx, artifact, env, build.Tags, "")
+		tags, err := processFlags(ctx, artifact, env, details.Tags, "")
 		if err != nil {
 			return cmd, err
 		}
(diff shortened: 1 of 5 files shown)


bflad:
Hey @caarlos0 👋  This looks like its going in the right direction, although I'm having a little trouble verifying it after installing from this branch. It seems to be losing all the build configuration in the `go build` command for me or not including the override configuration.

Configuration:

```yaml
builds:
  - flags:
      - -trimpath
    goos:
      - darwin
      - freebsd
      - linux
      - windows
    goarch:
      - '386'
      - amd64
      - arm
      - arm64
    ignore:
      - goarch: arm
        goos: windows
      - goarch: arm64
        goos: freebsd
      - goarch: arm64
        goos: windows
    ldflags:
      - -s -w -X main.Version={{.Version}}
    mod_timestamp: '{{ .CommitTimestamp }}'
    # overrides:
    #   - goos: darwin
    #     goarch: amd64
    #     tags:
    #       - netcgo
```

Environment:

```console
❯ go env GOOS          
darwin
❯ go env GOARCH
amd64
```

Command:

```shell
goreleaser build --snapshot --rm-dist --single-target --debug
```

v1.4.1 output:

```text
      • running                   cmd=[go build -trimpath -ldflags=-s -w -X main.Version=v0.1.3-next -o /Users/bflad/src/github.com/bflad/terraform-provider-framework/dist/terraform-provider-framework_darwin_amd64/terraform-provider-framework .]
```

buildconf output (no overrides):

```text
      • running                   cmd=[go build -trimpath -ldflags=-s -w -X main.Version=v0.1.3-next -o /Users/bflad/src/github.com/bflad/terraform-provider-framework/dist/terraform-provider-framework_darwin_amd64/terraform-provider-framework .]
```

buildconf output (with overrides uncommented, no goarm or gomips in override):

```text
     • running                   cmd=[go build -ldflags= -o /Users/bflad/src/github.com/bflad/terraform-provider-framework/dist/terraform-provider-framework_darwin_amd64/terraform-provider-framework .]
```

buildconf output (with overrides uncommented and goarm: 6 and gomips: hardfloat in override):

```text
      • running                   cmd=[go build -trimpath -ldflags=-s -w -X main.Version=v0.1.3-next -o /Users/bflad/src/github.com/bflad/terraform-provider-framework/dist/terraform-provider-framework_darwin_amd64/terraform-provider-framework .]
```

Not sure about the missing `tags` in the last one, but I'm wondering if https://github.com/goreleaser/goreleaser/blob/3f593d7a535fd747212d792a5fcadc24526cac8f/internal/builders/golang/build.go#L57-L62 has anything to do with the target matching.

caarlos0:
pushed some changes... it should now merge overrides into the build values, so it won't loose configs anymore

also, you can now template the goos/goarch/etc on overrides as well, in which you can for example use `{{ .Runtime.Goos }}`

internal/builders/golang/build.go
```diff
@@ -161,7 +193,7 @@ func buildGoBuildLine(ctx *context.Context, build config.Build, options api.Opti
 	// ldflags is not a repeatable flag
 	if len(build.Ldflags) > 0 {
```
bflad, line 194: Similarly here:

```suggestion
	if len(details.Ldflags) > 0 {
```

bflad (commented):
Looks like its now selecting and merging correctly 🎉  although I think the build command logic might still need some updates to ensure its checking the details/overrides 😄

internal/builders/golang/build.go
caarlos0: yes, I'm blind! hehe

will also ensure this is covered by tests

bflad (commented):
This is working as I would anticipate! 🎉  Thanks so much.

Totally fine as something in the future: I'm also wondering if `env` should allow similar overrides since I don't believe it can be templated and setting CGO_ENABLED between platforms (or other [go command environment variables](https://pkg.go.dev/cmd/go#hdr-Environment_variables)) is a bit awkward. I understand that is more difficult to merge/manage in the code though and happy to submit a followup issue/pull request. 👍

caarlos0:
> This is working as I would anticipate! 🎉 Thanks so much.

🚀🚀🚀🚀

> Totally fine as something in the future: I'm also wondering if `env` should allow similar overrides since I don't believe it can be templated and setting CGO_ENABLED between platforms (or other [go command environment variables](https://pkg.go.dev/cmd/go#hdr-Environment_variables)) is a bit awkward. I understand that is more difficult to merge/manage in the code though and happy to submit a followup issue/pull request. 👍

I was thinking the same... and maybe also other options... but I think that yes, at least for env, we should do it.

for envs the endeavor is slightly bigger, as we need to care for build hooks as well https://goreleaser.com/customization/build/#build-hooks

but I think we can work on it on a different PR once this one has better coverage and is merged

PR #2864, opened 2022-01-31
Two line "extended" description

staticdev:
Example on documentation of extended description to be compliant with Debian packaging.


caarlos0:
Thanks!

PR #2881, opened 2022-02-04
feat: replace cobra with coral

caarlos0:
replacing cobra with coral.

cobra is still an indirect dependency because of chglog. Will change that in the future.


Moulick:
Did goreleaser switch to `coral` to reduce the dependencies on Viper?  If so, Cobra 1.4.0 drops Viper.

https://github.com/muesli/coral#cobra-140 
https://github.com/spf13/cobra/releases/tag/v1.4.0

PR #2916, opened 2022-02-17
fix: do not copy binary by default

### LINKED ISSUE #2913: New --output flag should be optional

This PR https://github.com/goreleaser/goreleaser/pull/2701 introduced a new `--output` flag but it defaults to the local directory.
We already generate binaries into the `dist/` directory, it would be great to skip this new behavior.

Instead of defaulting to the local directory when `--output` is not specified it should present the previous behavior and not copy the binary.

Setting `--output=dist/` but it would still unnecessarily copy files around. 

### Search

### Code of Conduct

### Additional context

_No response_

caarlos0:
closes #2913


dirien (approved):
LGTM

PR #2921, opened 2022-02-19
fix: Homebrew README example formula

Justintime50:
This commit is a simple correction to the new style of Homebrew generate formulas based on a recent formula this tool generated for one of my projects. After checking the documentation, I found the README had not yet been updated.

This change is to bring the documentation inline with what the output of a Homebrew generated formula will look like.


caarlos0:
thanks!

PR #2922, opened 2022-02-20
feat: better homebrew unsupported arch error

caarlos0:
this adds a custom requirement for the architectures, so we can show a better error if the user tries to install on an unsupported architecture. 

closes #2917 
refs #2278  
refs https://github.com/goreleaser/goreleaser/issues/2159

this reimplements and old try I had at it, but since there was quite a bit of refactor since then, it was actually easier to rewrite instead of solving conflicts.

Anyway, there you have it.


ryancurrah:
We got the same error when trying to access a mixed architecture tap. You can see my example tap that has two formulas created by goreleaser: https://github.com/ryancurrah/brewtest using a build of goreleaser built from this branch.

```
brew tap --force-auto-update cdx/brewtest https://github.com/ryancurrah/brewtest
==> Tapping ryancurrah/brewtest
Cloning into '/opt/homebrew/Library/Taps/ryancurrah/homebrew-brewtest'...
remote: Enumerating objects: 15, done.
remote: Counting objects: 100% (15/15), done.
remote: Compressing objects: 100% (11/11), done.
remote: Total 15 (delta 2), reused 0 (delta 0), pack-reused 0
Receiving objects: 100% (15/15), done.
Resolving deltas: 100% (2/2), done.
Error: Invalid formula: /opt/homebrew/Library/Taps/ryancurrah/homebrew-brewtest/Formula/rcurrahtest.rb
Error: Cannot tap ryancurrah/brewtest: invalid syntax in tap!
```

So the change still doesn't allow us to use a tap with mixed architecture URLs.

ryancurrah:
This is how hashicorp handles multiple architectures: https://github.com/hashicorp/homebrew-tap/blob/master/Formula/nomad.rb#L6-L20.

It's not ideal but at least installing the tap works.

caarlos0:
interesting, the binary runs on rosetta then, I suppose?

caarlos0:
yeah, if we assume macos only, that seems like a good solution indeed, will work on it this weekend

ryancurrah:
> yeah, if we assume macos only, that seems like a good solution indeed, will work on it this weekend

I have a really dirty hack here: https://github.com/goreleaser/goreleaser/pull/2937/files.

Tested this out and it worked.

caarlos0:
oh great! 

just comment there, but yeah, the idea seems good

caarlos0:
superseeded by #2939

Issue #2920, opened 2022-02-19
.TagContents also includes the signature (along with the tag message body) if the tag is signed


Tuxdude:
### What happened?

`.TagContents` also includes the signature (along with the tag message body) if the tag is signed. You can see the example output in the release notes [here](https://github.com/Tuxdude/go-scratch-gh/releases/tag/v0.4.9).

My `.goreleaser.yml` includes the following snippet for the release notes header:

```
release:
  header: |
    ## Released {{ time "Mon, Jan 2 2006 03:04 PM UTC-0700" }}
    {{ .TagContents }}
```

## Issue

`.TagContents` is populated from the output of `git tag -l --format='%(contents)' <tag_name>`, and hence includes the full contents of the tag message (i.e. subject, body and the signature).

`.TagSubject` already provides a means to retrieve just the tag subject. However, there is no way to retrieve just the tag message body without the signature.

If the tag is not signed, `.TagContents` has the necessary data. So this is mainly an issue when using signed tags.

## Possible Solution

The output from `git tag -l --format='%(contents:body)' <tag_name>` includes just the tag message body and for most scenarios that is the useful data that users will want to include in the release notes.

If the current behavior is intended for `.TagContents`, it might then be a good idea to include another template variable that provides the output same as `git tag -l --format='%(contents:body)' <tag_name>`.

In addition, if someone does really want to access the signature information in a tag it is available through `git tag -l --format='%(contents:signature)' <tag_name>`, and you could consider exposing that as one more variable that can be used in templates.

### How can we reproduce this?

Repro steps with example snippets and output included in the `What happened?` section.

### goreleaser version

```bash
goreleaser version dev
goos: linux
goarch: amd64
module version: v1.5.0, checksum: h1:l4BxBt2Rrk5Tngmzyrcp7H/CszXSA0guzL4mPo4peAU=

https://goreleaser.com
```

### GoReleaser Check

- [X] goreleaser check shows no errors

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

zepatrik:
Example release containing the signature: https://github.com/ory/keto/releases/tag/v0.8.0-alpha.1

PR #2924, opened 2022-02-20
feat: add filter and reverseFilter template functions

caarlos0:
this adds `grep` and `reverseGrep` template functions, which should help dealing with text.  

refs #2920


caarlos0:
This should help remove, say, `Signed-by` lines from the `.TagContents` or `TagBody`.

PR #2930, opened 2022-02-23
fix: trim spaces from aur private key

caarlos0:
I had this issue on a repo, and it seems at least one more person had it too: ssh/git complaining that the key is invalid.  

My guess is that gh secret set from a file is now appending a new line - as it worked with the exact same key before, for me at least.  

anyway, if thats it, this pr should fix it. 

 closes https://github.com/goreleaser/goreleaser-action/issues/326


caarlos0:
probably not it in the issue reported, but won't hurt to do it anyway

PR #2854, opened 2022-01-28
docs(cookbooks): Add documentation for how to publish to JFrog artifa…

areese:
…ctory.

This commit adds a new cookbook that covers how to upload to JFrog Artifactory: https://jfrog.com/artifactory/ using the REST API: https://www.jfrog.com/confluence/display/JFROG/Artifactory+REST+API
or the jfrog cli: https://jfrog.com/getcli/

...

...

...

PR #2932, opened 2022-02-23
docs: Add a new cookbook that covers using the jfrog-cli as an alternative to the artifactory publisher to publish to artifactory.

areese:
The jfrog cli allows storing the artifactory configuration in a configuration file rather than directly in the environment.

This is a refactor of the previous PR I had and closed here:  https://github.com/goreleaser/goreleaser/pull/2854

I have found it easier to configure the jfrog cli to contain my artifactory credentials.


areese:
@jbaruch lemme know what you think of this cookbook, I ended up using the cli for my use case.

caarlos0:
Thanks!

PR #2937, opened 2022-02-25
always ensure there is a url specified

ryancurrah:


caarlos0:
probably woth going against `main`, and adding only a bool to the template data, something like `hasOnlyAmd64MacOsPkg`

PR #2938, opened 2022-02-25
fix: build breaking on windows 386

caarlos0:
For some reason the set lib was trowing an int overflow  error on snapshot when there are no binary artifacts to sign. 

This list is small, we don't need to worry about allocations et al, so I just switched it for a `map` for now, which seems to fix the issue...

cc @wagoodman 

Refs  #2839 
See https://github.com/goreleaser/goreleaser/runs/5334854323?check_suite_focus=true#step:21:50

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,6 @@ require (
 	github.com/muesli/coral v1.0.0
 	github.com/muesli/mango-coral v1.0.1
 	github.com/muesli/roff v0.1.0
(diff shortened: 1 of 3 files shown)


wagoodman:
curious! I've never seen an issue like that with `strset`. Nice / quick fix!

wagoodman (approved):
good fix (though my weight is +0 😎 )

PR #2939, opened 2022-02-25
improvement(pipe/brew): install amd64 binaries when no arm64 binaries present

ryancurrah:
Currently on a M1 macbook when adding a tap with formulas that only support `amd64` it fails to add the tap. With errors similar to the below. Which prevents new `arm64` users from using the tap.

```
brew tap --force-auto-update cdx/brewtest https://github.com/ryancurrah/brewtest
==> Tapping ryancurrah/brewtest
Cloning into '/opt/homebrew/Library/Taps/ryancurrah/homebrew-brewtest'...
remote: Enumerating objects: 15, done.
remote: Counting objects: 100% (15/15), done.
remote: Compressing objects: 100% (11/11), done.
remote: Total 15 (delta 2), reused 0 (delta 0), pack-reused 0
Receiving objects: 100% (15/15), done.
Resolving deltas: 100% (2/2), done.
Error: Invalid formula: /opt/homebrew/Library/Taps/ryancurrah/homebrew-brewtest/Formula/rcurrahtest.rb
Error: Cannot tap ryancurrah/brewtest: invalid syntax in tap!
```

This pull request attempts to resolve the issue by allowing M1 macbook users to install an `amd64` binary if no `arm64` binary is available and allows them to install the tap without error as well.

Users can try using Rosetta to run the `amd64` binary with their M1 macbook.

Apple support article showing how to install Rosetta https://support.apple.com/en-us/HT211861.

closes https://github.com/goreleaser/goreleaser/pull/2917
refs https://github.com/goreleaser/goreleaser/pull/2278
refs https://github.com/goreleaser/goreleaser/issues/2159

### DIFF
--- a/internal/pipe/brew/brew.go
+++ b/internal/pipe/brew/brew.go
@@ -374,6 +374,10 @@ func dataFor(ctx *context.Context, cfg config.Homebrew, cl client.Client, artifa
 		}
 	}
 
+	if len(result.MacOSPackages) == 1 && result.MacOSPackages[0].Arch == "amd64" {
+		result.HasOnlyAmd64MacOsPkg = true
+	}
+
 	sort.Slice(result.LinuxPackages, lessFnFor(result.LinuxPackages))
 	sort.Slice(result.MacOSPackages, lessFnFor(result.MacOSPackages))
 	return result, nil
--- a/internal/pipe/brew/brew_test.go
+++ b/internal/pipe/brew/brew_test.go
@@ -81,9 +81,10 @@ var defaultTemplateData = templateData{
 			Install:     []string{`bin.install "test"`},
 		},
 	},
-	Name:    "Test",
-	Version: "0.1.3",
-	Caveats: []string{},
+	Name:                 "Test",
+	Version:              "0.1.3",
+	Caveats:              []string{},
+	HasOnlyAmd64MacOsPkg: false,
 }
 
 func assertDefaultTemplateData(t *testing.T, formulae string) {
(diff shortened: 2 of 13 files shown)


internal/pipe/brew/brew_test.go
```diff
@@ -81,9 +81,10 @@ var defaultTemplateData = templateData{
 			Install:     []string{`bin.install "test"`},
 		},
 	},
-	Name:    "Test",
-	Version: "0.1.3",
-	Caveats: []string{},
+	Name:                 "Test",
```
ryancurrah, line 84: `gofmt` formatted this automatically.

internal/pipe/brew/brew.go
```diff
@@ -374,6 +374,10 @@ func dataFor(ctx *context.Context, cfg config.Homebrew, cl client.Client, artifa
 		}
 	}
 
+	if len(result.MacOSPackages) == 1 && result.MacOSPackages[0].Arch == "amd64" {
```
scarlson, line 377: does any consideration need to be made here for fat binaries?

ryancurrah: Hrm if by fat you mean all binaries in archive that arch name is `all` instead of `amd64` so I think we do not need to handle it.

caarlos0: we dont, it will not trigger this if in that case..

caarlos0:
thanks for the PR @ryancurrah 

and thanks @scarlson for helping with the review as well 

🤘

PR #2977, opened 2022-03-15
fix(ci): Pinned workflow by hash

naveensrinivasan:
- Pinned actions by SHA https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies
- Included permissions for some of the actions. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions


caarlos0:
Hey! 

Thanks for the PR!

Does dependabot updates deps based on hashes as well? If so, won't it flood the hell out of us with deps updates? 🤔

naveensrinivasan:
> Hey!
> 
> Thanks for the PR!
> 
> Does dependabot updates deps based on hashes as well? If so, won't it flood the hell out of us with deps updates? 🤔

dependabot is aware of these pinned versions and they upgraded pinned versions here is an example https://github.com/ossf/scorecard/pull/1698

naveensrinivasan:
> ah, sg then
> 
> although it might be noisier than usual, this seems like a good thing to do.
> 
> thanks!

Fixed the conflict! Thanks

PR #2993, opened 2022-03-19
feat: gitlab option to use_job_token

### LINKED ISSUE #2743: Use Gitlab's CI_JOB_TOKEN

So far, to push a release to Gitlab, one needs to setup a new API token, and configure the pipelines with it. However, there is already a job token available, via the `CI_JOB_TOKEN` env-var. It could be used to create a new release, but needs a special header to be set.

Right now, goreleaser uses [`gitlab.NewClient`](https://github.com/goreleaser/goreleaser/blob/167d95e67fe0266fd75bc5dffc949f5db550dc73/internal/client/gitlab.go#L46), which registers a [`PrivateToken`](https://github.com/xanzy/go-gitlab/blob/f8ceb0600dbec1eef4cb11d9779cc4804375dd0e/gitlab.go#L214-L222) which sets the [`PRIVATE-TOKEN`](https://github.com/xanzy/go-gitlab/blob/f8ceb0600dbec1eef4cb11d9779cc4804375dd0e/gitlab.go#L682) header value.

What should be possible, is to call [`NewJobClient`](https://github.com/xanzy/go-gitlab/blob/f8ceb0600dbec1eef4cb11d9779cc4804375dd0e/gitlab.go#L241-L249) instead if the `CI_JOB_TOKEN` is present (or via an option I guess), which would mostly remove the need to manage an API token.

caarlos0:
closes #2743


caarlos0:
closing for now, check https://github.com/goreleaser/goreleaser/issues/2743#issuecomment-1077833075

PR #3015, opened 2022-04-01
Update GitHub action docs, change secret to GH_TOKEN instead of GITHUB_TOKEN

barthr:
In actions it is not allowed to have a secret in your repository which starts with GITHUB


caarlos0:
`GITHUB_TOKEN` is the default actions token.

later on the docs explain how to use a custom one https://goreleaser.com/ci/actions/#token-permissions

PR #3045, opened 2022-04-16
WIP: feat: Support build using Kaniko on Docker

caalberts:
This commit will add a feature to build Docker image using Kaniko running on Docker.

The implementation adds a new kaniko api that implements `imager` interface. 

As kaniko builds and pushes in the same command, the `Push` method is a no-op. A no-push option is available within Kaniko through a CLI flag `--no-push`. So Goreleaser's `--skip_push` is translated into Kaniko's `--no-push`.

...

#2322

### DIFF
--- a/internal/pipe/docker/api_kaniko.go
+++ b/internal/pipe/docker/api_kaniko.go
@@ -1,0 +1,82 @@
+package docker
+
+import (
+	"fmt"
+	"github.com/goreleaser/goreleaser/pkg/config"
+	"os"
+
+	"github.com/goreleaser/goreleaser/pkg/context"
+)
+
+func init() {
+	registerImager(useKaniko, kanikoImager{})
+}
+
+const (
+	kanikoExecutorImage = "gcr.io/kaniko-project/executor:latest"
+	kanikoWorkspace     = "/workspace"
+)
+
+type kanikoImager struct{}
+
+// Push is a no-op and it returns nil because kaniko performs build and push in a single step.
+// https://github.com/GoogleContainerTools/kaniko#how-does-kaniko-work
+func (i kanikoImager) Push(_ *context.Context, _ string, _ []string) error {
+	return nil
+}
+
+func (i kanikoImager) Build(ctx *context.Context, docker config.Docker, root string, images, flags []string) error {
+	if err := runCommand(ctx, root, "docker", i.buildCommand(docker, images, flags)...); err != nil {
+		return fmt.Errorf("failed to build %s: %w", images[0], err)
+	}
+	return nil
+}
+
+func (i kanikoImager) buildCommand(docker config.Docker, images, flags []string) []string {
+	base := []string{"run"}
+
+	base = i.appendWorkspaceVolumeMount(base)
+	base = i.appendKanikoExectorImage(base)
+	base = i.appendContext(base)
+	base = i.appendDockerfile(base, docker.Dockerfile)
+	base = i.appendDestinations(base, images)
+	base = i.appendNoPush(base, docker.SkipPush == "true")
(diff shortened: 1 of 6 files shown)


internal/pipe/docker/api_kaniko.go
```diff
@@ -0,0 +1,82 @@
...
+
+type kanikoImager struct{}
+
+// Push is a no-op and it returns nil because kaniko performs build and push in a single step.
+// https://github.com/GoogleContainerTools/kaniko#how-does-kaniko-work
+func (i kanikoImager) Push(_ *context.Context, _ string, _ []string) error {
+	return nil
+}
+
+func (i kanikoImager) Build(ctx *context.Context, docker config.Docker, root string, images, flags []string) error {
```
caalberts, line 28: We need to pass `config.Docker` here so that the configuration values can be used to construct the build command.

caalberts:
@caarlos0 Could you take a look at this draft to see if it is in the right direction for #2322 please?

I have some questions:
1. Do we need to implement the `manifester`?
1. I tried to run `docker_test.go`, but the kaniko tests are failing because the registries use `localhost` which are not reachable from the kaniko executor image (see https://github.com/goreleaser/goreleaser/runs/6046907714?check_suite_focus=true#step:13:95). Do you have any idea how we can address this?

internal/pipe/docker/api_kaniko.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (i kanikoImager) buildCommand(docker config.Docker, images, flags []string) []string {
+	base := []string{"run"}
+
+	base = i.appendWorkspaceVolumeMount(base)
+	base = i.appendKanikoExectorImage(base)
+	base = i.appendContext(base)
+	base = i.appendDockerfile(base, docker.Dockerfile)
+	base = i.appendDestinations(base, images)
+	base = i.appendNoPush(base, docker.SkipPush == "true")
```
caarlos0, line 43: my understanding is that it will also push the image on build, correct?

if so, we need to think about another way, users might not want to publish on build, and the skip push option is used to skip the push when running the entire pipeline...

caarlos0:
looking good overall! thank you!

have a concern about the push on build part, and at some point we need to add tests and run `task fmt` as well

internal/pipe/docker/api_kaniko.go
caalberts: Yes by default it pushes on build.

caalberts:
> looking good overall! thank you!
> 
> have a concern about the push on build part, and at some point we need to add tests and run `task fmt` as well

I have been having some challenges with the `docker_test.go` tests where it automatically loads all the imager implementations and run the tests on each of them:

1. The current test image tags using registry address on `localhost` do not work for Kaniko because Kaniko runs within the docker network. Changing the registry address to fit Kaniko by using a docker network such as `alt_registry` would then break the tests for the other imagers.
1. Kaniko assumes secure registry by default and will push to `https` address. This breaks on the test registry which doesn't have SSL. The workaround is to have Kaniko run with insecure option. The default imager loaded within `init` then needs to be overridden during test.

Given these as well as the `skipPush` behaviour, I wonder if it should be implemented as a dedicated Kaniko pipeline, instead of adapting Docker.

caarlos0:
> Given these as well as the `skipPush` behaviour, I wonder if it should be implemented as a dedicated Kaniko pipeline, instead of adapting Docker.

closed by mistake, but yes, probably worth having another pipe just for kaniko then...

caarlos0:
closing for inactivity, feel free to ping if there's still interest in proceeding with this.

PR #3046, opened 2022-04-16
Change test docker registry port numbers

caalberts:
Port 5000 is used by MacOS Monterey for Airplay Receiver by default.

https://developer.apple.com/forums/thread/682332


caarlos0:
Thanks!

PR #3068, opened 2022-04-26
Update scoop.md

apzuk3:
Probably a typo in the docs.

It refers to homebrew instead of scoop 
https://goreleaser.com/customization/scoop/?h=scoop


caarlos0:
Thanks!

Issue #3072, opened 2022-04-28
goreleaser can't update existing draft release


yihuang:
### What happened?

When there's an existing draft release, gorelease won't update it, but create a new one.

https://docs.github.com/en/rest/releases/releases#get-a-release-by-tag-name
Because the api goreleaser use right now don't return draft release.
To find a exiting draft release, one need to iterate all the releases like this: https://github.com/softprops/action-gh-release/blob/fe9a9bd3295828558c7a3c004f23f3bf77d155b2/src/github.ts#L205

### How can we reproduce this?

Configure a goreleaser with `draft: true`, create a draft release first before trigger goreleaser in github action, after it finish, it'll create a duplicated draft release.

### goreleaser version

```bash
`ghcr.io/goreleaser/goreleaser-cross:v1.17.9`
```

### GoReleaser Check

- [X] goreleaser check shows no errors

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #3073, opened 2022-04-28
fix: update existing release drafts

yihuang:
find existing draft release by list all releases

otherwise it don't update exiting draft release, but create a duplicated one.

Closes: #3072


caarlos0:
thanks!

Issue #2911, opened 2022-02-15
Optionally run `go mod vendor` for source archives?


francislavoie:
### Is your feature request related to a problem? Please describe.

We were asked on the Caddy project https://github.com/caddyserver/caddy/issues/4573 to provide a source bundle including vendored dependencies, so that builds can be produced without additional downloads.

### Describe the solution you'd like

An optional flag like `vendor: true` in https://goreleaser.com/customization/source/ would be great. It should run `go mod vendor` to produce the `vendor/` dir in the project before zipping up the source.

### Describe alternatives you've considered

It would let us avoid having to write some custom shell script after the release to produce the extra tarball.

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

caarlos0:
not sure if its possible... we use `git archive` to create the source archive... so, if vendor is gitignored, it won't get picked - AFAIK

not sure if there's a way around this...

francislavoie:
Hmm. We could remove `vendor/` from `.gitignore` on the Caddy repo, I don't think there's any reason for it to be there. But we definitely don't want to commit the vendor to the repo anymore (we did for Caddy v1 and it was a total mess).

https://stackoverflow.com/a/54036395/846934 suggests maybe we can just throw `vendor` in afterwards?

caarlos0:
That's assuming tar, but it could be a myriad of formats.

You could, in theory, remove vendor from the gitignore, and run `go mod vendor` in a global before hook.

The vendor folder will be used to build - which shouldn't be a big deal, and will also be included in the source archive

mohammed90:
The `git archive` command only packs tracked and committed files. So unignoring the `vendor` directory won't help, neither will running `go mod vendor` to create the directory. It can pack extra/untracked files with the flag `--add-file`, but that'll need to be repeated for as many files in the vendor dir.

caarlos0:
> The `git archive` command only packs tracked and committed files. So unignoring the `vendor` directory won't help, neither will running `go mod vendor` to create the directory. It can pack extra/untracked files with the flag `--add-file`, but that'll need to be repeated for as many files in the vendor dir.

ah, indeed

caarlos0:
then, if we do this, we might need to implement the archiving ourselves instead of relying on `git-archive`...

mohammed90:
[...]

PR #3102, opened 2022-05-12
feat: add extra files to source archives

caarlos0:
closes https://github.com/goreleaser/goreleaser/issues/2911


caarlos0:
actually, this has a "small" issue: 

```
--add-file=<file>
Add a non-tracked file to the archive. Can be repeated to add multiple files. The path of the file in the archive is built by concatenating the value of the last --prefix option (if any) before this --add-file and the basename of <file>.
```

basename of file only... which means adding a `vendor` will not work... will rework this using my previous approach..

PR #3120, opened 2022-05-23
Remove "Windows" from quick start document

unasuke:
In default behavior, goreleaser will not build the app for windows.
So the sentence in the quick start is easy to misunderstand.

### DIFF
--- a/www/docs/quick-start.md
+++ b/www/docs/quick-start.md
@@ -78,7 +78,7 @@ goreleaser release
 
 That's all it takes!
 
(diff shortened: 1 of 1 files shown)


caarlos0:
ne of the previous steps is to run `goreleaser init`, which creates a config file with windows builds included... so, the docs seems to be correct to me...

unasuke:
Ah oh, indeed. I ran `goreleaser release` on a repository that doesn't have `.goreleaser.yaml` So I'll close this because the docs are proper, thanks.

PR #3149, opened 2022-06-09
fix: duplicate draft release in GitHub

### LINKED ISSUE #3148: Do not create duplicate draft release in GitHub

Our release workflow is like this:
1. Create a `vx.y.z` branch, then GitHub Actions create a draft release with the name `RELEASE vx.y.z`
2. In the next couple of day, we cherry-pick some bug-fix commit to the release branch as well as fill the content of the release note

The question is a new commit to the release branch will trigger goreleaser again, and create another draft release note (instead of appending to the existing one).

Seems like goreleaser only treat releases with the same tag as the same releases. While the draft release will be assigned a random tag like `untagged-1070a2429328d38d3638`.
 https://github.com/goreleaser/goreleaser/blob/f8e5bdac10a61271aee0ca58427fdb0432b3d10b/internal/client/github.go#L224-L230

Consider treating draft releases with the same name as the same release

N/A

### Search

### Code of Conduct

### Additional context

_No response_

qsliu2017:
For draft releases in GitHub, treat other draft releases with the same name as the same release.

Fix #3148

### DIFF
--- a/internal/client/github.go
+++ b/internal/client/github.go
@@ -245,7 +245,7 @@ func (c *githubClient) CreateRelease(ctx *context.Context, body string) (string,
 			opt.Page = resp.NextPage
 		}
 	} else {
-		release, _, err = c.client.Repositories.GetReleaseByTag(
+		release, _, _ = c.client.Repositories.GetReleaseByTag(
 			ctx,
 			ctx.Config.Release.GitHub.Owner,
 			ctx.Config.Release.GitHub.Name,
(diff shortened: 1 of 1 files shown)


caarlos0:
this is not going to work, unless we also remove all existing release artifacts first

qsliu2017:
updated. @caarlos0 please take a look

caarlos0:
I wonder if it isn't easier to just delete the draft release if it already exits before creating a new one? 

thoughts?

caarlos0:
fixed in https://github.com/goreleaser/goreleaser/pull/3318

PR #3152, opened 2022-06-11
[docs] Miscellaneous usage improvements

rfay:
This is a set of minor docs improvements.

I'm experimenting with goreleaser to see if I can use it with [DDEV](https://github.com/drud/ddev). I think I can. But I read through the (outstanding) docs a couple of times and thought that I could make some contributions as a native English speaker. There were a few phrases and usages that were not expected standard English. You can keep or not keep anything you like of course.


caarlos0:
hey @rfay! Thanks, lots of good input. As you guessed, I'm not a native english speaker, so any review in that sense helps a lot!

Also saw you update some outdated docs, all very good changes! 

Thanks!

caarlos0 (approved):

PR #3155, opened 2022-06-12
fix: allow wrap_in_directory in JSON Schema to be boolean

autopp:
This PR changes JSON schema of `.archives.wrap_directory`.
The current schema allows string only, but boolean should be also allowed.

VSCode's [YAML extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) seems to uses this schema and reports incorrect warning to user.
image

The document about of archives is [here](https://goreleaser.com/customization/archive/).

### DIFF
--- a/www/docs/static/schema-pro.json
+++ b/www/docs/static/schema-pro.json
@@ -196,7 +196,14 @@
 						"type": "array"
 					},
 					"wrap_in_directory": {
-						"type": "string"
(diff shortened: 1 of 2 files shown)


caarlos0:
those schemas are auto generated, this will have no effect.

autopp:
@caarlos0 Thanks for follow up!

PR #3167, opened 2022-06-18
docs: adds Koordinator to USERS.md

jasonliu747:
[Koordinator](https://koordinator.sh/) is a QoS based scheduling system for hybrid orchestration workloads on Kubernetes, bringing workloads the best layout and status. We uses GoReleaser to release containers, packages, and binaries.

So this PR would like to add Koordinator to USERS.md, thanks.


caarlos0:
nice, thanks!

PR #3190, opened 2022-06-23
docs: no-main error if running outside project root

rfay:
Minor fix; if you run goreleaser outside the root of the project you'll see the no-main error.

This may be a bug instead, maybe should be done that way.


caarlos0:
Thanks!

PR #3198, opened 2022-06-24
fix: run defaults pipe before before pipe

### LINKED ISSUE #3196: ProjectName doesn't get populated in before pipe
### What happened?

When `project_name` is not specified, Goreleaser guesses the project name from the git remote URL. However, not during the before pipe. During the before pipe, `{{.ProjectName}}` is empty and not guessed from the remote url.

A sample repository

```sh
mkdir test
cd test
git init
git remote add origin git@github.com:example/test.git
```

Goreleaser config

```yaml
before:
  hooks:
    - echo "{{.ProjectName}}"

release:
  footer: abc
```

### goreleaser version

```bash
goreleaser version 1.9.2-pro
commit: 7e532fd033dc7b29a6bd7633aa3ddcbc965f3d08
built at: 2022-05-27T16:27:23Z
built by: goreleaser
goos: darwin
goarch: amd64

https://goreleaser.com/pro
```

### GoReleaser Check

### Search

### Code of Conduct

### Additional context

_No response_

caarlos0:
otherwise, before might not have some template variables, for example,
the project name.

not sure if this will have adverse side effects though, need to think a
bit more about it.

fixes #3196

### DIFF
--- a/internal/pipeline/pipeline.go
+++ b/internal/pipeline/pipeline.go
@@ -50,8 +50,8 @@ var BuildPipeline = []Piper{
 	env.Pipe{},             // load and validate environment variables
 	git.Pipe{},             // get and validate git repo state
 	semver.Pipe{},          // parse current tag to a semver
-	before.Pipe{},          // run global hooks before build
 	defaults.Pipe{},        // load default configs
+	before.Pipe{},          // run global hooks before build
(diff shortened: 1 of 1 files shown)


caarlos0:
after some thought, I think it should be fine.

If you think I'm missing something, let me know and I'll revert.

PR #3203, opened 2022-06-25
feat: changelog group improvements

caarlos0:
This changes the changelog grouping feature a bit: it evaluates the
groups in the order they are declared in the config file, but renders
them following their `order` attribute.

This allows to have better filters, for instance, you might want to have
dependabot commits in another section, which does not need a high
priority in the changelog output, but does when evaluating, as it might
share the words with other groups: in goreleaser's case, the `feat`
prefix.

Giving that Go's regex seems to don't allow negative look-aheads (which
would help in this particular case), this seems like a good quality of
life improvement.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -52,17 +52,21 @@ changelog:
     - Merge branch
     - go mod tidy
   groups:
-    - title: 'New Features and updates'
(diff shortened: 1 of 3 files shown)


dirien:
Hi @caarlos0,

great improvement!

Issue #3202, opened 2022-06-25
Golang template actions aren't replaced in release.github.owner (environment variables)


rfay:
### What happened?

I need to replace `release.github.owner` with a template action (`{{ .Env.GITHUB_REPOSITORY_OWNER}}')`)

See https://github.com/rfay/junk/blob/cdc7e898179ae6857b3ae14e042fac22f613aa2b/.goreleaser.yml#L8

```yaml
  github:
    owner: "{{ .Env.GITHUB_REPOSITORY_OWNER }}"
    name: junk
```

`export GITHUB_REPOSITORY_OWNER=rfay` and then

```
$ goreleaser release --rm-dist
   • by using this software you agree with its EULA, available at https://goreleaser.com/eula
   • running goreleaser v1.9.2-pro
  ...
   • publishing
      • scm releases
         • creating or updating release repo={{ .Env.GITHUB_REPOSITORY_OWNER }}/junk tag=v0.3
         • release updated           url=
   ⨯ release failed after 0.79s error=scm releases: failed to publish artifacts: POST https://api.github.com/repos/%7B%7B%20.Env.GITHUB_REPOSITORY_OWNER%20%7D%7D/junk/releases: 404 Not Found []
```

Note that it's trying to push without replacing the environment variable, so it fails.

This same replacement does in fact work in `brews` and `aurs`.

### How can we reproduce this?

* Check out https://github.com/rfay/junk
* Change the details as necessary to something you have access to
* `export GITHUB_REPOSITORY_OWNER=you
* `goreleaser release --rm-dist`

### goreleaser version

```bash
goreleaser version 1.9.2-pro
```

### GoReleaser Check

- [X] goreleaser check shows no errors

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

rfay:
Is this just missing template replacement in https://github.com/goreleaser/goreleaser/blob/53b1ee07cfeb4895613ad37e39471a546ec71be9/internal/pipe/release/release.go#L92-L93 ?

caarlos0:
> Is this just missing template replacement in
> 
> https://github.com/goreleaser/goreleaser/blob/53b1ee07cfeb4895613ad37e39471a546ec71be9/internal/pipe/release/release.go#L92-L93
> 
> ?

essentially yes

done in https://github.com/goreleaser/goreleaser/pull/3205

rfay:
[...]

PR #3205, opened 2022-06-26
feat: allow to template release.repo

caarlos0:
this will allow to use templates in the release.github/gitea/gitlab owner and name fields.

closes https://github.com/goreleaser/goreleaser/issues/3202


rfay:
I built this and tested against my demo repository from #3202 and it worked fine. I couldn't test against my [real problem](https://github.com/rfay/ddev/blob/20220623_goreleaser_release/.goreleaser.yml) because I don't know how to build the pro version :). Thanks for the speedy response!

PR #3211, opened 2022-06-28
feat: allow multiple ids to be passed to build

juandiegopalomino:
This should be valid and build both build1 and build2 but no other
`goreleaser build --id build1 --id build2`

Because it's annoying that this feature is missing.

...

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -77,7 +77,7 @@ When using ` + "`--single-target`" + `, the ` + "`GOOS`" + ` and ` + "`GOARCH`" 
 	cmd.Flags().IntVarP(&root.opts.parallelism, "parallelism", "p", 0, "Amount tasks to run concurrently (default: number of CPUs)")
 	cmd.Flags().DurationVar(&root.opts.timeout, "timeout", 30*time.Minute, "Timeout to the entire build process")
 	cmd.Flags().BoolVar(&root.opts.singleTarget, "single-target", false, "Builds only for current GOOS and GOARCH")
-	cmd.Flags().StringArrayVar(&root.opts.id, "id", nil, "Builds only the specified build id(s)")
+	cmd.Flags().StringArrayVar(&root.opts.id, "id", nil, "Builds only the specified build ids")
 	cmd.Flags().BoolVar(&root.opts.deprecated, "deprecated", false, "Force print the deprecation message - tests only")
 	cmd.Flags().StringVarP(&root.opts.output, "output", "o", "", "Copy the binary to the path after the build. Only taken into account when using --single-target and a single id (either with --id or if config only has one build)")
 	_ = cmd.Flags().MarkHidden("deprecated")
@@ -184,7 +184,7 @@ func setupBuildID(ctx *context.Context, ids []string) error {
 	}
 
 	if len(keep) == 0 {
-		return fmt.Errorf("no builds with id(s) '%s'", ids)
+		return fmt.Errorf("no builds with ids '%s'", ids)
 	}
 
 	ctx.Config.Builds = keep
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -27,7 +27,7 @@ type buildCmd struct {
 
 type buildOpts struct {
 	config        string
-	id            string
+	id            []string
 	snapshot      bool
 	skipValidate  bool
 	skipBefore    bool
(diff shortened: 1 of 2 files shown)


cmd/build.go
```diff
@@ -27,7 +27,7 @@ type buildCmd struct {
 
 type buildOpts struct {
 	config        string
-	id            string
```
caarlos0, line 30: should probably rename this to `ids`

cmd/build.go
```diff
@@ -77,7 +77,7 @@ When using ` + "`--single-target`" + `, the ` + "`GOOS`" + ` and ` + "`GOARCH`"
 	cmd.Flags().IntVarP(&root.opts.parallelism, "parallelism", "p", 0, "Amount tasks to run concurrently (default: number of CPUs)")
 	cmd.Flags().DurationVar(&root.opts.timeout, "timeout", 30*time.Minute, "Timeout to the entire build process")
 	cmd.Flags().BoolVar(&root.opts.singleTarget, "single-target", false, "Builds only for current GOOS and GOARCH")
-	cmd.Flags().StringVar(&root.opts.id, "id", "", "Builds only the specified build id")
```
caarlos0, line 80: I would just day "ids" instead of "id(s)"

cmd/build.go
```diff
@@ -167,22 +167,24 @@ func setupBuildSingleTarget(ctx *context.Context) {
...
+		for _, id := range ids {
+			if build.ID == id {
+				keep = append(keep, build)
+				break
+			}
 		}
 	}
 
 	if len(keep) == 0 {
-		return fmt.Errorf("no builds with id '%s'", id)
```
caarlos0, line 185: only "ids" instead of "id(s)"

caarlos0 (commented):
a couple of minor tweaks.

would also be good to add a test with multiple IDs

cmd/build.go
juandiegopalomino: then it would be backwards incompatible and that's bad. More over since you add the ids one by one (--id id1 --id id2) it makes sense for it to be singular

caarlos0: the name of the arg can still be `id`, we can safely change name of the field in the struct

juandiegopalomino:
@caarlos0 done

cmd/build_test.go
```diff
@@ -205,7 +205,39 @@ func TestBuildFlags(t *testing.T) {
...
+					{
+						ID: "default",
+					},
+					{
+						ID: "foo",
+					},
+				},
+			})
+			require.NoError(t, setupBuildContext(ctx, buildOpts{
+				id: []string{"foo", "default"},
```
caarlos0, line 224:
```suggestion
				ids: []string{"foo", "default"},
```

cmd/build_test.go
```diff
@@ -205,7 +205,39 @@ func TestBuildFlags(t *testing.T) {
 				},
 			})
 			require.NoError(t, setupBuildContext(ctx, buildOpts{
-				id: "foo",
+				id: []string{"foo"},
```
caarlos0, line 208:
```suggestion
				ids: []string{"foo"},
```

cmd/build_test.go
```diff
@@ -106,7 +106,7 @@ func TestSetupPipeline(t *testing.T) {
 				context.New(config.Project{}),
 				buildOpts{
 					singleTarget: true,
-					id:           "foo",
+					id:           []string{"foo"},
```
caarlos0, line 109:
```suggestion
					ids:           []string{"foo"},
```

cmd/build_test.go
```diff
@@ -77,7 +77,7 @@ func TestSetupPipeline(t *testing.T) {
 			append(pipeline.BuildCmdPipeline, withOutputPipe{"foobar"}),
 			setupPipeline(context.New(config.Project{}), buildOpts{
 				singleTarget: true,
-				id:           "foo",
+				id:           []string{"foo"},
```
caarlos0, line 80:
```suggestion
				ids:           []string{"foo"},
```

cmd/build_test.go
```diff
@@ -66,7 +66,7 @@ func TestSetupPipeline(t *testing.T) {
 			pipeline.BuildCmdPipeline,
 			setupPipeline(context.New(config.Project{}), buildOpts{
 				singleTarget: true,
-				id:           "foo",
+				id:           []string{"foo"},
```
caarlos0, line 69:
```suggestion
				ids:           []string{"foo"},
```

cmd/build.go
```diff
@@ -167,22 +167,24 @@ func setupBuildSingleTarget(ctx *context.Context) {
...
+		return fmt.Errorf("no builds with ids '%s'", ids)
```
caarlos0, line 187: could probably look more readable with something like:

```suggestion
		return fmt.Errorf("no builds with ids: %s", strings.Join(ids, ", "))
```

caarlos0 (commented):
some suggestions

caarlos0:
all comments resolved in #3212

PR #3216, opened 2022-06-29
Not failing if there's the pre-built field in builds

juandiegopalomino:
Right now, goreleaser-pro adds the ability to specify the location of pre-built binaries, which is really cool: https://goreleaser.com/customization/build/#import-pre-built-binaries

The catch is that if you add them, then you can't use your yaml with regular goreleaser, like for the build step as it'll come up as incorrect schema. So I made this pr to accept the prebuilt field and just ignore it

...

...

...

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -307,6 +307,12 @@ func (a FlagArray) JSONSchema() *jsonschema.Schema {
 	}
 }
 
+// PreBuilt holds the data for prebuild binary specification. Currently a feature of goreleaser-pro, but back porting
+// this field to support yamls made for goreleaser-pro.
+type PreBuilt struct {
+	Path string `yaml:"path,omitempty"`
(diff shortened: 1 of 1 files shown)


caarlos0:
there are a lot of field that would cause this issue... for a while I thought about having the config part 100% shared, but my bet is that it would be very confusing for the users, as they would put seemingly valid configs in their yamls, which would be ignored...


I think I rather hard-fail.

As a workaround:
- if you wanna allow local dev without sharing a license, `goreleaser build` does not check for license
- you can also move the the pre build stuff into another file (say, `goreleaser.pro.yml`) and have it include the regular config. That way, oss can still read use the common config, and pro can specify the pro version, which would include the oss one. https://goreleaser.com/customization/includes/

if you have a different use case, I would love to hear it, please share :)

cheers!

PR #3223, opened 2022-07-04
chore: make skip messages more idiomatic

wesleimp:
Skip messages are not an error, in fact.

image


caarlos0:
ahh, that's a good one! 

thanks!

PR #3231, opened 2022-07-07
WIP: fix: propagate overridden env variables

### LINKED ISSUE #3230: `env` override is not applied to build env variables
### What happened?

I am attempting to cross-compile a CGO library, and thus have to set the toolchain for selected targets (through `CC` env variable).

When setting env variables in a build override - despite the arch/os matching - the overridden value is not set when actually performing the build.

I can see the build matching the override rule:
```
• overridden build details for linuxarm64        details={[-s -w] [] [-buildmode=c-shared] [] [] [CC=zig cc -target aarch64-linux]}
```

However, when running 

```
$ goreleaser build --debug
```

In the list of environment variables I do not see the overridden value:

```
[...] CGO_ENABLED=1 CC=zig cc GOOS=linux GOARCH=arm64 GOARM= GOMIPS= GOMIPS64= GOAMD64=
```

`CC` in particular is supposed to appear twice, with the default value, plus the overridden one as `CC=zig cc -target aarch64-linux`.

The `.goreleaser.yml` where I found the issue:

https://github.com/borgoat/farmfa/blob/ffi/.goreleaser.yml#L13-L42

### goreleaser version

```bash
goreleaser version dev
goos: linux
goarch: amd64
module version: v1.10.1, checksum: h1:NXCiqc4DJAG0y+iXDZKQew+0logeuZhYrCn4KZOrF0o=

https://goreleaser.com
```

### GoReleaser Check

### Search

### Code of Conduct

### Additional context

This is related to:

* #3201 
* #3119

borgoat:
This is to use the overridden env variables in the build step.

Right now, if providing environment variables as overrides, they are not used when invoking go build.

This fixes #3230 .

This is to get some early feedback, because while this effectively fixes the issue, the unit test is not properly implemented.

I'm wondering whether it'd be more effective to move the 3 steps that prepare the env variables currently in `Build` into a separate function?

### DIFF
--- a/internal/builders/golang/build.go
+++ b/internal/builders/golang/build.go
@@ -162,7 +162,13 @@ func (*Builder) Build(ctx *context.Context, build config.Build, options api.Opti
 		"GOAMD64="+options.Goamd64,
 	)
 
-	cmd, err := buildGoBuildLine(ctx, build, options, artifact, env)
+	details, err := withOverrides(ctx, build, options)
+	if err != nil {
+		return err
+	}
+	env = append(env, details.Env...)
+
+	cmd, err := buildGoBuildLine(ctx, build, details, options, artifact, env)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


caarlos0:
oh, sorry, did not see you were working on it... already done #3232

Issue #3284, opened 2022-08-03
HomeBrew Dependencies Version 


abebars:
### Is your feature request related to a problem? Please describe.

Currently, There is an option to mark a dependency as Optional but you can specify the version for the dependency. 
It would be great if we can also bake the version in the configuration because it's already supported in the [HomeBrew Formula ](https://docs.brew.sh/Formula-Cookbook#specifying-other-formulae-as-dependencies).

### Describe the solution you'd like

Update the config to have version string as an option

```
type HomebrewDependency struct {
	Name string `yaml:"name,omitempty"`
	Type string `yaml:"type,omitempty"`
        Version string `yaml:"version,omitempty"`
}
```
### Describe alternatives you've considered

There are no other options for that at the moment other than updating the Formula manually https://github.com/goreleaser/goreleaser/discussions/3283

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

caarlos0:
how does the ruby syntax look like for when there is both a type and a version?

fwiw, if anyone wants to work on this:

- edit config.go
- edit brew/template.go
- add a test in brew_test.go
- `go test -v ./internal/pipe/brew/... -update`
- commit :)

purush7:
@caarlos0  Can I work on this issue?

caarlos0:
> @caarlos0 Can I work on this issue?

sure think! Thanks

PR #3292, opened 2022-08-07
feat: add version for homebrew dependencies

purush7:
...

To specify the version of homebrew dependencies

...

Issue : #3284
...


internal/pipe/brew/template.go
```diff
@@ -49,6 +49,7 @@ class {{ .Name }} < Formula
   {{ range $index, $element := . }}
   depends_on "{{ .Name }}"
   {{- if .Type }} => :{{ .Type }}{{- end }}
+  {{- if .Version}} => :{{ .Version }}{{- end }}
```
caarlos0, line 52: how should it look like if it has both the version and type options set?

this will eval to something like:

```rb
depends_on "foo" => :optional => "v1.2.3"
```

Also I think the version shouldn't be a symbol here...

caarlos0 (changes requested):
Thanks for the PR, but I think this is not 100% correct:

- version should be a string, not a symbol https://docs.brew.sh/Formula-Cookbook#specifying-other-formulae-as-dependencies
- I am not sure what should it look like if it has both option and a version... this needs some research I'm afraid

caarlos0:
fixed https://github.com/goreleaser/goreleaser/issues/3284

internal/pipe/brew/template.go
caarlos0: I guess a sensible thing to do here is to use the version if given, and otherwise use the type, if both type and version are given, use just the version... I was looking into homebrew-core and did not find anything using both...

caarlos0: btw, I don't remember seeing anything with a version requirement tbh, also searching github rn and also can't find any examples...

https://github.com/search?l=Ruby&q=depends_on&type=Code

Issue #3269, opened 2022-07-31
Add sprig templates


andig:
### Is your feature request related to a problem? Please describe.

For historical reasons, our naming scheme is `0.<minor version>`. For publishing apt packages we need `0.<minor version>.<patch level>`. Since we've introduced bugfix releases, we can no longer append <patch level> als `.0` but instead need to do this optionally.

To build the required version number, we need to process `.Version` and determine the number of segments (i.e. check `len(split("."))`).

### Describe the solution you'd like

add `sprig` templates which provide the `split` function amongst other useful functions. Sprig is already a dependency.

### Describe alternatives you've considered

Fix the entire build process (which is a much larger effort). Other release tools like e.g. `seihon` already include sprig for that purpose.

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Additional context

Would be happy to provide a PR

andig:
@caarlos0 is this something you would consider?

caarlos0:
not sure if I really want to add that many new functions

you can probably use the `incpatch` function in your case though... https://goreleaser.com/customization/templates/

andig:
What I was thinking to do is making our version semver if it isn't due to patch level:

    {{ .Version }}{{ if lt(len(split .Version ".") 3) }}.0

Afaikt `incpatch` will panic since `.Version` is not semver.

> not sure if I really want to add that many new functions

Would you accept a minimal PR for `split` and `len` only?

caarlos0:
> Afaikt `incpatch` will panic since `.Version` is not semver.

it doesn't, it'll set the patch to 0, thus `incpatch "v2.2"` will eval to `v2.2.1`



> Would you accept a minimal PR for `split` and `len` only?

more likely yes

andig:
The docs for `incpatch` say:

> Will panic if not a semantic version.

I've created new PR for adding `split` only. `len` is available by default.

caarlos0:
> Will panic if not a semantic version.

yes, that means if the version is something like 2012-33-343, if it has at least 1 dot (e.g. 12.2) it works... I tested it :)

PR #3293, opened 2022-08-07
Add split function

andig:
Fix https://github.com/goreleaser/goreleaser/issues/3269 by adding the `split` template function for `strings.Split()`


caarlos0:
thanks!

PR #3302, opened 2022-08-10
Support mTLS with client certificate when configured.

scr-oath:
- Support artifactory servers that use mTLS to authorize
- Add configuration and re-run the schema generation
- Add a test case that uploads with a test certificate


pkg/config/config.go
```diff
@@ -864,8 +864,10 @@ type Upload struct {
 	Mode               string            `yaml:"mode,omitempty" json:"mode,omitempty"`
 	Method             string            `yaml:"method,omitempty" json:"method,omitempty"`
 	ChecksumHeader     string            `yaml:"checksum_header,omitempty" json:"checksum_header,omitempty"`
+	ClientX509Cert     string            `yaml:"client_x509_cert" json:"client_x509_cert"`
+	ClientX509Key      string            `yaml:"client_x509_key" json:"client_x509_key"`
 	TrustedCerts       string            `yaml:"trusted_certificates,omitempty" json:"trusted_certificates,omitempty"`
-	Checksum           bool              `yaml:"checksum,omitempty" json:"checksum,omitempty"`
+	Checksum           bool              `yaml:"checksum,omitqempty" json:"checksum,omitempty"`
```
caarlos0, line 870: typo

caarlos0 (commented):
lgtm, can you also update the docs related to this?

from the top of my head:

- artifactory.md
- upload.md

Thanks!

pkg/config/config.go
scr-oath: Fixed

scr-oath:
> lgtm, can you also update the docs related to this?
> 
> from the top of my head:
> 
> * artifactory.md
> * upload.md
> 
> Thanks!

Great, I added information to both of those, and also added a `Pipfile`, `Pipfile.lock`, and `README.md` to www to show how to iterate with `mkdocs serve` while documenting.

[Perhaps a task entry could be written to help with that too? But I'd love it if that could be a follow up as I'm not terribly familiar with task.](https://github.com/go-task/task)

caarlos0:
No need to commit those, I iterate locally with Docker... there is a task for it already as well:

https://github.com/goreleaser/goreleaser/blob/main/Taskfile.yml#L103

scr-oath:
> No need to commit those, I iterate locally with Docker... there is a task for it already as well:
> 
> https://github.com/goreleaser/goreleaser/blob/main/Taskfile.yml#L103

I just tried it - nice that it exists, but it's so very slow on M1 mac.

*WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested*

I'll just remove the files and/or rejigger to include info on using docs:serve task target - for a seldom task, being a little slow isn't the end of the world.

caarlos0:
hmm, I always use it on linux/amd64


hopefully this helps https://github.com/squidfunk/mkdocs-material/pull/4240

PR #3320, opened 2022-08-18
Chore: update discord webhook library

TopiSenpai:
This commit will update the discord webhook library. I deprecated https://github.com/DisgoOrg/disgohook a while ago and saw it being used here. The module was moved into https://github.com/disgoorg/disgo/tree/development/webhook

disgohook had a few bugs which are resolved in the new module and also the new module should be nicer to use in general!


caarlos0:
ah super cool, thanks for the lib and for the PR!

FWIW, It might be good to deprecate the module: https://tip.golang.org/ref/mod#go-mod-file-module-deprecation

TopiSenpai:
oh totally forgot about this ty!

PR #3330, opened 2022-08-21
feat: delay github tag creation

### LINKED ISSUE #3044: Delay creation of remote tag

In Hugo, there have been situations where the build fails for some reason and I need to make some adjustments to a script or similar and then try again.

What I do now is:

* Push a release tag to GitHub.
* Run Goreleaser.
* If it fails and I need to change something, I delete the remote tag and try again.

I know I should probably increment a version number, but that is just too much work. I don't mind doing the above, but with the new `gosumdb` validation in place and people running `go install github.com/gohugoio/hugo@v0.97.0` to install Hugo, we now end up with `checksum mismatch` errors that seem to be impossible to get out of.

https://github.com/gohugoio/hugo/issues/9785

I assume that the current workflow 

1. does not require the remote tag to exist when building the binaries/archives
2. does require the remote tag to exist when pushing the release to GitHub

The GitHub release API does allow a non-existing tag to be provided if you also provide the commit hash to be used for the tag (via `target_commitish`).

image

I'm not sure how feasible the above would be in Goreleaser, but it would be nice to delay the creation of the release tag as long as possible.

I have looked at the Goreleaser documentation for a way to split the release build in 2: 1) Build archives and 2) Push tag to GitHub 3) Release to Github, but that does not seem to be possible (there is a `--skip-publish` flag, but not a `--just-publish` flag).

### Search

### Code of Conduct

### Additional context

_No response_

caarlos0:
closes #3044

What happens is that, if the tag is created only locally, it'll create the tag in the remote upon creating the release. The tag still needs to be created locally though!

@bep let me know if that works for you.

### DIFF
--- a/internal/client/github.go
+++ b/internal/client/github.go
@@ -217,11 +217,12 @@ func (c *githubClient) CreateRelease(ctx *context.Context, body string) (string,
 	body = truncateReleaseBody(body)
 
 	data := &github.RepositoryRelease{
-		Name:       github.String(title),
-		TagName:    github.String(ctx.Git.CurrentTag),
-		Body:       github.String(body),
-		Draft:      github.Bool(ctx.Config.Release.Draft),
-		Prerelease: github.Bool(ctx.PreRelease),
+		Name:            github.String(title),
+		TagName:         github.String(ctx.Git.CurrentTag),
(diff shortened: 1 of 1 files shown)


caarlos0:
ok, I just tested this again (100 times) and it works. 
Will merge!

PR #3378, opened 2022-09-15
fix: add COSIGN_EXPERIMENTAL=1 for verify-blob

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

Cross-ref: https://github.com/fluxcd/source-controller/issues/899


caarlos0:
Thanks!

PR #3381, opened 2022-09-15
fix: build --single-target when using specific targets

caarlos0:
When using specific targets in the config, single target becomes ineffective. This should fix it.  This also removes other build options like gomips, goarm, goamd64, etc, as go will infer them from the current OS.  In theory we should one day support all the GO* flags go build supports.


agilgur5:
For reference (as a backlink, since GH doesn't generate those for discussions, and for any future readers like me), this bug was noted in https://github.com/goreleaser/goreleaser/discussions/3371#discussioncomment-3656182

PR #3401, opened 2022-09-25
fix: skip instead of hard fail

caarlos0:
This makes partially set up brews, krews, etc skip when no archives match, instead of hard-failing.

I'm not sure if this is better or not, but it would make sense for things like partial builds (e.g. build for linux only), so the config might get reused.

Moreover, IMHO, its better to skip than hard fail because people will likely have to create a new tag anyway, due to go mod caching et al... so with this at least they don't get a broken release.

tl;dr: still thinking about it.

### DIFF
--- a/internal/pipe/aur/aur.go
+++ b/internal/pipe/aur/aur.go
@@ -3,7 +3,6 @@ package aur
 import (
 	"bufio"
 	"bytes"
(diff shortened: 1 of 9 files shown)


caarlos0:
will leave this out for now

PR #3402, opened 2022-09-25
feat: docker meta images

caarlos0:
Until now, GoReleaser was not complaining when no artifacts were found, which could cause the actual docker build to fail if the `Dockerfile` expects them to be there.

So, from now on, this will be explicit, more or less like the `meta` option in the archives config.


caarlos0:
actually, probably better to leave as is

PR #3406, opened 2022-09-26
Error in the doc: archives.*.builds is type []string

xNok:
The example creates an error because builds is supposed to be an array here is shown as a string

...

...

...


caarlos0:
thanks!

PR #3407, opened 2022-09-27
Client key and cert should not be required.

scr-oath:
Fixing schema change from #3302

Signed-off-by: Sheridan C Rawlins <scr@yahooinc.com>

...

...

...

### DIFF
--- a/www/docs/static/schema.json
+++ b/www/docs/static/schema.json
@@ -2538,11 +2538,7 @@
 					}
 				},
 				"additionalProperties": false,
-				"type": "object",
-				"required": [
-					"client_x509_cert",
-					"client_x509_key"
-				]
+				"type": "object"
 			},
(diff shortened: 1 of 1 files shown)


caarlos0:
format fixed on #3409

caarlos0:
Thanks!

scr-oath:
FYI if curious, this cert stuff does work for mTLS protected sites; I was just surprised that I had made it required before (which could have broken others that use basic auth, or at least confused them in smart IDEs that look at the schema)

caarlos0:
> FYI if curious, this cert stuff does work for mTLS protected sites; I was just surprised that I had made it required before (which could have broken others that use basic auth, or at least confused them in smart IDEs that look at the schema)

it was just the yamlschema wrong, the code would work fine without the certs for non mtls stuff

PR #3414, opened 2022-09-28
feat: permanently remove buildpacks

caarlos0:
removing for good!


caarlos0:
(disable auto-merge because I wanna do a patch release before merging any big things into it)

PR #3419, opened 2022-09-30
docs: adds Meroxa to USERS.md

raulb:
One of the examples of where we use it https://github.com/meroxa/cli/blob/master/.github/workflows/release.yml#L31-L32.

Thank you for such fantastic tool!


caarlos0:
Thanks!

PR #3425, opened 2022-09-30
docs: add Goldenhand Software to users

MrWormHole:
This will add Goldenhand Software to users' list 

I would like to support GoReleaser as a solo contactor

https://www.goldenhandsoftware.co.uk/

Note: I love the announce functionality of GoReleaser and I do think anything other than Slack should be included in GoReleaser PRO. Announce requires a good time of development for change notes to be announced in different integrations. As a developer who do integrations every day, I do think announce section should require PRO (excluding slack and webhook one)

image

### DIFF
--- a/USERS.md
+++ b/USERS.md
@@ -20,6 +20,7 @@ Here's a running list of some organizations using GoReleaser[^1]:
 1. [GitHub](https://github.com)
 1. [Go Buffalo](https://gobuffalo.io)
 1. [GolangCI](https://golangci.com)
+1. [Goldenhand Software](https://www.goldenhandsoftware.co.uk)
 1. [Google](https://google.com)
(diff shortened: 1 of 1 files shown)


caarlos0:
Hey, thanks @MrWormHole, and thanks for the kind words!

Yes, a lot more features could be pro only, but they were already public when I launched the pro, and some were contributed by other people.

Changing it now would break people's workflow, so I rather leave as is. 

That said, I keep working on adding more things to the pro, it has been a bit quiet because I'm working in some bigger features now, that require a lot of thought on both how to implement and how to keep the OSS fork in sync, as well as the usability of the feature itself. 

Anyway, I really appreciate the support! And thanks again for the PR 🙏

PR #3434, opened 2022-10-04
fix: gomod.env not being used

### LINKED ISSUE #3426: Gomod pipe should include config env so it works with Go Workspaces and Go.mod replace directives
### What happened?

goreleaser recommends using `goreleaser build` instead of `go build` to keep a single source of truth which is useful in development environments.

When working with a Go Workspace `go.work`, `goreleaser build` needs to be combined with `env` `GOWORK=off` but is not included when doing `go list -m`, see below.

We have a goreleaser config with the following content:

```yaml
builds:
  -
    main: ./cmd/station-agent
    binary: station-agent
    env:
      - GOWORK=off
    goos:
      - linux
    goarch:
      - amd64
      - arm
      - arm64
    flags:
      - -mod=vendor
      - -tags="linux" # sqlite3 requirement https://github.com/mattn/go-sqlite3#linux
      - -v
```

In the step/pipe `loading go mod information`, `goreleaser build` tries to run `go list -m ` but it fails with the following error:

```
failed to get module path: exit status 1: go: open /tmp/sdk-go/go.mod: no such file or directory
```

That's because there are some `replace` directives in my `go.mod` file.

This is not a problem outside goreleaser if we set the env variable `GOWORK=off` as in `GOWORK=off go list -m -mod=vendor`.

I already include `- GOWORK=off` as part of my `env` but `internal/pipe/gomod/gomod.go` is not using to call `go list -m`.

1. Create a project with a go module
2. Use a `replace` directive in `go.mod`
3. use go workspaces (`go.work`)
4. Run `goreleaser build`

### goreleaser version

```bash
goreleaser version dev
goos: linux
goarch: amd64
module version: v1.11.4, checksum: h1:bmXfkrBa6xv3WzzZM7U0QCOmpFAanIm+91tFB4DYUnI=

https://goreleaser.com
```

### GoReleaser Check

### Search

### Code of Conduct

### Additional context

I can send a PR to make sure we execute the external command with `ctx.Config.Env`

caarlos0:
Fixes  #3426

### DIFF
--- a/internal/pipe/gomod/gomod.go
+++ b/internal/pipe/gomod/gomod.go
@@ -33,7 +33,9 @@ func (Pipe) Run(ctx *context.Context) error {
 	if ctx.Config.GoMod.Mod != "" {
 		flags = append(flags, "-mod="+ctx.Config.GoMod.Mod)
 	}
-	out, err := exec.CommandContext(ctx, ctx.Config.GoMod.GoBinary, flags...).CombinedOutput()
+	cmd := exec.CommandContext(ctx, ctx.Config.GoMod.GoBinary, flags...)
+	cmd.Env = append(ctx.Env.Strings(), ctx.Config.GoMod.Env...)
+	out, err := cmd.CombinedOutput()
 	result := strings.TrimSpace(string(out))
 	if result == go115NotAGoModuleError || result == go116NotAGoModuleError {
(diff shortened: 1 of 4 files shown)


djgilcrease (approved):
LGTM

PR #3436, opened 2022-10-05
feat: nfpm for ios

### LINKED ISSUE #3410: Add iOS builds to nFPM deb pkgs

No problem, but there is a community of people that pkg bins/apps for jailbroken phones etc and this would make it much easier for them to include my tools in their distros.

I believe that we could also pkg the GOOS ios GOARCH arm64 builds into deb pkgs if we set the "Architecture: field in the control file to `iphoneos-arm64`". 

Is that possible to add as a config?

I'm not sure how else to do this?

### Search

### Code of Conduct

### Additional context

I think from a UX point of view you would just add `ios` as a one of the `builds` to be packaged and it would auto set the `iphoneos-arm64` and apologies if you have already implemented this.

caarlos0:
Jailbroken iOS can install deb packages, and it seems the only change needed is to set the OS to `ios` instead of `linux`.

Closes #3410

cc/ @blacktop


blacktop:
https://github.com/goreleaser/goreleaser/issues/3410#issuecomment-1264787057

@caarlos0 does this also need to be added to nFPM?

blacktop:
cc/ @CRKatri @Arsen6331 just double checking we good?

CRKatri:
Sorry if I'm blind, but where is the `Architecture` field set? And is it customizable? Because for iOS we technically need both `iphoneos-arm` and `iphoneos-arm64` but both as 64bit (for compatibility reasons).

caarlos0:
> Sorry if I'm blind, but where is the `Architecture` field set? And is it customizable? Because for iOS we technically need both `iphoneos-arm` and `iphoneos-arm64` but both as 64bit (for compatibility reasons).

its the `platform` field in nfpm. 

can we support only the latest? adding support for both would a bit mesier

blacktop:
LGTM 👍

caarlos0:
so, is the iphoneos-arm64 only enough?

CRKatri:
I suppose.

Issue #3328, opened 2022-08-20
Add Fig integration


dirien:
### Is your feature request related to a problem? Please describe.

This enhancement will be in two steps:

1. Add Fig autocomplete function with the suggested way described here -> https://fig.io/docs/guides/integrating/getting-started

2. Will introduce the autopush function for the `goreleaser` spec in the fig autocomplete repository -> https://fig.io/docs/guides/integrating/push-to-autocomplete-action

### Describe the solution you'd like

Full automatic creation of the fig spec

### Describe alternatives you've considered

Currently I created the Spec by hand -> https://github.com/withfig/autocomplete/pull/1096

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

caarlos0:
can't goreleaser users do something like `goreleaser completion fig >somepath` like all other completions?

I really dislike having a PR for this every time I change anything, even if its automated... seems backwards, to say the least

dirien:
Good question, maybe need to ask @fedeci for help! Maybe I missed a part in the integration guide!

fedeci:
Thanks for the ping @dirien!
Hi @caarlos0, the PR is created against the `withfig/autocomplete` repo so apart from having an action running here (e.g. when you release a new version) everything would be handled by the fig team.

Of course the integration can be used as you mean. Essentially we expose two main methods:
- An "high-level" one that automatically adds the "generate-fig-spec" command to your cli so that when you run "goreleaser generate-fig-spec" it prints the completion file
- A "low-level" that instead of doing everything mentioned above just takes as input your root command and returns a stringified spec

We can bind the second one to "goreleased completion fig" and the in the github action use it easily.

caarlos0:
ok, so, is it possible to run `goreleaser completion fix >somepath`, and fig would pick that up?

if so, which path?

dirien:
I started to implement the "high level" version so is the "low level" version better as not inversive to the code base? How does the "low level" works? Any link to the docs, as I could find this in the current docs. But maybe it's me and i searched for the wrong subject!

fedeci:
@caarlos0  Fig does not pick local paths for specs but we rely on specs being served from a cdn, so everything that we want to be public needs to be pushed to the https://github.com/withfig/autocomplete repo. 
[...]

PR #3444, opened 2022-10-06
fix: annotate flags so they complete better

caarlos0:
refs https://github.com/withfig/autocomplete/pull/1584

cc/ @fedeci - thanks for the pointers :) 
refs #3328


fedeci (approved):
Great!

dirien (approved):

PR #3467, opened 2022-10-15
feat(deps): update nfpm to v2.20.0

caarlos0:
Signed-off-by: Carlos A Becker <caarlos0@users.noreply.github.com>


caarlos0:
will add the new archlinux packages only in goreleaser v1.13, probably

PR #3468, opened 2022-10-15
 feat: integrate ConventionalExtension from nfpm v2.20.0 

Arsen6331:
This PR upgrades nFPM to v0.20.0, and integrates the new `ConventionalExtension` method to use the correct extension for the packaging format that is being used.

Currently, goreleaser uses the name of the format to determine the extension. This has worked fine, but with the introduction of Archlinux packages, goreleaser has to handle packages with extensions that don't match the name of the format, since Archlinux uses `.pkg.tar.zst` as the extension.

goreleaser/nfpm#546
goreleaser/nfpm#543


caarlos0:
closes #3467

PR #3506, opened 2022-10-28
Remove `--skip-after` option

n8felton:
This option is not present in v1.12.3


caarlos0:
It is on [goreleaser pro](https://goreleaser.com/pro), although it probably needs to be specified here.

These docs are auto-generated.

PR #3511, opened 2022-10-30
feat: build of shared/static libraries

borgoat:
This PR improves the handling of shared or static libraries by GoReleaser. It uses the default behaviour of the Go compiler by appending the right extension to libraries.

* `.so` and `.a` for Linux shared libraries and static libraries respectively
* `.dylib` and `.a.` on Darwin
* `.dll` and `.lib` on Windows (pre-existent)

It does not add any configuration option to `.goreleaser.yml`, since it leverages the existing `buildmode` flag.

Additionally, this PR takes care of adding the generated header file into the archive.

Personally I would leverage this change to release some software both as a CLI and as a shared library. I believe others who use CGo or need interoperability with Go from other languages could benefit from this.

This was previously discussed in #3497.

I couldn't quite think of a proper way to add some tests to the header archiving feature. Any recommendation?

### DIFF
--- a/internal/builders/golang/build.go
+++ b/internal/builders/golang/build.go
@@ -215,11 +215,12 @@ func withOverrides(ctx *context.Context, build config.Build, options api.Options
 
 		if optsTarget == overrideTarget {
 			dets := config.BuildDetails{
-				Ldflags:  build.BuildDetails.Ldflags,
-				Tags:     build.BuildDetails.Tags,
-				Flags:    build.BuildDetails.Flags,
-				Asmflags: build.BuildDetails.Asmflags,
-				Gcflags:  build.BuildDetails.Gcflags,
+				Buildmode: build.BuildDetails.Buildmode,
+				Ldflags:   build.BuildDetails.Ldflags,
+				Tags:      build.BuildDetails.Tags,
+				Flags:     build.BuildDetails.Flags,
+				Asmflags:  build.BuildDetails.Asmflags,
+				Gcflags:   build.BuildDetails.Gcflags,
 			}
 			if err := mergo.Merge(&dets, o.BuildDetails, mergo.WithOverride); err != nil {
 				return build.BuildDetails, err
--- a/internal/builders/golang/build.go
+++ b/internal/builders/golang/build.go
@@ -192,6 +192,8 @@ func (*Builder) Build(ctx *context.Context, build config.Build, options api.Opti
 		}
 	}
 
+	addHeaderArtifactIfLibrary(ctx, build, options)
+
 	ctx.Artifacts.Add(artifact)
 	return nil
 }
@@ -365,4 +367,33 @@ func hasMain(file *ast.File) bool {
 		}
 	}
 	return false
+}
+
+func addHeaderArtifactIfLibrary(ctx *context.Context, build config.Build, options api.Options) {
+	for _, s := range build.BuildDetails.Flags {
+		if s == "-buildmode=c-shared" || s == "-buildmode=c-archive" {
+			fullPathWithoutExt := strings.TrimSuffix(options.Path, options.Ext)
+			basePath := filepath.Base(fullPathWithoutExt)
+			fullPath := fullPathWithoutExt + ".h"
+			headerName := basePath + ".h"
+
+			header := &artifact.Artifact{
+				Type:    artifact.Header,
+				Path:    fullPath,
+				Name:    headerName,
+				Goos:    options.Goos,
+				Goarch:  options.Goarch,
+				Goamd64: options.Goamd64,
+				Goarm:   options.Goarm,
+				Gomips:  options.Gomips,
+				Extra: map[string]interface{}{
+					artifact.ExtraBinary: headerName,
+					artifact.ExtraExt:    ".h",
+					artifact.ExtraID:     build.ID,
+				},
+			}
+
+			ctx.Artifacts.Add(header)
+		}
+	}
 }
--- a/internal/builders/golang/build.go
+++ b/internal/builders/golang/build.go
@@ -237,6 +237,9 @@ func withOverrides(ctx *context.Context, build config.Build, options api.Options
 func buildGoBuildLine(ctx *context.Context, build config.Build, details config.BuildDetails, options api.Options, artifact *artifact.Artifact, env []string) ([]string, error) {
 	cmd := []string{build.GoBinary, build.Command}
 
+	// tags, ldflags, and buildmode, should only appear once, warning only to avoid a breaking change
+	validateUniqueFlags(details)
+
 	flags, err := processFlags(ctx, artifact, env, details.Flags, "")
 	if err != nil {
 		return cmd, err
@@ -281,6 +284,20 @@ func buildGoBuildLine(ctx *context.Context, build config.Build, details config.B
 
 	cmd = append(cmd, "-o", options.Path, build.Main)
 	return cmd, nil
+}
+
+func validateUniqueFlags(details config.BuildDetails) {
+	for _, flag := range details.Flags {
+		if strings.HasPrefix(flag, "-tags") && len(details.Tags) > 0 {
+			log.WithField("flag", flag).WithField("tags", details.Tags).Warn("tags is defined twice")
+		}
+		if strings.HasPrefix(flag, "-ldflags") && len(details.Ldflags) > 0 {
+			log.WithField("flag", flag).WithField("ldflags", details.Ldflags).Warn("ldflags is defined twice")
+		}
+		if strings.HasPrefix(flag, "-buildmode") && details.Buildmode != "" {
+			log.WithField("flag", flag).WithField("buildmode", details.Buildmode).Warn("buildmode is defined twice")
+		}
+	}
 }
 
 func processFlags(ctx *context.Context, a *artifact.Artifact, env, flags []string, flagPrefix string) ([]string, error) {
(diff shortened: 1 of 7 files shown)


internal/builders/golang/build.go
```diff
@@ -192,6 +192,8 @@ func (*Builder) Build(ctx *context.Context, build config.Build, options api.Opti
 		}
 	}
 
+	addHeaderArtifactIfLibrary(ctx, build, options)
+
 	ctx.Artifacts.Add(artifact)
```
caarlos0, line 197: I think that if the header thing is added.. there's no binary to add, so this can probably be removed.

I would probably make the `addHeaderArtifact...etc` return an artifact, maybe either the binary or the library, and use it here. 

also probably remove the one defined in line 138

internal/builders/golang/build.go
```diff
@@ -366,3 +368,32 @@ func hasMain(file *ast.File) bool {
 	}
 	return false
 }
+
+func addHeaderArtifactIfLibrary(ctx *context.Context, build config.Build, options api.Options) {
+	for _, s := range build.BuildDetails.Flags {
+		if s == "-buildmode=c-shared" || s == "-buildmode=c-archive" {
```
caarlos0, line 374: one thing to be aware is that flags are templateable... so this might be an issue here...

internal/pipe/build/build.go
```diff
@@ -226,5 +226,25 @@ func extFor(target string, flags config.FlagArray) string {
...
+		for _, s := range flags {
+			if s == "-buildmode=c-shared" {
+				return ".so"
+			}
+			if s == "-buildmode=c-archive" {
+				return ".a"
+			}
+		}
+	}
+	if strings.Contains(target, "darwin") {
```
caarlos0, line 239: should we add windows dlls as well?

caarlos0:
looking good so far! Thank you!

I added a couple of comments, and regarding your question about the archive, you'll need to add the Header artifact type in the filter there: https://github.com/goreleaser/goreleaser/blob/e269a1d09599e8a1399393fc8230ed43a75bc851/internal/pipe/archive/archive.go#L91

There might be more places, so you can probably search for `artifact.Binary` and see what's there.

Another thing that comes to mind is that we have the `UploadableBinary` type, which is just an unarchived binary... I'm not sure that's a use case, but if people want unarchived libraries to be released, we'll need to change that too (also in the archive pipe).


That's all I remember for now about this. BTW feel free to ping me if you need any help, I'm mostly available on discord for quick chats if needed 🙏


And thanks again for your effort and work on this 🤘

internal/builders/golang/build.go
borgoat: This is actually the intended behaviour: when building a shared library, a C header always goes with it. The header documents all the exported functions that library users may access, the `so` file provides the actual implementation.

internal/pipe/build/build.go
borgoat: This was implemented before, in PR #1243 - I can reorder the blocks though if you like, maybe in alphabetical order?

internal/builders/golang/build.go
caarlos0: ah, got it... so maybe we could have both `artifacts.Header` and `artifacts.Library` or something? 

That would probably prevent some errors, like brew taps for libraries and stuff..

internal/pipe/build/build.go
caarlos0: yes, I think its good to have them all closer together

internal/builders/golang/build.go
borgoat: Fair point. I didn't think of that... How does this differ from the build pipe where we set the `so`, `dll`, etc extensions?
I guess the templates are resolved somewhere between here and then?
The only ideas I can think of right now would be to either:
* introduce a build-level setting to configure the buildmode
* have an archive setting for additional files, based on the binary output path

I guess you may have a better solution though!

borgoat:
Thanks for the review @caarlos0 !

I replied to the comments best I could. As for the other points:

> I added a couple of comments, and regarding your question about the archive, you'll need to add the Header artifact type in the filter there:
> 
> https://github.com/goreleaser/goreleaser/blob/e269a1d09599e8a1399393fc8230ed43a75bc851/internal/pipe/archive/archive.go#L91
> 
> There might be more places, so you can probably search for `artifact.Binary` and see what's there.

I added this line here: https://github.com/goreleaser/goreleaser/pull/3511/files#diff-4b69e49926d4bcaf08f5563b284ff33ac8f7a60a596dc62898ec333253014088R94

Did I miss it in other places though?

> Another thing that comes to mind is that we have the `UploadableBinary` type, which is just an unarchived binary... I'm not sure that's a use case, but if people want unarchived libraries to be released, we'll need to change that too (also in the archive pipe).

I need to think this through: the interesting exception in that scenario would be that the header file looks always the same, so it could be uploaded just once: it is not specific to any target os/arch combo.

internal/builders/golang/build.go
borgoat: Aha I see what you're saying. Yes then I guess it's reasonable to handle it separately at a higher level.

That does open up another question though! I didn't think of it before.. it would very much make sense to have this feature integrate with homebrew and nfpm, with libs and headers being deployed in the lib and include path respectively... But there again, I don't think I currently have the knowledge to make it happen..

internal/pipe/build/build.go
borgoat: I reordered them now, also in the tests for consistency

internal/builders/golang/build.go
caarlos0: > Fair point. I didn't think of that... How does this differ from the build pipe where we set the so, dll, etc extensions?

It seems that it is also wrong 🤔 

On a second thought, I think that's not something people will template, so it might be fine. I'll add an issue to refactor that at some point though.

internal/builders/golang/build.go
caarlos0: https://github.com/goreleaser/goreleaser/issues/3524

for now we can proceed as is. Future me will handle it :D

caarlos0:
> Did I miss it in other places though?

I think maybe nfpm if we want to create library packages... but that might be a bigger effort (and maybe worth doing in another pr)



> I need to think this through: the interesting exception in that scenario would be that the header file looks always the same, so it could be uploaded just once: it is not specific to any target os/arch combo.

hmmm that seems like a bigger change too.. maybe for now we just assume that libraries will always be archived?

borgoat:
> > Did I miss it in other places though?
> 
> I think maybe nfpm if we want to create library packages... but that might be a bigger effort (and maybe worth doing in another pr)
> 
> > I need to think this through: the interesting exception in that scenario would be that the header file looks always the same, so it could be uploaded just once: it is not specific to any target os/arch combo.
> 
> hmmm that seems like a bigger change too.. maybe for now we just assume that libraries will always be archived?

I guess that makes sense.. then the only point missing should then be to create 1 or 2 new Artefact types "DynamicLibrary"/"StaticLibrary" for c-shared and c-archive respectively (or just 1 "Library" for both? wdyt?) Anything else I may be missing?

caarlos0:
> > > Did I miss it in other places though?
> > 
> > 
> > I think maybe nfpm if we want to create library packages... but that might be a bigger effort (and maybe worth doing in another pr)
> > > I need to think this through: the interesting exception in that scenario would be that the header file looks always the same, so it could be uploaded just once: it is not specific to any target os/arch combo.
> > 
> > 
> > hmmm that seems like a bigger change too.. maybe for now we just assume that libraries will always be archived?
> 
> I guess that makes sense.. then the only point missing should then be to create 1 or 2 new Artefact types "DynamicLibrary"/"StaticLibrary" for c-shared and c-archive respectively (or just 1 "Library" for both? wdyt?) Anything else I may be missing?

I think we can create both 2 new types, in case we need some extra handling somewhere... 

I can't think of anything else missing besided that and the nfpm/brew/etc stuff (which we can handle in the future)

borgoat:
Given the latest comments, to more easily support the 2 possible artefacts, and considering that the field would not be templateable anyway, I tried to refactor things a bit. I have introduced a `buildmode` config option instead, and configured all logic based on that. What do you think?

internal/builders/golang/build.go
```diff
@@ -266,6 +275,10 @@ func buildGoBuildLine(ctx *context.Context, build config.Build, details config.B
 		cmd = append(cmd, "-ldflags="+strings.Join(ldflags, " "))
 	}
 
+	if details.Buildmode != "" {
+		cmd = append(cmd, "-buildmode="+details.Buildmode)
```
caarlos0, line 279: I think it would be good to check if cmd already has a `-buildmode`, and if so, error out... 

my though behind that: people might set different buildmodes (one via flags and another with the buildmode option introduced in this pr), and it could get a bit confusing.

a warning might be good enough too, though.

wdyt?

caarlos0 (approved):
just a single comment, I believe we are **very** close!

Thank you a ton for this!

internal/builders/golang/build.go
borgoat: While testing this, I realised the same could happen with ldflags and tags, which I guess could cause some problems too? I hope you don't mind I pushed a fix for it in this same MR, just for consistency.
As you suggested, I made it into a plain warning because erroring could be a breaking change.

caarlos0: ah, indeed, that looks nice, thanks!

borgoat:
So, I did now a rebase to main, and tested it further on the field.
I realised I was still missing some cases regarding the file extensions (I'm building with `GOOS=android`), so I hope you don't mind if I slightly refactored the logic...

Now when setting the buildmode we always fall back to `.so` and `.a` for shared and static libs respectively. With overrides for macOS and Windows. 

Additionally I found and fixed another bug with the override logic.

PR #3527, opened 2022-11-03
fix(changelog): group regexps

stevenh:
Fix the regular expressions used in changelog group processing to valid golang (RE2) regexps. These previously used PCRE character class `\w` which is not supported in RE2 which is what golang regexp uses.

Also document that format matches not just title as some may thing but the format `<abbrev-commit> <title-commit>`.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -53,16 +53,16 @@ changelog:
     - go mod tidy
   groups:
     - title: Dependency updates
-      regexp: '^[0-9a-h]+ (feat|fix)\(deps\)!?:.+$'
+      regexp: '^[[:xdigit:]]+ (feat|fix)\(deps\)!?:.+$'
       order: 300
     - title: 'New Features'
-      regexp: '^[0-9a-h]+ feat(\([[:word:]]+\))??!?:.+$'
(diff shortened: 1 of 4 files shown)


.goreleaser.yaml
```diff
@@ -53,16 +53,16 @@ changelog:
     - go mod tidy
   groups:
     - title: Dependency updates
-      regexp: "^.*(feat|fix)\\(deps\\)*:+.*$"
```
caarlos0, line 56: actually I don't think we really need the `^.*` prefix anyway

stevenh: Yer I thought that too, which is why I also added the comment to clarify that it's not just checking against the commit title but also the abbreviated commit reference too.

stevenh: If we did remove the `abbrev-commit` prefix then it might match the pattern anywhere in the title vs anchored to the stat, which is not the intention.

caarlos0: ahh, right! thanks

stevenh:
Had to roll back some changes which worked with the tests but didn't work in practice on github which has different output. I went with a conservative non greedy `.*?` as replacement for `.*` as the best we can do if the format varies per source control system.

caarlos0:
thank you!

PR #3528, opened 2022-11-04
docs: update docs to mention gitlab protected vars

marcjmiller:
add warning about gitlab-ci protected variables for both CI/gitlab and SCM/gitlab. 
Failing to run the pipelines on a protected ref will cause goreleaser to fail with `401: unauthorized` if for example, GITLAB_TOKEN is a protected variable and the CI Job is running on a tag that is not protected.

Add a warning about Gitlab CI and protected variables

To prevent future users from running into issues with protected variables.

[Discord Convo](https://discord.com/channels/890434333251362866/890434334803247126/1038130560293412955)


caarlos0:
awesome, thank you!

PR #3531, opened 2022-11-05
Update .goreleaser.yaml

frantjc:
URL in `.goreleaser.yaml` pointed to https://goreleser.com which doesn't exit, presumably it was intended to point to https://goreleaser.com


caarlos0:
oh, thank you so much! totally let that slip! :P

PR #3540, opened 2022-11-09
feat: add digest to artifacts info of published docker images

gal-legit:
Extract the digest (sha256) of docker images from the `docker push` command for dockers published to the docker registry.
Outputting the digest is required to avoid a race condition when referencing the image, where the image tag is being modified before the reference is done.
See this [blog post](https://github.com/goreleaser/goreleaser/issues/3496) for more info.
This PR fixes https://github.com/goreleaser/goreleaser/issues/3496.

Note that the 'publish' pipe now must run before the 'metadata' pipe, so that the information extracted during the 'publish' pipe would appear in the metadata.
Previously, the published docker images metadata wasn't printed (because of the order). It made sense because the content of the published image was just a subset of the local one.
Now that it is printed to the metadata, it should have a different name to avoid confusion.
As I mentioned, it wasn't printed before - so there shouldn't be any backward-compatibility issues.

---
Local tests:
```
go test -v .
=== RUN   TestVersion
=== RUN   TestVersion/only_version
=== RUN   TestVersion/version_and_date
=== RUN   TestVersion/version,_date,_built_by
=== RUN   TestVersion/all_empty
=== RUN   TestVersion/complete
--- PASS: TestVersion (0.00s)
    --- PASS: TestVersion/only_version (0.00s)
    --- PASS: TestVersion/version_and_date (0.00s)
    --- PASS: TestVersion/version,_date,_built_by (0.00s)
    --- PASS: TestVersion/all_empty (0.00s)
    --- PASS: TestVersion/complete (0.00s)
PASS
ok      github.com/goreleaser/goreleaser        0.764s
```

Output example:
```
  {
    "name": "gallegit/hello-world:latest",
    "path": "gallegit/hello-world:latest",
    "goos": "linux",
    "goarch": "amd64",
    "internal_type": 10,
    "type": "Published Docker Image",
    "extra": {
      "digest": "sha256:c3f7dd196a046dc061236d3c6ae1e2946269e90da30b0a959240ca799750e632"
    }
  }
```

PR #3548, opened 2022-11-13
feat: output checksums to artifacts info

gal-legit:
Following #3540, output artifacts' checksums to the artifact info.
This addition makes it easier to consume the checksums, especially when running from e.g. GitHub Actions.

New tests:
1. Add a check for the checksum in the extra field. 
2. Add a test for every checksum algorithm (to see that it doesn't break for any algo's output).
3. Add a case of a binary and an extra file (to see that the logic doesn't break when there's a mix).

p.s.
While working on that, I noticed that the convention for extra fields is actually to use UpperCamelCase rather than lower. 
I was mistaken because I looked at the subfields of the "DockerConfig" extra field.
I think it's a good idea to fix it quickly, before the next release rolls and it becomes a compatibility issue.
I took the liberty to fix it here as an extra commit. Please let me know if you want it to be in a separate PR.

---
Tests:
```
go test
  • refreshing checksums                             file=binary_bar_checksums.txt
  • refreshing checksums                             file=binary_bar_checksums.txt
  • refreshing checksums                             file=binary_bar_checksums.txt
PASS
ok  	github.com/goreleaser/goreleaser/internal/pipe/checksums	0.184s
```

### DIFF
--- a/internal/pipe/checksums/checksums.go
+++ b/internal/pipe/checksums/checksums.go
@@ -125,34 +125,6 @@ func refresh(ctx *context.Context, filepath string) error {
 		return err
 	}
 
-	artifactsMapping := make(map[string]int)
-	for i, a := range artifactList {
-		if a.Type == artifact.UploadableFile {
-			continue // skip extra files
-		}
-		// Path is the only unique field for an artifact
-		artifactsMapping[a.Path] = i
-	}
-
-	err = ctx.Artifacts.Visit(func(a *artifact.Artifact) error {
-		if index, ok := artifactsMapping[a.Path]; ok {
-			sumFields := strings.Fields(sumLines[index])
-			if len(sumFields) == 0 {
-				return fmt.Errorf("missing checksum in sumline of %s: %s", a.Path, sumFields)
-			}
-			sum := sumFields[0]
-			if a.Extra == nil {
-				a.Extra = make(artifact.Extras)
-			}
-			a.Extra[artifactChecksumExtra] = fmt.Sprintf("%s:%s", ctx.Config.Checksum.Algorithm, sum)
-		}
-
-		return nil
-	})
-	if err != nil {
-		return err
-	}
-
 	file, err := os.OpenFile(
 		filepath,
 		os.O_APPEND|os.O_WRONLY|os.O_CREATE|os.O_TRUNC,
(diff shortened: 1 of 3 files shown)


internal/pipe/checksums/checksums.go
```diff
@@ -121,6 +125,34 @@ func refresh(ctx *context.Context, filepath string) error {
...
+		if index, ok := artifactsMapping[a.Path]; ok {
+			sumFields := strings.Fields(sumLines[index])
+			if len(sumFields) == 0 {
+				return fmt.Errorf("missing checksum in sumline of %s: %s", a.Path, sumFields)
+			}
+			sum := sumFields[0]
+			if a.Extra == nil {
+				a.Extra = make(artifact.Extras)
+			}
+			a.Extra[artifactChecksumExtra] = fmt.Sprintf("%s:%s", ctx.Config.Checksum.Algorithm, sum)
```
caarlos0, line 147: can we put this in the `checksum` func instead? I think we then don't need most of the rest of the code and it should simplify things quite a bit...

gal-legit: That makes total sense. Not sure why I took the visit in the first place.
done (also applied fixes from the lint)

gal-legit:
p.s. do you have a plan for when the next release (tag) is gonna be? we actually want to use these new fields for https://github.com/Legit-Labs/legitify

caarlos0:
@gal-legit probably late this week/weekend

internal/pipe/checksums/checksums_test.go
```diff
@@ -304,47 +314,87 @@ func TestPipeCheckSumsWithExtraFiles(t *testing.T) {
...
 	}
+	checksumsMap := map[string]string{
+		"crc32":  "f94d3859",
+		"md5":    "5ac749fbeec93607fc28d666be85e73a",
+		"sha1":   "8b45e4bd1c6acb88bebf6407d16205f567e62a3e",
+		"sha224": "21bc225587d8768058837b68fe7e0341e87b972f02fd8fb0c236d1d3",
+		"sha256": "61d034473102d7dac305902770471fd50f4c5b26f6831a56dd90b5184b3c30fc",
+		"sha384": "f6055a96a105d2fb5941a616964ffda8294fd415730cc4154a602062bc3d00e99d3c6f4a11af8c965a343de4afca3c2b",
+		"sha512": "14925e01a7a0cf0801aa95fe52d542b578af58ae7997ada66db3a6eae68a329d50600a5b7b442eabf4ea77ea8ef5fe40acf2ab31d47311b2a232c4f64009aac1",
+	}
```
caarlos0, line 325: this is already tested in `artifacts_test.go`, I think we can revert to before.

gal-legit: I added these tests to make sure that the extraction from checksums.txt is correct,
i.e. in case it'll be e.g. `artifact-name  sha-val` instead of `sha-val artifact-name`,
and tbh I'm not sure it's covered in the other tests.

anyway, added a commit that does just that. I hope that I understood correctly and that it's the correct place to check it.

internal/pipe/checksums/checksums_test.go
```diff
@@ -344,6 +344,19 @@ func TestPipeCheckSumsWithExtraFiles(t *testing.T) {
 					require.Contains(t, string(bts), "3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855  "+want)
 				}
 			}
+
+			ctx.Artifacts.Visit(func(a *artifact.Artifact) error {
```
caarlos0, line 348:
```suggestion
			_ = ctx.Artifacts.Visit(func(a *artifact.Artifact) error {
```

caarlos0:
looks awesome, thank you!

PR #3549, opened 2022-11-13
Add Mercure to users

dunglas:
[Mercure](https://mercure.rocks), an open solution for real-time communications designed to be fast, reliable and battery-efficient is a proud user of GoReleaser.


caarlos0:
awesome, thank you!

PR #3553, opened 2022-11-14
docs: improve schema

caarlos0:
this should prevent yamlschema from complaining on some fields that accept an integer or string (goarm) and a bool or string (skip*)


pkg/config/config.go
```diff
@@ -300,7 +300,7 @@ type Build struct {
 	ID              string          `yaml:"id,omitempty" json:"id,omitempty"`
 	Goos            []string        `yaml:"goos,omitempty" json:"goos,omitempty"`
 	Goarch          []string        `yaml:"goarch,omitempty" json:"goarch,omitempty"`
-	Goarm           []string        `yaml:"goarm,omitempty" json:"goarm,omitempty"`
+	Goarm           []string        `yaml:"goarm,omitempty" json:"goarm,omitempty"nt`
```
caarlos0, line 303:
```suggestion
	Goarm           []string        `yaml:"goarm,omitempty" json:"goarm,omitempty"`
```

Issue #3525, opened 2022-11-02
Add option to push homebrew formula to arbitrary git repo


graytonio:
### Is your feature request related to a problem? Please describe.

We use an internal bitbucket server for hosting our source for tools as well as a private homebrew tap. Currently the workflow with gitreleaser requires us to manually move the generated formulas into the tap repository to update them.  It is a pain point in our workflow to have to move these formulas into place as an extra step to keep tools updated.

### Describe the solution you'd like

I would like the option to push the brew formula to any git repository not just a GitHub or GitLab one.  I have seen other publish targets like AUR have options for this with `git_url` and `private_key` to push artifacts.  I would imagine the config looking something like this.
```yaml
#.goreleaser.yaml
brews:
  - name: mytool
     tap:
       branch: main
       git_url: 'ssh://user@git.example.com/mytool.git'
       private_key: '{{ .Env.BREW_MYTOOL_SECRET }}'
```

So if a git_url is passed it uploads using those credentials.

### Describe alternatives you've considered

Currently the alternative is to have an extra step in the CI/CD to commit the artifact to the correct repo.  It creates another configuration to look at to resolve build issues which we are trying to reduce having more configurations in one place reduces points we have to check.

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

caarlos0:
seems valid, could possibly reuse some of the code from the aur pipe... 

if anyone wants to give it a shot, feel welcome to :)

caarlos0:
there's a PR for it already, if you want to help reviewing and testing it, that is also very much appreciated :) 

#3559

PR #3559, opened 2022-11-15
Allow brew repo to have arbitrary git target

graytonio:
Add ability for brew repos to have an arbitrary git repo as a publish target.

Adds configuration parameters in the RepoRef config for git urls
```
- name: "{{ .ProjectName }}"
    description: "My Tool Description"
    folder: Formulas
    tap:
      branch: main
      name: private-brew-tap
      git_url: "https://mygiturl.myurl.com/repo.git"
      private_key: "{{ .Env.SSH_KEY }}"
```

It does not effect the function of other pipes that use RepoRef without more work to explicitly use the added fields

To make it easier to push formulas for private git repositories

#3525
...


internal/client/config.go
```diff
@@ -27,10 +27,25 @@ func TemplateRef(apply func(s string) (string, error), ref config.RepoRef) (conf
 	if err != nil {
 		return ref, err
 	}
+	git_url, err := apply(ref.GitURL)
```
caarlos0, line 30: Go usually does not use `_` in variable names.

internal/git/git.go
```diff
@@ -51,6 +52,16 @@ func Run(ctx context.Context, args ...string) (string, error) {
 	return RunWithEnv(ctx, []string{}, args...)
 }
 
+func RunCmds(ctx context.Context, cwd string, env []string, cmds [][]string) error {
```
caarlos0, line 55: I don't think we need to move this here, it can probably live closer to where it's used.

graytonio:
Migrated the logic to its own client and have brew and aur pipes using it.

graytonio:
@caarlos0 whenever you get a chance can you look over the changes and provide some feedback

Rashid-rc:
> Add ability for brew repos to have an arbitrary git repo as a publish target.
> 
> Adds configuration parameters in the RepoRef config for git urls
> 
> ```
> - name: "{{ .ProjectName }}"
>     description: "My Tool Description"
>     folder: Formulas
>     tap:
>       branch: main
>       name: private-brew-tap
>       git_url: "https://mygiturl.myurl.com/repo.git"
>       private_key: "{{ .Env.SSH_KEY }}"
> ```
> 
> It does not effect the function of other pipes that use RepoRef without more work to explicitly use the added fields
> 
> To make it easier to push formulas for private git repositories
> 
> #3525 ...

internal/client/config.go
```diff
@@ -27,10 +30,25 @@ func TemplateRef(apply func(s string) (string, error), ref config.RepoRef) (conf
...
```
caarlos0, line 33: Go's convention is to name variables `likeThis` instead of `like_this`

internal/client/git.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	env := []string{fmt.Sprintf("GIT_SSH_COMMAND=%s", sshCmd)}
+
+	repoName := strings.TrimSuffix(url[strings.LastIndex(url, "/")+1:], ".git")
+
+	_, err = os.Stat(filepath.Join(ctx.Config.Dist, repoName))
+
+	// Only clone if this repo has not been cloned before
+	if os.IsNotExist(err) {
```
caarlos0, line 82: why do we need this?

internal/pipe/aur/aur.go
```diff
@@ -368,138 +365,25 @@ func doPublish(ctx *context.Context, pkgs []*artifact.Artifact) error {
...
-
-	path := key
-	if _, err := ssh.ParsePrivateKey([]byte(key)); err == nil {
-		// if it can be parsed as a valid private key, we write it to a
-		// temp file and use that path on GIT_SSH_COMMAND.
-		f, err := os.CreateTemp("", "id_*")
+		err = cl.CreateFile(ctx, cfg.CommitAuthor, repo, content, pkg.Name, cfg.CommitMessageTemplate)
 		if err != nil {
-			return "", fmt.Errorf("failed to store private key: %w", err)
+			return pipe.Skip(err.Error())
```
caarlos0, line 384: should be a hard err, not a skip

internal/pipe/brew/brew.go
```diff
@@ -22,7 +22,9 @@ import (
 	"github.com/goreleaser/goreleaser/pkg/context"
 )
 
-const brewConfigExtra = "BrewConfig"
+const (
+	brewConfigExtra = "BrewConfig"
+)
```
caarlos0, line 27:
```suggestion
const brewConfigExtra = "BrewConfig"
```

internal/pipe/brew/brew.go
```diff
@@ -142,6 +135,18 @@ func doPublish(ctx *context.Context, formula *artifact.Artifact, cl client.Clien
 		return err
 	}
 
+	cl, err = getClient(ctx, brew, cl, brew.Tap.Token)
+
```
caarlos0, line 139: dont need this empty line

internal/pipe/brew/brew.go
```diff
@@ -150,6 +155,14 @@ func doPublish(ctx *context.Context, formula *artifact.Artifact, cl client.Clien
 	return cl.CreateFile(ctx, author, repo, content, gpath, msg)
 }
 
+func getClient(ctx *context.Context, brew config.Homebrew, cl client.Client, token string) (client.Client, error) {
+	cl, err := client.NewIfToken(ctx, cl, brew.Tap.Token)
+	if len(brew.Tap.GitURL) > 0 && ctx.TokenType != context.TokenTypeGitLab {
+		cl, err = client.NewGitUploadClient(ctx)
+	}
+	return cl, err
```
caarlos0, line 163:
```suggestion
	if len(brew.Tap.GitURL) > 0 && ctx.TokenType != context.TokenTypeGitLab {
		return client.NewGitUploadClient(ctx)
	}
	return client.NewIfToken(ctx, cl, brew.Tap.Token)
```

internal/pipe/brew/brew.go
```diff
@@ -150,6 +155,14 @@ func doPublish(ctx *context.Context, formula *artifact.Artifact, cl client.Clien
 	return cl.CreateFile(ctx, author, repo, content, gpath, msg)
 }
 
+func getClient(ctx *context.Context, brew config.Homebrew, cl client.Client, token string) (client.Client, error) {
+	cl, err := client.NewIfToken(ctx, cl, brew.Tap.Token)
+	if len(brew.Tap.GitURL) > 0 && ctx.TokenType != context.TokenTypeGitLab {
```
caarlos0, line 160: why != gitlab? 🤔

internal/client/git.go
graytonio: You cannot git clone into a non-empty folder

internal/pipe/brew/brew.go
graytonio: A temporary bandaid I forgot to commit the real fix for.  Because the raw git client does not use tokens the current `NewIfToken` function would either give it a github client which would pass the test but fail in real world or if I didn't put the second expression in the if statement give it the gitclient during testing which would cause the test to fail. I have a better solution but forgot to commit it.

graytonio:
@caarlos0 went through and fixed up the mistakes you found should be good to go now.

scr-oath:
Hey… I'm looking forward to this but… I'm just curious if it works naturally with ssh-agent, or whether you *MUST* pass the PrivateKey… it would be really beneficial to be able to just git push in the end without passing the credential to git CLI and let the agent send the credentials.

caarlos0:
> Hey… I'm looking forward to this but… I'm just curious if it works naturally with ssh-agent, or whether you _MUST_ pass the PrivateKey… it would be really beneficial to be able to just git push in the end without passing the credential to git CLI and let the agent send the credentials.

I **think** if you don't pass a key, it'll use the SSH agent (or whatever you have in your `~/.ssh/config`), yes

caarlos0:
just to not leave you hanging here, I haven't made the time to properly review and test this yet, and its EOM so I'll release v1.15 without it.

Will make an extra effort to review it first for v1.16 - I promise :)

Djiit:
Hey @caarlos0 , thanks for the hard work. Any chance we could see this for 1.18 ?

internal/client/git.go
```diff
@@ -0,0 +1,190 @@
...
+	return ErrNotImplemented
+}
+
+func (gc *gitClient) GetDefaultBranch(ctx *context.Context, repo Repo) (string, error) {
+	return "", ErrNotImplemented
+}
+
+func (gc *gitClient) Changelog(ctx *context.Context, repo Repo, prev, current string) (string, error) {
+	return "", ErrNotImplemented
+}
```
caarlos0, line 190: TODO for me: segregate this interface, its too big and not all methods are needed everywhere its used...

caarlos0:
conflicts got out of hand, will probably re-create this and add you as co-author @graytonio, is that ok?

graytonio:
Sure I'm fine with that. Excited to see this get merged. 👍

caarlos0:
See #3961

PR #3591, opened 2022-11-24
fix: validate env templates

caarlos0:
ignore invalid environment variables (e.g. no key, no value, or no equal sign)


caarlos0:
not really sure this is needed btw

Issue #3582, opened 2022-11-22
support creating a reproducible tar archive


fho:
### Is your feature request related to a problem? Please describe.

I would like to create reproducible .tar archive for releases.
When running goreleaser with the same go version, on the same go source files on different systems it should produce the same archive.

### Describe the solution you'd like

Having a boolean `reproducible` option in the `archive` configuration file section.

By default i would be enabled and create a tar archive using the options described at:
https://reproducible-builds.org/docs/archives/

```sh
$ tar --sort=name \
      --mtime="@${SOURCE_DATE_EPOCH}" \
      --owner=0 --group=0 --numeric-owner \
      --pax-option=exthdr.name=%d/PaxHeaders/%f,delete=atime,delete=ctime \
      -cf product.tar build
```

### Describe alternatives you've considered

I tried to set owner and file mtime via:
```yml
info:
  owner: 0
  group: 0
  mode: 0755
  mtime: 2008-01-02T15:04:05Z
```

to create archives with a single binary.
I was not able to find a way to set these options for the binary created by `go build`. It seems that it is only possible use `info:` for other files then the compiled go binary. 

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

caarlos0:
have you tried following https://goreleaser.com/customization/build/?h=reprodu#reproducible-builds ?

fho:
@caarlos0 yes, that works  fine for building the binary. 
The issue is about making a created tar archive also reproducible

caarlos0:
if you only have the binary in the archive, and the binary is reproducible, it should be reproducible too... did you find something different?

also, the options you linked can't be used because goreleaser does not use `tar`.

fho:
> if you only have the binary in the archive, and the binary is reproducible, it should be reproducible too... did you find something different?

I had a deeper look, I think it's only missing to be able to enforce the uid/gid for the main binary in the tar archive.

I can't find a way to specify it, for additional files it can be set in `files.src.info:`.

This does not work:

```yml
project_name: "example"
builds:
  - goos:
      - linux
    goarch:
      - amd64
archives:
  - format: tar.xz
    files:
      - none*
      - src: 'example'
        strip_parent: true
        info:
          owner: 0
```

```sh
$ tree
.
├── go.mod
└── main.go

0 directories, 2 files
$ cat main.go
package main

import "fmt"

func main() {
[...]

PR #3618, opened 2022-12-03
feat: set file info for binaries inside archives

caarlos0:
Signed-off-by: Carlos A Becker <caarlos0@users.noreply.github.com>


ranoch (commented):
#3582

PR #3623, opened 2022-12-05
feat: update lock task parameter due to braking change

dirien:
Update lock task due to breaking changes 

https://github.com/dessant/lock-threads/blob/master/CHANGELOG.md#changelog

Updated the properties

...


caarlos0:
danke <3

PR #3628, opened 2022-12-08
doc: fix `skip_push` parameter to `skip_build`

j178:
Fix the wrong `skip_push` to `skip_build` in `docker.md`.


caarlos0:
ah, good catch, thanks

PR #3490, opened 2022-10-22
feat: build images with ko

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>
Co-authored-by: Furkan Türkal <furkan.turkal@trendyol.com>

/cc @Dentrax 

I have tested it with the following _.goreleaser.yaml_, and it worked as we expected:

```yaml
ko:
  repository: ghcr.io/developer-guy/goreleaser
  tags:
    - latest
  ldflags:
    - -X main.commitSHA={{.FullCommit}}
  platforms:
    - linux/amd64
  bare: true
```

and the output was like the following:

```shell
  • running ko
2022/10/23 01:24:03 Building github.com/goreleaser/goreleaser for linux/amd64
2022/10/23 01:24:20 Publishing ghcr.io/developer-guy/goreleaser:latest
2022/10/23 01:24:23 Published SBOM ghcr.io/developer-guy/goreleaser:sha256-8f57837403a9043c0e4e52b86be5a9da6d840d7ad7fc8df5ab2e23929812cd0d.sbom
2022/10/23 01:24:43 Published ghcr.io/developer-guy/goreleaser@sha256:8f57837403a9043c0e4e52b86be5a9da6d840d7ad7fc8df5ab2e23929812cd0d
ghcr.io/developer-guy/goreleaser@sha256:8f57837403a9043c0e4e52b86be5a9da6d840d7ad7fc8df5ab2e23929812cd0d

```

/cc @imjasonh @caarlos0  

closes #2556

PR #3653, opened 2022-12-21
feat: add ko support

### LINKED ISSUE #2556: [Feature Request]: add google/ko as brand new builder engine for building and pushing container images

caarlos0:
continuing the PR by @developer-guy 

- [x] should be a publisher, as it does publish the images it builds every time
- [x] `Default` method does not work
- [x] the `fromConfig` thing should probably be on the defaults, too
- [x] wire `--skip-ko`
- [x] documentation
- [x] more tests
- [x] use same registry as docker tests does
- [ ] see if we can make the log output match goreleaser's
- [ ] ??

closes #2556
closes #3490


developer-guy:
I'm going to take a look at it, looks amazing in advance 👋

caarlos0:
hey @developer-guy, can you take a look? especially the docs (https://feature-2556.goreleaser.pages.dev/customization/ko/)

www/docs/customization/ko.md
```diff
@@ -0,0 +1,106 @@
...
+options as the [build][] pipe when possible, so the results will probably be cached.
+
+!!! warning
+    Ko only runs on the publish phase, so it might be a bit hard to test — you
+    might need to push to a fake repository (or a fake tag) when playing around
+    with its configuration.
+
+```yaml
+# .goreleaser.yaml
+kos:
```
imjasonh, line 18: Can this be `ko:` instead?

www/docs/customization/ko.md
```diff
@@ -0,0 +1,106 @@
...
+
+  # Whether to use the base path without the MD5 hash after the repository name.
+  #
+  # Defaults to false.
+  base_import_paths: true
+```
+
+Refer to the [Ko Build][ko] project page for more information.
+
+[ko]: https://github.com/ko-build/ko
```
imjasonh, line 106:
```suggestion
[ko]: https://ko.build
```

pkg/config/config.go
```diff
@@ -916,6 +935,7 @@ type Project struct {
 	Brews           []Homebrew       `yaml:"brews,omitempty" json:"brews,omitempty"`
 	AURs            []AUR            `yaml:"aurs,omitempty" json:"aurs,omitempty"`
 	Krews           []Krew           `yaml:"krews,omitempty" json:"krews,omitempty"`
+	Kos             []Ko             `yaml:"kos,omitempty" json:"kos,omitempty"`
```
imjasonh, line 938:
```suggestion
	Ko             []Ko             `yaml:"ko,omitempty" json:"ko,omitempty"`
```

internal/pipe/ko/ko.go
```diff
@@ -0,0 +1,335 @@
...
+		if ko.Main == "" {
+			ko.Main = build.Main
+		}
+
+		if ko.WorkingDir == "" {
+			ko.WorkingDir = build.Dir
+		}
+
+		if ko.BaseImage == "" {
+			ko.BaseImage = chainguardStatic
```
imjasonh, line 92: We should try to reuse whatever const ko defines upstream, in case we ever change the default base image in a future ko release.

Luckily @developer-guy has already started toward that in https://github.com/ko-build/ko/pull/904, which we can reuse here when `ko.BaseImage` is not defined by the user.

internal/pipe/ko/ko.go
```diff
@@ -0,0 +1,335 @@
...
+			ko.Platforms = []string{"linux/amd64"}
+		}
+
+		if len(ko.Tags) == 0 {
+			ko.Tags = []string{"latest"}
+		}
+
+		if ko.SBOM == "" {
+			ko.SBOM = "spdx"
+		}
```
imjasonh, line 109: I think these should also be defaults that ko maintains upstream. If we ever change them in the future we don't want to have to go hunting for other places these have been set downstream.

imjasonh (commented):
This looks great! Thanks for keeping this moving. Just a few smallish comments.

www/docs/customization/ko.md
caarlos0: it's pluralized because it can be multiple KO configurations... is the pluralization a problem? If it is, I can change...

internal/pipe/ko/ko.go
caarlos0: cool, I can PR those there too

www/docs/customization/ko.md
developer-guy: We could build an OCI image with ko for different binaries built from the same repository, which is why I designed it that way.

internal/pipe/ko/ko.go
caarlos0: not merged/released yet, right? we can change this in the future then

caarlos0:
`TestPublishPipeError/publish_fail` started failing and I have no idea why... @imjasonh @developer-guy any guesses? It should fail to publish to a fake registry, but always succeeds, you can even see the log from KO:

CleanShot 2023-01-16 at 00 53 02@2x

I don't understand why it could succeed...

internal/pipe/ko/ko.go
developer-guy: yep, it is not released yet, it is waiting for merge right now

developer-guy:
> `TestPublishPipeError/publish_fail` started failing and I have no idea why... @imjasonh @developer-guy any guesses? It should fail to publish to a fake registry, but always succeeds, you can even see the log from KO:
> 
> CleanShot 2023-01-16 at 00 53 02@2x
> 
> I don't understand why it could succeed...

I found the problem and fixed it with cd78015785c1083af0fa890389ff59788e56c40e

developer-guy:
I've open a PR to fix this problem in ko project side: https://github.com/ko-build/ko/pull/932

caarlos0:
I did a few changes in your commit too, let me know what you think @developer-guy

caarlos0:
update docs, fixed the main (was being ignored :kek:), I think we are good to merge now!

PR #3656, opened 2022-12-22
feat: better archives relative paths

### LINKED ISSUE #3655: different behavior for extra file trees included in `archives` and `nfpms`.
### What happened?

1. `archives` config works differently from `npmfs` when including subtrees.
2. there's no way to replicate the same behavior.

This was also discussed on discord: https://discord.com/channels/890434333251362866/1035159881210146916/threads/1054929081990381588

Let's say we have a file structure:
```
build/webui/react/robots.txt
build/webui/docs/index.html
build/webui/docs/deploy/index.html
build/webui/docs/deploy/deploy.html
...
```

Case 1: I'd like to include it as is, stripping the `build/` prefix, i.e. include:
```
webui/react/robots.txt
webui/docs/index.html
webui/docs/deploy/index.html
webui/docs/deploy/deploy.html
...
```

With npmfs I can do:
```
npmfs:
    contents:
      - src: "build/**/*"
```

When I try the same with `archives`:
- `- "build/webui/**/*"` will output `build/webui/react/robots.txt`
- adding `- dst: webui` will output `webui/build/webui/react/robots.txt`
- using `strip_parent` option with dst will output flattened `webui/robots.txt`, `webui/index.html`, `webui/deploy.html` (and incidentally cause a filename collision with `index.html` in this example)

i.e. `archives` behaves differently, and there's no way to do what npmfs can do.

Case 2: If I want to strip `build/` prefix and include these files under a different path `/usr/share/derp`, e.g.

```
/usr/share/derp/webui/react/robots.txt
/usr/share/derp/webui/docs/index.html
/usr/share/derp/webui/docs/deploy/index.html
...
```

with npmfs I can do:

```
npmfs:
    contents:
      - src: "build/**/*"
         dst: "/usr/share/derp"
```

Similarly to case 1, I could not find any way to do that.

### goreleaser version

```bash
Tried this both on 1.1.0 and 1.13.1
```

### GoReleaser Check

### Search

### Code of Conduct

### Additional context

_No response_

caarlos0:
with this patch, a config like:

```yaml
archives:
  - format: tar.gz
    # this name template makes the OS and Arch compatible with the results of uname.
    name_template: >-
      {{ .ProjectName }}_
      {{- title .Os }}_
      {{- if eq .Arch "amd64" }}x86_64
      {{- else if eq .Arch "386" }}i386
      {{- else }}{{ .Arch }}{{ end }}
      {{- if .Arm }}v{{ .Arm }}{{ end }}
    rlcp: true
    files:
      - src: "build/**/*"
        dst: .

nfpms:
  - package_name: foo
    contents:
      - src: "build/**/*"
        dst: usr/share/foo
    formats:
      - apk

```

will eval this:

CleanShot 2022-12-21 at 22 21 00@2x

as much as I would like to make this the default, it would be a breaking change, so we really can't do it.

If `dst` is empty, it'll have the same behavior as before (no rlcp), and if `strip_parent` is set, it will also still have the same behavior. Finally, if the format is binary, `rlcp` is ignored too (as it doesn't make sense).

So, this only changes if:
- your format is not binary; and
- you have files with `src` and `dst` set

Then, goreleaser will warn you to set `rlcp: true`.

## todo

- [x] docs
- [x] more tests probably
- [x] any ideas for a better name for the new config option?

fixes #3655

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -142,6 +142,7 @@ archives:
     builds_info:
       group: root
       owner: root
+    rlcp: true
     files:
       - README.md
       - LICENSE.md
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -18,8 +18,8 @@ builds:
     - CGO_ENABLED=0
   goos:
     - linux
-    - darwin
-    - windows
+    # - darwin
+    # - windows
   goarch:
     - "386"
     - amd64
@@ -256,17 +256,17 @@ nfpms:
         - statically-linked-binary
         - changelog-file-missing-in-native-package
 
-snapcrafts:
-  - name_template: '{{ .ProjectName }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}'
-    summary: Deliver Go binaries as fast and easily as possible
-    description: |
-      GoReleaser builds Go binaries for several platforms, creates a GitHub
-      release and then pushes a Homebrew formula to a repository. All that
-      wrapped in your favorite CI.
-    grade: stable
-    confinement: classic
-    publish: true
-
+# snapcrafts:
+#   - name_template: '{{ .ProjectName }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}'
+#     summary: Deliver Go binaries as fast and easily as possible
+#     description: |
+#       GoReleaser builds Go binaries for several platforms, creates a GitHub
+#       release and then pushes a Homebrew formula to a repository. All that
+#       wrapped in your favorite CI.
+#     grade: stable
+#     confinement: classic
+#     publish: true
+#
 sboms:
   - artifacts: archive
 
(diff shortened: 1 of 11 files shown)


caarlos0:
One idea:

- make it warns that `relative: true` will be default soon (maybe specify a version, e.g. v1.21.0)
- wait 6mo (or until the specified version)
- make it the default
- deprecation warning for 6 months so people can remove the option from their configs
- remove it completely

would take 1 year to phase it all, but IMHO is the safest way to do so, if we decide to go that way.

Thoughs?

ioga:
> One idea:
> 
>     * make it warns that `relative: true` will be default soon (maybe specify a version, e.g. v1.21.0)
> 
>     * wait 6mo (or until the specified version)
> 
>     * make it the default
> 
>     * deprecation warning for 6 months so people can remove the option from their configs
> 
>     * remove it completely
> 
> 
> would take 1 year to phase it all, but IMHO is the safest way to do so, if we decide to go that way.
> 
> Thoughts?

since `relative: true` will be a new default global behavior, I'd suggest making it an option on `archives` (or at the top level), instead of having it to be an option on each entry under `archives -> files`, because you probably don't want to let people mix old and new behavior in their configurations, as it'll make the migration harder.

caarlos0:
yes, that makes sense @ioga.

Will do it! Thank you!

caarlos0:
fwiw, renamed to `rlcp` (relative longest common path) so its less confusing, and purposefully require a little bit of reading.

Issue #3669, opened 2022-12-29
Nil pointer dereference


ddddddO:
### What happened?

I tried to release using goreleaser/goreleaser-action@v4 on GitHub Actions and got a panic.
See below for specific logs.

```console
Run goreleaser/goreleaser-action@v4
Downloading https://github.com/goreleaser/goreleaser/releases/download/v1.14.0/goreleaser_Linux_x86_64.tar.gz
Extracting GoReleaser
/usr/bin/tar xz --warning=no-unknown-keyword --overwrite -C /home/runner/work/_temp/038[59](https://github.com/ddddddO/kaisekisan/actions/runs/3799388086/jobs/6461904856#step:4:60)883-20b0-4b95-8bdc-4744d1de4185 -f /home/runner/work/_temp/b91883f7-ed78-43bb-80e8-d1dfd0d25715
GoReleaser latest installed successfully
/opt/hostedtoolcache/goreleaser-action/1.14.0/x64/goreleaser release --rm-dist
  • starting release...
  • loading config file                              file=.goreleaser.yml
  • loading environment variables
  • getting and validating git state
    • couldn't find any tags before "v0.0.2"
    • building...                                    commit=8872978d902369b7ac10b08fe3acb0934133[60](https://github.com/ddddddO/kaisekisan/actions/runs/3799388086/jobs/6461904856#step:4:61)a9 latest tag=v0.0.2
    • running against a shallow clone - check your CI documentation at https://goreleaser.com/ci
  • parsing tag
  • setting defaults
      • DEPRECATED: `archives.rlcp` will be the default soon, check https://goreleaser.com/deprecations#archivesrlcp for more info
      • DEPRECATED: `archives.replacements` should not be used anymore, check https://goreleaser.com/deprecations#archivesreplacements for more info
  • running before hooks
    • running                                        hook=go mod tidy
    • took: 2s
  • checking distribution directory
  • loading go mod information
  • build prerequisites
  • writing effective config file
    • writing                                        config=dist/config.yaml
  • building binaries
    • building                                       binary=dist/kaisekisan_linux_arm64/kaisekisan
    • building                                       binary=dist/kaisekisan_linux_amd64_v1/kaisekisan
    • building                                       binary=dist/kaisekisan_linux_386/kaisekisan
    • building                                       binary=dist/kaisekisan_darwin_amd64_v1/kaisekisan
    • building                                       binary=dist/kaisekisan_darwin_arm64/kaisekisan
[...]

PR #3670, opened 2022-12-29
fix: Check that the variable resp is not nil

ddddddO:
# If applied, this commit will

Linked issues will be resolved.

# Why is this change being made?

Because the program will no longer have to refer to nil.

# Provide links to any relevant tickets, URLs or other resources
- https://github.com/goreleaser/goreleaser/issues/3669


caarlos0:
good catch, thanks!

ddddddO:
thank you!👍

PR #3680, opened 2023-01-04
docs: rename `.goreleaser.yml` to `.goreleaser.yaml` in intro

craigpastro:
Every other instance I have seen so far uses `.yaml` so, for consistency, let's rename this one?


caarlos0:
good catch, thanks!

PR #3685, opened 2023-01-07
WIP feat: notify goreleaser-cross about new release

troian:
don't merge yet, doing some local test to ensure event works

Signed-off-by: Artur Troian <troian.ap@gmail.com>


.github/workflows/release.yml
```diff
@@ -115,3 +115,14 @@ jobs:
           DISCORD_WEBHOOK_TOKEN: ${{ secrets.DISCORD_WEBHOOK_TOKEN }}
           AUR_KEY: ${{ secrets.AUR_KEY }}
         run: task goreleaser
+      - name: get version
+        if: success() && startsWith(github.ref, 'refs/tags/v')
+        run: echo "RELEASE_TAG=${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
+      - name: notify goreleaser-cross with new release
+        if: success() && startsWith(github.ref, 'refs/tags/v')
+        uses: peter-evans/repository-dispatch@v2
```
caarlos0, line 123: maybe put in another job, like `trigger-generate` on line 16?

also, we could use the same action for both triggers, the other one is using `benc-uk/workflow-dispatch`... I don't have strong feelings about any of them, just think it reduces our burden to depend on only one of them

troian: makes sense, need to test how `benc-uk/workflow-dispatch` work with `gorealeser-cross` tho

troian: yeah, it works, i forced pushed changes below

caarlos0:
awesome, thank you!

PR #3758, opened 2023-02-05
feat: finally deprecate the single build option

caarlos0:
This option was still being supported, even though undocumented, for many years now. 

I think it's finally time to sunset it for good, in 6 months :)


www/docs/deprecations.md
```diff
@@ -37,6 +37,30 @@ Description.
 -->
 
 
+### build
+
+> since 2023-02-05
```
caarlos0, line 42:
```suggestion
> since 2023-02-09 (v1.16.0)
```

PR #3765, opened 2023-02-07
fix(gitlab): correctly prepend/append/keep releases notes

christophwitzko:
Hi, I found a bug in the GitLab client that leads to not correctly prepend/append/keep releases notes.

This will use the original `Description`  instead of the pre-rendered `DescriptionHTML`. Furthermore, as `include_html_description` is not enabled, the `DescriptionHTML` field is always empty.

[GitLab documentation](https://docs.gitlab.com/ee/api/releases/index.html#get-a-release-by-a-tag-name)

### DIFF
--- a/internal/client/gitlab.go
+++ b/internal/client/gitlab.go
@@ -319,7 +319,7 @@ func (c *gitlabClient) CreateRelease(ctx *context.Context, body string) (release
 	} else {
 		desc := body
 		if release != nil {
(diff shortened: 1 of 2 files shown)


caarlos0:
> Hi, I found a bug in the GitHub client that leads to not correctly prepend/append/keep releases notes.

I guess typo github -> gitlab

---

Ahh good catch, thanks for the PR!

christophwitzko:
@caarlos0 ahhh, yes. GitLab, of course. ☺️

Issue #3069, opened 2022-04-26
Customize Content-Disposition attribute for glob extra files


mbamber:
### Is your feature request related to a problem? Please describe.

As part of my release I would like to upload some simple files that should be viewed inline rather than be downloaded by the browser.

### Describe the solution you'd like

I'd like to be able to control if any extra files are for download or for inline-viewing, something like the below as a contrived example:
```yaml
extra_files:
  - glob: myfile.html
    isAttachment: false
```

### Describe alternatives you've considered

The only solution I can see right now is either uploading files and then modifying them after upload as part of a hook, or not upload these files with goreleaser (which seems like a shame)

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Additional context

Line of interest:
https://github.com/goreleaser/goreleaser/blob/69cf1aa88754454d074626d49969d0751f2afa39/internal/pipe/blob/upload.go#L229

caarlos0:
I would say to allow to set the whole content disposition instead, e.g.

```yaml
extra_files:
  - glob: myfile.html
    content_disposition: attachment
```

is it possible to set the content disposition to something that would allow it to be viewed in the browser or should it be empty instead?

another (similar) idea: set the content disposition only to files generated by goreleaser (archives, packages, etc) and do not set it for other files

mbamber:
Yep that works too!

> is it possible to set the content disposition to something that would allow it to be viewed in the browser or should it be empty instead?

Yes I believe it can be set to `inline`. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#as_a_response_header_for_the_main_body

> another (similar) idea: set the content disposition only to files generated by goreleaser (archives, packages, etc) and do not set it for other files

Personally I don't like this idea; it feels more restrictive for no real benefit. In my case, I am actually generating the file to upload as part of a post hook, so I definitely wouldn't want this to be affected.

caarlos0:
this will be a rather big refactory, might get only on 1.10+

till:
Is the idea to support _arbitrary_ writer opts? Or specific features?

caarlos0:
initally, arbitrary content-types.

Do you think it should support arbitrarty opts? If so, can you share your usecase?

till:
> initally, arbitrary content-types.
> 
> 
> 
> Do you think it should support arbitrarty opts? If so, can you share your usecase?

[...]

PR #3772, opened 2023-02-10
feat: extend s3 compatible config

warjiang:
1.  if user set the s3 endpoint,`goreleaser` will set `s3ForcePathStyle=true` in the query :
image
this pr want to allow user make choice whether to set `s3ForcePathStyle=true`

2.  allow user set s3 upload policy, wants goreleaser support s3 acl policy 

In short, user will use goreleaser  s3 provider like this:
image

### DIFF
--- a/internal/pipe/blob/upload.go
+++ b/internal/pipe/blob/upload.go
@@ -1,12 +1,16 @@
 package blob
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"net/url"
 	"os"
 	"path"
 	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 
 	"github.com/caarlos0/log"
 	"github.com/goreleaser/goreleaser/internal/artifact"
@@ -53,7 +57,11 @@ func urlFor(ctx *context.Context, conf config.Blob) (string, error) {
 	}
 	if endpoint != "" {
 		query.Add("endpoint", endpoint)
-		query.Add("s3ForcePathStyle", "true")
+		if conf.S3ForcePathStyle == nil {
+			query.Add("s3ForcePathStyle", "true")
+		} else {
+			query.Add("s3ForcePathStyle", fmt.Sprintf("%t", *conf.S3ForcePathStyle))
+		}
 	}
 
 	region, err := tmpl.New(ctx).Apply(conf.Region)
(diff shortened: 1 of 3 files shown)


caarlos0:
those options would apply to all the files... I think, specially the ACLs, people will want to customize per file


check #3069

warjiang:
> those options would apply to all the files... I think, specially the ACLs, people will want to customize per file
> 
> check #3069

As you mentioned, the ACL policy will apply to all ids in the config file。But there is a trick way can specify differen acl for artifact.  If the original config like this:
```
blobs:
  - provider: s3
    endpoint: {{s3-endpoint}}
    region: {{s3-region}}
    disableSSL: true
    bucket:{{s3-bucket}}
    s3ForcePathStyle: false
    ids:
      - a
      - b
    folder: {{folder}}
```

if we want `a` be `public-read` acl, and `b` be `private` acl, we can do it like this:
```
blobs:
 - provider: s3
    endpoint: {{s3-endpoint}}
    region: {{s3-region}}
    disableSSL: true
    bucket:{{s3-bucket}}
    s3ForcePathStyle: false
    acl: public-read
    ids:
      - a
    folder: {{folder}
 - provider: s3
    endpoint: {{s3-endpoint}}
    region: {{s3-region}}
    disableSSL: true
    bucket:{{s3-bucket}}
    s3ForcePathStyle: false
    acl: public-read
    ids:
      - b
    folder: {{folder}}
```

warjiang:
By the way `s3ForcePathStyle` cannot specify for each file,`s3ForcePathStyle` should come with s3 endpoint,  more info can visit https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html

image

internal/pipe/blob/upload.go
```diff
@@ -230,15 +239,27 @@ func (u *productionUploader) Open(ctx *context.Context, bucket string) error {
...
 
 func (u *productionUploader) Upload(ctx *context.Context, filepath string, data []byte) error {
 	log.WithField("path", filepath).Info("uploading")
-
 	opts := &blob.WriterOptions{
 		ContentDisposition: "attachment; filename=" + path.Base(filepath),
 	}
+	if u.conf.ACL != "" {
+		opts.BeforeWrite = func(asFunc func(interface{}) bool) error {
+			req := &s3manager.UploadInput{}
```
caarlos0, line 253: what if the bucket is not on s3? wouldn't this fail?

PR #3788, opened 2023-02-20
fix: better handle password protected keys

caarlos0:
password-protected keys would give path errors, which was simply not right.


abhimanyu003:
This was super quick, thanks for this. I think we can also add this line in our docs as a note.

Issue #3790, opened 2023-02-21
Docker extra_files fails if path begins with ctx.Config.Dist 


FGummesson:
### What happened?

Adding docker.extra_files in directory `dist_files/*` fails with  `invalid docker.files: can't be . or inside dist folder: dist_files`. 

The problem occurs because the docker.extra_files only checks strings.HasPrefix() between `dist` folder and `files_extra` folder. 

### How can we reproduce this?

Create a config file and build a release. The build task will fail.
```yaml
# .goreleaser.yaml

dist: ./dist
dockers:     
    extra_files:
       - dist_files    #Fails
       - dist_files/   #Fails
       - ./dist_files/   #Works
```

### goreleaser version

```bash
$> goreleaser --version
goreleaser version 1.15.1
commit: 7c6bd86b286ee76cfb0a0b4ec17d18d4ee9daeec
built at: 2023-02-02T02:25:36Z
built by: goreleaser
goos: darwin
goarch: amd64
module version: v1.15.1, checksum: h1:GWI+6mpdyEsGU0rEtKvTA7/10L9JZhkHPh9wqXg7kSk=

https://goreleaser.com
```

### GoReleaser Check

- [X] goreleaser check shows no errors

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #3791, opened 2023-02-21
fix: docker extra_files fails if path starts with dist path

FGummesson:
Fix the handling of docker.extra_files to allow for folders that stars with the same pattern as `dist` folder, but isn't the a subfolder of `dist`
...

...

Issue reference: #3790
...

### DIFF
--- a/internal/pipe/docker/docker.go
+++ b/internal/pipe/docker/docker.go
@@ -61,8 +61,10 @@ func (Pipe) Default(ctx *context.Context) error {
 		if err := validateImager(docker.Use); err != nil {
 			return err
 		}
+		distPath := filepath.Clean(ctx.Config.Dist) + "/"
 		for _, f := range docker.Files {
-			if f == "." || strings.HasPrefix(f, ctx.Config.Dist) {
+			file := filepath.Clean(f) + "/"
+			if file == "./" || strings.HasPrefix(file, distPath) {
 				return fmt.Errorf("invalid docker.files: can't be . or inside dist folder: %s", f)
 			}
 		}
--- a/internal/pipe/docker/docker_test.go
+++ b/internal/pipe/docker/docker_test.go
@@ -1246,32 +1246,40 @@ func TestDefaultNoDockers(t *testing.T) {
 	require.Empty(t, ctx.Config.Dockers)
 }
 
-func TestDefaultFilesDot(t *testing.T) {
-	ctx := &context.Context{
-		Config: config.Project{
-			Dist: "/tmp/distt",
-			Dockers: []config.Docker{
-				{
-					Files: []string{"./lala", "./lolsob", "."},
+func TestForbiddenFilesDestinations(t *testing.T) {
+	testcases := []string{
+		".", "./", "./tmp/dist", "./tmp/dist/", "tmp/dist/", "tmp/dist/asdasd/asd",
+	}
+
+	for _, tc := range testcases {
+		ctx := &context.Context{
+			Config: config.Project{
+				Dist: "./tmp/dist",
+				Dockers: []config.Docker{
+					{Files: []string{tc}},
 				},
 			},
-		},
+		}
+		require.Errorf(t, Pipe{}.Default(ctx), `invalid docker.files: can't be . or inside dist folder: %s`, tc)
 	}
-	require.EqualError(t, Pipe{}.Default(ctx), `invalid docker.files: can't be . or inside dist folder: .`)
 }
 
-func TestDefaultFilesDis(t *testing.T) {
-	ctx := &context.Context{
-		Config: config.Project{
-			Dist: "/tmp/dist",
-			Dockers: []config.Docker{
-				{
-					Files: []string{"./fooo", "/tmp/dist/asdasd/asd", "./bar"},
+func TestAllowedFilesDestinations(t *testing.T) {
+	testcases := []string{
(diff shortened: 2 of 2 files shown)


internal/pipe/docker/docker.go
```diff
@@ -61,8 +61,10 @@ func (Pipe) Default(ctx *context.Context) error {
 		if err := validateImager(docker.Use); err != nil {
 			return err
 		}
+		distPath := filepath.Clean(ctx.Config.Dist) + "/"
```
caarlos0, line 64: I think we should probably use os.Separator instead of `/` 🤔

caarlos0: another idea is to try to use `filepath.Rel` to check if one path is within dist or not 🤔

FGummesson: I started to use the filepath.Rel() function. I couldn't figure out how to write the if statement if using it. 

```
filepath.Rel(ctx.Config.Dist, ".") == ../../.
filepath.Rel(ctx.Config.Dist, "./tmp/dist") == "."
filepath.Rel(ctx.Config.Dist, "./tmp/dist/something") == "something"
filepath.Rel(ctx.Config.Dist, "./tmp/dist_files/something") == ../dist_files/something
```

caarlos0: hmmm true, I guess your current solution is better...

do you think we should use os.separator instead of forward slash?

FGummesson: Yes, I think so. 

Should it be possible to use like ".." as extras_file path?

caarlos0: tbh I don't quite remember why we have this check... it might be some old thing, as once upon a time we symlinked the files, and I think this was to avoid circular linking... maaaaybe this check can be removed... 

do you want to test it out?

like, building a sample project adding files that would be refused in the current impl, and seeing what happens... I'm 95% sure it will work now...

FGummesson: I just tested it. I removed the check and added this in my .goreleaser.yaml

```
#.goreleaser.yaml
dockers:
  - image_templates:
    - fgummesson/myrepo/extra-files-test:{{.Version}}
    extra_files:
      - "."
`
``


then ran 
```
 $> goreleaser release --snapshot --clean
  • starting release...
  • loading config file                              file=.goreleaser.yaml
.....
  • docker images
  ⨯ release failed after 0s                  error=docker build failed: failed to copy extra file '.': failed to open 'dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/Dockerfile': open dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/dist/goreleaserdocker2114537310/Dockerfile: file name too long
```

caarlos0: Ahhh, yeah, it stills create the temp inside dist, maybe we can simply create it elsewhere... I'll take a look later 🙏

caarlos0: indeed, this is not needed anymore https://github.com/goreleaser/goreleaser/pull/3823

PR #3795, opened 2023-02-22
build: test on windows too

caarlos0:
this ought to be painful, but its time, probably...


caarlos0:
being continued in https://github.com/goreleaser/goreleaser/pull/4293

Issue #3078, opened 2022-04-29
Gitlab Login on Docker Image


djarbz:
### Is your feature request related to a problem? Please describe.

I wish to deploy my image to both the Gitlab Registry for my project as well as the Docker Hub.
Currently, the only way I see this possible, when using the goreleaser docker image, is to run the build twice passing different DOCKER_... env variables.

### Describe the solution you'd like

I would like to be able to pass the following variable directly to the goreleaser docker image so that I can login to both Docker Hub and the Gitlab registry. I do see that an option is already added to the entrypoint script for Github.

```
CI_REGISTRY
CI_REGISTRY_USER
CI_REGISTRY_PASSWORD
```

### Describe alternatives you've considered

I could run the build twice and just pass the variable as the existing docker ones.
```
DOCKER_REGISTRY: "$CI_REGISTRY"
DOCKER_USERNAME: "$CI_REGISTRY_USER"
DOCKER_PASSWORD: "$CI_REGISTRY_PASSWORD"
```

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

djarbz:
Another option would be to add logins in the `.goreleaser.yaml` configuration that we can set our own environment variable to.
This would allow us to add an arbitrary amount of registries to login to.

For example:
```
docker_registries:
  - registry: '{ .Env.DOCKER_HUB_REGISTRY }'
    username: '{ .Env.DOCKER_HUB_USERNAME }'
    password: '{ .Env.DOCKER_HUB_PASSWORD }'
  - registry: '{ .Env.CI_REGISTRY }'
    username: '{ .Env.CI_REGISTRY_USER }'
    password: '{ .Env.CI_REGISTRY_PASSWORD }'
  - registry: '{ .Env.GITHUB_REGISTRY }'
    username: '{ .Env.GITHUB_REGISTRY_USER }'
    password: '{ .Env.GITHUB_TOKEN }'
```

caarlos0:
can you maybe login before running the docker image, and then mount `~/.docker/config.json`?

djarbz:
Ok, so I tried and I keep running up empty.
```
release failed after 556.31s error=docker images: failed to publish artifacts: failed to push notmycloud/smtpsmartrelay:0.1.1-amd64: exit status 1: The push refers to repository [docker.io/notmycloud/smtpsmartrelay]
77b3a3588e44: Preparing
[...]

PR #3806, opened 2023-02-26
refactor: improve context on tests

caarlos0:
using `context.New` is the correct way, as it set up everything needed in the context.

using `context.Context{}` might break things eventually, or not test the same thing goreleaser does.

There's a lot more places to fix, those are the easy ones :P


caarlos0:
Refs #3078

PR #3812, opened 2023-02-28
test: add identified failing condition in sourcearchive

mohammed90:
We have recently enabled Source Archives in caddyserver/caddy#5403, which works fine when run on macOS. We found out it fails on Linux with this error:
```
  • creating source archive
    • creating source archive                        file=caddy_2.6.4-SNAPSHOT-096971e3_src.tar.gz
  ⨯ release failed after 20s                 error=globbing failed for pattern "modules/caddyhttp/fileserver/testdata/\331\205\331\204\331\201.txt": matching "./"modules/caddyhttp/fileserver/testdata/331205331204331201.txt"": file does not exist
```
It refers to a file with non-ASCII name, i.e. https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/fileserver/testdata/ملف.txt. Indeed, adding a file with non-ASCII name to the `source_test.go` file passes with flying colors on macOS but fails on Linux. I've run it on both machines and confirmed the discrepancy.

I don't know how to fix it yet because I haven't delved into goreleaser's codebase nor am I familiar with it. I need to investigate further and trace the execution path.

Thanks @IndeedNotJames for alerting me for the failure on Linux systems, which enabled me to find the issue.

### DIFF
--- a/internal/pipe/sourcearchive/source_test.go
+++ b/internal/pipe/sourcearchive/source_test.go
@@ -21,6 +21,7 @@ func TestArchive(t *testing.T) {
 			testlib.GitInit(t)
 			require.NoError(t, os.WriteFile("code.rb", []byte("not really code"), 0o655))
 			require.NoError(t, os.WriteFile("code.py", []byte("print 1"), 0o655))
+			require.NoError(t, os.WriteFile("ملف.txt", []byte("محتوى عربي"), 0o655))
 			require.NoError(t, os.WriteFile("README.md", []byte("# my dope fake project"), 0o655))
 			require.NoError(t, os.WriteFile(".gitignore", []byte(`
 added-later.txt
@@ -88,6 +89,7 @@ subfolder/
 				"foo-1.0.0/code.py",
 				"foo-1.0.0/code.rb",
 				"foo-1.0.0/code.txt",
+				"foo-1.0.0/ملف.txt",
 				"foo-1.0.0/added-later.txt",
 				"foo-1.0.0/subfolder/file.md",
 			}


caarlos0:
hmmm, cool find, let's see the failure in github actions... 

on the top of my head, I'm not sure where the issue could be either... but I'll investigate.... or if you want to dig into it, I'm on [discord](https://goreleaser.com/discord) most of the day

francislavoie:
For ref:

```
2023-03-01T03:44:55.7288110Z === RUN   TestArchive
2023-03-01T03:44:55.7298301Z === RUN   TestArchive/tar.gz
2023-03-01T03:44:55.7699904Z     source_test.go:51: 
2023-03-01T03:44:55.7701073Z         	Error Trace:	/home/runner/work/goreleaser/goreleaser/internal/pipe/sourcearchive/source_test.go:51
2023-03-01T03:44:55.7701763Z         	Error:      	Received unexpected error:
2023-03-01T03:44:55.7702937Z         	            	globbing failed for pattern "\331\205\331\204\331\201.txt": matching "./"331205331204331201.txt"": file does not exist
2023-03-01T03:44:55.7703550Z         	Test:       	TestArchive/tar.gz
```

caarlos0:
there's this issue on sourcearchives and also another one in which if you have things like this in your gitattributes:

```gitattributes
some-path export-subst
```

it won't get subst, because, well, our implementation doesn't even look at this.

Maybe a better approach would be to run git-archive and actually edit the resulting files 🤔

caarlos0:
changed the source archive code to use `git-archive` under the hood, and edit it later... and it still happens there... 100% no idea why

mohammed90:
Here's something curious. This indicates the file name consists of 6 characters:
```
\331\205\331\204\331\201
```
But the file name is actually `ملف`, which is 3 characters. Moreover, those unicode escapes/codepoints don't appear to be correct. Where does this string of characters come from? Where is this globbing happening and where is it receiving this value from?

caarlos0:
the flow is more or less:

- https://github.com/goreleaser/goreleaser/blob/main/internal/archivefiles/archivefiles.go#L20
- https://github.com/goreleaser/fileglob/blob/main/glob.go#L103

q: are those chars utf-8?

mohammed90:
> q: are those chars utf-8?

Yep. They're regular Arabic letters.

internal/pipe/sourcearchive/source_test.go
```diff
@@ -21,6 +21,7 @@ func TestArchive(t *testing.T) {
 			testlib.GitInit(t)
 			require.NoError(t, os.WriteFile("code.rb", []byte("not really code"), 0o655))
 			require.NoError(t, os.WriteFile("code.py", []byte("print 1"), 0o655))
+			require.NoError(t, os.WriteFile("ملف.txt", []byte("محتوى عربي"), 0o655))
```
caarlos0, line 24: will have to move this to after the git commit to get the "right" error

caarlos0:
okay, I figured it out, PR incoming!

caarlos0:
see #3925

PR #3816, opened 2023-03-01
typo: document number error

CC11001100:
Hello, I found a number error in the document when I consulted the document, so I propose a pr to fix it


caarlos0:
Thanks!

PR #3827, opened 2023-03-03
feat: infer package name from go.mod

### LINKED ISSUE #3825: Improve the "Quick start" documentation to avoid run failure after goreleaser init

The quick start documentation [here](https://goreleaser.com/quick-start/) implies that after running `goreleaser init`, `goreleaser release --snapshot --clean` should succeed.

However, `goreleaser release --snapshot --clean` invocation ends in an error:

```error=couldn't guess project_name, please add it to your config```

This is because the default generated `.goreleaser.yaml` does not have a `project_name` name set in the yaml file. 

Adding the `project_name` makes the run to succeed.

There are two options:

- Add the `project_name` attribute to the generated yaml file when the `goreleaser init` is run
- Add an extra step in the documentation asking the user to set the `project_name` attribute

I added the `project_name` attribute myself and it worked, but had to dig around for what's possibly wrong since the Quick start does not mention it.

### Search

### Code of Conduct

### Additional context

_No response_

caarlos0:
If all other strategies fail, try to infer the `package_name` property from the `go.mod` file, using its last segment as the actual package name.

This is not perfect, but usually this will only be used when running against a new project, with no git url, empty/default config, etc... so, in reality, it'll rarely be used.

closes #3825

### DIFF
--- a/internal/pipe/project/project.go
+++ b/internal/pipe/project/project.go
@@ -3,6 +3,8 @@ package project
 
 import (
 	"fmt"
+	"os/exec"
+	"strings"
 
 	"github.com/goreleaser/goreleaser/pkg/context"
 )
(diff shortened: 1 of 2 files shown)


ryancurrah:
Did this take into account major version modules? Where they end in /v2 or /v3?

caarlos0:
at the time I thought about that and thought it would not be important, as this only affects repos with no `project_name` and no git remotes set... if you think it should I can fix it 🤔

ryancurrah:
> at the time I thought about that and thought it would not be important, as this only affects repos with no `project_name` and no git remotes set... if you think it should I can fix it 🤔 

Meh was just curious.

PR #3841, opened 2023-03-06
feat: include prerelease suffix in git

vandot:
This commit will fix bad version tag sort if there is a prerelease on the same commit as a release tag. Current output is shown below
```
❯ git tag --points-at HEAD --sort=-version:refname --format='%(creatordate)%09%(refname)'
Thu Mar 2 21:38:51 2023 +0300   refs/tags/v1.13.0-rc3
Thu Mar 2 21:38:51 2023 +0300   refs/tags/v1.13.0
```
Test is changed to match current default value so it will fail without this fix.

Default value `-` is set to the one that is described inside [docs](https://goreleaser.com/how-it-works/?h=prerelease#how-it-works), but people are still allowed to change it.

Output with fix applied
```
❯ git -c versionsort.suffix=- tag --points-at HEAD --sort -version:refname --format='%(creatordate)%09%(refname)'
Thu Mar 2 21:38:51 2023 +0300   refs/tags/v1.13.0
Thu Mar 2 21:38:51 2023 +0300   refs/tags/v1.13.0-rc3
```

More info about `versionsort.suffix` can be found [here](https://github.com/git/git/blob/master/Documentation/config/versionsort.txt#L5)

Docs as well both schemas are updated as well.

I am not sure if users should be allowed to change this option at all.

### DIFF
--- a/internal/pipe/git/git.go
+++ b/internal/pipe/git/git.go
@@ -288,24 +288,22 @@ func getPreviousTag(ctx *context.Context, current string) (string, error) {
 }
 
 func gitTagsPointingAt(ctx *context.Context, ref string) ([]string, error) {
-	args := []string{
+	args := []string{}
+	if ctx.Config.Git.PrereleaseSuffix != "" {
+		args = append(
+			args,
+			"-c",
+			"versionsort.suffix="+ctx.Config.Git.PrereleaseSuffix,
+		)
+	}
+	args = append(
+		args,
 		"tag",
 		"--points-at",
 		ref,
 		"--sort",
 		ctx.Config.Git.TagSort,
-	}
-	if ctx.Config.Git.PrereleaseSuffix != "" {
-		args = []string{
-			"-c",
-			"versionsort.suffix=" + ctx.Config.Git.PrereleaseSuffix,
-			"tag",
-			"--points-at",
-			ref,
-			"--sort",
-			ctx.Config.Git.TagSort,
-		}
-	}
+	)
 	return git.CleanAllLines(git.Run(ctx, args...))
 }
 
--- a/internal/pipe/git/git.go
+++ b/internal/pipe/git/git.go
@@ -28,9 +28,6 @@ func setDefaults(ctx *context.Context) {
 	if ctx.Config.Git.TagSort == "" {
 		ctx.Config.Git.TagSort = "-version:refname"
 	}
-	if ctx.Config.Git.PrereleaseSuffix == "" {
-		ctx.Config.Git.PrereleaseSuffix = "-"
-	}
 }
 
 // Run the pipe.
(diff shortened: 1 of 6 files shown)


internal/pipe/git/git_test.go
```diff
@@ -180,7 +180,7 @@ func TestTagSortOrder(t *testing.T) {
 	testlib.GitTag(t, "v0.0.1")
 	ctx := testctx.NewWithCfg(config.Project{
 		Git: config.Git{
-			TagSort: "-version:creatordate",
```
caarlos0, line 183: probably best to create more subtests for each config we need to test...

www/docs/customization/git.md
```diff
@@ -12,4 +12,9 @@ git:
   #
   # Default: `-version:refname`
   tag_sort: -version:creatordate
+  # What should be used to specify prerelease suffix while sorting tags when gathering
```
caarlos0, line 15: should add an empty line before

internal/pipe/git/git.go
```diff
@@ -290,6 +293,8 @@ func getPreviousTag(ctx *context.Context, current string) (string, error) {
 func gitTagsPointingAt(ctx *context.Context, ref string) ([]string, error) {
 	return git.CleanAllLines(git.Run(
 		ctx,
+		"-c",
+		"versionsort.suffix="+ctx.Config.Git.PrereleaseSuffix,
```
caarlos0, line 297: I think we shouldn't assume `-` to be the default... maybe its better to only apply those when the config is not empty?

caarlos0 (commented):
Many thanks for the PR!


I made a couple of comments, let me know what you think!

vandot:
@caarlos0 thank you for your comments. I don't like my approach with `args`, that's why I set it to have default value in the first place :)

internal/pipe/git/git.go
```diff
@@ -291,16 +288,25 @@ func getPreviousTag(ctx *context.Context, current string) (string, error) {
...
+			"-c",
+			"versionsort.suffix=" + ctx.Config.Git.PrereleaseSuffix,
+			"tag",
+			"--points-at",
+			ref,
+			"--sort",
+			ctx.Config.Git.TagSort,
+		}
+	}
+	return git.CleanAllLines(git.Run(ctx, args...))
```
caarlos0, line 309: can use less repetition:

```go
args := []string{}
if ctx.Config.Git.PrereleaseSuffix != "" {
 			 args = append(
 			  			 args, 
 			 			"-c", 
 			 			"versionsort.suffix=" + ctx.Config.Git.PrereleaseSuffix,
 			)
}
args = append(
 			 args, 
 			"tag",
 			"--points-at",
 			ref,
 			"--sort",
 			ctx.Config.Git.TagSort,
)
return git.CleanAllLines(git.Run(ctx, args...))
```

vandot: Done, thanks

caarlos0:
awesome, thanks for the work!

this will be in goreleaser 1.17, so, by the end of march, if everything goes well. 
Will leave this open for a couple of weeks in case we need to do patch releases of 1.16. 
If conflicts happen, don't worry about it, I'll fix them.

Thanks, again!

caarlos0 (approved):

Issue #3844, opened 2023-03-08
Brew tap fails if `binary` format for archives are used.


caarlos0:
### Discussed in https://github.com/orgs/goreleaser/discussions/3843

<div type='discussions-op-text'>

<sup>Originally posted by **fogfish** March  7, 2023</sup>
If project uses the binary archives

```
archives:
  - format: binary
```

it fails to install brewtap
```
Errno::ENOENT: No such file or directory - ...
```

The rootcause seems to be `bin.install`. The brew downloads the binary file but `bin.install` assumes tarball

```
on_macos do
    if Hardware::CPU.intel?
      url "https://github.com/abc/abc/releases/download/v1.0.5/abc_1.0.5_darwin_amd64"
      sha256 "883eea2c809bae2b6d49b7420f1349c9066f2bd4bbc993806e0586a3fad7efe0"

      def install
        bin.install "abc"
      end
    end
   end
``` 

Is this a know limitations, bug or user error?

Thank you in advanced.</div>

caarlos0:
@fogfish could you share your full config file?

also, if possible, the goreleaser output log.

---

if not, a couple of questions:

- how many builds in your `build` section?
- how many archives? are they all with format `binary`?
- do you have a custom `brews.install`, maybe?
- in the logs, please look for warns like "guessing install to be XYZ"

fogfish:
Nothing special has been done for the config. It is really out-of-the-box file generated by goreleaser:
https://github.com/assay-it/assay-it/blob/v1.0.4/.goreleaser.yaml#L5
https://github.com/assay-it/assay-it/releases/tag/v1.0.4

and outbound tap
https://github.com/assay-it/homebrew-tap/blob/ade7afb2faba430bb2712fc1681cc51f60643cb9/Formula/assay-it.rb

No errors by goreleaser, everything work fine. Just to empathize again, it seems that the binary "archives" cannot be handler by `bin.install` because the name of downloaded file is not matching expectation.

caarlos0:
okay, you actually probably just need to remove this line:

https://github.com/assay-it/assay-it/blob/bbded6cb5eb926f979a7ce61c92e373aa305c74a/.goreleaser.yaml#L43

goreleaser will then guess the correct install, hopefully. 

You can try it out with `goreleaser --clean --snapshot`, and investigating the `dist/*.rb` file.

fogfish:
Excellent! Thank you a lot for support! It has fixed it. 

```
def install
[...]

PR #3851, opened 2023-03-08
Update archive.md

fogfish:
The changes inspired by #3844

### DIFF
--- a/www/docs/customization/archive.md
+++ b/www/docs/customization/archive.md
@@ -253,5 +253,12 @@ archives:
 - format: binary
 ```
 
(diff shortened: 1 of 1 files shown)


caarlos0:
Thanks for the pr!

I almost think it will be best to add a documentation for that in the build section, so we can have like "binary builds" subtitle in there... wdyt?

fogfish:
Yes, it would be easier to discover it under the build section. I've added `Command line application and utilities` section to `build.md`

I am closing this PR, a new changes are proposed here #3862

PR #3852, opened 2023-03-09
feat(ko): support labels

maxbrunet:
Add a `labels` key-value map to the `kos` config.

My interest is to be able to label the built images: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#labelling-container-images


www/docs/customization/ko.md
```diff
@@ -39,6 +39,12 @@ kos:
   # Defaults to cgr.dev/chainguard/static.
   base_image: alpine
 
+  # Labels for the built image.
+  #
+  # Defaults to null.
```
caarlos0, line 44:
```suggestion
  # Defaults to null.
  # Since v1.17.
```

caarlos0 (commented):
looks good overall.

I guess we should also check that the actual end image has the labels?

On Docker tests, line 54, we have a `shouldFindImagesWithLabels` which I think we can copy/edit as needed here...

Other than that, great work! Thanks for the PR 🙏

maxbrunet:
> I guess we should also check that the actual end image has the labels?
>
> On Docker tests, line 54, we have a `shouldFindImagesWithLabels` which I think we can copy/edit as needed here...

Good call, I think I'll need to query the registry directly instead of using the Docker CLI, and from a quick check with `curl` what I have done does not work :disappointed: 

```shell
$ curl -sS -H "Accept: application/vnd.oci.image.manifest.v1+json" \
  http://127.0.0.1:5052/v2/goreleasertest/testapp-0d1db77ee324390a18cda6b3112df9c6/manifests/labels \
  | jq
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "config": {
    "mediaType": "application/vnd.oci.image.config.v1+json",
    "size": 1007,
    "digest": "sha256:bce047be9b1c7f361ed1ac3224631e0949626a856955b0c0afe72c2fc419b11d"
  },
  "layers": [
    {
      "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
      "size": 465725,
      "digest": "sha256:4a5ee0fcd8d009284fcc50837518ef3b50df57a2ef7a941340c94b72e7b8087c"
    },
    {
      "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
      "size": 127,
      "digest": "sha256:250c06f7c38e52dc77e5c7586c3e40280dc7ff9bb9007c396e06d96736cf8542"
    },
    {
      "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
      "size": 409938,
      "digest": "sha256:a39365476c57e6a1714d5e6c0d58b510c22f9dff66b484f59fd6e3bba780c328"
    }
  ],
  "annotations": {
    "org.opencontainers.image.base.digest": "sha256:3348108bb659829a4876b39cc07ae309383cc2e63eb67e640b7551591c232ffb",
    "org.opencontainers.image.base.name": "cgr.dev/chainguard/static:latest",
    "org.opencontainers.image.revision": "36a1664e1821de8e8986390f78a403bd98139b10",
    "org.opencontainers.image.source": "https://github.com/chainguard-images/images"
  }
}
```

developer-guy:
lgtm, great catch, thanks for making this!

internal/pipe/ko/ko.go
```diff
@@ -195,7 +199,7 @@ func (o *buildOptions) makeBuilder(ctx *context.Context) (*build.Caching, error)
 	case "go.version-m":
 		buildOptions = append(buildOptions, build.WithGoVersionSBOM())
 	case "none":
-		// don't do anything.
+		buildOptions = append(buildOptions, build.WithDisabledSBOM())
```
maxbrunet, line 202: The bug (unrelated to labels), let me know if you'd like it in a separate PR to release a hotfix

maxbrunet (commented):
> what I have done does not work

nvm, I confused labels which are in the config with manifest annotations

---

I added more tests for the labels, but also for the platforms and SBOM, and caught a bug!

---

I tried to use `Bare: true` for the tests to make the image reference easier to build, but I get an error:

```
publish: GET http://localhost:5052/v2/goreleasertest/testapp/blobs/uploads/: BLOB_UPLOAD_UNKNOWN: blob upload unknown to registry
```

No idea why it makes a `GET` request and not a `POST` :confused:

internal/pipe/ko/ko.go
caarlos0: if you don't mind, that would be awesome, yes!

thank you <3

maxbrunet: Here it is: #3861 :slightly_smiling_face:

internal/pipe/ko/ko_test.go
```diff
@@ -166,20 +214,89 @@ func TestPublishPipeSuccess(t *testing.T) {
 				},
 				Kos: []config.Ko{
 					{
-						ID:         "default",
-						Build:      "foo",
-						WorkingDir: "./testdata/app/",
-						BaseImage:  table.BaseImage,
-						Repository: fmt.Sprintf("%s/goreleasertest", registry),
```
maxbrunet, line 173: > I tried to use `Bare: true` for the tests to make the image reference easier to build, but I get an error

Found the issue, there is a double slash in the `Repository`, `registry` already contains a trailing slash, this is going to be fixed

maxbrunet:
@caarlos0 I added creation times too, let me know if it is ok with you :slightly_smiling_face:

www/docs/customization/ko.md
```diff
@@ -58,6 +65,18 @@ kos:
   - latest
   - '{{.Tag}}'
 
+  # Creation time given to the image.
```
caarlos0, line 68: this needs to be seconds since epoch, right? should we make that more explicit here?

maxbrunet: Correct, I have clarified that in the comment, let me know if that's enough :)

caarlos0:
Thank you!

PR #3861, opened 2023-03-10
fix(ko): disable SBOM when `sbom` is `none`

maxbrunet:


caarlos0:
Many thanks @maxbrunet !

developer-guy:
thanks @maxbrunet !

PR #3862, opened 2023-03-11
Update builds.md

fogfish:
The changes inspired by #3844


caarlos0:
okay, I tested it, nothing creates the `bin.install` default in the config file, if you leave it empty on binary releases, its guessed correctly:


```yaml
builds:
  - env:
      - CGO_ENABLED=0
    goos:
      - linux
      - darwin
    goarch:
      - amd64
      - arm64

archives:
  - format: binary

brews:
  - tap:
      owner: caarlos0
      name: homebrew-tap
    folder: Formula
```

This already does what's its supposed to. 🤔

Issue #3890, opened 2023-03-25
GPG signing failing when using Gitbash


go-compile:
### What happened?

When using Gitbash on Windows within VSCode, gpg signing fails, giving the error: `gpg: no default secret key: No secret key`, despite having a valid GPG key.

### Suspected Cause of the Error
GoRealeaser is using the wrong installation of GPG (Gitbash GPG instead of GnuPG). Both are available via the %PATH%.

```bash
$ where gpg
C:\Program Files\Git\usr\bin\gpg.exe
C:\Program Files (x86)\GnuPG\bin\gpg.exe
```

### Proposed Solution
Before executing GPG:
1. Check `git config gpg.program`
2. If string empty, break sequence and use default path.
3. Use `git config gpg.program` as the gpg binary path.

```bash
$ git config gpg.program
C:\Program Files (x86)\GnuPG\bin\gpg.exe
```

### How can we reproduce this?

1. Install: VScode, Gitbash, GnuPG
2. Import a key into Kleopatra.
3. Set `goreleaser.yml` to appendix A
4. Open a Gitbash terminal in VSCode and execute `goreleaser release`.

#### Appendix A
> .goreleaser.yml
```bash
signs:
  - artifacts: checksum
    output: true
```

### goreleaser version

```bash
goreleaser version dev
goos: windows
goarch: amd64
module version: v1.14.1, checksum: h1:b+cW4BSDL5Ii32scSX8IoXrnpwHIItXgZyMskBhNK8s=

https://goreleaser.com
```

### GoReleaser Check

- [X] goreleaser check shows no errors

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Additional context

There is a temporary hacky fix by setting:
```yaml
signs:
  - artifacts: checksum
    output: true
    cmd: C:\Program Files (x86)\GnuPG\bin\gpg.exe
```

The problem with this is, different users and CI/CD systems have different configurations, thus different GPG paths. Thus, builds will fail depending on the build environment. Hence, the need for a more robust solution, e.g. accessing the `git config gpg.program` variable for a seamless developer experience.

PR #3891, opened 2023-03-25
fix(gpg sign): issue #3890

go-compile:
Fixes issue #3890 by checking the user's git `gpg.program` configuration. If the user didn't set this value it will use the default "gpg", just like before this PR.

No need to add a additional unit test as your existing tests cover it (mostly); however, a comment has been added to that check, informing the reader that the test environment assumes `git config gpg.program` is not be set.

### DIFF
--- a/internal/pipe/git/git.go
+++ b/internal/pipe/git/git.go
@@ -337,3 +337,15 @@ func getFromEnv(s string) func() ([]string, error) {
 		return nil, nil
 	}
 }
+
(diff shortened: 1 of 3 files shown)


caarlos0:
I like this, thanks!

caarlos0:
for some reason github won't let me merge your commit 🤔

caarlos0:
https://www.githubstatus.com/incidents/52z0j6phhnjs

oh, ok, that explains it

PR #3901, opened 2023-04-02
fix: set parallelism to match Linux container CPU

marukaz:
Currently Goreleaser uses `runtime.NumCPU()` as the default value if `--parallelism` is not set.
However, this will get the number of CPUs on the host even when Goreleaser is run in a container with a limit on the maximum number of CPUs that can be used (typically in a Kubernetes pod).
Actually, `docker run --cpus=1 goreleaser/goreleaser --debug` shows `parallelism: 4` on my machine.
This behavior causes CPU throttling, which increases execution time and, in the worst case, terminates with an error.
I ran into this problem with Jenkins where the agent runs on pod ([Kubernetes plugin for Jenkins](https://plugins.jenkins.io/kubernetes/)).

This commit introduces [automaxprocs](https://github.com/uber-go/automaxprocs) to fix this issue.
This library sets `GOMAXPROCS` to match Linux container CPU quota.
I have also looked for a library that can get CPU quota more directly, but this seems to be the best I could find.
The reason it is set in a different notation from the automaxprocs README is to prevent logs from being displayed ([comment](https://github.com/uber-go/automaxprocs/issues/18#issuecomment-511330567)).

I would have liked to write a test, but this change is dependent on the number of CPUs in the execution environment, so I could not.
Instead, I wrote a Dockerfile for testing

```Dockerfile
FROM golang:1.20.2

WORKDIR /go/app
RUN sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b /usr/local/bin
COPY . .
RUN task build
```

and confirmed built binary shows expected parallelism by following commands:

```sh
docker build --file Dockerfile.test . -t test-goreleaser
docker run --cpus=1 test-goreleaser ./goreleaser build --snapshot --debug # parallelism: 1
docker run test-goreleaser ./goreleaser build --snapshot --debug # parallelism: 4
```

I also ran the built binary on my Macbook and it was fine.


caarlos0:
I add this burried in a todo list, thanks for fixing it!

PR #3910, opened 2023-04-05
feat: allow to forcefully use a token when multiple are set

caarlos0:
This would allow to, when multiple tokens are set in the environment, force which one you want to use.

The need for this comes from the fact gitea sets both `GITHUB_TOKEN` and `GITEA_TOKEN`, and doesn't allow to easily disable either.

With this, users can add a `GORELEASER_FORCE_TOKEN=gitea` to force the gitea client to be used.

I'm not sure what's the best name for this env yet, happy to hear suggestions.

Also improved the `env_test.go` file a bit, as it was kinda messy with env vars...

refs https://github.com/orgs/goreleaser/discussions/3900


GoryMoon:
Looks good to me, I've not tested it but it looks like it should provide a solution.

PR #3924, opened 2023-04-10
fix(jsonschema): add binary to Archive format enum

maxbrunet:
Add `binary` to Archive `format` enum in the JSON schema.

JSON schema validation currently fails on `archives[0].format: binary`:

```
Value is not accepted. Valid values: "tar", "tgz", "tar.gz", "zip", "gz", "tar.xz", "txz".
``` 

Follow up to c41d6de833114f52759e6c6b70349d960a8ae293


caarlos0:
Ahh, right! Thanks!

PR #3925, opened 2023-04-11
test: ensure utf-8 sources/archives filenames work properly

caarlos0:
This now works for files added to the source archive **after** the `git archive` command is run. During the `git archive`, it seems that the `tar` is still using some format that doesn't support utf-8 by default.

Tomorrow I'll look more into it.

see https://pkg.go.dev/archive/tar#Format

closes #3812

### DIFF
--- a/internal/testlib/archive.go
+++ b/internal/testlib/archive.go
@@ -136,7 +136,7 @@ func doLsTar(f io.Reader) []string {
 		if h == nil || err == io.EOF {
 			break
 		}
-		if h.Format == tar.FormatPAX {
+		if h.Name == "pax_global_header" {
 			continue
 		}
 		paths = append(paths, h.Name)
--- a/pkg/archive/tar/tar.go
+++ b/pkg/archive/tar/tar.go
@@ -34,9 +34,8 @@ func Copying(source io.Reader, target io.Writer) (Archive, error) {
 		if err == io.EOF || header == nil {
 			break
 		}
-		// needed for git archive
-		if len(header.PAXRecords) == 0 {
-			header.Format = tar.FormatGNU
+		if err != nil {
+			return Archive{}, err
 		}
 		w.files[header.Name] = true
 		if err := w.tw.WriteHeader(header); err != nil {
@@ -76,8 +75,6 @@ func (a Archive) Add(f config.File) error {
 		return fmt.Errorf("%s: %w", f.Source, err)
 	}
 	header.Name = f.Destination
-	// https://pkg.go.dev/archive/tar#Format
-	header.Format = tar.FormatGNU
 	if !f.Info.ParsedMTime.IsZero() {
 		header.ModTime = f.Info.ParsedMTime
 	}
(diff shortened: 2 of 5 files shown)


caarlos0:
@mohammed90 no idea tbh

caarlos0:
I'm trying to come up with a reproducible for the Linux issue, so far no look...

if you wanna try it too: https://go.dev/play/p/57CO6WXxoPN

caarlos0:
that said, forcing GNU headers seems to fix it... not sure if its the right fix though

mohammed90:
I figured it out 🕵🏼‍♂️ 

It comes mangled from the output of `git ls-files`, because Git is weird: [[How to make Git properly display UTF-8 encoded pathnames in the console window?](https://stackoverflow.com/questions/22827239/how-to-make-git-properly-display-utf-8-encoded-pathnames-in-the-console-window)](https://stackoverflow.com/questions/22827239/how-to-make-git-properly-display-utf-8-encoded-pathnames-in-the-console-window).

The issue was git all along. Goreleaser used to call `git ls-files`, then parsing those lines to figure out which files to add, appending them to the list of files. The output coming from `git ls-files` was messed up due to the description detailed in the StackOverflow link. Part of your refactor removed the call to `git ls-files`, which side-stepped git shenanigans of printing non-ASCII characters in octal. The fix was a side-effect of removing that line, not due to tar headers. Why Linux? I don't know yet, but I'll chalk it up to Git being itself 💁🏼‍♂️ 

All in all, I think this change is still good because it expands the compatibility, but it's up to you. I can rest in peace now.

caarlos0:
oh, awesome you find it out!

yeah, I was already convinced that the issue would either be git or somehow Go's tar package... I was digging into tar trying to find something, but everything seemed fine.

One thing that still bothers me (because I don't understand why it happens yet):

- I create the archive with `git archive`
- I scan the archive and add the same headers/files to another archive - and I checked, they are all in PAX format
- the files with emojis and etc, even though they are added to the new archive, are not present in the final result
- if I force the all headers to GNU, it works

so, I'll investigate that as well before merging, just to make sure its not hiding some other issuel...

mohammed90:
I did a quick search and couldn't find anything meaningful. However, it seems like Gnu recommends moving to PAX format: https://www.gnu.org/software/tar/manual/html_section/Formats.html

caarlos0:
okay, this last part was on me, I feel dumb even, but figured it out...

tests were failing because my method that lists files in a tar file was ignoring pax headers 👎🏻 

anyway, fixed that, now it works!

pkg/archive/tar/tar.go
```diff
@@ -30,13 +30,15 @@ func Copying(source io.Reader, target io.Writer) (Archive, error) {
...
-		h, err := r.Next()
-		if err == io.EOF || h == nil {
+		header, err := r.Next()
+		if err == io.EOF || header == nil {
 			break
 		}
-
-		w.files[h.Name] = true
-		if err := w.tw.WriteHeader(h); err != nil {
+		if err != nil {
```
caarlos0, line 37: missing error handling, only real change in this pr

internal/testlib/archive.go
```diff
@@ -136,7 +136,7 @@ func doLsTar(f io.Reader) []string {
 		if h == nil || err == io.EOF {
 			break
 		}
-		if h.Format == tar.FormatPAX {
+		if h.Name == "pax_global_header" {
```
caarlos0, line 139: something git-archive does, apparently

PR #3930, opened 2023-04-11
feat(snapcraft): support title, icon, assumes, environment & hooks

abemedia:
This PR adds support for a few missing snapcraft fields which were required for my project.
Reference here: https://snapcraft.io/docs/snapcraft-yaml-reference


internal/pipe/snapcraft/snapcraft.go
```diff
@@ -38,16 +38,21 @@ var ErrNoSummary = errors.New("no summary provided for snapcraft")
...
+	Icon          string
 	Base          string `yaml:",omitempty"`
 	License       string `yaml:",omitempty"`
 	Grade         string `yaml:",omitempty"`
 	Confinement   string `yaml:",omitempty"`
 	Architectures []string
+	Assumes       []string                  `yaml:",omitempty"`
 	Layout        map[string]LayoutMetadata `yaml:",omitempty"`
 	Apps          map[string]AppMetadata
+	Environment   map[string]interface{} `yaml:",omitempty"`
```
caarlos0, line 54: I think this needs be in the AppMetadata struct...

abemedia: Ah indeed! Weirdly this did actually work but I guess that's an undocumented feature that should not be relied upon.

I'm currently on holiday but I'll update this PR when I'm back in 10 days.

abemedia: Just FYI it is in the schema: https://github.com/snapcore/snapcraft/blob/eb0bf1276cf2ac22ee8db5dc27dfadc0d0433cdd/schema/snapcraft.json#L490

caarlos0:
hey, @abemedia , still working on this?

abemedia:
Hi @caarlos0, sorry about the delay in response.
Will fix this up tomorrow evening.

caarlos0:
ah cool, thankss

abemedia:
Sorry about the delay @caarlos0
I just did some testing on my application with the environment config moved to the app metadata and it builds and runs fine.
Have removed this field so I think we're good now.

caarlos0:
thank you!

PR #3936, opened 2023-04-13
feat: add force_token to config

jolheiser:
As a follow-up to #3910, this adds `force_token` to the config.


caarlos0 (approved):
this looks great, thank you!


will hold the merge a bit because I want to release a patch release first, but it'll get merged, don't worry :D

PR #3938, opened 2023-04-13
fix: source archive not being added when no extra-files

### LINKED ISSUE #3937: Source archive are not added as an artifact
### What happened?

In the last version, the following source archive is not added to the list of artifacts because `len(ctx.Config.Source.Files) == 0`:

```yaml
source:
  enabled: true
  rlcp: true
  name_template: "{{ .ProjectName }}_{{ .Version }}"
```

The problem seems to be these lines
https://github.com/goreleaser/goreleaser/blob/ffefd6c4ae257d2e9aeccb5786ba66401cc1bd78/internal/pipe/sourcearchive/source.go#L64-L66

Added in a recent commit https://github.com/goreleaser/goreleaser/commit/0eb3e7975c7a134713c7e377dcb8696cb0294357

Adding a file in the `files` property works properly, the file cannot be present in the archive and needs to be in the .gitignore, if not you will get the dirty git state error.

Run `goreleaser release --clean --skip-publish --debug` on a file with a source archive:

You are supposed to see something like this:
```
  • creating source archive
    • creating source archive                        file=foo_0.25.0.tar.gz
    • running git                                    args=[-c log.showSignature=false archive -o dist/foo_0.25.0.tar.gz 1434cdcc350ace025abfd12c13eea25d521cdbc0]
    • git result                                     stderr= stdout=
    • copying file                                   dst=dist/foo_0.25.0.tar.gz.bkp src=dist/foo_0.25.0.tar.gz
    • added new artifact                             name=foo_0.25.0.tar.gz path=dist/foo_0.25.0.tar.gz type=Source
```

And instead, you see:
```
  • creating source archive
    • creating source archive                        file=foo_0.24.0.tar.gz
    • running git                                    args=[-c log.showSignature=false archive -o dist/step-ca_0.24.0.tar.gz 1434cdcc350ace025abfd12c13eea25d521cdbc0]
    • git result                                     stderr= stdout=
```

### goreleaser version

```bash
$ goreleaser --version
goreleaser version 1.17.0
commit: ffefd6c4ae257d2e9aeccb5786ba66401cc1bd78
built by: homebrew
goos: darwin
goarch: amd64

https://goreleaser.com
```

### GoReleaser Check

### Search

### Code of Conduct

### Additional context

_No response_

caarlos0:
closes #3937

### DIFF
--- a/internal/pipe/sourcearchive/source_test.go
+++ b/internal/pipe/sourcearchive/source_test.go
@@ -42,65 +42,122 @@ subfolder/
 			require.NoError(t, os.MkdirAll("subfolder", 0o755))
 			require.NoError(t, os.WriteFile("subfolder/file.md", []byte("a file within a folder, added later"), 0o655))
 
-			ctx := testctx.NewWithCfg(
-				config.Project{
-					ProjectName: "foo",
-					Dist:        "dist",
-					Source: config.Source{
-						Format:         format,
-						Enabled:        true,
-						PrefixTemplate: "{{ .ProjectName }}-{{ .Version }}/",
-						Files: []config.File{
-							{Source: "*.txt"},
-							{Source: "subfolder/*"},
+			t.Run("with extra files", func(t *testing.T) {
+				ctx := testctx.NewWithCfg(
+					config.Project{
+						ProjectName: "foo",
+						Dist:        "dist",
+						Source: config.Source{
+							Format:         format,
+							Enabled:        true,
+							PrefixTemplate: "{{ .ProjectName }}-{{ .Version }}/",
+							Files: []config.File{
+								{Source: "*.txt"},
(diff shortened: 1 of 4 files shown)


internal/pipe/sourcearchive/source_test.go
```diff
@@ -42,69 +43,104 @@ subfolder/
...
-				"foo-1.0.0/code.rb",
-				"foo-1.0.0/code.txt",
-				"foo-1.0.0/added-later.txt",
-				"foo-1.0.0/subfolder/file.md",
-			}
-
-			// zips wont have the parent dir
-			if format == "zip" {
-				expected = expected[1:]
-			}
```
caarlos0, line 98: this was actually another bug, fixed it too

PR #3946, opened 2023-04-17
fix: do not fatal on automaxprocs error

### LINKED ISSUE #3945: goreleaser 1.17.x broken in WSL with Docker Desktop installed
### What happened?

Since `1.17.0`, goreleaser does not work anymore under Windows Subsystem for Linux (WSL) with Docker Desktop installed.
The reason is this recent PR in goreleaser:
* #3901

Which introduced the [automaxprocs](https://github.com/uber-go/automaxprocs) library into goreleaser, which does not currently work under WSL + Docker Desktop:
* https://github.com/uber-go/automaxprocs/issues/65

Goreleaser refuses to run because it cannot set GOMAXPROCS. See below in the reproduction case.

The problem should be solved in the automaxprocs library, but in the meantime goreleaser is unusable since 1.17.0 out of the box.

A workaround is to provide an explicit GOMAXPROCS value as an environment variable. For example:
```
$ GOMAXPROCS=0 ./goreleaser  --version
goreleaser version 1.17.0
commit: ffefd6c4ae257d2e9aeccb5786ba66401cc1bd78
built at: 2023-04-10T11:56:39Z
built by: goreleaser
goos: linux
goarch: amd64
module version: v1.17.0, checksum: h1:CcA7EOcljdYhpEY2MUI1je6swCTCAlubxZk1VCbPKBY=

https://goreleaser.com
```

To reproduce, just run this into a WSL session (with Docker Desktop also installed in the machine):
```
$ mkdir tmp
$ cd tmp
$ curl -sSL https://github.com/goreleaser/goreleaser/releases/download/v1.17.0/goreleaser_Linux_x86_64.tar.gz | tar zxvf -
$ ./goreleaser --version
  ⨯ failed to set GOMAXPROCS                         error=invalid format for MountPoint: "991 77 0:252 / /Docker/host rw,noatime - 9p drvfs rw,dirsync,aname=drvfs;path=C:\\Program Files\\Docker\\Docker\\resources;symlinkroot=/mnt/,mmap,access=client,msize=262144,trans=virtio"
```

### goreleaser version

```bash
goreleaser version 1.17.0
commit: ffefd6c4ae257d2e9aeccb5786ba66401cc1bd78
built at: 2023-04-10T11:56:39Z
built by: goreleaser
goos: linux
goarch: amd64
module version: v1.17.0, checksum: h1:CcA7EOcljdYhpEY2MUI1je6swCTCAlubxZk1VCbPKBY=

https://goreleaser.com
```

### GoReleaser Check

### Search

### Code of Conduct

### Additional context

_No response_

caarlos0:
This might fail on windows/wsl (see https://github.com/uber-go/automaxprocs/issues/65), and its not too bad if it fails, actually, so, let's just warn the error instead of exiting 1

closes #3945

### DIFF
--- a/main.go
+++ b/main.go
@@ -29,7 +29,8 @@ func init() {
 	// automatically set GOMAXPROCS to match available CPUs.
 	// GOMAXPROCS will be used as the default value for the --parallelism flag.
 	if _, err := maxprocs.Set(); err != nil {
-		log.WithError(err).Fatal("failed to set GOMAXPROCS")
+		// might fail on WSL...
+		log.WithError(err).Error("failed to set GOMAXPROCS")
 	}
 }
 
--- a/main.go
+++ b/main.go
@@ -30,7 +30,7 @@ func init() {
 	// GOMAXPROCS will be used as the default value for the --parallelism flag.
 	if _, err := maxprocs.Set(); err != nil {
 		// might fail on WSL...
-		log.WithError(err).Error("failed to set GOMAXPROCS")
+		log.WithError(err).Warn("failed to set GOMAXPROCS")
 	}
 }
 


hhromic (commented):
@caarlos0 thanks for working out a solution so quick. Appreciated!
My only suggestion would be to log a warning-level message instead of an error.

PR #3966, opened 2023-05-01
feat: improve output

caarlos0:
- log keys will be ordered as intended instead of sorted
- paths always relative to cwd

### DIFF
--- a/internal/client/github.go
+++ b/internal/client/github.go
@@ -105,11 +105,10 @@ func (c *githubClient) Changelog(ctx *context.Context, repo Repo, prev, current 
 func (c *githubClient) getDefaultBranch(ctx *context.Context, repo Repo) (string, error) {
 	p, res, err := c.client.Repositories.Get(ctx, repo.Owner, repo.Name)
 	if err != nil {
-		log.WithFields(log.Fields{
-			"projectID":  repo.String(),
-			"statusCode": res.StatusCode,
-			"err":        err.Error(),
-		}).Warn("error checking for default branch")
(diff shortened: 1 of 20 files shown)


internal/client/github.go
```diff
@@ -105,11 +105,10 @@ func (c *githubClient) Changelog(ctx *context.Context, repo Repo, prev, current
 func (c *githubClient) getDefaultBranch(ctx *context.Context, repo Repo) (string, error) {
 	p, res, err := c.client.Repositories.Get(ctx, repo.Owner, repo.Name)
 	if err != nil {
-		log.WithFields(log.Fields{
-			"projectID":  repo.String(),
-			"statusCode": res.StatusCode,
-			"err":        err.Error(),
-		}).Warn("error checking for default branch")
+		log.WithField("projectID", repo.String()).
+			WithField("statusCode", res.StatusCode).
```
davidnewhall, line 109: if `err` is not `nil` is `res` also guaranteed to be not `nil`?

https://github.com/Unpackerr/xt/actions/runs/7597537660/job/20692451367#logs

davidnewhall, line 109: If `err` is not `nil` there's no guarantee `res` is not `nil`. This code may panic when an error is returned. I've witnessed the panic in GitHub actions. This bug appears to be present in multiple files in this contribution.

caarlos0: hey, thanks for the heads up, fixed them at 6097ea50f50b937f5f9385c803ee95cf4c2b0cd8

PR #3987, opened 2023-05-05
Fixes for prebuilt binaries docs

brianstrauch:
Fixes documentation for the "prebuilt binaries" section of the docs


caarlos0:
oh, good catch, thank you!

PR #4012, opened 2023-05-15
feat: nix support

### LINKED ISSUE #3537: nix overlay support
would be nice if goreleaser could create/manage a nix overlay... I've been told! 

I have no experience with nix, but this sounds like a cool idea given the little I understand about it...

caarlos0:
very, very, very WIP implementation of nixpkgs for GoReleaser.

**Decisions made for this first version:**
- only linux and darwin, arm64, 386 and amd64
- only support pkgs from goreleaser-generated archives
- no support to push into default nixpkgs repository
- no support to automatically add the _maybe_ new pkg to the root `default.nix`
- the generated nixpkg will be rather verbose, which shouldn't be too much of an issue as it is autogenerated anyway

**TODOs**:
- [x] macos universal binary support
- [x] custom pkg path (e.g. pkgs/misc/foo/bar/default.nix)
- [x] handle archives with a folder in them
- [x] add more options: postInstall, ??

**Will be handled in future versions**:
- [ ] archives.format=binary support
- [ ] compile from source
- [ ] PR-ing into nixpkgs
- [ ] armv6l-linux & armv7l-linux support

closes #3537


caarlos0:
example of the final nixpkg: https://github.com/caarlos0/nur/blob/master/pkgs/mods/default.nix

figsoda:
What is the scope of GoReleaser's Nix support? Is this only aiming to generate binary packages?

Just curious because [nix-init](https://github.com/nix-community/nix-init), a tool I wrote that generates Nix packages also has some basic GoReleaser support (just ldflags atm). I'm wondering if some level of integration would be possible, or are we just doing two different things.

Here is the generated Nix expression feeding `https://github.com/goreleaser/goreleaser` to nix-init The expression is currently invalid without any manual tweaking, because nix-init doesn't yet know how to translate `{{ .IsGitDirty }}` to Nix

```nix
{ lib
, buildGoModule
, fetchFromGitHub
}:

buildGoModule rec {
  pname = "goreleaser";
  version = "1.18.2";

  src = fetchFromGitHub {
    owner = "goreleaser";
    repo = "goreleaser";
    rev = "v${version}";
    hash = "sha256-OvGPE6U/OeY1KcUFjR9CCBd0WOo2Rd4349wBm9SN8bo=";
  };

  vendorHash = "sha256-0hT7wraXTUAGMJdAw3xkGzojpXnwaEOoHnW28DrA1QQ=";

  ldflags = [ "-s" "-w" "-X" "main.version=${version}" "-X" "main.commit=${src.rev}" "-X" "main.date=1970-01-01T00:00:00Z" "-X" "main.builtBy=goreleaser" "-X" "main.treeState=${isGitDirty}" ];

  meta = with lib; {
    description = "Deliver Go binaries as fast and easily as possible";
    homepage = "https://github.com/goreleaser/goreleaser";
    license = licenses.mit;
    maintainers = with maintainers; [ ];
  };
}
```

caarlos0:
@figsoda for v0 (this PR) I wanted to tackle the binary distributions only, but I wanted to generate the build from source stuff too (in another PR, in the same goreleaser release, hopefully).

will definitely look into your project and see how it works, maybe its worth just integrating with it instead of rebuilding, although I usually prefer not to rely on having something installed unless strictly necessary.

caarlos0:
I think I'm gonna leave some things out on the first version:

- [ ] binary archives support
- [ ] compile from source
- [ ] pring into nixpkgs

caarlos0:
PS: final product will look like these: https://github.com/goreleaser/nur/tree/master/pkgs

PR #4038, opened 2023-05-26
feat: allow to sign KO manifests

### LINKED ISSUE #4027: Is this possible to sign images built with ko ?
### Discussed in https://github.com/orgs/goreleaser/discussions/4026

<div type='discussions-op-text'>

<sup>Originally posted by **eddycharly** May 23, 2023</sup>
We use goreleaser with ko and tried to sign images with cosign but it looks like cosign sees nothing to sign.

GitHub [logs](https://github.com/kyverno/playground/actions/runs/5054823339/jobs/9070223468):

```shell
    ....
    • signing docker images
      • no artifacts matching the given filters found
    ....
```

Our [goreleaser config](https://github.com/kyverno/playground/blob/main/backend/.goreleaser.yaml):

```yaml
kos:
  - repository: ghcr.io/kyverno/playground
    tags:
      - '{{.Tag}}'
    bare: true
    preserve_import_paths: false
    sbom: none
    platforms:
      - all
    ldflags:
      - -s -w

docker_signs:
  - artifacts: all
    args:
      - 'sign'
      - '${artifact}'
      - "--yes"
```</div>

### PR DESCRIPTION
**Pull request by caarlos0**
add ko-generated manifests to the artifact list, this way they can be signed later.

closes #4027

### DIFF
--- a/internal/pipe/ko/ko.go
+++ b/internal/pipe/ko/ko.go
@@ -23,6 +23,7 @@ import (
 	"github.com/google/ko/pkg/build"
 	"github.com/google/ko/pkg/commands/options"
 	"github.com/google/ko/pkg/publish"
+	"github.com/goreleaser/goreleaser/internal/artifact"
 	"github.com/goreleaser/goreleaser/internal/ids"
 	"github.com/goreleaser/goreleaser/internal/semerrgroup"
 	"github.com/goreleaser/goreleaser/internal/tmpl"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by eddycharly** [comment]
@caarlos0 sure, let me try it out.

**Comment by eddycharly** [comment]
@caarlos0 it doesn't seem to fix the issue https://github.com/kyverno/playground/actions/runs/5088591208/jobs/9145187651#step:9:321

**Comment by eddycharly** [comment]
But i have:

```yaml
docker_signs:
  - cmd: cosign
    env:
      - COSIGN_EXPERIMENTAL=1
    artifacts: images
    output: true
    args:
      - sign
      - ${artifact}
      - --yes
```

And it looks like you are adding `manifest` artifacts, should I change to `artifacts: manifests` ?

BTW, why is the artifact type `manifest` ? `ko` is producing images right ?

eddycharly:
@caarlos0 it seems to work with `artifacts: all` see https://github.com/kyverno/playground/actions/runs/5089708242/jobs/9147666590

eddycharly:
> Yeah you'll need to use manifests instead of images

Can you please briefly the difference between `manifests` and `images` ?

caarlos0:
> A manifest list is a list of image layers that is created by specifying one or more (ideally more than one) image names. It can then be used in the same way as an image name in docker pull and docker run commands, for example.
> 
> Ideally a manifest list is created from images that are identical in function for different os/arch combinations. For this reason, manifest lists are often referred to as “multi-arch images”. However, a user could create a manifest list that points to two images -- one for windows on amd64, and one for darwin on amd64.

[Source](https://docs.docker.com/engine/reference/commandline/manifest/)

KO creates and publishes manifests :)

eddycharly:
> KO creates and publishes manifests :)

Thanks !

I'll have to read more about it, it's still not crystal clear.
I guess I sort of understand what a manifest is but it looks like a manifest references image (layers) so `ko` produces both manifests and layers ?

caarlos0:
yeah, I think so.

although I think the layers added on top of manifests are usually signing and sbom, whereas the layers added to images are usually the actual software - but can be signatures and sboms too

PR #4058, opened 2023-05-30
docs: prettier www

caarlos0:
probably a good idea to have some standard formatting for markdown and yaml to prevent issues 🤔


caarlos0:
will do this another time, probably...

PR #4065, opened 2023-06-01
set docker context instead of builder context

zhiyanfoo:


caarlos0:
why?

zhiyanfoo:
> why?

Apologizes I didn't realize I made this to the upstream goreleaser branch instead of my own fork.

I'll create a more detailed writeup as an issue tonight. Sorry for the noise.

PR #4070, opened 2023-06-04
docs: add missing dash in ldflags default

scop:


caarlos0:
good catch, thank you!

PR #4112, opened 2023-06-15
docs: runner script fixes

scop:
Primary motivation here is to fix the runner script on Linux arm64; see the first commit. The other two contain some while-at-it cleanups.

### DIFF
--- a/www/docs/static/run
+++ b/www/docs/static/run
@@ -23,13 +23,16 @@ test -z "$VERSION" && {
 TMP_DIR="$(mktemp -d)"
 # shellcheck disable=SC2064 # intentionally expands here
 trap "rm -rf \"$TMP_DIR\"" EXIT
-export TAR_FILE="$TMP_DIR/${FILE_BASENAME}_$(uname -s)_$(uname -m).tar.gz"
+OS="$(uname -s)"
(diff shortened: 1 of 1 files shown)


www/docs/static/run
```diff
@@ -23,14 +23,16 @@ test -z "$VERSION" && {
 TMP_DIR="$(mktemp -d)"
 # shellcheck disable=SC2064 # intentionally expands here
 trap "rm -rf \"$TMP_DIR\"" EXIT
-export TAR_FILE="$TMP_DIR/${FILE_BASENAME}_$(uname -s)_$(uname -m).tar.gz"
+OS="$(uname -s)"
+ARCH="$(uname -m)"
+test "$ARCH" = "aarch64" && ARCH="arm64"
```
caarlos0, line 28: I think we can instead replace `arm64` with `aarch64` in our archives, like we do with the others:

https://github.com/goreleaser/goreleaser/blob/a76db83da0fe0a2694bf348224ad08cc775b4f46/.goreleaser.yaml#L153-L155

thoughts?

caarlos0: oh, actually on macos it returns `arm64`... so I think this is good

PR #4114, opened 2023-06-15
fix: runner tarball extract path

scop:
Sigh, yet another regression from 4b7827829298c2f4a23dfcdc79bd4fcb685ac1bf. I actually _did_ test already with 10a627c1967a09810473c182d6e1c1dcb8748f33, but for some reason failed to notice the error. Tested better now :/


caarlos0:
thanks!!

PR #4133, opened 2023-06-21
fix: do runner cleanups also on `SIGINT` and `SIGTERM`

scop:
Behavior of the `EXIT` condition is not consistent between shells with regards to abnormal exits and signals. This cleanup should occur always.

https://austingroupbugs.net/view.php?id=621


caarlos0:
thanks!

PR #4135, opened 2023-06-21
build: cache ~/.cache/go-build

caarlos0:
In theory this should cache test results too


caarlos0:
oh, in theory this is already happening :P

PR #4140, opened 2023-06-25
test: use `t.Setenv` to set env vars

Juneezee:
We have been using `t.Setenv` is most of our tests. This PR replaces the remaining `os.Setenv` with `t.Setenv`.

Reference: https://pkg.go.dev/testing#T.Setenv

```go
func TestFoo(t *testing.T) {
	// before
	os.Setenv(key, "new value")
	defer os.Unsetenv(key)
	
	// after
	t.Setenv(key, "new value")
}
```

### DIFF
--- a/cmd/build_test.go
+++ b/cmd/build_test.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"os"
 	"runtime"
 	"testing"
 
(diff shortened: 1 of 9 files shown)


caarlos0 (approved):

caarlos0:
Thanks!

PR #4150, opened 2023-06-28
fix: Scoop releases being skipped via disabled SCM releases

radujipa:
The check to skip Scoop releases when SCM ones are disabled seems to be the only place where this is happening and in conflict to Brew releases. These two need not be linked as users can configure the `url_template` to point elsewhere.

--- 

Consider the following use case:

```
 * goreleaser/my-app.yml
   - builds: spec for building a Go binary for my-app
   - dockers: spec for packaging my-app in a container and pushing to Artifactory & GCP
   - archives: spec for zipping up my-app binary
   - artifactories: spec for pushing archives of my-app to a private jfrog artifactory
   - brews: spec for publishing my-app through brew on macOS with URLs from artifactory
   - scoop: spec for publishing my-app through scoop on Windows with URLs from artifactory
   - release: disabled SCM (e.g. GitHub) release
```

Running `goreleaser release -f goreleaser/my-app.yml` with the above setup results in the correct artifacts (archives & containers) being published to Artifactory, GCP, Brew, but no Scoop.

Given that Scoop's `skip_upload` cannot be templated in 1.18 and I do not wish to have SCM releases & tags on every release, one can be rather stuck.

### DIFF
--- a/internal/pipe/scoop/scoop.go
+++ b/internal/pipe/scoop/scoop.go
@@ -241,14 +241,6 @@ func doPublish(ctx *context.Context, manifest *artifact.Artifact, cl client.Clie
 		return pipe.Skip("release is prerelease")
 	}
 
-	relDisabled, err := tmpl.New(ctx).Bool(ctx.Config.Release.Disable)
-	if err != nil {
-		return err
-	}
-	if relDisabled {
-		return pipe.Skip("release is disabled")
-	}
-
 	commitMessage, err := tmpl.New(ctx).Apply(scoop.CommitMessageTemplate)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


caarlos0:
yes, this makes sense... they diverged in impl over the years...

thanks for the fix!

radujipa:
@caarlos0 Thanks for the quick review & merge! 🙏 

While this may be a sensitive question, is there an ETA on 1.19 being released or at least this fix?

caarlos0:
> While this may be a sensitive question, is there an ETA on 1.19 being released or at least this fix?

this week! maybe even today... still running some tests et al, since this will be a rather big release

radujipa:
Great stuff, much appreciated!

PR #4153, opened 2023-06-29
fix(nix): extra chars in vimline

caarlos0:
removed typo from vimline


caarlos0:
oh right not actually a typo

PR #4161, opened 2023-07-03
Support `make_latest` for GitHub release

### LINKED ISSUE #4159: Allow configuration of "make latest" for GitHub releases

At present, this is automatically set to `true` due to GitHub's [default API behavior around `make_latest`](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release), and goreleaser not [providing any value](https://github.com/goreleaser/goreleaser/blob/c57cdab0c6f30be07979669a465a17062345c3ef/internal/client/github.go#L346-L352).

This can be problematic when you are maintaining a series of minor release ranges, as any time a release is made it is marked as "latest" while this may not actually be true in terms of SemVer ordering.

```yaml
# .goreleaser.yaml
release:
  # If set to false, will not mark the release as "latest".
  # Not compatible with "prerelease". Default is true.
  latest: true|false|legacy
```

One way around this is to draft the release before publishing, but this takes the pleasure around fully automated releases away.

### Search

### Supporter

### Code of Conduct

### Additional context

_No response_

hiddeco:
This commit adds a `make_latest` boolean to the release configuration,
to allow signaling to GitHub if the release should be marked as latest.

Albeit being a boolean, the internal Go type is a string to allow
to distinguish an empty string (default behavior: `true`) from an
explicit `false`.

For more information around the GitHub API field, see
https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release

I did not include the `legacy` option, to not adopt something which
appears to be scheduled for removal in the future.

In addition, I opted for `make_latest` over `latest` because the
option is only available for GitHub. Which keeps the latter key
reserved for e.g. future use of a config option which is used across
Git providers.

Fixes #4159

### DIFF
--- a/internal/client/github.go
+++ b/internal/client/github.go
@@ -349,6 +349,7 @@ func (c *githubClient) CreateRelease(ctx *context.Context, body string) (string,
 		Body:       github.String(body),
 		Draft:      github.Bool(ctx.Config.Release.Draft),
 		Prerelease: github.Bool(ctx.PreRelease),
+		MakeLatest: github.String("true"),
 	}
 
 	if ctx.Config.Release.DiscussionCategoryName != "" {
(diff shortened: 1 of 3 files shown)


caarlos0 (approved):

caarlos0:
thanks @hiddeco !

PR #4164, opened 2023-07-06
fix(winget): add missing fields, more templates

caarlos0:
This adds missing fields:

- LicenseURL
- CopyrightURL
- Tags

As well as make several more fields templateable.

cc/ @twpayne


twpayne:
Looks perfect!

PR #4170, opened 2023-07-07
docs: fix yaml formatting on scoop deprecations

nicks:
docs: fix yaml formatting on scoop deprecations

right now, the formatting at 
https://goreleaser.com/deprecations#scoop
is slightly broken when you click 'after'


caarlos0:
good catch, thank you!

PR #4178, opened 2023-07-11
fix: Delete temporary dir used for docker build

jimmidyson:
Delete the temporary directory used for building Docker images.

Leaving the temporary directory leads to wasted disk space.


caarlos0:
thanks!

PR #4186, opened 2023-07-13
Use dagger for build and test pipeline

kpenfound:
This PR changes the test CI workflow to use a [dagger](https://github.com/dagger/dagger) pipeline. The net result is that dependencies required to run the full test suite locally are reduced to just Go and Docker. Runtimes locally should be significantly reduced because of Dagger's caching abilities, and the same can be applied in CI with some infrastructure changes.

This work was previously seen in a demo of this from [equinix demo day](https://twitter.com/marcosnils/status/1671593182335754248?s=20)

TODO:
- [ ] tidy up [builder.go](ci/builder.go)
- [ ] feedback on structure
- [ ] tests for the ci pipeline


go.mod
```diff
@@ -55,6 +56,19 @@ require (
...
+	github.com/Khan/genqlient v0.5.0 // indirect
+	github.com/adrg/xdg v0.4.0 // indirect
+	github.com/iancoleman/strcase v0.2.0 // indirect
+	github.com/vektah/gqlparser/v2 v2.5.1 // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20230530153820-e85fd2cbaebc // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20230530153820-e85fd2cbaebc // indirect
+)
+
+// DO NOT UPDATE THIS!! needs to be v0.0.0-20210512092938-c05353c2d58c
+require github.com/ProtonMail/go-crypto v0.0.0-20230626094100-7e9e0395ebec // indirect
```
caarlos0, line 70: can ignore the `// DO NOT UPDATE` thing, its outdated...

ci/main.go
```diff
@@ -0,0 +1,47 @@
...
+	fmt.Println("Executing CI...")
+	builder, err := builder(client, src).Sync(ctx)
+	if err != nil {
+		fmt.Printf("Error creating CI environment: %+v\n", err)
+		panic(err)
+	}
+
+	builder = builder.WithExec([]string{"go", "mod", "tidy"}).
+		WithExec([]string{"go", "build"}).
+		WithExec([]string{"go", "test", "./..."})
```
caarlos0, line 32: in the original pipeline, this would call `./script/test.sh`, which also appends to `$GITHUB_STEP_SUMMARY`... not sure if that's possible with dagger too..

Taskfile.yml
```diff
@@ -31,14 +31,9 @@ tasks:
...
   test:
     desc: Run tests
-    env:
-      LC_ALL: C
     vars:
-      TEST_OPTIONS: '{{default "" .TEST_OPTIONS}}'
-      SOURCE_FILES: '{{default "./..." .SOURCE_FILES}}'
-      TEST_PATTERN: '{{default "." .TEST_PATTERN}}'
     cmds:
-      - go test {{.TEST_OPTIONS}} -failfast -race -coverpkg=./... -covermode=atomic -coverprofile=coverage.txt {{.SOURCE_FILES}} -run {{.TEST_PATTERN}} -timeout=5m
```
caarlos0, line 41: I would leave this as is, maybe add a `ci` task, that calls `go run ./ci`

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.20
 
 require (
 	code.gitea.io/sdk/gitea v0.15.1
+	dagger.io/dagger v0.7.2
```
caarlos0, line 7: can we make `./ci` its own module? so whoever `go install` goreleaser don't need to download dagger and all this other deps too...

gerhard:
Thank you for reviewing this @caarlos0!

We gave you a shout-out in today's Dagger Community Call (https://dagger.io/events fwiw). Recording will be uploaded to YouTube in a day (or a few at most). Will drop a link when it's public. cc @jpadams @mircubed @d3rp3tt3

Thank you @marcosnils for helping kick this off 💪

ci/main.go
kpenfound: Yeah that should be possible, I'll try it out

go.mod
kpenfound: Yeah good call

go.mod
caarlos0: PS: I'll fix this on main later to not conflict with this pr :)

ci/builder.go
```diff
@@ -0,0 +1,45 @@
...
+		WithMountedCache("/gocache", client.CacheVolume("gobuildcache")).
+		WithExec([]string{"chown", "nonroot", "/gocache"}).
+		WithEnvVariable("GOCACHE", "/gocache").
+		WithExec([]string{"mkdir", "/nix"}).
+		WithExec([]string{"chown", "nonroot", "/nix"}).
+		WithMountedDirectory("/src", source).
+		WithExec([]string{"chown", "-R", "nonroot", "/src"}).
+		WithNewFile("/usr/local/sbin/docker", dagger.ContainerWithNewFileOpts{
+			Contents: `#!/bin/sh
+DOCKER_HOST=tcp://localhost:2375 /usr/bin/docker $@`,
```
marcosnils, line 28: @kpenfound given that @caarlos0 seems to be quite responsive to merge this maybe we can take the opportunity to make this change in goreleaser so this hack is not necessary?

@caarlos0 FWIW we had to do this beacuse `goreleaser` tests seem to overriding env variables when calling docker

kpenfound: Agreed. Originally this code ran outside of the goreleaser repo so it was running the tests without modifications. A few workarounds were necessary to accommodate that. In this case, there's a few tests where `localhost` is hardcoded and overriding any `DOCKER_HOST` env

caarlos0: hmmm, totally, feel free to make the necessary changes! <3

marcosnils: @caarlos0 basically `goreleaser` tests are overriding `go test` envs since they're being set here (https://github.com/goreleaser/goreleaser/blob/7c6201f5f588fa1a8504f5cffa6264646fdee1eb/internal/pipe/docker/docker_test.go#LL992C17-L992C17) and then used here (https://github.com/goreleaser/goreleaser/blob/7c6201f5f588fa1a8504f5cffa6264646fdee1eb/internal/pipe/docker/api.go#L51). 

The change that would help us to remove this hack is if `goreleaser`  appended `ctx.Env` envs to the current existing variables here https://github.com/goreleaser/goreleaser/blob/7c6201f5f588fa1a8504f5cffa6264646fdee1eb/internal/pipe/docker/api.go#L51 instead of overriding them altogether. WDYT?

caarlos0: I think it's probably the right thing to do, and what most people would expect to happen.

Honestly kinda surprised no one reported that as a bug yet 🤔

marcosnils: > I think it's probably the right thing to do, and what most people would expect to happen.
> 
> Honestly kinda surprised no one reported that as a bug yet thinking

https://github.com/goreleaser/goreleaser/pull/4187

caarlos0: merged!

mircubed:
Thanks again @caarlos0 ! You can check out the recording for the shout out here (13:20 timestamp) :) https://www.youtube.com/watch?v=dRfcSIIAvG4

caarlos0:
> Thanks again @caarlos0 ! You can check out the recording for the shout out here (13:20 timestamp) :) https://www.youtube.com/watch?v=dRfcSIIAvG4

Pretty cool! Thanks for the mention and the PR 💜

caarlos0:
one Q about this: can we automate updates, like we do with dependabot? 

if not, is there any plans to do so?

caarlos0:
hey, thanks for all the effort on this, but I think I'll leave it as-is for now.

marcosnils:
:(

sent from mobile

Em dom., 14 de jan. de 2024 15:38, Carlos Alexandro Becker <
***@***.***> escreveu:

kpenfound:
> hey, thanks for all the effort on this, but I think I'll leave it as-is for now.

sounds good @caarlos0 ! Maybe we can take another run at this with a future version of Dagger :)

PR #4187, opened 2023-07-13
ci: append envs when calling docker instead of overriding

marcosnils:
This is causing issues when trying to run tests in Dagger.
ref: https://github.com/goreleaser/goreleaser/pull/4186#discussion_r1262977616

Signed-off-by: Marcos Lilljedahl <marcosnils@gmail.com>

### DIFF
--- a/internal/pipe/docker/api.go
+++ b/internal/pipe/docker/api.go
@@ -48,7 +48,7 @@ func runCommand(ctx *context.Context, dir, binary string, args ...string) error 
 	/* #nosec */
 	cmd := exec.CommandContext(ctx, binary, args...)
 	cmd.Dir = dir
-	cmd.Env = ctx.Env.Strings()
+	cmd.Env = append(cmd.Env, ctx.Env.Strings()...)
 
 	var b bytes.Buffer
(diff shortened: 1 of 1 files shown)


caarlos0:
test failure is unrelated

internal/pipe/docker/api.go
```diff
@@ -69,7 +69,7 @@ func runCommandWithOutput(ctx *context.Context, dir, binary string, args ...stri
 	/* #nosec */
 	cmd := exec.CommandContext(ctx, binary, args...)
 	cmd.Dir = dir
-	cmd.Env = ctx.Env.Strings()
+	cmd.Env = append(cmd.Env, ctx.Env.Strings()...)
```
caarlos0, line 72: I think this should be cmd.Environ(), no?

caarlos0: https://github.com/goreleaser/goreleaser/commit/cbb09b89fd3297129aaba89f57240c32bda6aa41

marcosnils: Yep, good catch. Thx for fixing

PR #4195, opened 2023-07-18
fix(winget): replace forward slashes in Winget `RelativeFilePath`

tashian:
If an archive filename contains `/` characters, they can sneak into Winget's `RelativeFilePath`.
In this PR, I make sure that `RelativeFilePath` only uses `\` directory separators.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -56,6 +56,7 @@ require (
 )
 
 require (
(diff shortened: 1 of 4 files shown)


caarlos0:
ahh good catch, thank you!

caarlos0 (approved):

caarlos0 (approved):

PR #4201, opened 2023-07-20
chore: remove comment about WSL failure

stevenh:
Remove the comment about WSL failing to set max procs as this was fixed by v1.5.3 of automaxprocs.


caarlos0:
Thanks! 🙏

PR #4218, opened 2023-07-30
make docker build stage files same as archive

tg123:
`docker` packaging does not honor the path structure defined in builds/binary
it copies all binaries to temp docker build folder

this PR fixes the behavior and make it same as `archive` 

for example

```
builds:
- id: xxx 
  binary: plugins/xxx
```

before:

```
/tmp/goreleaserdocker2210262014/xxx
```

after:

```
/tmp/goreleaserdocker2210262014/plugins/xxx
```

### DIFF
--- a/internal/pipe/docker/docker.go
+++ b/internal/pipe/docker/docker.go
@@ -192,7 +192,12 @@ func process(ctx *context.Context, docker config.Docker, artifacts []*artifact.A
 		}
 	}
 	for _, art := range artifacts {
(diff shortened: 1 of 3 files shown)


caarlos0:
hey, thanks for the PR... can you add a test for it too?
the build failure is unrelated btw, some weirdness in github actions I guess..

tg123:
test added

caarlos0:
awesome, thank you!

PR #4254, opened 2023-08-21
feat(dependencies): upgrade cosign

PG2000:
Bump cosign to version 2.1.1
...

To use latest and greates features of cosign

...


caarlos0:
thank you!

PR #4255, opened 2023-08-22
feat: git.ignore_tags

caarlos0:
Allows to ignore tags that match the given regex expressions.


www/docs/customization/git.md
```diff
@@ -18,4 +18,14 @@ git:
   #
   # Since: v1.17
   prerelease_suffix: "-"
+
+  # Regex expressions for tags to be ignored by GoReleaser.
```
caarlos0, line 22:
```suggestion
  # Regular expressions for tags to be ignored by GoReleaser.
```

PR #4265, opened 2023-08-28
feat(docker): retry push if 503

caarlos0:
refs https://github.com/orgs/goreleaser/discussions/4263

### DIFF
--- a/internal/pipe/docker/docker.go
+++ b/internal/pipe/docker/docker.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/caarlos0/log"
 	"github.com/goreleaser/goreleaser/internal/artifact"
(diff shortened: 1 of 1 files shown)


internal/pipe/docker/docker.go
```diff
@@ -328,3 +329,24 @@ func dockerPush(ctx *context.Context, image *artifact.Artifact) error {
...
 }
+
+func doPush(ctx *context.Context, img imager, name string, flags []string) (string, error) {
+	var try int
+	for try < 10 {
+		digest, err := img.Push(ctx, name, flags)
+		if err == nil {
+			return digest, nil
+		}
+		if strings.Contains(err.Error(), "received unexpected HTTP status: 503 Service Unavailable") {
```
grzesuav, line 340: I believe all 5xx errors should be treated in that manner, is there any chance to check status code ?

caarlos0: not really, that's just the output of `docker push`... but we can probably add a regex there

PR #4293, opened 2023-09-14
ci: test on windows

caarlos0:
adding windows tests so I can better investigate some issues


lincolnthalles:
Just in case you didn't noticed yet, there's a long-standing issue with the Go testing package on Windows.

Whenever a test uses t.TempDir(), the cleanup fails, and so does the test.

This still affects the latest go 1.21.3.

https://github.com/golang/go/issues/51442
https://github.com/golang/go/issues/52986
https://github.com/golang/go/issues/44919

PR #4298, opened 2023-09-15
feat(winget): update manifest schema version & links

vedantmgoyal2009:
Use the latest schema `1.5.0` available on winget-pkgs.


caarlos0 (commented):
hey, thanks for the PR!

Doesn't this version update change anything on the fields we use? Whats different from 1.4 to 1.5?

vedantmgoyal2009:
As far as I know, no changes have been made to existing fields. Some new fields were introduced though, but they aren't  available for public use in winget-pkgs. I updated it because I think it's good to use latest version.

caarlos0:
gotcha, what effects could this have? people with older winget versions not being able to install things maybe? 

if it can have negative effects, I would probably hold it a little

vedantmgoyal2009:
No negative effects it should have. WinGet is fully backwards compatible:)

PR #4317, opened 2023-09-20
feat: add `map` and `indexOrDefault` template functions

j178:
Add a new function `map` to template functions.

`map` creates a map from a list of pairs of key values, then you can convert a key to a value using `.Get`. This simplifies writing  key-mapping templates. For example, the defaule `archives.name_template`

```yaml
name_template: >-
    {{ .ProjectName }}_
    {{- title .Os }}_
    {{- if eq .Arch "amd64" }}x86_64
    {{- else if eq .Arch "386" }}i386
    {{- else }}{{ .Arch }}{{ end }}
    {{- if .Arm }}v{{ .Arm }}{{ end }}
```

becomes

```yaml
name_template: >-
  {{ $arch := map "amd64" "x86_64" "386" "i386" -}}
  {{ .ProjectName }}_
  {{- title .OS }}_
  {{- $arch.Get .Arch .Arch }}
  {{- if .Arm }}v{{ .Arm }}{{ end }}
```


internal/tmpl/tmpl.go
```diff
@@ -346,3 +347,28 @@ func mdv2Escape(s string) string {
 		"!", "\\!",
 	).Replace(s)
 }
+
+type stringMap map[string]string
+
+// Get returns the value for the given key, or the default value if the key is
+// not present.
+func (m stringMap) Get(key string, def ...string) string {
```
caarlos0, line 355: Can the user use the `index` template fn for this?

j178: They can. However, `index` returns zero value if the key is not present. If the use want to fallback to some default value, the have to use `index` combined with `or`, which may be a little cubersome. Take `name_template` for an example:
```yaml
name_template: >-
  {{ $arch := map "amd64" "x86_64" "386" "i386" -}}
  {{ .ProjectName }}_
  {{- title .OS }}_
  {{- or (index $arch .Arch) .Arch }}
  {{- if .Arm }}v{{ .Arm }}{{ end }}
```

j178: Do you recommend making it a method of map, or a function?
`$arch.IndexOrDefault .Arch .Arch` or `indexOrDefault $arch .Arch .Arch`?

caarlos0: A function is more consistent I think

caarlos0:
awesome, thank you!

PR #4335, opened 2023-09-29
fix: decouple project_name guessing from the release pipe

caarlos0:
the release's defaults run before the project's does, so, usually the github/gitlab/gitea names are set.

however, in some cases, the release's defaults might be skipped, in which case they'll be empty.

this breaks things like `goreleaser changelog`, especially on non-go repositories.

this pr tries to extract the project name from the git remote url in the project's defaulter.

it might be possible now to move it to run before the release defaulter, even.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -6,6 +6,7 @@ Only the last stable version at any given point.
 
 ## Reporting a Vulnerability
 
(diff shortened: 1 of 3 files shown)


harshsingh32 (commented):
lgtm

PR #4340, opened 2023-10-03
feat!: removing deprecated options

caarlos0:
preparing for v2!


caarlos0:
will do this when I finally branch out v2, for now it'll be kept

PR #4356, opened 2023-10-08
refactor(tmpl): avoid unnecessary byte/string conversion

Juneezee:
We can use `(*regexp.Regexp).MatchString` instead of `(*regexp.Regexp).Match([]byte(...))` to avoid unnecessary `[]byte` conversions and reduce allocations. A one-line change for free performance gain.

Benchmark:

```go
func BenchmarkMatch(b *testing.B) {
	for i := 0; i < b.N; i++ {
		if match := envOnlyRe.Match([]byte("{{ .Env.FOO }}")); !match {
			b.Fail()
		}
	}
}

func BenchmarkMatchString(b *testing.B) {
	for i := 0; i < b.N; i++ {
		if match := envOnlyRe.MatchString("{{ .Env.FOO }}"); !match {
			b.Fail()
		}
	}
}
```

Result:

```
goos: linux
goarch: amd64
pkg: github.com/goreleaser/goreleaser/internal/tmpl cpu: AMD Ryzen 7 PRO 4750U with Radeon Graphics
BenchmarkMatch-16          	 4320873	       381.2 ns/op	      16 B/op	       1 allocs/op
BenchmarkMatchString-16    	 5973543	       203.9 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/goreleaser/goreleaser/internal/tmpl	3.366s
```

### DIFF
--- a/internal/tmpl/tmpl.go
+++ b/internal/tmpl/tmpl.go
@@ -260,6 +260,8 @@ func (e ExpectedSingleEnvErr) Error() string {
 	return "expected {{ .Env.VAR_NAME }} only (no plain-text or other interpolation)"
 }
 
+var envOnlyRe = regexp.MustCompile(`^{{\s*\.Env\.[^.\s}]+\s*}}$`)
+
 // ApplySingleEnvOnly enforces template to only contain a single environment variable
 // and nothing else.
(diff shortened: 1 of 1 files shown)


caarlos0 (approved):

caarlos0:
awesome, thanks!

PR #4378, opened 2023-10-19
build: release for riscv64

michalbiesek:


caarlos0:
no explanation on why this is needed, no merge

Issue #3746, opened 2023-02-01
Add support for depot.dev


flemzord:
### Is your feature request related to a problem? Please describe.

I would like to improve the build time of my Docker images

### Describe the solution you'd like

I wish I could tell GoReleaser to use depot.dev instead of buildx or podman

### Describe alternatives you've considered

No longer use GoReleaser to build my Docker images

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jacobwgillespie:
Hey! 👋 I'm one of the cofounders of Depot, I'd love to see an integration with GoReleaser!

For some background, Depot provides hosted BuildKit/buildx builders as a service. The [depot CLI](https://github.com/depot/cli) wraps buildx build with auth for the remote builder, so `depot build ...` accepts all the same CLI flags as `docker buildx build` would.

I think the direct way to integrate Depot and GoReleaser would be to offer a `use: depot` backend that calls `depot build --load ...` in place of the current `docker buildx build`. That would behave similarly to the `use: buildx` backend.

A more optimized integration would be to integrate with buildx's `--push` flag - you can skip downloading the built container to the local Docker image store by not including `--load` and instead calling `depot build --push`. That buildx flag both builds and pushes the image directly from the remote buildx builder, which saves both the network time and the time tarballing the image into Docker's format. This might be more involved though, since my understanding is that GoReleaser expects build and push to be two separate steps today.

Let me know your thoughts and what I can do to help.

caarlos0:
if anyone wants to try to pr it, I can review.

as far as the build and push in a single go, its how we implemented Ko, maybe its worth having its own implementation like Ko does too...

caarlos0:
Is anyone interested in working on this integration?

caarlos0:
I'm not interested in working on this, so, if no one volunteers, I'll close the issue.

That said, I think a better integration would be to make the `depot` cli configure auth on the client, and then the user can use the `buildx` builder as usual.

WitoDelnat:
Hey! We'd like to use goreleaser with Depot.dev and I noticed that there is a draft PR. I'm wondering if something is blocking the pull request from being merged? The integration would be much appreciated.

caarlos0:
sorry, I'm closing this. I just don't have the energy to maintain this, and it doesn't seem to be asked for a lot, so I'm leaving it out for now.

[...]

PR #4397, opened 2023-11-01
feat(depot): add depot image builder and pusher

goller:
Depot is a fast way to build and push docker images.

This is my draft to add depot with minimal changes to the docker build and push.
The key difference compared to the existing docker code is that depot can both build
and push multi-platform images in one step.

This PR does some refactoring of the `imager` interface to allow the `build` function to optionally return
a digest that signifies the pushed image.  I'm not sure if this is the right approach; let me know!

( no tests yet ... I'm not sure yet how to proceed as I think the tests
may need a refactor to accommodate simultaneous build and push)

So, this commit adds `use: depot` to the docker builder allowing images to be build and pushed using depot.dev.
Additionally, a new optional docker setting, `depot_project`, can explicitly set the project. 
Note that the project can also be set via the env var, `DEPOT_PROJECT_ID` or with a local file, `depot.json`.

depot.dev is a hosted buildkit that is a fast way build and push multi-platform images.

Here is the ticket about depot: https://github.com/goreleaser/goreleaser/issues/3746

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1040,6 +1040,7 @@ type Docker struct {
 	BuildFlagTemplates []string `yaml:"build_flag_templates,omitempty" json:"build_flag_templates,omitempty"`
 	PushFlags          []string `yaml:"push_flags,omitempty" json:"push_flags,omitempty"`
 	Use                string   `yaml:"use,omitempty" json:"use,omitempty" jsonschema:"enum=docker,enum=buildx,default=docker"`
+	DepotProject       string   `yaml:"depot_project,omitempty" json:"depot_project,omitempty"`
 }
 
 // DockerManifest config.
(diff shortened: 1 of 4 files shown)


pkg/config/config.go
```diff
@@ -1040,6 +1040,7 @@ type Docker struct {
 	BuildFlagTemplates []string `yaml:"build_flag_templates,omitempty" json:"build_flag_templates,omitempty"`
 	PushFlags          []string `yaml:"push_flags,omitempty" json:"push_flags,omitempty"`
 	Use                string   `yaml:"use,omitempty" json:"use,omitempty" jsonschema:"enum=docker,enum=buildx,default=docker"`
+	DepotProject       string   `yaml:"depot_project,omitempty" json:"depot_project,omitempty"`
```
caarlos0, line 1043: Assuming we might add more things later (either for depot or for something else), maybe something like:

```yaml
dockers:
- use: depot
  depot:
    project: foo
```

wdyt?

caarlos0 (commented):
one comment, but looks to me overall

pkg/config/config.go
goller: @caarlos0 Sounds great !  I've added that in [27eba41](https://github.com/goreleaser/goreleaser/pull/4397/commits/27eba41e969db2c8cd3f0a3112af7a52bfe7261a)

caarlos0:
sorry, I'm closing this. I just don't have the energy to maintain this, and it doesn't seem to be asked for a lot, so I'm leaving it out for now.

Thanks for the suggestion either way.

PR #4428, opened 2023-11-18
fix: linkedin announce api changes

### LINKED ISSUE #4421: `linkedin: could not get profile id: could not find 'id' in result`
### What happened?

Trying to use the LinkedIn announce feature, I've supplied my access token in my workflow and passed it to the Github Action. The releases all pass, then the announce fails with the error in the title. Am I doing something wrong, or is there any further setup that needs to be done on my side for this to work? The docs for this were pretty lacking

https://github.com/stefanlogue/meteor

My public repo that reproduces the issue currently

### goreleaser version

```bash
1.22.1
```

### GoReleaser Check

### Search

### Supporter

### Code of Conduct

### Additional context

_No response_

gabrielcipriano:
Closes #4421 

I chose to keep `getProfileID` as `getProfileIDLegacy` and use it as a fallback if `getProfileSub` fails because of permission scope.

In this way, it's not a breaking change because one that has only a deprecated permissions such as `r_liteprofile` will still be able to hit `v2/me`

this logic is encapsulated in the new function `getProfileURN`, that resolves the user identifier and returns it formatted as a URN

### DIFF
--- a/internal/pipe/linkedin/client.go
+++ b/internal/pipe/linkedin/client.go
@@ -3,13 +3,17 @@ package linkedin
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
 
+	"github.com/caarlos0/log"
 	"github.com/goreleaser/goreleaser/pkg/context"
 	"golang.org/x/oauth2"
 )
+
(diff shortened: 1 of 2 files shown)


caarlos0:
awesome, thank you!

PR #4441, opened 2023-11-24
fix: allow homebrew to use tar.xz format

jftuga:
In a `brews` section, goreleaser will fail when using `format: tar.xz` even though homebrew supports installing binaries bundled in a `.tar.xz` archive.

I use `.tar.xz` instead of `.tar.gz` and would like goreleaser to support this when used in conjunction with `brews` sections.

With this patch, I created a test [homebrew formulae](https://github.com/jftuga/homebrew-tap/blob/main/awswho.rb) and successfully installed it under macOS.

### DIFF
--- a/internal/pipe/brew/brew.go
+++ b/internal/pipe/brew/brew.go
@@ -215,7 +215,7 @@ func doRun(ctx *context.Context, brew config.Homebrew, cl client.ReleaseURLTempl
 		),
 		artifact.Or(
 			artifact.And(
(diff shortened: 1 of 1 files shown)


jftuga:
@caarlos0 What do I need to do in order to resolve the `golangci-lint` errors?  FWIW, this is a very small patch.

caarlos0:
@jftuga I need to fix the settings to ignore older errors actually, don't worry about it.

caarlos0:
thanks for the PR :)

PR #4473, opened 2023-12-12
fix: add -c flags when building go test

fl0Lec:
when building a go test without the `-c` flag like so:
```yaml
 builds:
- command: test
  binary: env.test
  dir: ./internal/builders/golang
  no_main_check: true
  ```

will result in the error: `exit status 1: fork/exec : exec format error`

adding the -c flags when not present in the flags
adding unit test

https://github.com/goreleaser/goreleaser/issues/4462

### DIFF
--- a/internal/builders/golang/build_test.go
+++ b/internal/builders/golang/build_test.go
@@ -1175,7 +1175,7 @@ func TestBuildGoBuildLine(t *testing.T) {
 		}, strings.Fields("go test -c -o foo.test ."))
 	})
 
-	t.Run("build test alwasy as c flags", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: internal/builders/golang/build_test.go [hunk 1]
```diff
@@ -1175,6 +1175,15 @@ func TestBuildGoBuildLine(t *testing.T) {
 		}, strings.Fields("go test -c -o foo.test ."))
 	})
 
+	t.Run("build test alwasy as c flags", func(t *testing.T) {
```
caarlos0, line 1178:
```suggestion
	t.Run("build test always as c flags", func(t *testing.T) {
```

caarlos0:
thanks!

PR #4490, opened 2023-12-21
chore: update the link to the go wiki page on first-class ports

smlx:
This change updates broken links to the old Go wiki to point to the new location.


caarlos0:
thanks!

PR #4504, opened 2023-12-26
[Issue 3759] Support direct buildkit builds without requiring a docker daemon on the build host

valerian-roche:
This PR is currently a draft and does not aim at being merged as is, but I'd like feedback on:
- is this feature potentially going to be merged?
- is this okay to add it as a separate entry from docker? The docker part is strictly written as a distinct build/push/manifest system which is not relevant here (as `buildx build` covers all three). I initially wanted to merge the configuration part but it ended up far too confusing

If this is the correct direction, I'll go on with:
- fix the run when running `goreleaser build` (it's currently simply skipped). Build can still be tested through `release --snapshot`. I'll potentially end up with `release --snapshot` not using `--load` and `build` using `--load` and failing if a daemon is not present
- add unittests
For now I was able to test:
- successful build only through `release --snapshot`, including when setting multiple architectures
- successful push with manifest when running `release`, including with multiple platforms
- successful signing of images and manifests


caarlos0:
Hey, thanks for the PR! 🤘

> This PR is currently a draft and does not aim at being merged as is, but I'd like feedback on:
> 
> * is this feature potentially going to be merged?

Yes! It's something I've been meaning to work in for a while now as well.

> * is this okay to add it as a separate entry from docker? The docker part is strictly written as a distinct build/push/manifest system which is not relevant here (as `buildx build` covers all three). I initially wanted to merge the configuration part but it ended up far too confusing

Yes. I would probably call it `containers` instead of `docker_buildkits` though, as it would allow us to later use the same configuration for other daemonless builders (e.g. kaniko #2322)

I see some code is copied from the docker pipe, maybe we should put it in a lib in the internal package for reuse?

valerian-roche:
> Yes. I would probably call it containers instead of docker_buildkits though, as it would allow us to later use the same configuration for other daemonless builders (e.g. kaniko https://github.com/goreleaser/goreleaser/issues/2322)

I updated the PR to use containers and allow the definition of more builders along the way. 
I also now support three buildkit modes (the ones I can test):
- basic docker builder, though it can't do multi-arch pushes for now
- local docker builder with experimental `containerd` image store. It's able to do single-step, multi-arch manifests pushes
- remote kubernetes builder

A missing step today is how to [distinguish](https://github.com/goreleaser/goreleaser/pull/4504/files#diff-c49935bb0dd01f5bde137fc5fc125fc222295507dc0c15e8d00233f92a2408f3R93) a `goreleaser build` run from a `goreleaser release` run, as I want to always build and import the images for the first one, and only build if not pushing (snapshot) in the second case.

If this looks okay I'll work on the unit tests

caarlos0:
> A missing step today is how to [distinguish](https://github.com/goreleaser/goreleaser/pull/4504/files#diff-c49935bb0dd01f5bde137fc5fc125fc222295507dc0c15e8d00233f92a2408f3R93) a `goreleaser build` run from a `goreleaser release` run, as I want to always build and import the images for the first one, and only build if not pushing (snapshot) in the second case.

I don't think we have that now... but seems like something good to have. Will PR it ASAP.

valerian-roche:
@caarlos0 Thanks for the commit. I updated the PR with:
- Common code between `docker` and `container` for the build context and platform part
- Proper handling of multi-arch builds. This requires supporting a new binary path when running multi-arch builds. For compatibility the current root artifacts are still present for single arch builds

If this looks okay to you I'll start working on the tests once you confirm this would be going this direction

caarlos0:
hey, sorry for the delay, this does look good to me so far, yes!

thanks for all the work and effort here btw 🙏

valerian-roche:
Hey, sorry for the delay. I have had conflicting priorities and haven't had anytime to work on unit testing this sadly. 
From my manual testing it works:
 - local run on my laptop with a daemon and remote build
 - local run with local daemon and local build
 - CI run with no daemon and remote build

If you have some bandwidth to work on testing that'd be great, otherwise I plan on coming back to the PR later on when I'll have more capacity

caarlos0:
closing due to lack of activity

PR #4508, opened 2023-12-29
feat: store which action is being taken in the context

caarlos0:
refs #4504


valerian-roche:
Thanks for the quick delivery on this!

PR #4537, opened 2024-01-13
feat: make scm pre-release a template field

vedantmgoyal2009:
```yaml
# .goreleaser.yaml
release:
  # If set to auto, will mark the release as not ready for production
  # in case there is an indicator for this in the tag e.g. v1.0.0-rc1
  # If set to true, will mark the release as not ready for production.
  # Templates: allowed (since v1.24.0)
  # Default is false.
  prerelease: auto
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -49,7 +49,7 @@ require (
 	golang.org/x/oauth2 v0.16.0
 	golang.org/x/sync v0.6.0
 	golang.org/x/text v0.14.0
-	golang.org/x/tools v0.16.1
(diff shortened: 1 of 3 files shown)


caarlos0:
I'm curious what's the usecase for this?

also, can you update the docs and add tests?

Thanks!

vedantmgoyal2009:
The usecase for this is little odd; allows to workaround nightly releases functionality from GoReleaser Pro.

caarlos0:
okay, I rather not merge this then.

thanks for the pr though

PR #4540, opened 2024-01-15
docs: add instructions for setting up local documentation

zimeg:
This PR adds a few notes on installing mkdocs and the extensions needed to run the documentation locally. The list of extensions was added to until build errors went away and the site was served successfully!


caarlos0:
thanks!

PR #4550, opened 2024-01-18
fix(nix): sourceRoot when using archives.wrap_in_directory

caarlos0:
If `archives.[*].wrap_in_directory` is set, it'll create a folder inside the archive file, usually something like `app_goos_goarch`.

In those cases, the root of the archive is not constant, so we create a `sourceRootMap` and use it instead.

In cases where the `sourceRoot` is constant, the generated derivation will be the same.

refs https://github.com/orgs/goreleaser/discussions/4549

### DIFF
--- a/internal/pipe/nix/tmpl.nix
+++ b/internal/pipe/nix/tmpl.nix
@@ -62,6 +62,32 @@ let
     aarch64-darwin = "{{ . }}";
     {{- end }}
   };
+
+  {{- if not .SourceRoot }}
+  sourceRootMap = {
+    {{- with  .SourceRoots.linux386 }}
+    i686-linux = "{{ . }}";
+    {{- end }}
+    {{- with  .SourceRoots.linuxamd64 }}
+    x86_64-linux = "{{ . }}";
+    {{- end }}
+    {{- with  .SourceRoots.linuxarm6 }}
+    armv6l-linux = "{{ . }}";
+    {{- end }}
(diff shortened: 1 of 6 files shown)


internal/pipe/nix/tmpl.nix
```diff
@@ -62,6 +62,32 @@ let
     aarch64-darwin = "{{ . }}";
     {{- end }}
   };
+
+  {{- if not .SourceRoot }}
```
till, line 66: 👍🏼

internal/pipe/nix/tmpl.nix
```diff
@@ -71,7 +97,7 @@ pkgs.stdenvNoCC.mkDerivation {
     sha256 = shaMap.${system};
   };
 
-  sourceRoot = "{{ .SourceRoot }}";
+  sourceRoot = {{ with .SourceRoot }}"{{ . }}"{{ else }}sourceRootMap.${system}{{ end }};
```
till, line 100: 🎉

till (commented):
Looks good, trying this out on localhost.

till:
Can confirm, it works:

```nix
...
  sourceRootMap = {
...
  };
in
pkgs.stdenvNoCC.mkDerivation {
  pname = "foo";
  version = "1.3.8-next";
  src = fetchurl {
    url = urlMap.${system};
    sha256 = shaMap.${system};
  };

  sourceRoot = sourceRootMap.${system};
```

PR #4563, opened 2024-01-19
refactor(http): remove redundant `nil` check

Juneezee:
From the Go specification [^1]:

> "3. If the map is nil, the number of iterations is 0."

Therefore, it is not required to do a nil check for the map before the loop.

[^1]: https://go.dev/ref/spec#For_range


caarlos0:
thanks!

PR #4575, opened 2024-01-25
fix(config): handle relative git repos

aauren:

If applied this change will allow goreleaser to handle relative remotes when attempting to parse a repo URL from git.

To fix the error that I recently came across while trying to test my goreleaser configuration:

```
% goreleaser check
  • checking                                 path=
  ⨯ configuration is invalid                 error=invalid scm url: .
  ⨯ .goreleaser.yml                                  error=configuration is invalid: invalid scm url: .
  ⨯ command failed                                   error=1 out of 1 configuration file(s) have issues
```

This change happened while on a branch doing some development. As part of that development I needed to test a change to my goreleaser config.

My git config at the time looked like (repo obfuscated):

```
% cat .git/config
[core]
        repositoryformatversion = 0
        filemode = true
        bare = false
        logallrefupdates = true
[remote "origin"]
        url = git@gitlab.com:some/repo
        fetch = +refs/heads/*:refs/remotes/origin/*
[branch "main"]
        remote = origin
        merge = refs/heads/main
[branch "release_fixes"]
        remote = .
        merge = refs/heads/main
```

It is fairly common for git to add remotes with a `.` when branch tracking is enabled.

While, in general, there aren't many use cases that require a user to need to release from a non-primary branch, there are cases where the user may want to test their configuration with `goreleaser check` and the error of `invalid scm url: .` isn't very helpful.

### DIFF
--- a/internal/git/config_test.go
+++ b/internal/git/config_test.go
@@ -2,6 +2,7 @@ package git_test
 
 import (
 	"context"
+	"strings"
 	"testing"
 
 	"github.com/goreleaser/goreleaser/internal/git"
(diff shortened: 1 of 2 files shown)


internal/git/config_test.go
```diff
@@ -22,6 +22,22 @@ func TestNoRemote(t *testing.T) {
...
+	ctx := context.Background()
+	testlib.Mktmp(t)
+	testlib.GitInit(t)
+	testlib.GitRemoteAddWithName(t, "upstream", "git@github.com:goreleaser/goreleaser.git")
+	_, err := git.Run(ctx, "pull", "upstream", "main")
+	require.NoError(t, err)
+	_, err = git.Run(ctx, "branch", "--set-upstream-to", "upstream/main")
+	require.NoError(t, err)
+	_, err = git.Run(ctx, "checkout", "--track", "-b", "relative_branch")
+	require.NoError(t, err)
```
caarlos0, line 35: maybe also assert that `.git/config` is how we want it to be?

caarlos0 (approved):
just one comment

internal/git/config_test.go
aauren: I added some testing right below this.

I also changed the Git URL to be HTTPS so that it would work on pull via the CI.

internal/git/config_test.go
```diff
@@ -22,6 +23,24 @@ func TestNoRemote(t *testing.T) {
...
+	testlib.Mktmp(t)
+	testlib.GitInit(t)
+	testlib.GitRemoteAddWithName(t, "upstream", "https://github.com/goreleaser/goreleaser.git")
+	_, err := git.Run(ctx, "pull", "upstream", "main")
+	require.NoError(t, err)
+	_, err = git.Run(ctx, "branch", "--set-upstream-to", "upstream/main")
+	require.NoError(t, err)
+	_, err = git.Run(ctx, "checkout", "--track", "-b", "relative_branch")
+	require.NoError(t, err)
+	gitCfg, err := git.Run(ctx, "config", "--local", "--list")
```
caarlos0, line 37:
```suggestion
	gitCfg, err := git.Run(ctx, "config", "--local", "--list")
	require.NoError(t, err)
```

caarlos0:
thankss 🙏

PR #4587, opened 2024-01-30
feat(nfpm): support libraries

### LINKED ISSUE #4346: Unable to build rpms for c-shared library
### What happened?

I have a repository with an configuration to build an binary and a c-shared library. nfpm is confiured to provide deb/rpm packages.

https://github.com/jkroepke/openvpn-auth-oauth2/tree/v1.8.0-rc.4

I triggered a release using the configuration

https://github.com/jkroepke/openvpn-auth-oauth2/blob/v1.8.0-rc.4/.goreleaser.yaml

and got the error

>   ⨯ release failed after 46s                 error=no linux binaries found for builds [openvpn-auth-oauth2.so]

https://github.com/jkroepke/openvpn-auth-oauth2/actions/runs/6410470783/job/17403879409

Removing the nfpm configuration for the second build gives no error back

clone https://github.com/jkroepke/openvpn-auth-oauth2
checkout v1.8.0-rc.4
run github action on tag v1.8.0-rc.4 OR run goreleaser release. (build only works on a linux system)

### goreleaser version

```bash
v1.21.2 (from gh action output)
```

### GoReleaser Check

### Search

### Supporter

### Code of Conduct

### Additional context

The nfpm configuration for the binary build works fine.

Mentionable is that in .goreleaser I define the name of the binary (https://github.com/jkroepke/openvpn-auth-oauth2/blob/8e9c0830779dfc644f3309bc7c8b54d7801e62fc/.goreleaser.yaml#L13), but since its an c-shared build, the .so suffix is attach to the file. Maybe thats the root cause

caarlos0:
This adds `nfpm.libdirs` to allow to set where to put libraries built, as well as include them in the search for artifacts when building the package.

closes #4346

---

PS: I'm not sure about the default dirs, let me know what you think!

### DIFF
--- a/internal/pipe/nfpm/nfpm.go
+++ b/internal/pipe/nfpm/nfpm.go
@@ -209,9 +209,6 @@ func create(ctx *context.Context, fpm config.NFPM, format string, artifacts []*a
 		&fpm.Homepage,
 		&fpm.Description,
 		&fpm.Maintainer,
-		&fpm.Libdirs.Header,
-		&fpm.Libdirs.CShared,
-		&fpm.Libdirs.CArchive,
 		&overridden.Scripts.PostInstall,
 		&overridden.Scripts.PreInstall,
 		&overridden.Scripts.PostRemove,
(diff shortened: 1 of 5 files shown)


jkroepke:
> PS: I'm not sure about the default dirs, let me know what you think!

I think `/usr/include` and  `/usr/lib` are safe defaults. Sometimes, we have `/usr/lib64` or `/usr/lib/aarch64-linux-gnu/`, there are too my variants.

Since templating is supported, multi-arch packages should be a problem if one arch has `/usr/lib/aarch64-linux-gnu`, the other one `/usr/lib/x86_64-linux-gnu` ?

caarlos0:
> Since templating is supported, multi-arch packages should be a problem if one arch has `/usr/lib/aarch64-linux-gnu`, the other one `/usr/lib/x86_64-linux-gnu` ?

ah, yes, might need to apply those per-target. will fix it later today, thanks for the feedback 💜

PR #4617, opened 2024-02-09
Fix sort order of brew artifacts

brianstrauch:
Fix the order in which brew artifacts are sorted.

* The order occasionally switches, which results in a larger diff: https://github.com/confluentinc/homebrew-tap/pull/41
* The artifacts are already grouped by OS before `lessFnFor()` is called, so `list[i].OS > list[j].OS` always evaluates to `false` and the order remains unchanged. This PR removes that statement.
* It looks like a `map` is used earlier, while filtering the artifacts, which might explain why the order occasionally switches.
* Update the remaining statement in `lessFnFor()` to actually use `<` as the function suggests.

### DIFF
--- a/internal/pipe/brew/testdata/TestRunPipeForMultipleAmd64Versions/v1.rb.golden
+++ b/internal/pipe/brew/testdata/TestRunPipeForMultipleAmd64Versions/v1.rb.golden
@@ -20,8 +20,8 @@ class V1 < Formula
   end
 
   on_linux do
-    if Hardware::CPU.arm? && Hardware::CPU.is_64_bit?
-      url "https://dummyhost/download/v1.0.1/arm64.tar.gz"
+    if Hardware::CPU.intel?
+      url "https://dummyhost/download/v1.0.1/amd64v2.tar.gz"
       sha256 "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
 
       def install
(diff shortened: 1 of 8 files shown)


internal/pipe/brew/testdata/TestRunPipeForMultipleAmd64Versions/v1.rb.golden
```diff
@@ -20,17 +20,17 @@ class V1 < Formula
   end
 
   on_linux do
-    if Hardware::CPU.arm? && Hardware::CPU.is_64_bit?
-      url "https://dummyhost/download/v1.0.1/arm64.tar.gz"
+    if Hardware::CPU.intel?
```
brianstrauch, line 23: amd64v2 < arm64

internal/pipe/brew/testdata/TestRunPipeForMultipleArmVersions/multiple_armv5.rb.golden
```diff
@@ -30,16 +30,16 @@ class MultipleArmv5 < Formula
   end
 
   on_linux do
-    if Hardware::CPU.arm? && Hardware::CPU.is_64_bit?
-      url "https://dummyhost/download/v1.0.1/arm64.tar.gz"
+    if Hardware::CPU.arm? && !Hardware::CPU.is_64_bit?
```
brianstrauch, line 33: armv5 < arm64

brianstrauch:
`ruleguard` and `test` failures look unrelated to this PR

PR #4626, opened 2024-02-12
feat(pipe/release): Mark GitHub releases as non-draft only after all artifacts are uploaded.

nichtverstehen:
Previously end-users would see missing artifacts if trying to use latest version while artifacts are being uploaded.

This currently applies only to GitHub releases. GitLab does not support drafts, and I don't dare to make the change for Gitea since I don't use it (and can't test easily).

### DIFF
--- a/www/docs/customization/release.md
+++ b/www/docs/customization/release.md
@@ -290,10 +290,6 @@ release:
   # You can disable this pipe in order to not upload any artifacts.
   disable: true
 
-  # If set to true, will not auto-publish the release.
-  # Available only for GitHub and Gitea.
-  draft: true
-
   # What to do with the release notes in case there the release already exists.
   #
   # Valid options are:
--- a/www/docs/customization/release.md
+++ b/www/docs/customization/release.md
@@ -26,6 +26,7 @@ release:
     - bar
 
   # If set to true, will not auto-publish the release.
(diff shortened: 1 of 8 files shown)


nichtverstehen:
Hi! Any concerns with this change?

www/docs/customization/release.md
```diff
@@ -289,6 +290,10 @@ release:
   # You can disable this pipe in order to not upload any artifacts.
   disable: true
 
+  # If set to true, will not auto-publish the release.
+  # Available only for GitHub and Gitea.
+  draft: true
+
```
caarlos0, line 296: this prop is already documented above

caarlos0, line 296:
```suggestion
```

caarlos0 (commented):
one small comment, lgtm otherwise

thanks for the pr

www/docs/customization/release.md
nichtverstehen: The duplicate piece above was under GitHub section, but perhaps repeating it for Gitea explicitly is not worth it. Applied your suggestion!

caarlos0 (approved):

caarlos0:
refs https://github.com/goreleaser/goreleaser/pull/4744

caarlos0:
refs https://github.com/goreleaser/goreleaser/pull/4745

PR #4638, opened 2024-02-18
docs: bump setup-go version

ddymko:
This commit updates the GitHub Actions GoReleaser example by upgrading setup-go from version 4 to version 5. Additionally, it includes cleanup of the tables to address linting issues that flagged them as "incorrect".

I updated the setup-go version in the GitHub Actions GoReleaser example because right after I committed the exampled dependabot flagged the old version as outdated and suggested an update. I figured it'd be simple to just bump up the version to keep things current.

...


caarlos0:
thanks

PR #4641, opened 2024-02-21
wip: checksums

### LINKED ISSUE #4618: Separate checksum file for each release artifact

The checksum feature currently lumps all the checksums for a release into a single `checksums.txt` file. If we're using automation to download and verify our release binaries, some tools may not be able to parse the `checksums.txt` file correctly.

For our use-case, we're using gorelease to build and release a Vault plugin. To register the vault plugin, we need to pass the binary and sha256 has to vault: https://developer.hashicorp.com/vault/docs/plugins/plugin-architecture#plugin-catalog Because the hash is store in `checksums.txt`, rather than in an individual checksum file per file, it is a bit hard to automate.

I would like to be able to produce checksums for each released asset individually in self-contained files.
For example, if my release has 2 artifacts, I'd like to have 2 checksums:
- `my_plugin.exe`
- `my_plugin.exe.sha256`
- `my_plugin_linux`
- `my_plugin_linux.sha256`

None

### Search

### Supporter

### Code of Conduct

### Additional context

_No response_

caarlos0:
This fundamentally changes the way checksums work.

Historically, we would checksum everything when the pipe ran.
Later on, we added the ability to refresh the checksum at any given time, which worked fine for cases in which a file could be modified after it was first checksummed.

In #4618, @F21 requested an interesting (and IMHO valid) new feature: one checksum file per artifact, instead of a single `checksums.txt`.

That feature was simply not possible to implement with the old implementation of checksums.

---

This PR does a couple of things:

- [ ] removes `artifacts.Checksum`
- [ ] removes the artifacts refresh feature
- [x] adds an `artifacts.Checksummer` interface
- [x] adds an `artifacts.ChecksummingArtifacts`, with a new `List` method that will in fact eval the checksums every time it's called 
- [x] `pipe/checksum.Run` now sets the checksummer impl into `ctx.Artifacts`
- [x] update docs
- [x] adds more tests 
- [ ] updates usage of `artifacts.Checksum`

I'm not sure yet if this is the best approach, but after some tries with different ideas, this is def the best one I got so far.

Happy to hear any feedback you might have 🙏

---

closes #4618


caarlos0:
not really liking this implementation either 🤔

PR #4675, opened 2024-03-08
feat: apply template to gobinary

nekohasekai:
Useful for people who want to customize go but don't want to use golang.org/dl wrappers


nekohasekai:
I see I'm a duplicate of #4454 but his code doesn't work, I'll open a new PR.

PR #4685, opened 2024-03-12
fix: template is not actually applied to gobinary

nekohasekai:


caarlos0:
hey, went with a different approach in c9068b4

thanks for the PR and for looking into it 🙏

PR #4691, opened 2024-03-13
fix: do not log `release published` if it is a draft

nekohasekai:
it's a bit scary

### DIFF
--- a/internal/pipe/release/release.go
+++ b/internal/pipe/release/release.go
@@ -106,9 +106,7 @@ func (Pipe) Publish(ctx *context.Context) error {
 	if err := doPublish(ctx, c); err != nil {
 		return err
 	}
-	if !ctx.Config.Release.Draft {
-		log.WithField("url", ctx.ReleaseURL).Info("published")
-	}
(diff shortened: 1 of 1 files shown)


caarlos0:
hey, thanks for the PR
I think this was already fixed in https://github.com/goreleaser/goreleaser/pull/4626 (not released yet)

caarlos0:
oh, disregard my previous comment, that PR doesn't change this

internal/pipe/release/release.go
```diff
@@ -106,7 +106,9 @@ func (Pipe) Publish(ctx *context.Context) error {
 	if err := doPublish(ctx, c); err != nil {
 		return err
 	}
-	log.WithField("url", ctx.ReleaseURL).Info("published")
+	if !ctx.Config.Release.Draft {
+		log.WithField("url", ctx.ReleaseURL).Info("published")
+	}
```
caarlos0, line 111: maybe instead do something like this:
```suggestion
	log.WithField("url", ctx.ReleaseURL).WithField("published", !ctx.Config.Release.Draft).Info("release created/updated")
```

caarlos0:
thanks!

Issue #4664, opened 2024-03-05
Support a generic path, rather than just github's owner/name repository naming scheme


bradleyjames:
### Is your feature request related to a problem? Please describe.

Not all git hosting platforms support the owner/name scheme.  This is naming scheme is github only.  goreleaser shouldn't care what the naming scheme is, only that it can build and access the path of the project.  Specifically, gitlab allows for a repo path that has multiple "/"s.

This request is to add a `path` or some other variable, when referring to a repository rather than expecting it to be defined as owner/name. I'm specifically referring to the homebrew repo definition that expects owner/name when using gitlab.

### Describe the solution you'd like

Provide a solution that isn't github specific and is intuitive to the user.

### Describe alternatives you've considered

Provide a solution that isn't github specific and is intuitive to the user.

### Search

- [X] I did search for other open and closed issues before opening this

### Supporter

- [ ] I am a [sponsor](https://github.com/sponsors/caarlos0/) or a [Pro](https://goreleaser.com/pro) customer

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

caarlos0:
I think you can put the owner as the gitlab owner, and name as the rest of the repo path, and it'll work

bradleyjames:
I ran into an issue last night where that didn't appear to be the case.  I stumbled upon this https://docs.gitlab.com/ee/api/rest/#namespaced-path-encoding which seems to state the project namespace should be URL encoded.  I was surprised when I found it but it was the result of trying to debug something that seemed simple.  The current goreleaser design also prevents using just the project ID as the repository path.  In my case, using the project ID, resolved the issue but I can't use that in goreleaser because of the current design.  I get that may be an issue on gitlab's side but I'm currently prevented from even trying it in goreleaser.

I also get this is probably a small thing to most.  I can't volunteer to make a PR at the moment but am curious if open to the idea.

bradleyjames:
And here's some proof.  Unfortunately the projects are private for the company I work for and thus not accessible by anyone outside our company.

## URL unencoded path
 curl -i --header "PRIVATE-TOKEN: $GITLAB_TOKEN" https://gitlab.com/api/v4/projects/shipium-corp/engineering/homebrew-shipium
HTTP/2 404 

## URL encoded path
curl -i --header "PRIVATE-TOKEN: $GITLAB_TOKEN" https://gitlab.com/api/v4/projects/shipium-corp%2Fengineering%2Fhomebrew-shipium
[...]

PR #4705, opened 2024-03-17
feat: support projectID for gitlab CreateFile (brew, nix, etc)

bradleyjames:
Allow for using a projectID for a gitlab homebrew tap destination.  There are still some issues here because I can't use our full project name but one step at a time.

According to the goreleaser [gitlab doc](https://goreleaser.com/customization/release/#gitlab), gitlab's project IDs are supported.  Before this change that isn't true for homebrew formulas.

Some conversation here: https://github.com/goreleaser/goreleaser/issues/4664#issuecomment-1986056251.


caarlos0:
thank you!

PR #4706, opened 2024-03-17
feat: support netbsd/arm64

maxbrunet:
Add support for `goos=netbsd`,`goarch=arm64` build target

Supported since golang/go@d23cba683e59f6092c1e3f676c8e83e7973f27dd. Like `freebsd/arm64`, it is not documented in the [official list](https://go.dev/doc/install/source#environment)


caarlos0:
thanks!

PR #4718, opened 2024-03-21
Allow the use of absolute paths in release.extra_files globs

smlx:
This change allows the `release.extra_files` globs to reference absolute paths.

I want to make this change because I am using the Github [sbom-generator-action](https://github.com/advanced-security/sbom-generator-action) to download the Github SBOM and attach it to the Github release. I'm saving the downloaded SBOM to a location outside the git checkout so that `goreleaser` doesn't fail due to a dirty git state.

As per [this issue](https://github.com/goreleaser/fileglob/issues/36), the `MaybeRootFS` option is needed in the call to `Glob()` for absolute paths, otherwise the prefix defaults to `./`, and so `Glob()` tries to match `.//my/absolute/path` and fails with an error.

### DIFF
--- a/internal/extrafiles/extra_files.go
+++ b/internal/extrafiles/extra_files.go
@@ -25,7 +25,7 @@ func Find(ctx *context.Context, files []config.ExtraFile) (map[string]string, er
 			log.Warn("ignoring empty glob")
 			continue
 		}
-		files, err := fileglob.Glob(glob)
+		files, err := fileglob.Glob(glob, fileglob.MaybeRootFS)
 		if err != nil {
 			return result, fmt.Errorf("globbing failed for pattern %s: %w", extra.Glob, err)
 		}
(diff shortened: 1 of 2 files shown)


caarlos0:
I think its best to add a `cp /dir ./dir` in your before hooks... globbing outside the current dir can be a breaking change (if someone is using `/dir-relative-to-cwd` instead of `./dir-relative-to-cwd`).

caarlos0:
okay, after thinking more about it, sorry, not merging this.

the full workaround is:

In either a before hook in goreleaser.yml or something else, copy the files you need somewhere within the CWD. Need to make sure that dir is gitignored and I probably best of not putting it inside dist either.

---

FWIW: goreleaser can also generate sboms: https://goreleaser.com/customization/sbom/

smlx:
Okay, thanks for the response.

Yes, my current workaround is to put the SBOM into a `.gitignore` location temporarily during the build.

FWIW I'm using the Github SBOM because Syft SBOMs generated by goreleaser miss a bunch of transitive dependencies compared to Github.

PR #4725, opened 2024-03-25
feat(git): retry git clone on retriable error

Zebradil:
This PR adds retry logic to the process of cloning a git repository.
Currently, it retries only if the output of the git clone command contains the string `Connection reset`.
Probably, there are more cases where retry is reasonable, but I'm not sure what they are.

The number of retries is hardcoded to 10 with increasing delay between retries — in the same way as it is done in #4265, which served me as an example.

The initial use case is described in #4724.


caarlos0:
thanks!

Zebradil:
Awesome, thanks!

PR #4729, opened 2024-03-29
fix: support dir in gomod

caarlos0:
closes https://github.com/orgs/goreleaser/discussions/4728


mbamber (approved):
Can confirm this works

mbamber:
@caarlos0 - I appreciate you're likely very busy, but I'd be grateful if you could give an estimate on when you are planning to release 1.25.0, just so we can plan around that. Thanks in advance!

caarlos0:
Released it ~8h ago 😀

https://goreleaser.com/blog/goreleaser-v1.25/

PR #4739, opened 2024-04-01
fix `run` not being truly POSIX-compliant

bartekpacia:
I'm sorry, but my previous PR #4736 contained a mistake and the script wasn't truly POSIX compliant.

With this PR I fix that mistake.

See relevant shellcheck warning: https://www.shellcheck.net/wiki/SC3010


caarlos0:
ahh, good catch! thanks!

PR #4750, opened 2024-04-04
docs: fix deprecation typos

j178:
The Before part  should be "folder"


caarlos0:
ah, yes, thank you! totally missed that

PR #4756, opened 2024-04-04
fix: use correct path variable for logging in check command

hhromic:
This trivial PR fixes a minor cosmetic issue in the `check` command.

Before this fix, GoReleaser would not show the auto-detected config path under check when no args are provided:
```
$ goreleaser check
  • checking                                 path=
  • 1 configuration file(s) validated
  • thanks for using goreleaser!
```

After this check, GoReleaser correctly shows any auto-detected config paths:
```
$ goreleaser check
  • checking                                 path=.goreleaser.yaml
  • 1 configuration file(s) validated
  • thanks for using goreleaser!
```


caarlos0:
thanks! 🙏

PR #4779, opened 2024-04-11
feat(announce): add BlueSky support

jghiloni:
If applied, this commit will allow users to create BlueSky posts (skeets) about their Goreleaser-built projects

Because I wanted to post to BlueSky when projects I work on relating to BlueSky are built!

Example post made during unit testing (requires an account to see): https://bsky.app/profile/jaygles.bsky.social/post/3kpv573c2pc2k


caarlos0:
I wonder if we need to support different PDS from the default... would be simpler without it I think... not sure how common it is to use custom PDSs 🤔

jghiloni:
> I wonder if we need to support different PDS from the default... would be simpler without it I think... not sure how common it is to use custom PDSs 🤔

I don't have the stats on federation yet, I believe it's in an open beta. I can remove it for now, and we'll see how quickly folks ask for it (I suspect the Venn diagram of people who want to announce to Bluesky and the people who self-host PDSs has more overlap than usual 😆 )

I'm also trying to figure out why the grype test is failing -- it's complaining about a version of the `cosign` module that gets brought in, but `go mod why` isn't giving me any info.

caarlos0 (approved):

internal/pipe/bluesky/bluesky.go
```diff
@@ -0,0 +1,112 @@
...
+						},
+					},
+				},
+			},
+		}
+	}
+
+	httpClient := butil.RobustHTTPClient()
+
+	userAgent := fmt.Sprintf("goreleaser/%s", ctx.Version)
```
caarlos0, line 78: fwiw, ctx.version is the versions of the software being released, not goreleaser's 🤔

www/docs/customization/announce/bluesky.md
```diff
@@ -0,0 +1,29 @@
+# Bluesky
+
+To use [Bluesky](https://bsky.app/), you need to create an account, and set the following
+environment variable on your pipeline:
+
+- `BLUESKY_ACCOUNT_PASSWORD`
```
caarlos0, line 6: this is an app password, right?

maybe we could clarify that in the previous statement?

caarlos0 (commented):
looks good, just a couple of comments

caarlos0:
fixed the comments above on main.


thanks again for the PR :D

jghiloni:
Great, sorry I wasn’t able to get to it, my free time tanked in the last few days!

caarlos0:
> Great, sorry I wasn’t able to get to it, my free time tanked in the last few days!

no worries!

PR #4781, opened 2024-04-12
Update SBOM generation to add .json as file type

cpanato:
- add the file type in the end of the sbom generated file 
- fix wrong value attribution in the doc example


caarlos0:
thanks!

PR #4784, opened 2024-04-14
Add mdv2escape for ProjectName in default telegram message template

EpicStep:
This pull request fix case when ProjectName contains any not escaped symbols


caarlos0:
thanks!

Issue #4543, opened 2024-01-15
gitlab: publish a brew formula throws an 400: "You can only create or edit files when you are on a branch" when creating brew formula  


metalmolly:
### What happened?

While running our GitLab CI/CD tag (GitLab Enterprise Edition v16.4.5) pipeline it executes the command: `goreleaser release --clean`. But the publishing step for the homebrew tap formula always fails with the following error:

`⨯ error creating brew formula file             fileName=Formula/formula.rb branch=brew-releases/0.1.0 projectID=xxx/tooling/homebrew-xxx statusCode=400 error=POST https://gitlab.test.de/api/v4/projects/xxx/tooling/homebrew-repository/repository/files/Formula/formula.rb: 400 {message: You can only create or edit files when you are on a branch}`

### How can we reproduce this?

Use the goreleaser configuration below and run the command `goreleaser release --clean`
```
brews:
  - name: formula
    # Repository to push the generated files to.
    repository:
      owner: xxx/tooling
      name: homebrew-repository
      branch: brew-releases/{{ .Version }}
    folder: Formula
    commit_author:
      name: user
      email: "user@email.com"
    homepage: "{{ .GitURL }}"
    description: "Support tool"
    install: |
      bin.install "xxx"
    test: |
      system "#{bin}/xxx --version"

```

### goreleaser version

```bash
____       ____      _
 / ___| ___ |  _ \ ___| | ___  __ _ ___  ___ _ __
| |  _ / _ \| |_) / _ \ |/ _ \/ _` / __|/ _ \ '__|
| |_| | (_) |  _ <  __/ |  __/ (_| \__ \  __/ |
 \____|\___/|_| \_\___|_|\___|\__,_|___/\___|_|
goreleaser: Deliver Go Binaries as fast and easily as possible
https://goreleaser.com

GitVersion:    1.23.0
GitCommit:     3ec68fbf8c3a1b16ca0f69aeccfb93765685b643
GitTreeState:  dirty
BuildDate:     2023-12-26T14:29:46
BuiltBy:       homebrew
GoVersion:     go1.21.5
Compiler:      gc
ModuleSum:     unknown
Platform:      darwin/amd64
```

### GoReleaser Check

- [X] goreleaser check shows no errors

### Search

- [X] I did search for other open and closed issues before opening this

### Supporter

- [ ] I am a [sponsor](https://github.com/sponsors/caarlos0/) or a [Pro](https://goreleaser.com/pro) customer

[...]

PR #4792, opened 2024-04-19
fix(gitlab): Use start_branch for file creation if branch doesn't exist already

baurmatt:
Creating files on a new branch is only possible with the API if `start_branch`
is given otherwise the API returns:

> You can only create or edit files when you are on a branch

Fixes https://github.com/goreleaser/goreleaser/issues/4543


baurmatt:
Sorry, this doesn't work yet because the file existence check might return 404 but that's because the new branch doesn't exist already not the file itself :(

baurmatt:
This turned out to be more complicated/bigger change than I've anticipated in the beginning. The problem was:
* start_branch needs to be handle for update _and_ create
* Non existing branches need to be taken care of when checking if the file already exists
* Hardcoded default branch fallback of `master` complicated the situation even more

internal/client/gitlab.go
```diff
@@ -135,57 +157,75 @@ func (c *gitlabClient) CreateFile(
...
 	if repo.Owner != "" {
 		projectID = repo.Owner + "/" + projectID
 	}
 
-	// Use the project default branch if we can get it...otherwise, just use
-	// 'master'
-	var branch, ref string
+	log.
+		WithField("projectID", projectID).
+		Debug("projectID at brew")
```
caarlos0, line 170:
```suggestion
		Debug("projectID")
```

internal/client/gitlab.go
```diff
@@ -135,57 +157,75 @@ func (c *gitlabClient) CreateFile(
...
+			if err != nil {
+				return err
+			}
+		}
+
+		log.
+			WithField("projectID", projectID).
+			WithField("branch", branch).
+			WithField("branchExists", branchExists).
+			Debug("useing given branch")
```
caarlos0, line 195:
```suggestion
			Debug("using given branch")
```

internal/client/gitlab.go
```diff
@@ -135,57 +157,75 @@ func (c *gitlabClient) CreateFile(
...
+
+	// If the branch doesn't exist, we need to check the default branch
+	// because that's what we use as `start_branch` later if the file needs
+	// to be created.
+	opts := &gitlab.GetFileOptions{}
+	if branchExists {
+		opts = &gitlab.GetFileOptions{Ref: &branch}
+	} else {
+		opts = &gitlab.GetFileOptions{Ref: &defaultBranch}
+	}
```
caarlos0, line 220:
```suggestion
	opts := &gitlab.GetFileOptions{Ref: &defaultBranch}
	if branchExists {
		opts.Ref = &branch
	}
```

internal/client/gitlab.go
```diff
@@ -196,24 +236,32 @@ func (c *gitlabClient) CreateFile(
 	}
 
 	log.
-		WithField("fileName", fileName).
-		WithField("branch", branch).
 		WithField("projectID", projectID).
-		Debug("found already existing brew formula file")
+		WithField("branch", branch).
+		WithField("fileName", fileName).
+		Info("pushing brew formula")
```
caarlos0, line 242:
```suggestion
		Info("pushing file")
```

internal/client/gitlab.go
```diff
@@ -196,24 +236,32 @@ func (c *gitlabClient) CreateFile(
...
 	if res.StatusCode == 404 {
+		// Create a new file because it's not already there
 		log.
-			WithField("fileName", fileName).
-			WithField("ref", ref).
 			WithField("projectID", projectID).
-			Debug("creating brew formula")
+			WithField("branch", branch).
+			WithField("fileName", fileName).
+			Debug("formula file doesn't exist, creating it.")
```
caarlos0, line 250:
```suggestion
			Debug("file doesn't exist, creating it")
```

internal/client/gitlab.go
```diff
@@ -235,45 +283,52 @@ func (c *gitlabClient) CreateFile(
...
 			Debug("created brew formula file")
 		return nil
-	}
+	} else {
+		// Update the existing file
+		log.
+			WithField("fileName", fileName).
+			WithField("branch", branch).
+			WithField("projectID", projectID).
+			Debug("formula file exists, updating it.")
```
caarlos0, line 292:
```suggestion
			Debug("file exists, updating it")
```

internal/client/gitlab.go
```diff
@@ -235,45 +283,52 @@ func (c *gitlabClient) CreateFile(
...
+		if err != nil {
+			log := log.
+				WithField("fileName", fileName).
+				WithField("branch", branch).
+				WithField("projectID", projectID)
+			if res != nil {
+				log = log.WithField("statusCode", res.StatusCode)
+			}
+			log.WithError(err).
+				Error("error updating brew formula file")
```
caarlos0, line 317:
```suggestion
				Error("error updating file")
```

internal/client/gitlab.go
```diff
@@ -235,45 +283,52 @@ func (c *gitlabClient) CreateFile(
...
-	}
-
-	log := log.
-		WithField("fileName", fileName).
-		WithField("branch", branch).
-		WithField("projectID", projectID).
-		WithField("filePath", updateFileInfo.FilePath)
-	if res != nil {
-		log = log.WithField("statusCode", res.StatusCode)
+		log.Debug("updated brew formula file")
```
caarlos0, line 329:
```suggestion
		log.Debug("updated file")
```

internal/client/gitlab.go
```diff
@@ -235,45 +283,52 @@ func (c *gitlabClient) CreateFile(
 			WithField("filePath", fileInfo.FilePath).
 			Debug("created brew formula file")
 		return nil
-	}
+	} else {
```
caarlos0, line 286: don't need this else as previous block ends with return

caarlos0 (commented):
thank you so much for this!

made a couple of small comments, mostly log messages!

mavogel:
Let me try got get a review done early next week here as well :)

baurmatt:
@caarlos0 Thanks! :) I've applied all suggestions!

Looking forward to your review @mavogel! :)

internal/client/gitlab.go
```diff
@@ -98,6 +98,28 @@ func (c *gitlabClient) getDefaultBranch(_ *context.Context, repo Repo) (string,
...
+		log.WithError(err).
+			Error("error verify branch existence")
+		return false, err
+	}
+
+	if res.StatusCode == 404 {
+		return false, nil
+	} else {
+		return true, nil
+	}
```
caarlos0, line 120:
```suggestion
	return res.StatusCode != 404 
```

internal/client/gitlab.go
```diff
@@ -135,57 +157,73 @@ func (c *gitlabClient) CreateFile(
...
+
+	// If the branch doesn't exist, we need to check the default branch
+	// because that's what we use as `start_branch` later if the file needs
+	// to be created.
+	opts := &gitlab.GetFileOptions{Ref: &defaultBranch}
+	if branchExists {
+		opts.Ref = &branch
+	}
+	castedContent := string(content)
```
caarlos0, line 219: probably better to name this stringContents, or something like that.

would also move it closer to where it's used

internal/client/gitlab.go
```diff
@@ -272,8 +325,9 @@ func (c *gitlabClient) CreateFile(
 	if res != nil {
 		log = log.WithField("statusCode", res.StatusCode)
 	}
-	log.Debug("updated brew formula file")
+	log.Debug("updated file")
 	return nil
+
```
caarlos0, line 330:
```suggestion
```

caarlos0 (commented):
second pass, a couple more small things

internal/client/gitlab.go
baurmatt: Are you sure about this one? Because the function has `(bool, error)` as return type. Should this be:

```
return res.StatusCode != 404, nil
```

internal/client/gitlab.go
baurmatt: Done! :) Though it was there before 🙃

internal/client/gitlab.go
caarlos0: yes, forgot to add the `, nil`

internal/client/gitlab.go
```diff
@@ -98,6 +98,24 @@ func (c *gitlabClient) getDefaultBranch(_ *context.Context, repo Repo) (string,
...
+	// Verify if branch exists
+	_, res, err := c.client.Branches.GetBranch(projectID, branch)
+	if err != nil && res.StatusCode != 404 {
+		log.WithError(err).
+			Error("error verify branch existence")
+		return false, err
+	}
+
+	return res.StatusCode != 404
```
caarlos0, line 116:
```suggestion
	return res.StatusCode != 404, nil
```

PR #4813, opened 2024-04-28
fix: system conditionals

willswire:
Fix brews releases to homebrew core

To satisfy tap syntax rules

[view breaking CI job](https://github.com/Homebrew/homebrew-core/actions/runs/8870312909/job/24351959087?pr=170289#step:5:22)

### DIFF
--- a/internal/pipe/brew/templates/macos_packages.rb
+++ b/internal/pipe/brew/templates/macos_packages.rb
@@ -42,10 +42,10 @@
   end
   {{- else }}
   {{- if eq $element.Arch "amd64" }}
-  if Hardware::CPU.intel?
+  on_intel do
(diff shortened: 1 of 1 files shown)


caarlos0:
I'm ok with this change, but please note that it was never the intent of goreleaser-generated formulas to be "core compatible" (for lack of better term).

Happy to merge this though, will just regenerate the golden files first.

Thanks for the PR!

willswire:
Thanks @caarlos0! I'm realizing that about the self-taps being the intended use case. Planning to spin mine up over contributing to HB Core. Is there any desire to include support for that in the future? If so, I'd be happy to start building that out. Just LMK - thank you!

PR #4821, opened 2024-05-01
Fix telegram default message

EpicStep:
Hi, there is another fix of default message. I tried this message, now everything works

### DIFF
--- a/internal/pipe/telegram/telegram.go
+++ b/internal/pipe/telegram/telegram.go
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	defaultMessageTemplate = `{{ mdv2escape .ProjectName }} {{ mdv2escape .Tag }} is out! Check it out at {{ mdv2escape .ReleaseURL }}`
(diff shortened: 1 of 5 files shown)


internal/pipe/telegram/telegram.go
```diff
@@ -6,13 +6,13 @@ import (
...
 	"github.com/caarlos0/log"
-	api "github.com/go-telegram-bot-api/telegram-bot-api"
+	api "github.com/go-telegram-bot-api/telegram-bot-api/v5"
 	"github.com/goreleaser/goreleaser/internal/tmpl"
 	"github.com/goreleaser/goreleaser/pkg/context"
 )
 
 const (
-	defaultMessageTemplate = `{{ mdv2escape .ProjectName }} {{ mdv2escape .Tag }} is out! Check it out at {{ mdv2escape .ReleaseURL }}`
+	defaultMessageTemplate = `{{ mdv2escape .ProjectName }} {{ mdv2escape .Tag }} is out\! Check it out at {{ mdv2escape .ReleaseURL }}`
```
caarlos0, line 15: can we handle this in the `mdv2escape` function instead?

EpicStep: Yes, we can

caarlos0:
awesome, thank you so much! 🙏

PR #4832, opened 2024-05-03
blob: permit skipping the configuration of the Content-Disposition header

mvisonneau:
This change would allow users to disable the `Content-Disposition` header that is set for blob storage operations. The application will continue to set a default value for `Content-Disposition` of `attachment; filename={{.Filename}}` if no value was provided by the user. However, with this change, users can now specifically disable this header by setting the value to "-" in the configuration.

We feel this would be a nice solution for this issue: https://github.com/Homebrew/brew/issues/15604


caarlos0:
this seems like a good compromise, thanks!

PR #4839, opened 2024-05-06
Update USERS.md

xcloudscript:
Adopter.

...

...

...


xcloudscript:
Thanks, the project is very nice. I'm using a lot with Gitlab, but Kos dependency needs docker dind for publish images. It took us a while to get it working here.

caarlos0:
ah yes, on gitlab ci you'll probably need it indeed

thanks!

PR #4850, opened 2024-05-09
feat(checksum): supports BLAKE2 and SHA-3

sorairolake:
If applied, these commits will allow users to use BLAKE2 (BLAKE2b-512 and BLAKE2s-256) and SHA-3 (SHA3-{224,256,384,512}) as checksum algorithms.

This is because I think it would be useful if these algorithms could be used as an alternative to SHA-1 and SHA-2. These algorithms are standardized as [RFC 7693](https://datatracker.ietf.org/doc/html/rfc7693) (BLAKE2) and [FIPS PUB 202](https://www.nist.gov/publications/sha-3-standard-permutation-based-hash-and-extendable-output-functions) (SHA-3).

- <https://en.wikipedia.org/wiki/BLAKE_(hash_function)#BLAKE2>
- <https://en.wikipedia.org/wiki/SHA-3>


caarlos0:
this is amazing! Thank you!

PR #4858, opened 2024-05-13
feat(chocolatey): allow template in copyright, provide ctx.ReleaseNotes

bradenhilton:
This PR allows `chocolateys.copyright` to be templated, and provides `ctx.ReleaseNotes` as `.Changelog`.

Allowing templates for the copyright will enable, for example:

```yaml
chocolateys:
- ...
  copyright: Copyright (c) 1970-{{ .Now.Format "2006" }} Foo Bar
  ...
```

Without this change:

```xml
<?xml version="1.0" encoding="utf-8"?>
<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
  <metadata>
    ...
    <copyright>Copyright (c) 1970-{{ .Now.Format &#34;2006&#34; }} Foo Bar</copyright>
    ...
  </metadata>
  ...
</package>
```

With this change:

```xml
<?xml version="1.0" encoding="utf-8"?>
<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
  <metadata>
    ...
    <copyright>Copyright (c) 1970-2024 Foo Bar</copyright>
    ...
  </metadata>
  ...
</package>
```

---

Providing the release notes means they can be placed directly in the package page instead of linking to them, for example:

```yaml
chocolateys:
- ...
  release_notes: '{{ .Changelog }}'
  ...
```

Without this change:

```console
❯ goreleaser release --skip=sign,announce,publish --clean
  • starting release...
  ...
  • chocolatey packages
    ⨯ release failed after 4s                  error=template: failed to apply "{{ .Changelog }}": template: failed to apply "{{ .Changelog }}": map has no entry for key "Changelog"
```

With this change:

```xml
<?xml version="1.0" encoding="utf-8"?>
<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
  <metadata>
    ...
    <releaseNotes>## Changelog&#xA;### Other&#xA;* cfa5cbfa5 test&#xA;</releaseNotes>
    ...
  </metadata>
  ...
</package>
```

---

These are already supported for WinGet.

There is probably a better way of doing this, and I've probably also missed something, so feedback is welcome!

### DIFF
--- a/www/docs/customization/chocolatey.md
+++ b/www/docs/customization/chocolatey.md
@@ -70,6 +70,8 @@ chocolateys:
     icon_url: "https://rawcdn.githack.com/foo/bar/efbdc760-395b-43f1-bf69-ba25c374d473/icon.png"
 
     # Your app's copyright details.
+    #
+    # Templates: allowed (since v2.1)
     copyright: 2022 Drummer Roll Inc
 
     # App's license information url.
--- a/internal/pipe/chocolatey/chocolatey.go
+++ b/internal/pipe/chocolatey/chocolatey.go
@@ -219,7 +219,10 @@ func doPush(ctx *context.Context, art *artifact.Artifact) error {
 func buildNuspec(ctx *context.Context, choco config.Chocolatey) ([]byte, error) {
 	tpl := tmpl.New(ctx)
 
-	if err := tpl.ApplyAll(
+	if err := tpl.WithExtraFields(tmpl.Fields{
+		"Changelog": ctx.ReleaseNotes,
+	}).ApplyAll(
+		&choco.Copyright,
 		&choco.Summary,
 		&choco.Description,
 		&choco.ReleaseNotes,


caarlos0:
hey! thanks for the PR!

for full transparency: I plan to release v2 as 1.26 minus the deprecations, so this will probably land in v2.1 only.

I'm aiming to release v2 either this weekend or the next, depending if I get any bug reports during this time

www/docs/customization/chocolatey.md
```diff
@@ -70,6 +70,7 @@ chocolateys:
     icon_url: "https://rawcdn.githack.com/foo/bar/efbdc760-395b-43f1-bf69-ba25c374d473/icon.png"
 
     # Your app's copyright details.
+    # Templates: allowed
```
caarlos0, line 73:
```suggestion
    #
    # Templates: allowed (since v2.1)
```

PR #4870, opened 2024-05-16
feat(archive): support `.tzst` suffix

sorairolake:
See <https://lists.gnu.org/archive/html/info-gnu/2019-01/msg00001.html>:

> When '-a' option is in effect, zstd compression is selected if the
> destination archive name ends in '.zst' or '.tzst'.

### DIFF
--- a/www/docs/customization/archive.md
+++ b/www/docs/customization/archive.md
@@ -29,8 +29,8 @@ archives:
     # - `tgz`
     # - `tar.xz`
     # - `txz`
-    # - `tar.zst` (Since v1.26)
-    # - `tzst`
+    # - `tar.zst` (since v1.26)
+    # - `tzst` (since v2.1)
     # - `tar`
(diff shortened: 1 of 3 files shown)


caarlos0:
hey! thanks for the PR!

for full transparency: I plan to release v2 as 1.26 minus the deprecations, so this will probably land in v2.1 only.

I'm aiming to release v2 either this weekend or the next, depending if I get any bug reports during this time

www/docs/customization/archive.md
```diff
@@ -30,6 +30,7 @@ archives:
     # - `tar.xz`
     # - `txz`
     # - `tar.zst` (Since v1.26)
+    # - `tzst`
```
caarlos0, line 33:
```suggestion
    # - `tar.zst` (since v1.26)
    # - `tzst` (since v2.1)
```

PR #4875, opened 2024-05-17
Update winget.md path docs

lionello:
Update docs for `winget` to show the default past contains the name as well.

...

Current doc (and sample value) don't contain the name and this doesn't match the implementation.

...

https://github.com/goreleaser/goreleaser/blob/a82491afe446295c8756a2c5f46950a11d717aa0/internal/pipe/winget/winget.go#L165
...


caarlos0:
good catch! thank you!

PR #4878, opened 2024-05-18
fix(docs): version typo

nullswan:
According to https://github.com/goreleaser/goreleaser-action?tab=readme-ov-file#workflow, there is a typo `varsion` -> `version`


caarlos0:
ouch, thanks!

Issue #4920, opened 2024-06-07
When `blobs: ids: []` is set, no artifacts should be uploaded


brianstrauch:
### What happened?

I'm attempting to upload a single text file to S3 with the following configuration:
```
blobs:
  - acl: public-read
    bucket: <bucket>
    ids: []
    extra_files:
      - glob: release-notes.txt
    provider: s3
    region: us-west-2
```

I would expect no artifacts to be uploaded (only release-notes.txt), but instead GoReleaser treats `ids: []` the same as if you were to not specify the `ids` field. This causes all artifacts to be uploaded in addition to release-notes.txt.

### How can we reproduce this?

Our repository is private 😢 but the snippet above should work just fine!

### goreleaser version

```bash
1.25.1-pro
```

### GoReleaser Check

- [X] goreleaser check shows no errors

### Search

- [X] I did search for other open and closed issues before opening this

### Supporter

- [X] I am a [sponsor](https://github.com/sponsors/caarlos0/) or a [Pro](https://goreleaser.com/pro) customer

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

brianstrauch:
https://github.com/goreleaser/goreleaser/blob/88a49c5c06756f2f47fdec74fd05f393c438c6d8/internal/pipe/blob/upload.go#L114
I think this line should say `if conf.IDs != nil` instead!

PR #4921, opened 2024-06-07
Do not upload artifacts if `blobs: ids: []` is set

brianstrauch:
Closes https://github.com/goreleaser/goreleaser/issues/4920

...

...


caarlos0:
Oh good catch, yes I think this fixes it!

thanks!

caarlos0:
although... thinking more about it... this would be a breaking change to someone not declaring any ids and expecting everything to get upload.

I'm gonna revert and add another field, maybe a `meta`, mimicking what we have in other places.

caarlos0:
see #4925

PR #4922, opened 2024-06-08
docs: added the missing deprecation notices deep link to v2 section

cafferata:
Added the missing deprecation notices deep link to v2 section.

> If you keep up with the [deprecation notices][notices], it's likely you don't need to do anything.

_<https://goreleaser.com/blog/goreleaser-v2/#upgrading>_


caarlos0:
thanks!

PR #4925, opened 2024-06-09
feat(blob): allow to upload only extra_files

caarlos0:
Not happy with the option name tbh... happy to hear suggestions :) 

refs #4921
refs https://github.com/goreleaser/goreleaser/issues/4920


brianstrauch (commented):
How about `extra_files_only: true`?

brianstrauch (approved):

PR #4949, opened 2024-06-18
Fixed missing closing quote

fkorotkov:
With the current example build fails with:

> notarize: macos: template: failed to apply "{{ isEnvSet \"MACOS_SIGN_P12 }}": unterminated quoted string


caarlos0:
good catch! thank you!

PR #4953, opened 2024-06-23
docs: improving home page

caarlos0:
tried to refresh the home a little bit, not sure if I like it or not.

let me know what you think :)


jghiloni:
Your alt text on the theme switcher is backward -- it suggests switching to the mode you're already in. otherwise, it looks really nice!

caarlos0:
thank you @jghiloni, just fixed it 💜

PR #4964, opened 2024-06-28
docs: Fix deprecation keyword in `.gitlab-ci.yml`

sue445:
`only` keyword in `.gitlab-ci.yml` is now deprecated.

c.f. https://docs.gitlab.com/ee/ci/yaml/#only--except

So I migrated because GitLab now recommends using `rules:if`

c.f. https://docs.gitlab.com/ee/ci/jobs/job_rules.html#run-jobs-only-in-specific-pipeline-types


caarlos0:
awesome, thank you!

PR #4969, opened 2024-07-02
Daggerize test pipeline

kpenfound:
## What is this?

This daggerizes the lint, test, and build pipelines for Goreleaser.

## Why?

For context, the previous pass at this can be found here https://github.com/goreleaser/goreleaser/pull/4186 . Since that time, the DX for using Dagger has been considerably improved.

The benefit this brings to the Goreleaser project is that the test pipeline can be run locally the same as it is run in CI without requiring contributors to configure additional tools in their developer environments. Additionally, by codifying the test and build execution environments, you no longer need to be concerned with changing or outdated Github Actions runner environments.

## How?

As a contributor, you can simply clone/fork Goreleaser and run:

`dagger functions` to see which commands are available.

To lint local code:

`dagger call --source . lint`

To run tests against local code:

`dagger call --source . test output`

To run tests against local code and get the coverage report:

`dagger call --source . test coverage-report -o ./coverage.txt`

To run tests on the main branch on Github:

`dagger call --source=https://github.com/goreleaser/goreleaser test output`

To run tests against a PR branch on Github:

`dagger call --source=https://github.com/goreleaser/goreleaser#pull/4958/head test output`

To run tests against a PR branch using the dagger pipeline committed to the main branch, without checking out goreleaser:

`dagger -m github.com/goreleaser/goreleaser call --source=https://github.com/goreleaser/goreleaser#pull/4958/head test output`

And so on 😃 

## Also

In addition to the Dagger code, I've updated the build.yml workflow to use the test pipeline and updated CONTRIBUTING.md with the command to run tests with Dagger.
Note that I did not update the Taskfile.yml to avoid breaking anything for contributors comfortable with their existing workflows.

Do you feel that this will benefit the Goreleaser project? Would you like to see the Dagger functions doing more/less?


kpenfound:
Additional question for maintainers: There's a unit test that verifies the git info of the project. This fails on forks because my git org is `kpenfound` instead of `goreleaser`. I could fix this in the test environment in dagger if that is desired!

marcosnils:
@caarlos0 this can be closed in favor of: https://github.com/goreleaser/goreleaser/pull/4981

levlaz:
Hey @caarlos0, when you get a chance could you please approve the workflows to run from this fork so that we can test and make sure the proposed changes will work within GitHub Actions?

levlaz:
Hey @caarlos0 I've updated everything to the latest version of Dagger and this is now ready for your review and feedback. 

I have the pipeline flowing on my own fork here: https://github.com/levlaz/goreleaser/actions/runs/10205880694/job/28237611037?pr=2 but a lot of the failures have to do with the build not running under the `goreleaser` org. 

It would be great to hear your feedback on the overall approach here. We're happy to continue to iterate based on what you think.

kpenfound:
I think this is in a good spot for review @caarlos0 ! The full test suite passes in about a minute locally, aside from the scm_test mentioned in my first comment

Screenshot 2024-08-28 at 4 33 10 PM

kpenfound:
Updated this to use the new feature - context directories. This makes it so that by default, the `source` will be the context where the module is loaded from. So in general, the `--source .:default` flag can be dropped.
Now, I can do:
Test my local checkout: `dagger call test output`
Test the remote main branch: `dagger call test --source github.com/goreleaser/goreleaser\#main output`
Test a remote PR: `dagger call --source https://github.com/goreleaser/goreleaser\#pull/4969/head test output`

caarlos0:
I chatted with @marcosnils on private, and I think we can try this out as following:

- instead of replacing the pipeline, we leave the old one, and add make the dagger version a new pipeline
- we let them both run for a while 
- eventually, if everything works well, we remove the old one

wdyt?

kpenfound:
Sounds great @caarlos0 , I'll make the changes!

caarlos0:
Finally merged this! Thanks for all the work, and sorry for the delay!

PR #4991, opened 2024-07-12
Update blob.md `neq` -> `ne`

tashian:
I was looking at this example and looking at the text/template docs, and I got confused.
Turns out there is no `neq` function available.


caarlos0:
good catch, thanks!

PR #5005, opened 2024-07-17
ko: rework error message for relative path for .go suffix

Dentrax:
...

Even though `^\.?(\.\/[^\/]?.*)?$` was matching with `./cmd/main.go`, I spent ~10m to figure this out why it was complaining. And noticed that, we were returning the same error on different case, which is `HasSuffix` check. Let's clarify the error message a bit for better UX.

...


caarlos0:
ahh, good catch! Thanks!

Issue #4974, opened 2024-07-07
GitLab releases broken


abemedia:
### What happened?

I believe GitLab changed something with their uploads API as all of a sudden the release assets uploaded to GitLab are returning 404 errors.

### How can we reproduce this?

Here is a test repo which demonstrates the issue: https://gitlab.com/abemedia/test_1720080953016023000

### goreleaser version

```bash
GitVersion:    2.0.1
GitCommit:     684c1805864e5f29acc204e34e1770eb74918d15
GitTreeState:  clean
BuildDate:     2024-06-11T01:45:52
BuiltBy:       homebrew
GoVersion:     go1.22.4
Compiler:      gc
ModuleSum:     unknown
Platform:      darwin/amd64
```

### GoReleaser Check

- [X] goreleaser check shows no errors

### Search

- [X] I did search for other open and closed issues before opening this

### Supporter

- [ ] I am a [sponsor](https://github.com/sponsors/caarlos0/) or a [Pro](https://goreleaser.com/pro) customer

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

abemedia:
Just did some digging and turns out that uploads now no longer have a URL like `https://gitlab.com/{user}/{repo}/uploads/{hash}/file.ext` but instead they are like `https://gitlab.com/-/project/{projectID}/uploads/{hash}/file.ext`.

I've added a PR to `go-gitlab` to get the full path from the API instead of having to build a custom one which might break again: https://github.com/xanzy/go-gitlab/pull/1968

Once that is merged I'll push the fix here.

caarlos0:
ah, good find! thank you!

caarlos0:
cool!

do we need to change anything else on goreleaser's side?

abemedia:
Yes, but fairly simple. Happy to create a PR after work today.

elee1766:
does the aur template need to change?

abemedia:
Sorry but I didn't get around to fixing this before going on holiday. Have created the PR now: https://github.com/goreleaser/goreleaser/pull/5035

caarlos0:
no worries :D thanks for the PR <3

PR #5035, opened 2024-07-30
fix(gitlab): release file URL broken

abemedia:
Fix the GitLab upload URL.

Gitlab changed upload URLs from 
`https://gitlab.com/{user}/{repo}/uploads/{hash}/file.ext` to `https://gitlab.com/-/project/{projectID}/uploads/{hash}/file.ext`.

To avoid future issues like this the PR https://github.com/xanzy/go-gitlab/pull/1968 was created upstream which adds the `full_path` field to `go-gitlab` which this PR uses. 

Fixes https://github.com/goreleaser/goreleaser/issues/4974

### DIFF
--- a/internal/client/gitlab.go
+++ b/internal/client/gitlab.go
@@ -523,12 +523,7 @@ func (c *gitlabClient) Upload(
 			return fmt.Errorf("templating GitLab Download URL: %w", err)
 		}
 
-		// search for project details based on projectID
-		projectDetails, _, err := c.client.Projects.GetProject(projectID, nil)
-		if err != nil {
-			return err
-		}
-		linkURL = gitlabBaseURL + "/" + projectDetails.PathWithNamespace + baseLinkURL
+		linkURL = gitlabBaseURL + "/" + projectFile.FullPath
 	}
 
 	log.WithField("file", file.Name()).


caarlos0:
will this still work for old versions of gitlab?

abemedia:
@caarlos0 the `full_path` field was already there in the API response, it just was not in go-gitlab.
I'm not entirely sure in what GitLab version it was added but it is available on the oldest supported version (v15): https://archives.docs.gitlab.com/15.11/ee/api/projects.html#upload-a-file
I also downloaded the v14 docs and the field is available there too. How far back do you want Goreleaser to support? If Gitlab no longer supports a version, should Goreleaser?

caarlos0:
> If Gitlab no longer supports a version, should Goreleaser?

definitely not... I'm leaning towards supporting latest + maybe the previous version

that said, there's no promises made about this anywhere, so we can decide what makes more sense.

caarlos0:
thanks for the PR btw

Reprazent:
It looks like we got caught ourselves by that change 🙈, see https://gitlab.com/gitlab-com/runbooks/-/merge_requests/7690#note_2032388070. @caarlos0 do you have an idea when this will make it into a release? Thanks!

caarlos0:
probably next week, but you can use the nightly builds meanwhile:

- https://github.com/goreleaser/goreleaser/releases/tag/nightly
- https://github.com/goreleaser/goreleaser-pro/releases/tag/nightly

PR #5067, opened 2024-08-11
docs: use "~> v2" in GitHub actions docs

idebeijer:
Simple PR to edit the docs for GitHub actions to use GoReleaser `"~> v2"`, as v2 is now the default for `goreleaser init` and the GoReleaser GitHub action.


caarlos0:
Thanks

PR #4898, opened 2024-05-29
Add support for Platform and RPM.Arch, plus AIX

Dylan-M:
First pass at adding support for AIX rpm files by allowing platform and rpm.arch to be set

If applied, this commit will add the ability to manually specify the platform and rpm.arch fields for nfpm. This brings parity to the product with the experience of standalone nfpm.

Additionally, it brings in support for creating AIX rpm files, for which the 2 fields above are necessary.

This change is being made, due to a desire from my team to add support for AIX for our customers.

Addresses #4853
Based on findings from goreleaser/nfpm#821

PR #5075, opened 2024-08-14
feat(nfpm): better support aix

caarlos0:
building upon the great work done at #4898

closes #4898


internal/pipe/nfpm/nfpm.go
```diff
@@ -338,6 +360,16 @@ func create(ctx *context.Context, fpm config.NFPM, format string, artifacts []*a
...
+	// If platform is manually set, prefer it
+	if len(fpm.Platform) > 0 {
+		infoPlatform = fpm.Platform
+	}
+
+	// If rpmArch is manually set, prefer it
+	if len(overridden.RPM.Arch) > 0 {
+		rpmArch = overridden.RPM.Arch
+	}
+
```
caarlos0, line 372: i wonder how needed these really are

Dylan-M (changes requested):
I've put one inline comment in the doc. Otherwise, this looks good to me. I'm not super familiar with the Golang testing yet. However, the tests all pass.


I presume more can be added later, should we find any issues.

internal/pipe/nfpm/nfpm.go
Dylan-M: With another look at the way overrides are handled, the rpmArch may not be needed. I'm not familiar enough with the codebase to be 100% certain. I've always taken the approach of "better safe than sorry", and remove it later if feedback says it isn't needed.

That said, it was done because of this:
https://github.com/goreleaser/goreleaser/pull/5075/files#diff-6cbd845fb048a5924cf8ea025ab328901f9b3efcb79523861a512472a42d42c1R195

Trying to set "intelligent" defaults for AIX, if someone skips setting the platform and arch.

www/docs/customization/nfpm.md
```diff
@@ -362,6 +368,12 @@ nfpms:
         # Templates: allowed.
         key_file: "{{ .Env.GPG_KEY_PATH }}"
 
+      # Set the arch.
+      #
+      # Default: inferred from binary's GOARCH based on the format.
+      # Since: v1.22.
+      arch: arm64
```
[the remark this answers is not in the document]
caarlos0: the configs in the docs are not meant to be correct, only valid... each option should be thought about in its on merit.. if we aim to make it consistent... it would be a loooooooot of work and a lot more docs I think

Dylan-M (approved):
Based on PR discussion, I believe this is good to go.

internal/pipe/nfpm/nfpm.go
caarlos0: yeah, I think I might remove it.


to clarify, AIX is only ever ppc, right?

caarlos0: ah, gotcha... so I think we can probably remove these two new options (for now, at least), and handle it gracefully

caarlos0: also, Go's seems to only support `aix ppc64` 🤔

Dylan-M: Right, Go only supports recent versions of AIX, which are all ppc64 for goarch, but need to be ppc for rpm arch due to AIX weirdness

internal/pipe/nfpm/nfpm_test.go
```diff
@@ -425,7 +425,7 @@ func TestRunPipe(t *testing.T) {
 		case "android":
 			require.Equal(t, "foo_1.0.0_android_"+arch+"-10-20"+ext, pkg.Name)
 		case "aix":
-			require.Equal(t, "foo_1.0.0_aix_ppc-10-20"+ext, pkg.Name)
+			require.Equal(t, "foo_1.0.0_aix_ppc64-10-20"+ext, pkg.Name)
```
Dylan-M, line 428: This is incorrect now. As I explained in my comment, this needs to be ppc, even though the system is ppc64. The RPM arch **must** be set to ppc for AIX, regardless of the system being ppc64.

Dylan-M (changes requested):
Your current changes won't work. You've removed setting the RPM arch, which absolutely has to be set for AIX because of the ppc64/ppc mismatch between OS arch and RPM arch that is specific to AIX.

internal/pipe/nfpm/nfpm_test.go
caarlos0: the name template used in the test uses goarch, so this is correct. there's another test using `ConventionalFilename` below

Dylan-M: Okay, that makes sense. However, you're setting infoArch to ppc, which also doesn't work in my mind. I separated infoArch and RPMarch on purpose. My understanding is that infoArch needs to remain ppc64 while the rpmArch needs to be ppc.

internal/pipe/nfpm/nfpm.go
```diff
@@ -191,6 +192,30 @@ func create(ctx *context.Context, fpm config.NFPM, format string, artifacts []*a
...
+	// https://developer.ibm.com/articles/au-aix-build-open-source-rpm-packages/
+	// https://developer.ibm.com/articles/configure-yum-on-aix/
+	if infoPlatform == "aix" {
+		if artifacts[0].Goarch != "ppc64" {
+			log.Debugf("skipping aix/%s as its not supported", infoArch)
+			return nil
+		}
+		if format == "rpm" {
+			infoPlatform = "aix7.2"
+			infoArch = "ppc"
```
Dylan-M, line 212: Copying from my comment in the other thread:

Okay, that makes sense. However, you're setting infoArch to ppc, which also doesn't work in my mind. I separated infoArch and RPMarch on purpose. My understanding is that infoArch needs to remain ppc64 while the rpmArch needs to be ppc.

Dylan-M (changes requested):

internal/pipe/nfpm/nfpm.go
Dylan-M: Specifically, this is what it takes in raw NFPM to work:
```
name: "aix-somefile"
arch: "ppc64"
platform: "aix7.2"
version: "1.0.0"
vendor: Company
maintainer: Company Maintainer Email
description: Package Description
homepage: https://some-url.com
rpm:
  arch: ppc
```

Dylan-M: You'll notice Platform is also aix7.2, but that can be 7.3 also. Which is why I made it settable in my initial pass.

caarlos0: it doesn't, its only used to package this one package.

the usage of rpm.arch is more like "i use this same config for every format, but on rpm i want arch to be different", which is not our case

caarlos0:
> Your current changes won't work. You've removed setting the RPM arch, which absolutely has to be set for AIX because of the ppc64/ppc mismatch between OS arch and RPM arch that is specific to AIX.

that's not correct I think

inside nfpm, it will simply copy info.rpm.arch over info.arch, so this will end up doing the same thing.

internal/pipe/nfpm/nfpm.go
caarlos0: yeah got that, its being done here :)

internal/pipe/nfpm/nfpm.go
Dylan-M: Does the arch not need to match the actual binary arch?

What about situations where we're doing multiple OSes with the ppc64 arch, and it only needs to be swapped to ppc for AIX?

caarlos0: > Does the arch not need to match the actual binary arch?

no, the arch set into `nfpm.Info` is used only in the package... if it doesn't match (i.e. if its wrong), you'll get a package that installs but a binary that doesn't execute.

but this is handled here, we only change to `aix/ppc64` to `aix72/ppc` when format is `rpm`. 

> What about situations where we're doing multiple OSes with the ppc64 arch, and it only needs to be swapped to ppc for AIX?

its handled, we create a new `nfpm.Info` for each os/arch we package to... so it should all be fine.

Dylan-M: Okay. Thanks for explaining. I was under the impression the top level arch needed to match the binary, and the rpm level matches whatever weirdness for package names.

Dylan-M: One more question, what if they set their platform to aix7.3, and arch to ppc64. The arch will still need swapped to ppc. That doesn't appear to be done currently.

caarlos0: there's no way to do that inside goreleaser...

the platform for aix binaries will always be `aix`, because that's what the build gives it (the `GOOS` basically)

if we eventually want to allow to customize the platform, that's another problem, but right now its not possible, so this is a non-issue

Dylan-M (approved):

Dylan-M (approved):

Issue #5086, opened 2024-08-17
Generated archives contain a mix of user and root file uids/gids


twpayne:
### What happened?

I noticed that the generated archive contained entries with both `root` and local-user owners.

### How can we reproduce this?

Checkout goreleaser at a recent commit, e.g. 17a955458ca085b010a75127d2a0572ea89d22eb, and run goreleaser on itself:

```console
$ git clone https://github.com/goreleaser/goreleaser.git
...
$ cd goreleaser
$ git checkout 17a955458ca085b010a75127d2a0572ea89d22eb
...
$ go run . release --clean --snapshot --skip=chocolatey,sign,publish,sbom,docker
...
```

Then inspect the contents of `dist/goreleaser_Linux_x86_64.tar.gz`:

```console
$ tar tvzf dist/goreleaser_Linux_x86_64.tar.gz                                  
-rw-rw-r-- twp/twp        1085 2024-08-17 09:19 LICENSE.md
-rw-rw-r-- twp/twp       12982 2024-08-17 09:19 README.md
-rw-rw-r-- twp/twp       11975 2024-08-17 18:10 completions/goreleaser.bash
-rw-rw-r-- twp/twp       10061 2024-08-17 18:10 completions/goreleaser.fish
-rw-rw-r-- twp/twp        7892 2024-08-17 18:10 completions/goreleaser.zsh
-rw-rw-r-- twp/twp        1594 2024-08-17 18:10 manpages/goreleaser.1.gz
-rwxrwxr-x root/root  66003096 2024-08-16 19:49 goreleaser
```

Note that the `goreleaser` entry is owned by `root/root`, whereas all other entries are owned by the local user (`twp/twp` in my case).

### goreleaser version

```bash
17a955458ca085b010a75127d2a0572ea89d22eb
```

### GoReleaser Check

- [X] goreleaser check shows no errors

### Search

- [X] I did search for other open and closed issues before opening this

### Supporter

- [ ] I am a [sponsor](https://github.com/sponsors/caarlos0/) or a [Pro](https://goreleaser.com/pro) customer

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Additional context

Running `go releaser check` warns about goreleaser's own `.goreleaser.yaml`:

```console
$ go run . check
  • checking                                 path=.goreleaser.yaml
  • DEPRECATED:  snapshot.name_template  should not be used anymore, check https://goreleaser.com/deprecations#snapshotnametemplate for more info
[...]

PR #5087, opened 2024-08-17
chore: use snapshot.version_template in own config

twpayne:
If applied, this commit will fix the self-check failure reported in https://github.com/goreleaser/goreleaser/issues/5086#issue-2471620837.

tl;dr goreleaser v2.2 switched from `snapshot.name_template` to `snapshot.version_template`. This PR updates goreleaser's own config to use the new variable.


caarlos0:
yup, thanks, forgot to do that

PR #5099, opened 2024-08-22
chore(deps): upgrade cosign to 2.4.0

suprememoocow:
goreleaser currently uses `cosign` `v2.1.1`, this change switches it to `v2.4.0`.

While there may be other useful updates, I'd like this update to workaround a bug which I'm experiencing: https://github.com/sigstore/cosign/issues/3614#issuecomment-2012521670, and which is solved by upgrading the `cosign` version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ RUN apk add --no-cache bash \
 	tini
 
 # install cosign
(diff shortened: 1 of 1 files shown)


suprememoocow:
@caarlos0 here's a tiny MR to upgrade the `cosign` binary in the Goreleaser docker image. 

As an extra datapoint, I noticed that `goreleaser-cross` already uses `cosign` `v2.4.0` https://github.com/goreleaser/goreleaser-cross/blob/master/.env#L11

Thanks for maintaining this wonderful product!

caarlos0:
thanks!

PR #5103, opened 2024-08-26
internal/builders/golang: ignore empty flags after templating

nicolasparada:
Ignore empty flags after templating is applied for final Go build line.

This caused us some problems since we had an `if` without `else`, resulting in an empty flag, causing the whole build to fail with an obscure error message like:

```
malformed import path "-myflag": leading dash
```


caarlos0:
thanks!

PR #5111, opened 2024-08-30
chore: fix linter errors

twpayne:
If applied, this commit will fix the current linter errors, which are currently causing all Github Actions to fail.

See, for example, the latest `main` run: https://github.com/goreleaser/goreleaser/actions/runs/10607661596/job/29400451251.


caarlos0:
thanks, had the same thing done over here, but forgot to push it 😆

PR #5131, opened 2024-09-11
feat(docker): install upx in docker container for using in the build pipeline

giesan:
I like to use the goreleaser for my private project and would like to have the binary scaled down with the UPX in the GitLab build pipeline.
Therefore, it would be nice if the UPX tool is pre-installed in the Docker container.


caarlos0:
thanks!

giesan:
Thank you for accepting my suggestion.

PR #5159, opened 2024-09-26
fix: remove filetype exe from chocolatey template

robinovitch61:
When submitting a new project to chocolatey using goreleaser, a reviewer told me that a line in the template is incorrect:

image

image

> since that is not a valid parameter to the Install-ChocolateyZipPackage function: https://docs.chocolatey.org/en-us/create/functions/install-chocolateyzippackage/

This PR aims to correct that by removing the line from the template as requested


gep13 (approved):
LGTM!

caarlos0:
thanks

caarlos0:
@gep13 one question: if the file is a MSI, should we set `fileType = 'MSI'`, or is `fileType` not needed in any case?

caarlos0:
awesome, thanks for the explanation! 🙏

PR #5161, opened 2024-09-30
perf(changelog): improve changelog sort performance

Juneezee:
This commit removes the unnecessary slice copy in `sortEntries`, and replaces `sort.Slice` [^1] with the new `slices.SortFunc` [^2].

As recommended by the Go documentation, `slices.SortFunc` is generally faster because it uses generic, whereas `sort.Slice` relies on reflection, which incurs additional allocations

The benchmark result from the newly added `Benchmark_sortEntries` show approximately a 64% performance improvement.

Benchmark result:

```
                     │   old.txt    │               new.txt               │
                     │    sec/op    │   sec/op     vs base                │
_sortEntries/asc-16    16.458µ ± 1%   5.958µ ± 1%  -63.80% (p=0.000 n=10)
_sortEntries/desc-16   17.675µ ± 1%   6.020µ ± 0%  -65.94% (p=0.000 n=10)
geomean                 17.06µ        5.989µ       -64.89%

                     │   old.txt    │               new.txt                │
                     │     B/op     │     B/op      vs base                │
_sortEntries/asc-16    3.164Ki ± 0%   1.164Ki ± 0%  -63.21% (p=0.000 n=10)
_sortEntries/desc-16   3.422Ki ± 0%   1.164Ki ± 0%  -65.98% (p=0.000 n=10)
geomean                3.290Ki        1.164Ki       -64.62%

                     │  old.txt   │              new.txt               │
                     │ allocs/op  │ allocs/op   vs base                │
_sortEntries/asc-16    68.00 ± 0%   25.00 ± 0%  -63.24% (p=0.000 n=10)
_sortEntries/desc-16   72.00 ± 0%   25.00 ± 0%  -65.28% (p=0.000 n=10)
geomean                69.97        25.00       -64.27%
```

[^1]: https://pkg.go.dev/sort#Slice
[^2]: https://pkg.go.dev/slices#SortFunc

### DIFF
--- a/internal/pipe/changelog/changelog.go
+++ b/internal/pipe/changelog/changelog.go
@@ -2,12 +2,13 @@
 package changelog
 
 import (
+	"cmp"
 	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 	"regexp"
-	"sort"
+	"slices"
 	"strings"
 
 	"github.com/caarlos0/log"
@@ -211,7 +212,7 @@ func formatChangelog(ctx *context.Context, entries []string) (string, error) {
 		}
 	}
 
-	sort.Slice(groups, groupSort(groups))
+	slices.SortFunc(groups, groupSort)
 	for _, group := range groups {
 		if len(group.entries) > 0 {
 			result = append(result, group.title)
@@ -221,10 +222,8 @@ func formatChangelog(ctx *context.Context, entries []string) (string, error) {
 	return strings.Join(result, newLineFor(ctx)), nil
 }
 
-func groupSort(groups []changelogGroup) func(i, j int) bool {
-	return func(i, j int) bool {
-		return groups[i].order < groups[j].order
-	}
+func groupSort(i, j changelogGroup) int {
+	return cmp.Compare(i.order, j.order)
 }
 
 func filterAndPrefixItems(ss []string) []string {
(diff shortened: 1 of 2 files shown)


internal/pipe/changelog/changelog.go
```diff
@@ -306,17 +305,16 @@ func sortEntries(ctx *context.Context, entries []string) []string {
 	if direction == "" {
 		return entries
 	}
-	result := make([]string, len(entries))
-	copy(result, entries)
```
Juneezee, line 310: The slice copy here is unnecessary because `sortEntries` is the last function to be called in `buildChangelog`. So it is okay to sort the `entries` slice in-place.

https://github.com/goreleaser/goreleaser/blob/d2469666b8fe4c14bbce4c7a58f557350fd19d64/internal/pipe/changelog/changelog.go#L274-L279

internal/pipe/changelog/changelog_test.go
```diff
@@ -814,12 +840,23 @@ func TestGroup(t *testing.T) {
...
+### Others
+\* \w+ first
+\* \w+ this is not a Merge pull request
+\* \w+ chore: something about cArs we dont need
+\* \w+ docs: whatever
+\* \w+ fix: whatever
+\* \w+ ignored: whatever
+\* \w+ fixed bug 2
+\* \w+ added feature 1
+`, ctx.ReleaseNotes)
```
Juneezee, line 859: I updated this section to make sure the sorting of changelog groups is tested. `require.Contains` does not ensure the order of the changelog groups.

https://github.com/goreleaser/goreleaser/blob/d2469666b8fe4c14bbce4c7a58f557350fd19d64/internal/pipe/changelog/changelog.go#L214-L221

Juneezee (commented):
Self code review

caarlos0:
thanks!

PR #5176, opened 2024-10-06
point to GHCR instead of GCR for cosign container

bobcallaway:
This changes the Dockerfile to pull the cosign container image from GHCR instead of Google Cloud. This helps the Sigstore team manage their cloud spend (as GHCR is provided for free and Google Cloud Artifact Registry is not).

Note the container hash does not change and images are posted to both locations upon cosign's release process.


caarlos0:
thanks!!

PR #5186, opened 2024-10-08
feat(build): add GO386, GOMIPS64, GORISCV64, and GOPPC64 support

caarlos0:
continuing the work of #5153

closes #5153


peterdeme:
FYI, this is a breaking change for anyone who uses the folder structure of the build.

For example, we're uploading the binaries to S3 after goreleaser and this broke our build.

caarlos0:
yes, you can probably rely on `dist/artifacts.json` instead, which will have the full paths regardless of any future changes in them

PR #5196, opened 2024-10-12
ci: add riscv64 in release

mengzhuo:


.goreleaser.yaml
```diff
@@ -39,11 +39,16 @@ builds:
...
+      - riscv64
     goarm:
       - "7"
     ignore:
       - goos: windows
         goarch: arm
+      - goos: dwarin
+        goarch: riscv64
+      - goos: windows
+        goarch: riscv64
```
caarlos0, line 51: i don't think we need to add these lines

caarlos0:
anyone using goreleaser on riscv? how? CI?

mengzhuo:
> anyone using goreleaser on riscv? how? CI?

CI that runs on real RISCV box, https://riscv.builders, requires quick download on binary of goreleaser

PR #5225, opened 2024-10-25
[Hooks] option to ignore hook failure

sks:
- Adds an option to ignore the hook failures
...

- Some of the optional pipeline cleanup steps added in hooks could fail, I would like to ignore those fails 
Example of the steps

```
hooks:
      post:
        - cmd: env GOOS={{ .Os }} GOARCH={{ .Arch }} go clean -cache
        - cmd: env GOOS={{ .Os }} GOARCH={{ .Arch }} go clean -modcache
```


caarlos0:
i think its best to use something like:

```yaml
before:
  hooks:
    - sh -c 'aaaaaa || true'
```

sks:
> ```yaml
> sh -c 'aaaaaa || true'
> ```

Thanks @caarlos0 

Closing the PR

PR #5273, opened 2024-11-13
Add glob support to multiply builds

### LINKED ISSUE #4927: Building Multiple Binaries with a Single Command
see discussion: https://github.com/orgs/goreleaser/discussions/4886

egorse:
Instead of defining 'main' you have to define 'glob', and during configuration load the builds would be multiplied with settings id, main and binary.
Thats allows instead of having few builds definitions just specify one glob (as filepath.Glob accepts) i.e. ./cmd/*/main.go

Not fixes #4927


egorse:
This is just an idea which locally working. Probably its not even close to how maintainers see it implemented. No bad filling

caarlos0:
hey! thanks for the PR and the idea.

Not sure if this is something I want to add in though, I think you can do the same thing with yaml anchors and what not. Also, this might conflict with #4927

All in all, I also think its not that bad to repeat a bit of yaml for the sake of better clarity of what's going on.

In any case, I'll reflect about this one a bit more before I make a decision!

Thanks again for the time and effort - and for using GoReleaser :)

egorse:
@caarlos0 you are right - I am not sure it would conflict with #4927 but for sure will not end up running only one ```go build ...```

Not sure yaml anchors would help either - I guess those has no access to fs to enumerate files like make can do $(shell fine cmd/*/main.go)/$(cmd/*/main.go)

caarlos0:
after giving it more thought, this is not something I have interest in merging. I rather have a little config repetition.

as mentioned before, this would also conflict with #4927, which seems to make more sense to me

that said, closing this. thanks for the suggestion/PR, though!

cheers!

PR #5276, opened 2024-11-14
fix(jsonschema): add custom JSONSchema type to umask

ldez:
Currently, the JSONSchema defines the type of `umask` as an `integer` because the type is inherited from `fs.FileMode` (`uint32`).

https://github.com/goreleaser/goreleaser/blob/a7721dc1fed1168b71a06384d6a8136f9bae31ff/www/docs/static/schema.json#L1830-L1832

https://github.com/goreleaser/goreleaser/blob/a7721dc1fed1168b71a06384d6a8136f9bae31ff/pkg/config/config.go#L948

The documentation says octal values (`0o002`) can be used.

https://github.com/goreleaser/goreleaser/blob/a7721dc1fed1168b71a06384d6a8136f9bae31ff/www/docs/customization/nfpm.md?plain=1#L79-L80

But as this value contains a letter (`o`), so this field becomes a `string` and the JSON validation fails.

Screenshot_20241114_044553

This PR adds a struct tag on the field `Umask` to allow `integer` and `string`.

I generated the JSONSchema by using the following command and I copy-pasted the changes to the pro schema.

```bash
go run . schema -o ./www/docs/static/schema.json
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -945,7 +945,7 @@ type NFPMOverridables struct {
 	Recommends       []string       `yaml:"recommends,omitempty" json:"recommends,omitempty"`
 	Suggests         []string       `yaml:"suggests,omitempty" json:"suggests,omitempty"`
 	Conflicts        []string       `yaml:"conflicts,omitempty" json:"conflicts,omitempty"`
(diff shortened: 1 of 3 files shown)


caarlos0 (approved):
Thank you 🙏

PR #5277, opened 2024-11-14
ci: release build using too much space

caarlos0:
I'm not sure what changed, I guess something in the runner, but anyway, it seems now we're hitting the limits of it.

To work around it, I'm doing the following:

- removing some unused things that take a lot of space and not used (codeql, haskell, android, dotnet)
- prune docker images
- clone only the last 1000 commits + tags instead of the entire history (which is getting kinda big)

Hopefully that is enough. The cleanups alone should increase the free space from ~21GB to ~38GB. I doubt we're using that much.


anotherChowdhury:
Thanks for writing up the blog, just stole your code xD

PR #5279, opened 2024-11-14
fix(aur): missing arm v6 filter

ldez:
The AUR pipe can handle arm v6 but there is a missing filter.

https://github.com/goreleaser/goreleaser/blob/82fd112f7774669e5997a9e7513274686ddd6cdc/internal/pipe/aur/aur.go#L275-L276


ldez:
Sorry I missed #4695

Issue #5282, opened 2024-11-14
Slash in tag not handled correctly for git + homebrew


alapidas:
### What happened?

I'm using goreleaser to publish a homebrew tap.  I am also using monorepo support to use tags of the format `spa-explorer/xxxxxxxx`.  Relevant stanza from `.goreleaser.yaml` below:
```
monorepo:
  tag_prefix: spa-explorer/
  dir: .

brews:
  - name: spa-explorer
    repository:
      owner: <myOrg>
      name: spa-tools
      branch: master
```

It appears to work correctly, but after tapping and trying to download the binary, it fails:

```
at 13:09:36 ➜ brew install spa-explorer
==> Fetching <myOrg>/spa-tools/spa-explorer
==> Downloading https://github.com/<myOrg>/spa-tools/releases/download/spa-explorer/v0.72.0/spa-explorer_Darwin_arm64.tar.gz
curl: (56) The requested URL returned error: 404

Error: spa-explorer: Failed to download resource "spa-explorer"
Download failed: https://github.com/<myOrg>/spa-tools/releases/download/spa-explorer/v0.72.0/spa-explorer_Darwin_arm64.tar.gz
```

If I pop over to my releases page, and look for that URL, I see this:
https://github.com/confluentinc/spa-tools/releases/download/spa-explorer%2Fv0.72.0/spa-explorer_Darwin_arm64.tar.gz

NOTE the `%2f` in place of a slash in the URL.  I _think_ github is URL encoding it because it's not part of the path.  From [this doc](https://docs.github.com/en/rest/using-the-rest-api/troubleshooting-the-rest-api?apiVersion=2022-11-28#404-not-found-for-an-existing-resource):
> Additionally, any path parameters must be URL encoded. For example, any slashes in the parameter value must be replaced with %2F. If you don't properly encode any slashes in the parameter name, the endpoint URL will be misinterpreted.

### How can we reproduce this?

I don't have this set up in a public repo, but if you use monorepo support + a tag with a slash in it, and homebrew integration, I _think_ this would be repro

### goreleaser version

```bash
at 14:22:38 ➜ curl -sfL https://goreleaser.com/static/run | VERSION=v2.4.5-pro bash -s -- --version 
Using Pro distribution...
Downloading GoReleaser v2.4.5-pro...
Verifying checksums...
Verifying signatures...
Verified OK
  ____       ____      _                            ____
[...]

PR #5283, opened 2024-11-15
fix: URL escape tags in github, gitlab, and gitea

caarlos0:
refs #5282


alapidas:
Thanks for the quick turnaround @caarlos0!  When do you anticipate a new point release being available?

caarlos0:
@alapidas probably over the weekend! will trigger a nightly build in the meantime though!

alapidas:
OK thanks, I'll keep an eye out for it to unblock me.

PR #5298, opened 2024-11-19
fix(build): overrides without specifying goarm64 et al

caarlos0:
this will use the default value for GOMIPS, GOARM, GO386, and etc in build overrides if they are not specified.


caarlos0:
refs https://github.com/orgs/goreleaser/discussions/5297

PR #5304, opened 2024-11-22
fix(aur): description with quotes

ldez:
Currently, the field `pkgdesc` is wrapped by simple quotes, but if the description also contains simple quotes it breaks the package.

So it's not possible to publish a package that contains "Let's Encrypt" inside the description.

This PR wraps the description with the right quotes depending on the description content.

If there is a mix of quotes, it replaces, arbitrarily, double quotes with single quotes.

Example with double quotes inside the description: https://gitlab.archlinux.org/archlinux/packaging/packages/cargo-release/-/blob/main/PKGBUILD?ref_type=heads

Example with simple quotes inside the description: https://gitlab.archlinux.org/archlinux/packaging/packages/cargo-modules/-/blob/main/PKGBUILD?ref_type=heads


caarlos0:
good find! thanks!

PR #5307, opened 2024-11-25
refactor(build): preparing to support multiple languages

caarlos0:
This starts laying the foundation for supporting more languages, the first of which will probably be Zig, and then Rust.

I already have a zig prototype working in another branch, just raw dogged it to see if it would work, and since it does, now I'll do it piece by piece but with hopefully slightly better code.

### DIFF
--- a/internal/tmpl/tmpl.go
+++ b/internal/tmpl/tmpl.go
@@ -211,6 +211,19 @@ func buildOptsToFields(opts build.Options) Fields {
 		ext:    opts.Ext,
 		name:   opts.Name,
 		path:   opts.Path,
+
+		// set them all to empty, which should prevent breaking templates.
+		// the .Fields() call will override whichever values are actually
+		// available.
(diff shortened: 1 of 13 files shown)


internal/tmpl/tmpl.go
```diff
@@ -198,21 +206,16 @@ func (t *Template) WithBuildOptions(opts build.Options) *Template {
...
-		arm:     opts.Goarm,
-		arm64:   opts.Goarm64,
-		mips:    opts.Gomips,
-		ppc64:   opts.Goppc64,
-		riscv64: opts.Goriscv64,
+	f := Fields{
+		target: opts.Target.String(),
+		ext:    opts.Ext,
+		name:   opts.Name,
+		path:   opts.Path,
```
caarlos0, line 213: should probably still set all the keys from before, but empty, so it doesn't break anyone... might help template reuse...

PR #5315, opened 2024-11-28
docs: fix I386 variable naming mismatch

mowies:
This PR fixes a small docs issue where the `GO386` has 2 different names, but only one of them is correct.

### DIFF
--- a/www/docs/customization/templates.md
+++ b/www/docs/customization/templates.md
@@ -93,7 +93,7 @@ You should be able to use all its fields on each item:
 - `.Gomips64` (since v2.4)
 - `.Goppc64` (since v2.4)
 - `.Goriscv64` (since v2.4)
-- `.Go386` (since v2.4)
+- `.I386` (since v2.4)
 - `.Target` (Since v2.5)
 - `.Type`
 - `.Extra`


www/docs/customization/templates.md
```diff
@@ -93,7 +93,7 @@ You should be able to use all its fields on each item:
 - `.Gomips64` (since v2.4)
 - `.Goppc64` (since v2.4)
 - `.Goriscv64` (since v2.4)
-- `.Go386` (since v2.4)
```
caarlos0, line 96: i think it should actually be Go386 to match the others, let me see where I fucked it up

caarlos0:
OK, admittedly, this is a little bit confusing, and should probably be fixed at some point...

When you use `.Artifacts`, it'll use `Go/.*/`, as the field names are also named like that in the `artifact.Artifact` struct.

In other places, however, `.I386` is the correct option. It is documented in section above the one you were looking.

We should probably unify this at some point, or make the template engine smart about it, at least.

That said, the documentation is currently correct I believe

mowies:
Hm ok good I guess. But yeah this could use a bit more consistency :)

caarlos0:
see https://github.com/goreleaser/goreleaser/pull/5323

mowies:
i guess this is not needed anymore

PR #5323, opened 2024-11-30
feat: make `.Artifacts` Os/Arch/etc template name be the same as in other places

caarlos0:
this should allow templates in goreleaser pro to do `.Artifacts.*.Os`, much like the other templates were a single artifact is enabled.

On v3 we should probably drop the old names, rename the fields, etc

closes https://github.com/goreleaser/goreleaser/pull/5315


caarlos0:
unsure if this is the right move... gonna let this sit here for a while

PR #5325, opened 2024-12-01
feat(build): rust support

caarlos0:
TODO:

- [x] tests
- [x] docs
- [ ] on future PR: [`cargo publish`](https://doc.rust-lang.org/cargo/reference/publishing.html)
- [ ] more examples in the documentation


www/docs/customization/builds.md
```diff
@@ -477,6 +477,33 @@ GoReleaser:
 - Remove uses of the `time` template function. This function returns a new value
   on every call and is not deterministic.
 
+## Build Rust binaries
+
+<!-- md:version v2.5-unreleased -->
+
+<!-- md:alpha -->
+
+You can now build Zig binaries using `rust zigbuild` and GoReleaser!
```
vedantmgoyal9, line 486:
```suggestion
You can now build Rust binaries using `cargo zigbuild` and GoReleaser!
```

internal/builders/rust/all_targets.txt
vedantmgoyal9, line 1: https://doc.rust-lang.org/rustc/platform-support.html
I don't think GoReleaser should include Tier 3 targets as of now. GoReleaser probably won't be used in very low-level projects. Also, they can be re-added anytime in future if a request comes.

caarlos0: yeah that sounds good

caarlos0:
cc/ @raphamorim

vedantmgoyal9 (changes requested):
1. Use `cargo-xwin` instead of `cargo-zigbuild` for cross-compiling to `*-pc-windows-msvc` targets. A bit confusion here is that what how should goreleaser behave:
    - Just log a simple message and use `cargo-xwin` internally.
    - Throw error and require user to specify windows msvc builds separately in config using `cargo-xwin`.
2. Automatically read config from `(.)rustreleaser.y(a)ml` if present.
3. Documentation-related suggestions:
    - Mention that SDKROOT is needed when cross-compiling to macOS.
    - Zig doesn't come pre-installed on GitHub Actions so people should install it beforehand, and it should be available in `PATH` env. variable.

internal/builders/rust/targets.go
```diff
@@ -0,0 +1,91 @@
...
+	"strings"
+	"sync"
+
+	_ "embed"
+
+	"github.com/goreleaser/goreleaser/v2/internal/tmpl"
+)
+
+// tier 1 and tier 2
+// https://doc.rust-lang.org/rustc/platform-support.html
```
vedantmgoyal9, line 14:
```suggestion
// tier 1 and tier 2
// aarch64-pc-windows-gnullvm is the only tier 3 target added
// https://doc.rust-lang.org/rustc/platform-support.html
```

vedantmgoyal9 (changes requested):

caarlos0:
> * Use `cargo-xwin` instead of `cargo-zigbuild` for cross-compiling to `*-pc-windows-msvc` targets. A bit confusion here is that what how should goreleaser behave:
>   
>   * Just log a simple message and use `cargo-xwin` internally.
>   * Throw error and require user to specify windows msvc builds separately in config using `cargo-xwin`.


I would leave that for the user. I feel cargo-zigbuild is a good enough default for the average case, other cases might need different things. We may add examples of these in the docs though.

> * Automatically read config from `(.)rustreleaser.y(a)ml` if present.

I'm not inclined to do this honestly... its already too hard to get any sort of usage metric out of GitHub search...

> * Documentation-related suggestions:
>   
>   * Mention that SDKROOT is needed when cross-compiling to macOS.
>   * Zig doesn't come pre-installed on GitHub Actions so people should install it beforehand, and it should be available in `PATH` env. variable.

yeah those are good points!

internal/builders/rust/testdata/proj/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "proj"
+version = "0.1.0"
+edition = "2021"
```
raphamorim, line 4: nit and not related to this PR: you may want to track 2024 edition that's coming soon (available at nightly).

raphamorim (approved):
Looks good in the rust part, just a minor concern about cargo workspaces but I think might not be a blocker for this PR.

caarlos0:
> Looks good in the rust part, just a minor concern about cargo workspaces but I think might not be a blocker for this PR.

handling it now!

internal/builders/rust/testdata/proj/Cargo.toml
caarlos0: we can add more tests later, I figured it shouldn't change the implementation we have so far though.

raphamorim: no no, should be okey 👌

PR #5329, opened 2024-12-03
feat: handle healthcheck for builders

caarlos0:
this handles `goreleaser healthcheck` taking into account the builders being used.


vedantmgoyal9 (approved):
i haven't looked into why ci is failing, but rest everything seems to work as expected

PR #5330, opened 2024-12-03
feat: add rustup and zig to packages optional dependencies

caarlos0:
this will affect goreleaser packages themselves.

go was already optional, as people might install it from source, and so are now zig and rust.


vedantmgoyal9 (approved):
:shipit:

PR #5385, opened 2024-12-20
Update documentation links

alexbozhenko:
Update documentation links.
This doc: https://go.dev/doc/install/source#environment
is about compiling go from source.

This doc https://pkg.go.dev/cmd/go#hdr-Environment_variables
is about go tool (including `go build`).

I feel like the latter is more likely to stay up to date. 
E.g. goamd64 is not documented in the former, and is documented in the latter.

Also, add extra links for relevant build flags documentation.

### DIFF
--- a/www/docs/customization/builds.md
+++ b/www/docs/customization/builds.md
@@ -117,6 +117,7 @@ builds:
 
     # GOARM to build for when GOARCH is arm.
     # For more info refer to: https://pkg.go.dev/cmd/go#hdr-Environment_variables
+    # and https://go.dev/wiki/MinimumRequirements#microarchitecture-support
     #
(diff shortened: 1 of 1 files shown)


alexbozhenko:
go help build says:
>> The -asmflags, -gccgoflags, -gcflags, and -ldflags flags accept a
space-separated list of arguments to pass to an underlying tool
during the build.
the (undocumented?) trick from stackoverflow allowed me to figure out which tool flags correspond to:
`go build -gcflags -help` or `go build -ldflags -help`

alexbozhenko:
cc @caarlos0 could you please take a look when you get a chance?

caarlos0:
thanks!

PR #5392, opened 2024-12-24
feat(nfpm): allow to set mtime on packages

caarlos0:
- allow to set `nfpm.mtimes`
- use that value when adding the binaries to the package
- also use it in our own configuration
- adds docs as well
- allows to template `file_info.{owner,group,mtime}`


alexbozhenko:
Thank you!
Maybe add a note here as well:
https://goreleaser.com/customization/builds/#reproducible-builds

alexbozhenko:
FYI:
I did `go install github.com/goreleaser/goreleaserv2@006ef06aa322f34fe0bd8d543650df8020781695` and was able to produce reproducible rpms and debs with `    mtime: "{{ .CommitDate }}"`. Thank you!

alexbozhenko:
@caarlos0 thank you! Could you please cut a release with this?

caarlos0:
Will do eventually this month, in the meantime, could you use the nightly builds?

alexbozhenko:
Ack, thanks again!

PR #5393, opened 2024-12-25
fix(aur_sources): missing new line

ldez:
There is a missing new line between `.Sources` and the next section.


caarlos0:
good catch!

PR #5410, opened 2025-01-03
fix: don't assume username in tests

twpayne:
If the repo is not set then the repo owner is set to the user's name. The tests assume that this username is `goreleaser`, which is not true for most users. This PR removes the assumption that the username is `goreleaser`.

If applied, this commit will help goreleaser's tests pass on other people's machines.

This change being made so I can run goreleaser's tests successfully on my machine, so I can be confident that any failing tests in future PRs that I propose are due to any changes that I made in the PR.

Refs https://github.com/goreleaser/goreleaser/pull/3412.


caarlos0:
good one! thanks

PR #5435, opened 2025-01-10
wip: nodejs support

vedantmgoyal9:
A separate independent project is being developed at https://github.com/vedantmgoyal9/nodejs-sea-creator to ease building Node.js SEA (https://nodejs.org/api/single-executable-applications.html). This project aims to simplify the process, which is currently cumbersome and relies on platform-specific tools like `signtool` (Windows) and `codesign` (macOS). The tool will implement the necessary functionality of these tools in pure Rust 🦀 to ensure Windows and macOS binaries can be built on any platform.


caarlos0:
fwiw the macOS signing we already do natively with anchorre/quill, so I think you can not add it there if you don't want to 

the windows native signing needs work on our side still though

PR #5443, opened 2025-01-13
fix: build --single-target infer target from runtime

caarlos0:
This works with Go already, but not with other languages. This PR adds a best-effort thingy to try to figure it out, based on common target patterns and names.

This should allow users to run `goreleaser build --single-target` with any language supported by goreleaser on most machines!


CosmicJesterX:
they can still start over again. Always make their own workflows also =)

PR #5455, opened 2025-01-15
feat(archive): allow to specify multiple formats

### LINKED ISSUE #5436: Add archive format instead of overriding it

In [editorconfig-checker](https://github.com/editorconfig-checker/editorconfig-checker) we recently started using goreleaser. 

One of the historic manual release process caveats is that we used to only provide .tar.gz archives, but later got asked to provide .zip archives for our windows users, since this obscure distribution has (or at least had at the time) no native support for .tar.gz. Using our previous manual process it was fairly easy to just create another compression step to build an additional archive with zip. This meant, that for the last couple releases we provided .tar.gz archives for all platforms, including windows, plus an additional archive with the windows build as a zip. 

In the process of migrating to goreleaser I opted for the well documented pattern of building an archive defaulting to .tar.gz and overriding the format for `goos: windows`. This of course means, that we no longer generate a .tar.gz for windows. Unfortunately we have integrations into multiple language infrastructures to run our (Javascript, PHP, Python, GitHub Actions) which all will need adjusting. Since I missed this during the first release using goreleaser, our users rightfully complained that the integrations no longer work, since they expected a .tar.gz under windows, which was not present in our releases.

So I resorted to duplicating the archive definition in order to provide both archives types.
But with this we now also have releases in zip format for all non-windows platforms, which use disk space and make the already extensive list of archives more cumbersome for users not knowing exactly what they are looking for.

Ideally I'd like specify for a `format_override` that the override should be generated in addition, not as a replacement. This way one would only need one instance under `archive:` , adding `format_overrides:` for each additional format needed.

Example how a  config doing this could look like (the last line is the example):
```yaml
archives:
  - id: archive-ec
    builds:
      - ec-compat
    format: tar.gz
    name_template: >-
      ec-{{- .Os }}-{{ .Arch }}{{- if .Arm }}v{{ .Arm }}{{ end }}
    format_overrides:
      - goos: windows
        format: zip
        add_format: true
```

If that approach is not suitable, I think an option to restrict archives to specific platforms would be nice option too.

I have not seen an option to restrict archives to a singular platform.
I did consider to add an additional build restricted to the windows platform we would like a zip file for, duplicating the build but reducing the build matrix to the GOOS and GOARCH needed in the zip, then duplicating the archive for only this build, and excluding it from the nfpms and docker steps.

### Search

### Supporter

### Code of Conduct

### Additional context

Our .goreleaser.yaml: https://github.com/editorconfig-checker/editorconfig-checker/blob/main/.goreleaser.yaml
As probably valuable information for reading it: we are currently gearing up to replace the binary name we distribute from `ec` to `editorconfig-checker` in our next major release. So we are currently building the binary twice, once with each name, and are building archives for both names using a different name template.

caarlos0:
Pluralize `format` and allow create multiple archives with the same configuration.

Also allow to override the formats for a GOOS.

Closes #5436

TODO:
- [x] improve docs
- [x] check examples and etc still using singular `format`


klaernie:
Awesome, didn't expect this to materialize so quickly!

I just gave it a try, and it works nicely and behaves exactly as I'd expect.

internal/pipe/archive/archive_test.go
```diff
@@ -808,15 +796,45 @@ func TestDefaultSet(t *testing.T) {
...
+		Archives: []config.Archive{
+			{
+				Formats: []string{"tar.gz", "binary"},
+			},
+		},
+	})
+	require.Error(t, Pipe{}.Default(ctx))
+}
+
+func TestDefaultDeprecatd(t *testing.T) {
```
klaernie, line 814:
```suggestion
func TestDefaultDeprecated(t *testing.T) {
```

klaernie:
Just out of curiosity: why does the `binary` archive format not mix with the other formats? Is it just the different default naming template?

caarlos0:
> Just out of curiosity: why does the `binary` archive format not mix with the other formats? Is it just the different default naming template?

yes, that and the fact that pretty much every other option is irrelevant at that point 🤔

I guess we could downgrade it to a warning probably... playing safe for the time being

klaernie:
> I guess we could downgrade it to a warning probably... playing safe for the time being

I think this would be what I'd expect as a user. One could reduce it to being only emitted, if the name template does not include `{{.Binary}}` - since the user then has definitely considered this problem and solved it in the template.

caarlos0:
to be fair that error only happens when there's not name template set, so I guess we can use the binary name template if any of the formats is binary and that's it :)

PR #5509, opened 2025-01-31
docs(sbom): fix default artifact info

scop:
Default is archive, correct in the options listing below.


www/docs/customization/sbom.md
```diff
@@ -11,13 +11,8 @@ GoReleaser.
 The artifact cataloging step can analyze one or more artifacts generated by
 GoReleaser and output one or more SBOM files into the dist directory.
 
-The default is configured to create an SBOM for each binary produced with
-[Syft](https://github.com/anchore/syft). To enable artifact cataloging just add:
-
-```yaml title=".goreleaser.yaml"
-sboms:
-  - artifacts: archive
-```
```
scop, line 20: I'm not sure if the way to switch to binary is important enough so it needs to be documented via an example here, I think the config options listing below suffices for it.

caarlos0:
thanks!

PR #5516, opened 2025-02-01
fix(build): --output completion usage

scop:
Passing the empty string as an extension is a functional no-op, but causes actual problems: https://github.com/spf13/cobra/pull/2228

Remove marking the flag as taking filename altogether, as there's no need to do that; cobra comments doing so as a mistake: https://github.com/spf13/cobra/blob/41b26ec8bb59dfba580f722201bf371c4f5703dd/completions.go#L387-L390

Leave MarkFlagFilename in comments so it's clear it's not inadvertently left out.


caarlos0:
Thanks!

PR #5535, opened 2025-02-08
fix(init): prevent init from creating config files with deprecated fields

yardenshoham:
I noticed the default example config files were specifying `format` instead of `formats` so I fixed that and added a test to ensure this issue never happens again.

The test checks the default configuration files for deprecated fields.


caarlos0 (approved):
good one! thanks!

PR #5557, opened 2025-02-15
docs: Replace deprecated key kos.repository

trallnag:
The key `kos.repository` has been deprecated in 2.5.

<https://goreleaser.com/deprecations/#kosrepository>

Replaced by `kos.repositories`, which is a list of strings instead of a string.

 But it is still mentioned in this document as a valid example. 

<https://goreleaser.com/customization/ko/>


caarlos0:
thank1

PR #5568, opened 2025-02-18
feat: better support calver

caarlos0:


caarlos0:
this has some extra implications: e.g. `.Minor` will return the month without the prefixed 0 - which is probably what the user would expect.


not sure if this is worth it tbh 

I use calver in one of my repos and have been just trimming the leading 0... it works kind of fine 🤔

PR #5599, opened 2025-02-26
docs: correct redirect link for the `Name Templates` article

pPrecel:
Set the correct redirect link name for the `contains` description for the Functions description in the `Name Templates` article.

The description for the `contains` func should point to the `Contains` func documentation, not `ToLower`.

Article link:
- https://goreleaser.com/customization/templates/#functions


caarlos0:
Good catch! Thanks!

PR #5607, opened 2025-03-01
feat: generate casks instead of formulae

caarlos0:
refs #5594 

hit some walls on this though:

- cant install completions and manpages with it 
- seems more focused at desktop apps? 

so maybe our initial plan is kinda wrong and we might want to have both casks and our hack-ish use of formulae?


caarlos0:
superseded by https://github.com/goreleaser/goreleaser/pull/5780

PR #5643, opened 2025-03-13
fix(winget): better handle publisher name

caarlos0:
- use package identifier for the filename if set 
- properly build the package identifier if not set


caarlos0:
The commands come from the binary names... we can't simply remove the spaces I'm afraid.

vedantmgoyal9:
> The commands come from the binary names... we can't simply remove the spaces I'm afraid.

my bad, i didn't see the previous commit. previously, it used to be `project_name`.

vedantmgoyal9 (approved):

PR #5652, opened 2025-03-15
feat: uv builder

caarlos0:
Initial impl of the UV support.

### Questions/commentary

- Do we need the `sdist` build as well, specially considering we already have the source archive feature - which I suspect we can setup to add `dir.egg-info` and `setup.cfg` and it would be basically the same thing? (the more I look at it, the more I think yes)
- The default binary comes from the build pipe, and is set to the project name - should we use that, or require the user to inform the py3-none-any.whl part in the config?

### TODOs

- [ ] docs


internal/pipe/nfpm/nfpm.go
```diff
@@ -111,6 +117,8 @@ func doRun(ctx *context.Context, fpm config.NFPM) error {
 			artifact.ByType(artifact.Header),
 			artifact.ByType(artifact.CArchive),
 			artifact.ByType(artifact.CShared),
+			artifact.ByType(artifact.PyWheel),
+			artifact.ByType(artifact.PySdist),
```
caarlos0, line 121: TODO: add tests for this

internal/pipe/checksums/checksums.go
```diff
@@ -183,6 +183,8 @@ func buildArtifactList(ctx *context.Context) ([]*artifact.Artifact, error) {
 		artifact.ByType(artifact.UploadableSourceArchive),
 		artifact.ByType(artifact.LinuxPackage),
 		artifact.ByType(artifact.SBOM),
+		artifact.ByType(artifact.PyWheel),
+		artifact.ByType(artifact.PySdist),
```
caarlos0, line 187: TODO: add tests for this

internal/pipe/build/build.go
```diff
@@ -253,6 +255,10 @@ func extFor(target string, build config.BuildDetails) string {
 	// Configure the extensions for shared and static libraries - by default .so and .a respectively -
 	// with overrides for Windows (.dll for shared and .lib for static) and .dylib for macOS.
 	switch build.Buildmode {
+	case "wheel":
+		return ".whl"
+	case "sdist":
+		return ".tar.gz"
```
caarlos0, line 261: TODO: add tests for this

ryancurrah:
If it helps, some time this weekend I can take look at your open questions and add feedback.

caarlos0:
> If it helps, some time this weekend I can take look at your open questions and add feedback.

yes it would help a lot!

Thank you @ryancurrah 🙏🏻

JenspederM:
FWIW, it seems that tests are currently failing because setup-uv is missing in the workflow.

In regards to running in docker, you'd probably do `uv pip install my-module` and then invoke it with `uv run python -m my-module`.

ryancurrah:
- Do we need the sdist build as well, specially considering we already have the source archive feature - which I suspect we can setup to add dir.egg-info and setup.cfg and it would be basically the same thing?
  - Yes, generating and uploading source distributions (sdists) to both PyPI and GitHub Releases is recommended. Sdists allow users to build the package on their own systems when a compatible wheel isn't available, ensuring broader compatibility. Unlike generic source archives, sdists integrate directly with Python's packaging tools.
- The default binary comes from the build pipe, and is set to the project name - should we use that, or require the user to inform the py3-none-any.whl part in the config? (current impl mandates that binary from config is ignored)
  - Seems ok to use project name or define the name in the builds section. A note this change only targets only pure Python packages. Wheel filename structure is `{distribution}-{version}-{python tag}-{abi tag}-{platform tag}.whl`. In the future it may nice to handle compiled Python project wheel names.
- Assuming I want to run it in a Docker container, I know I need to pip install file.whl, but how do I run it afterwards? Might be skill issue, but things like python -m mymod gives me module not found errors...
 - You need to rename `__main__.py` to `example_uv.py` then `python -m example_uv` works. If you want to use `__main__.py` you have to put in a sub-folder named `example_uv`.
- Where inside a deb or rpm package does the .whl and sdist files usually go? Can't find any references for that anywhere - is that even a thing?
  - Yeah it doesn't seem to common to distribute the wheel in a deb or rpm. Though plenty of python3-<name> packages exist they are not wheels but source files. Example: https://packages.ubuntu.com/oracular/all/python3-pandas/filelist. Its almost like the sdist file contents is used. 
- Is the the sdist file at all useful inside a docker image if you already have the wheel?
  - I don't think so unless they want to use that to install package.
- This is not a common implementation of a builder in GoReleaser. GoReleaser builders usually build binaries, whereas .wheel and sdist are more like packages. Maybe we should actually move them out of the build api restraints?
  - Yeah if build imposes a lot of constraints on configuration you may be right, how will creating python packages work without a build section?

caarlos0:
thanks for all the good feedback @ryancurrah 

do you think its too bad to have to do something like this if you need to build both wheel and sdist?

```yaml
builds:
  - id: wheel
    builder: uv
    buildmode: wheel
  - id: sdist
    builder: uv
    buildmode: sdist
```

if you only need wheel, it'll be simply:

```yaml
builds:
  - builder: uv
```


gonna also look into compiled wheels as well

JenspederM:
Hey @caarlos0,

Love what you are doing here. After using goreleaser (for a go project) for the first time a a few weeks ago I've been completely smitten with your work.

I work as a staff engineer at Halfspace.ai and do a lot of python projects, so this feature would be highly useful for me.

Let me know if I can do anything to help - code, docs, or just some sparring. 😊

PR #5654, opened 2025-03-18
fix: ExtraOr

caarlos0:
I'm still not sure about this.

The problem is: In GoReleaser Pro, you can split and merge the build process.

When merging, artifacts.json is loaded, and extra gets loaded as `map[string]any`.

So, specially if for things like `[]string`, the previous implementation would not work, as it will be of type `[]any` instead of `[]string`. 

The `Extra` function handles that by marshaling it back to JSON and then trying to unmarshal it back to the needed type.

So, I just changed the `ExtraOr` function to call it instead.

The only problem with all this is: many places we're using `ExtraOr` should probably properly handle the missing key instead of blindingly ignoring it.

On the other hand, this has never been a problem because the keys are in fact always there.

Another possibly better alternative is to change `map[string]any` to a specific struct - but that would make it so we have to add every field there, which is probably overkill, but on the other hand, better for correctness.


caarlos0:
going with #5685 instead

PR #5664, opened 2025-03-21
fix(winget): make tags lowercase

vedantmgoyal9:


caarlos0 (commented):
Can you add it to a test as well?

Can add to an existing test and run go test -update on the winget pkg

caarlos0:
awesome, thank you!

PR #5683, opened 2025-03-27
feat(winget): update manifest version to 1.10.0

vedantmgoyal9:
Use the latest schema `1.10.0` available on winget-pkgs.

The new schema adds some new fields and is fully backwards compatible with the previous version, but they do not affect goreleaser.


caarlos0:
since it doesnt' really affect much, we could merge it as a `fix`, so we could include it in the v2.8.2 release next week 🙏🏻 

wdyt?

vedantmgoyal9:
@caarlos0 done 👍🏻

PR #5699, opened 2025-04-03
fix: set GOMAXPROCS=1 on each go build

caarlos0:
By default, we'll run goreleaser with `--parallelism=$GOMAXPROCS`, and then run `go build` for each target, which will also have `GOMAXPROCS` set.

This makes `--parallelism` actually not be respected, as it would create `GOMAXPROCS*GOMAXPROCS` threads.

This sets `GOMAXPROCS=1` in the `go build` process to address that.

### DIFF
--- a/internal/builders/golang/build.go
+++ b/internal/builders/golang/build.go
@@ -320,6 +320,7 @@ func (*Builder) Build(ctx *context.Context, build config.Build, options api.Opti
 	if v := os.Getenv("GOCACHEPROG"); v != "" {
 		env = append(env, "GOCACHEPROG="+v)
 	}
+	env = append(env, "GOMAXPROCS=1")
 
 	if len(testEnvs) > 0 {
 		a.Extra["testEnvs"] = testEnvs


internal/builders/golang/build.go
```diff
@@ -320,6 +320,7 @@ func (*Builder) Build(ctx *context.Context, build config.Build, options api.Opti
 	if v := os.Getenv("GOCACHEPROG"); v != "" {
 		env = append(env, "GOCACHEPROG="+v)
 	}
+	env = append(env, "GOMAXPROCS=1")
```
ryancurrah, line 323: This change could slow builds down for any project that has been used to all cores for every build. I'm not sure what the right answer is. What if I want 5 parallel builds but each build use 2 cores? I'm thinking there might need to be a second configuration option for processes per build.

ryancurrah (changes requested):

internal/builders/golang/build.go
ryancurrah: Worded another way: GoReleaser parallelism defines how many tasks run concurrently. However, each individual task can potentially use all available CPU cores unless we explicitly limit its resource usage. If a task is CPU-intensive and not constrained, it may compete with other tasks for processing power, leading to inefficiencies or system overload (Like we're seeing on our nodes). To manage this, we may need additional configurations or tools to control per-task CPU consumption.

ryancurrah: Another example is, uv the python package manager has settings for concurrency https://docs.astral.sh/uv/reference/settings/#compile-bytecode.

I'm wondering if GoReleaser should ignore GOMAXPROCS for itself and instead pass it to all go commands it launches and strictly use parallelism flag,

ryancurrah: The ideal value for `GOMAXPROCS` is likely different between a developer’s laptop and a CI build environment, so hardcoding it in the GoReleaser config might not be practical.

caarlos0: Maybe caller can set an Env and use it in the builds section?

Btw if you have any suggestions I would love to hear them as well 

Btw the current behavior has been the default since v0.0.1 I think 😄

ryancurrah: Could we use env like below in include files? https://goreleaser.com/customization/includes. It will be set in CI but devs would not set it. Would that cause an error locally if its not set?

```yaml
env:
  - GOMAXPROCS ={{ .Env. GOMAXPROCS }}
```

ryancurrah: That absolutely works for me.

caarlos0: awesome, tomorrow I'll document it as well

caarlos0: thanks btw <3

ryancurrah: This is a timely proposal https://github.com/golang/go/issues/73193

caarlos0: nice

caarlos0:
docs pr: https://github.com/goreleaser/goreleaser/pull/5712

Issue #5722, opened 2025-04-15
bad artifact grouping when building zig with multiple abis


caarlos0:
### Discussed in https://github.com/orgs/goreleaser/discussions/5721

we should consider the abi when grouping by platoform

this requires adding the abi as an extra field in the binaries, as well as updating the grouping function

caarlos0:
also applies to `deno`.

caarlos0:
see https://github.com/goreleaser/goreleaser/pull/5723

PR #5723, opened 2025-04-15
fix: group by platform should consider abi if available

caarlos0:
closes https://github.com/goreleaser/goreleaser/issues/5722


internal/pipe/archive/archive.go
```diff
@@ -132,7 +132,7 @@ func (Pipe) Run(ctx *context.Context) error {
 			return fmt.Errorf("invalid archive: %d: %w", i, ErrArchiveDifferentBinaryCount)
 		}
 		for group, artifacts := range artifacts {
-			log.Debugf("group %s has %d binaries", group, len(artifacts))
+			log.Infof("group %s has %d binaries", group, len(artifacts))
```
caarlos0, line 135:
```suggestion
			log.Debugf("group %s has %d binaries", group, len(artifacts))
```

PR #5743, opened 2025-04-26
feat: add checksum template function

skycaptain:
Closes #5742.

When used with the configuration example shown in #5742, Artifactory will not complain about missing checksums anymore:

Screenshot 2025-04-26 at 11 51 46


internal/tmpl/tmpl_test.go
```diff
@@ -341,6 +341,21 @@ func TestFuncMap(t *testing.T) {
...
+	folder := t.TempDir()
+	file := filepath.Join(folder, "subject")
+	require.NoError(t, os.WriteFile(file, []byte("lorem ipsum"), 0o644))
+
+	artifact := artifact.Artifact{
+		Path: file,
+	}
+
+	tp := New(testctx.New()).WithArtifact(&artifact)
+	sum, err := tp.Apply(`{{ checksum .ArtifactPath "sha256" }}`)
```
caarlos0, line 354: I wonder if it would be better to make it more akin to the command line stuff, e.g.

```
{{ sha1 "foo/bar" }}
{{ sha256 "etc" }}
```

what do you think?

skycaptain: You're absolutely right - I had the same idea. I can't remember why I did it differently. anyway, I've applied your suggestion.

caarlos0:
awesome, thank you! 🙏

PR #5759, opened 2025-05-07
Add blog post about packaging and publishing Go binaries on NPM which use GoReleaser as build tool

thetnaingtn:
...

...

...


caarlos0:
thanks!

PR #5775, opened 2025-05-15
fix(init): handle CRLF line endings in gitignore

Ansurfen:
This PR fixes two issues with the `goreleaser init` command when modifying `.gitignore` files:

1. When a `.gitignore` file contains CRLF line endings, the command would incorrectly add duplicate `dist/` entries
2. When a `.gitignore` file doesn't end with a newline, the new entries would be appended on the same line, causing incorrect formatting

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -115,16 +115,26 @@ func newInitCmd() *initCmd {
 	return root
 }
 
+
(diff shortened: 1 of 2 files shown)


caarlos0:
This looks good, thanks!

Gonna wait for CI and merge if good :)

PR #5783, opened 2025-05-23
fix(gitlab): use CI_SERVER_VERSION when available

rbqvq:
In some cases (e.g. security reasons), the CI triggerer may not have permission to access the GitLab Version API (401 Unauthorized), and we should prioritize reading from the CI value

This problem seems to be common in self-hosted instances (but official instances sometimes have problems)


caarlos0 (approved):
looks good, maybe worth adding a test for it as well?

mavogel:
Yes, IMO it also makes sense to use the predefine variables in the pipeline (https://docs.gitlab.com/ci/variables/predefined_variables/).

PR #5793, opened 2025-05-26
feat(docker/v2): initial implementation

### LINKED ISSUE #5733: New dockers configuration section
Currently, our Docker configuration is too repetitive, and the implementation is kinda old as well.

Since then, we added Ko support, which included the Docker engine and etc into our dependencies.

So, in theory, we can probably recreate the Docker stuff, without using the Docker CLI, at least for the manifest creation.

We can also take this time to improve the configuration format to be less repetitive.
Here's how it could look like:

```yaml
dockers:
  - images:
      - "goreleaser/goreleaser"
      - "ghcr.io/goreleaser/goreleaser"
    tags:
      - "v{{ .Version }}"
      - "{{ if not .IsNightly }}latest{{ end }}"
      - "{{ if .IsNightly }}nightly{{ end }}"
    dockerfile: Dockerfile # default
    use: buildx # default
    platforms:
      - linux/amd64
      - linux/arm64
    labels:
      io.artifacthub.package.readme-url: "https://raw.githubusercontent.com/goreleaser/goreleaser/main/README.md"
      io.artifacthub.package.logo-url: "https://goreleaser.com/static/avatar.png"
      io.artifacthub.package.maintainers: '[{"name":"Carlos Alexandro Becker","email":"carlos@carlosbecker.dev"}]'
      io.artifacthub.package.license: "MIT"
      org.opencontainers.image.description: "Release engineering, simplified"
      org.opencontainers.image.created: "{{.Date}}"
      org.opencontainers.image.name: "{{.ProjectName}}"
      org.opencontainers.image.revision: "{{.FullCommit}}"
      org.opencontainers.image.version: "{{.Version}}"
      org.opencontainers.image.source: "{{.GitURL}}"
    build_flag_templates:
      - "--pull" 
    extra_files:
      - scripts/entrypoint.sh
```

This would replace both `dockers` and `docker_manifests` in the current configuration.

This would mean deprecating many options in `dockers`, and the entire `docker_manifests` as well.

caarlos0:
This is the v2 of the docker feature!

- better DX (less yaml repetition and templates)
- simpler implementation 

### So now, what happens:

- if `--snapshot`, it'll build the manifest, but not push it (only useful to test the actual docker image build, not the image itself)
- otherwise, on publish, it'll run a single `docker buildx build` to build and publish the manifest + SBOM 
- `docker_signs` should still work
- `docker_digests` should still work

Discussion for after the merge: https://github.com/orgs/goreleaser/discussions/6005

### TODO from the top of my head:

- [x] ~~pro: split/merge wiring~~[^1]
- [x] tests
- [x] signing
- [x] sbom
- [x] attestations https://github.com/goreleaser/goreleaser/pull/6001

---

closes #5733 

[^1]: Not really needed, will only run in publish, which runs in the merging step.


polarathene:
> * At build time, images are created but stored as `.tar` files. This helps specially if split/merge builds and all that, as you can use the GitHub actions cache (for example). It also helps keeping local docker registry clean

One potential concern might be with larger images? In the AI space, these can be many GB in size, even for runtime images (not just the builder stages).

Are you outputting a single `.tar` per image, instead of individual layers? That could add up quite a bit and be less beneficial to cache. CI runners also have limited disk space, so it might not play well there either, and GH cache last I checked is 10GB, so if there is any control over what to cache you offer less flexibility there.

Instead it may be better to match what the `docker/build-push-action` supports, a configurable `output`? That'd allow for using the GHA cache output, which avoids issues with local cache output (_not sure if that'd apply to tar, possibly_). That said, as cache additionally does it's own compression prior to upload IIRC, that might add too much time to a CI run vs avoiding cache entirely 😅

The output is also only layer cache, it excludes any cache mounts used on RUN instructions. In GHA there is a separate action that can grab that, I'm not too familiar with it, but it might be something else to keep in mind, as that type of cache is often more valuable to persist when used.

Having support for other output types does allow to control what you'd selectively cache/upload, and one output is direct to registry rather than the local registry of the build host.

polarathene:
> * Might be worth exploring doing things on a single pass, but pointing to a local registry, and then copying to production regiestries in the publish phase?

By local registry, I assume you're referring to the default with `docker pull`? (or equivalent for podman) or do you mean a self-hosted registry like Zot?

Some registries like Docker provides as it's default is single platform only. So outputting to that would overwrite any prior platform. There is an experimental containerd backend storage instead which is multi-platform capable, but as that's not standard yet you can't really rely on it.

In GHA, you tend to rely on buildx to build both platforms and publish them, and only load one to the build host local registry if you need to run the image (such as perform some tests).

While others may instead have workflows that build on separate runners, such as using the native x86_64 and aarch64 nodes, avoiding the need for QEMU to emulate a platform (much slower), however that does require the manifest merging approach, unless you only do a build and upload an artifact (or rely on cache) and have another job that pulls in that cache or artifacts to publish with.

That's again something you have flexibility with when using `docker/build-push-action`. Ideally GoReleaser would also allow for supporting different use-cases like that. It's convenient on a single build host, but some projects are far too slow to build ARM64 images via QEMU.

polarathene:
> * Test if we can do all of this with crane instead of shelling out to docker

You may also be interested in oras.

caarlos0:
> > * At build time, images are created but stored as `.tar` files. This helps specially if split/merge builds and all that, as you can use the GitHub actions cache (for example). It also helps keeping local docker registry clean
> 
> One potential concern might be with larger images? In the AI space, these can be many GB in size, even for runtime images (not just the builder stages).
> 
> Are you outputting a single `.tar` per image, instead of individual layers? That could add up quite a bit and be less beneficial to cache. CI runners also have limited disk space, so it might not play well there either, and GH cache last I checked is 10GB, so if there is any control over what to cache you offer less flexibility there.
> 
> Instead it may be better to match what the `docker/build-push-action` supports, a configurable `output`? That'd allow for using the GHA cache output, which avoids issues with local cache output (_not sure if that'd apply to tar, possibly_). That said, as cache additionally does it's own compression prior to upload IIRC, that might add too much time to a CI run vs avoiding cache entirely 😅
> 
> The output is also only layer cache, it excludes any cache mounts used on RUN instructions. In GHA there is a separate action that can grab that, I'm not too familiar with it, but it might be something else to keep in mind, as that type of cache is often more valuable to persist when used.
> 
> Having support for other output types does allow to control what you'd selectively cache/upload, and one output is direct to registry rather than the local registry of the build host.

those are good points, maybe not a good strategy at all 🤔

caarlos0:
> > * Might be worth exploring doing things on a single pass, but pointing to a local registry, and then copying to production regiestries in the publish phase?
> 
> By local registry, I assume you're referring to the default with `docker pull`? (or equivalent for podman) or do you mean a self-hosted registry like Zot?
> 
> Some registries like Docker provides as it's default is single platform only. So outputting to that would overwrite any prior platform. There is an experimental containerd backend storage instead which is multi-platform capable, but as that's not standard yet you can't really rely on it.
> 
> In GHA, you tend to rely on buildx to build both platforms and publish them, and only load one to the build host local registry if you need to run the image (such as perform some tests).
> 
> While others may instead have workflows that build on separate runners, such as using the native x86_64 and aarch64 nodes, avoiding the need for QEMU to emulate a platform (much slower), however that does require the manifest merging approach, unless you only do a build and upload an artifact (or rely on cache) and have another job that pulls in that cache or artifacts to publish with.
> 
> That's again something you have flexibility with when using `docker/build-push-action`. Ideally GoReleaser would also allow for supporting different use-cases like that. It's convenient on a single build host, but some projects are far too slow to build ARM64 images via QEMU.

my idea here was to run a local registry within docker, use the docker build command passing all platforms et al (which creates the manifest and pushes it), and then, on publish, copy that manifest from the local registry to whatever other registries the user wants.

my understanding is that you can't create a manifest without pushing the images, so that would probably work (I think, haven't tried it yet).

polarathene:
> my idea here was to run a local registry within docker, use the docker build command passing all platforms et al (which creates the manifest and pushes it), and then, on publish, copy that manifest from the local registry to whatever other registries the user wants.

That does not clarify what you meant by local registry.
- The one that you have by default from Docker when you run `docker build`? That doesn't support storing more than one platform (unless you tag each uniquely), it will in future once containerd storage backend is used.
- A self-hosted private registry like Zot. I don't think this is what you meant as such a choice like that should be optional. `docker buildx build` with a container driver can build multi-platform images and push to public registries without needing any local registry (only required if you want to run the image locally).

---

> my understanding is that you can't create a manifest without pushing the images, so that would probably work (I think, haven't tried it yet).

A manifest is just JSON and is pushed to a registry to associate some images by digest. You can create these locally. I suggested `oras` to you because it's rather nice to work with for this sort of thing if you really need it, it's Golang based like `crane`.

For `oras` you can see [`oras manifest push`](https://oras.land/docs/commands/oras_manifest_push) and how it pushes a JSON file. They have a [Go client library](https://oras.land/docs/client_libraries/go) so you can skip interacting via a CLI command.

But really you can just have buildx build all platforms and push, the only reason you need to manage the manifest yourself to stitch images together is if you have a different build strategy like some do on Github CI with native ARM runner for a faster ARM build, and then need to update the registry manifest for tags to have a multi-platform manifest.

In the past before buildx supported multi-platform image build + publishing, you would do individual platform builds and tag those with their platform. At some point people were doing that and managing manifests, which seems to be similar to what GoReleaser has been doing?

Personally I'd just align with buildx `--platforms` to build and push through that. No need to worry about manifest management, unless you want to implement that as a separate feature (such as through `oras`). This will require the user to be setup for multi-platform builds though (_on Github CI this is handled via two complimentary actions to install QEMU with additional platform support and setup buildx with the container driver to use that_). On a local system you'd need to match that expectation, but that's not really the responsibility of GoReleaser beyond documentation?

This way GoReleaser would still support the equivalent of `docker/build-push-action` where you could build and push single platforms on different runners (x86_64 + aarch64 native runner nodes), and then 
have a post-step that pushes a manifest to create the multi-platform image. For the time being though, I think it's better to keep it simple and focused on the one happy path with a single build system 😅

caarlos0:
Oh, sorry for not being clear - again 😶‍🌫️

I mean running this locally: https://hub.docker.com/_/registry
Then we can `docker buildx build`, and test the images locally (for example, for `--snapshot` builds), and later use crane to copy them to the public registries.

Again, not sure if that will even work, but sounds like it should? 

--- 



> For the time being though, I think it's better to keep it simple and focused on the one happy path with a single build system

Agreed, the way it works currently is way too complicated and requires a lot of code repetition. I'm striving to find the simplest way possible to make this work as natively as possible 🙏

polarathene:
> I mean running this locally: https://hub.docker.com/_/registry
> Then we can `docker buildx build`, and test the images locally (for example, for `--snapshot` builds), and later use crane to copy them to the public registries.

I don't quite follow why that is necessary? If you're doing it all on the same Docker host, you'd just build the platforms via buildx like so:

```bash
docker buildx build --platforms linux/amd64,linux/arm64 ...
```

You only run into the single platform concern when you want to actually run these images locally, which would require adding `--load` to the options there. As mentioned going forward with the containerd storage backend in future, that will probably allow `--load` to handle multiple platforms at which point I'm not sure what advantage you'd be getting from self-hosting a registry like you suggest?

If you're not running `docker run` or similar after the build, then none of this concern should really matter. Just have the `docker buildx build` push directly to a registry?

It differs when like I mentioned earlier you are using CI with two separate runners/jobs, each on a native platform to build separately so that the non-native platform isn't considerably slower to build. Your self-hosted registry proposal isn't ideal then though, it'd require additional config, config that isn't really going to help in an environment like Github Actions?

---

> Agreed, the way it works currently is way too complicated and requires a lot of code repetition. I'm striving to find the simplest way possible to make this work as natively as possible 🙏

I don't personally use GoReleaser yet, but I have seen a few projects I contribute to use it for their docker image build/publish in CI. I do have a fair amount of experience with the equivalent locally and with Github Actions though.

I think for now it should be like `docker/build-push-action`, just focused on YAML config that maps to CLI options. Anything else that'd involve more fancy logic could come after that, or defer to separate tooling? (_like how [`docker/build-push-action`](https://github.com/docker/build-push-action) only has `tags` input to match buildx `--tags`, the separation of `images` + `tags` as inputs like you've proposed is good, while in GHA you'd use a separate action [`docker/metadata-action`](https://github.com/docker/metadata-action) with expressions to generate the actual values for `--tags`_)

If you look at `docker/build-push-action` implementation, like I documented in the below collapsed section for you, it's rather simple YAML to CLI mapping for the most part. That should be easy to maintain, I wouldn't worry about the manifest management concern for now. If anything it might be better suited as a manual post-step until you have enough user engagement/discussion and demand to improve support with how GoReleaser could simplify that use-case better. For now, focus should be on improving the more common OCI image build config.

<details>
<summary>Collapsed for brevity</summary>

**EDIT:** As mentioned at the end, I have noticed we already discussed most of this before 😅 You can probably ignore it 👍

Personally I think `docker/build-push-action` should be a good reference to follow given it's popularity over the years, it'll potentially benefit users adopting GoReleaser for OCI images from being familiar with a well established action that's validated their approach which is also YAML config based?

I would focus on a subset of what that can do configuration wise, adapted to GoReleaser. If you have feedback about differences there from the GoReleaser perspective of what should be different, I'm happy to help discuss what configuration should look like before another implemention attempt?

---

For starters, rather than `dockers_v2`, if it's any easier you might be able to consider [integrating with BuildKit directly](https://github.com/moby/buildkit) which I believe is more about building OCI images in general and buildx wraps?

If it's simpler to call out to buildx instead, aligning with `docker/build-push-action` makes sense 😅 I was just thinking about GoReleaser being more agnostic here as offering an OCI image builder, not specific to Docker, such that instead of `dockers_v2` it might be named `oci` :shrug:

Stay with the happy path of one build host, the other platform (if building multi-platform) should just be deferred to QEMU + buildx, GoReleaser doesn't need to be concerned with that. Support customizing the output type, along with cache from/to options and it'd be just like it is with `docker/build-push-action` or running `docker buildx build` manually should someone need to stray from the defaults.

No need to worry about self-hosting a registry, leave that up to the user of GoReleaser to decide/configure if they need that. You can add a follow-up feature later with `crane`/`oras` if it's deemed useful to simplify managing multi-platform manifests for images with different build hosts (_typically these can both just publish to a registry with no tags, adjusting the output type, and acquire the published image digests for referencing in a multi-platform manifest, no need to juggle storage/cache_).

Does this sound more viable? Having GoReleaser mirror config support for [`docker/build-push-action` inputs](https://github.com/docker/build-push-action?tab=readme-ov-file#inputs) as needed? (_no need to cover everything there, only what users request_)

- At a glance, these inputs would be important to support: `annotations`, `build-args`, `cache-from`, `cache-to`, `context`, `file`, `labels`, `load`, `outputs`, `platforms`, `push`, `target`, `tags`. I don't perform any builds that require secrets in the `Dockerfile` itself to be used, there's probably demand for it but just the mentioned inputs alone is sufficient for many OSS projects.
- The [usage section](https://github.com/docker/build-push-action?tab=readme-ov-file#usage) references additional steps/preparation for using that action, QEMU if needing to build non-native platforms, buildx itself to call, and login to a registry if publishing an image.
- The actions inputs are defined [in this config file](https://github.com/docker/build-push-action/blob/master/action.yml) that shows default values, what's required by minimum, and a descriptive text for each.
- The actions inputs are [parsed here](https://github.com/docker/build-push-action/blob/master/src/context.ts), with some additional logic but it's mostly for converting YAML inputs into CLI args. Any defaults are effectively the defaults of Buildx itself which GoReleaser could do as well?

Where it differs from [your proposal config](https://github.com/goreleaser/goreleaser/issues/5733#issue-3015614452) is there is no `images`, there is only `tags`, just like when you use the CLI. There is a separate action `docker/metadata-action` that helps configure this with the same separation of [`images`](https://github.com/docker/metadata-action/?tab=readme-ov-file#images-input) + [`tags`](https://github.com/docker/metadata-action/?tab=readme-ov-file#tags-input), along with labels (_see [this example for reference](https://github.com/docker/metadata-action/?tab=readme-ov-file#semver)_). So I think your separate `images` + `tags` approach is fine 👍

If you're not familiar with `annotations`, that is effectively `labels`, but instead of being added to the image metadata, it's associated to a manifest for example. Since it should be the same CLI mapping logic for buildx, that shouldn't be a worry either. There are image indexes (_manifest lists that point each image platform to their respective image manifest_) and manifests based on single or multi-platform image published. Depending on that type, annotations or labels would be used at the registry like GHCR for presenting information at the web UI.

Whoops just realized we [already discussed this](https://github.com/orgs/goreleaser/discussions/5442#discussioncomment-11844747) months ago 😓 

</details>

EpicStep:
Hi, I'm interested in this feature because I want to use GoReleaser to build amd64 and arm64 images simultaneously. Do you have an estimated timeline for completing it? I'd be happy to help if there's anything I can contribute to.

caarlos0:
no timeline, as I don't yet know which strategy to take.

jkroepke:
I would recommend to keep this simple, similar how simple ko is.

In my opionon, gorelease should fully depend against the build tool here (buildx) where takes care of caching and publishing (images + manifests), similar how ko works. builx can also take care of SBOM.

Basicly a command line wrapper around docker buildx.

On top, the project root could be the build context, including the dist dir and use the capability of buildx to create multi-platform docker images. A Dockerfile like

```
FROM scratch

ARG TARGETARCH=amd64
ARG TARGETOS=linux

ENTRYPOINT ["/access-log-exporter"]
COPY packaging/etc/access-log-exporter/config.yaml /config.yaml
COPY dist/access-log-exporter_${TARGETOS}_${TARGETARCH}*/access-log-exporter /
```

should work fine in that context. podman, which uses buildah support the builtin args since buildah v4, too. 

This should works for Linux and FreeBSD containers. For Windows, a different FROM is necessary.

caarlos0:
yep, what I was trying to do is having a way to build the images without publishing them as well... I think the only way would be to start up a local docker registry and publish to it instead of a remote registry. Not sure how good of an idea that is, though. 🤔

caarlos0:
FWIW most of this PR is garbage and should be rewritten

caarlos0:
This is getting into shape now:

- if `--snapshot`, it'll start a registry and build the manifest there (only useful to test the actual docker image build)
- otherwise, on publish will run a single `docker buildx build` to build and publish the manifest + sboms

caarlos0:
fyi: hid previous comments as they were about a version of the implementation that no longer exists, and quite long/distracting for the current version

.goreleaser.yaml
```diff
@@ -112,74 +112,27 @@ changelog:
...
-      - "ghcr.io/goreleaser/goreleaser:v{{ .Version }}-arm64"
-  - name_template: "{{ if .IsNightly }}goreleaser/goreleaser:nightly{{ end }}"
-    image_templates:
-      - "goreleaser/goreleaser:v{{ .Version }}-amd64"
-      - "goreleaser/goreleaser:v{{ .Version }}-arm64"
-  - name_template: "{{ if .IsNightly }}ghcr.io/goreleaser/goreleaser:nightly{{ end }}"
-    image_templates:
-      - "ghcr.io/goreleaser/goreleaser:v{{ .Version }}-amd64"
-      - "ghcr.io/goreleaser/goreleaser:v{{ .Version }}-arm64"
+    labels:
```
jkroepke, line 125: Would it be possible to support annotations as well?

Also keep note of the annotations level: https://docs.docker.com/build/metadata/annotations/#specify-annotation-level

pkg/config/config.go
```diff
@@ -1054,6 +1056,20 @@ type DockerManifest struct {
 	Retry          Retry    `yaml:"retry,omitempty" json:"retry,omitempty"`
 }
 
+// DockerV2 is the new Docker build pipe options.
+type DockerV2 struct {
```
jkroepke, line 1060: Any reason for not exposing `build_flag_templates` ?

Use-cases are `--build-context`, `--cache-from`, `--cache-to`, `--no-cache`.

Not sure if it worth to expose each single CLI options. Something like `extra_args` for advanced usage would be great.

www/docs/customization/dockers_v2.md
```diff
@@ -0,0 +1,199 @@
...
+├── linux/amd64/myprogram.apk
+└── linux/amd64/myprogram.deb
+```
+
+`myprogram` would actually be your binary name, and the Linux package names
+would follow their respective configuration's names.
+
+## Setting up a builder
+
+For buildx to work, you'll need to have a builder that supports multi-platform
```
jkroepke, line 189: Note sure, if you would like to mention

https://github.com/docker/setup-buildx-action

for GHA as well.

www/docs/customization/dockers_v2.md
```diff
@@ -0,0 +1,199 @@
...
+
+Note that we are not building any go files in the Docker build phase, we are
+merely copying the binary to a `scratch` image and setting up the `entrypoint`.
+
+The idea is that you reuse the previously built binaries instead of building
+them again when creating the Docker image.
+
+The build context itself is a temporary directory which contains the binaries
+and packages for the target platforms, which you can `COPY` into your image
+(mind the use of `$TARGETPLATFORM` above).
```
jkroepke, line 163: Edge Case scenario:

ARM32 builds (for older RPIs) `GOARCH=arm GOARM=7` have `TARGETPLATFORM=linux/arm/v7`. I'm not checked the code here, but I would like to keep this in your mind.

www/docs/deprecations.md
```diff
@@ -41,6 +41,157 @@ PS: Don't forget to add it to cmd/mcp.go as well!
...
+artifacts for the given target platform will be located within
+`$TARGETPLATFORM/`.
+
+One side effect of this new feature is that builds with `--snapshot` will
+produce images that cannot be ran - it only tests that the `docker build`
+succeeds.
+This is something we'll need to improve in future versions.
+
+Feel free to suggest improvements
+[here](https://github.com/orgs/goreleaser/discussions/6005).
```
jkroepke, line 172: What about having a push option, which has the value `push: {{ not .IsSnapshot }}` by default. If an end-user really wants to have usage images from a snapshot build, he can opt-in. He also can use templated `image_templates` to configure different push location.

My use case here: I'm running `--snapshot` on main branches and I would like to push the main branch to the `main` tag. Thats one way, how I can quickly get feedback from reporters without publish a full release.

www/docs/customization/dockers_v2.md
jkroepke, line 163: Great Idea!

caarlos0: this will work too yes!

pkg/config/config.go
caarlos0: yeah I think we can add the flags yes

will probably do in a subsequent PR

.goreleaser.yaml
caarlos0: yes!

will probably do in a subsequent PR

www/docs/customization/dockers_v2.md
caarlos0: good catch! We should probably have a complete actions example

www/docs/deprecations.md
caarlos0: That's the use case for nightly builds... if we allow to pass random flags, though, you could still pass a `{{ if .IsSnapshot }}--push{{ end }}` then 🤔

.goreleaser.yaml
caarlos0: #6011

ryancurrah:
> we will now run a single docker buildx build with the given platforms, which will build and publish the manifest and SBOM.

For teams that only use Go Releaser to build artifacts (and not to publish them), will this switch to `docker buildx build` + manifest publishing break our workflow?

This is how we currently build artifacts.

```console
goreleaser release --clean --skip=validate,publish
```

caarlos0:
Hmmm this might need some extra work - can you open an issue for this particular case? 

If you run with --snapshot it will create the local images btw, but with skip publish it won't

ryancurrah:
We will look at an internal project to modify our 1000+ Go Releaser files to also publish. So I will hold off on the ticket for now. Thanks for the response.

PR #5798, opened 2025-05-29
feat: mcp server

caarlos0:
I'm playing with the idea of having a MCP server for goreleaser, so things like Cursor and can use it to figure things out about the project.

Some ideas of tools we can add:

- [x] check configuration file
- [x] run a snapshot build
- [x] health check 
- [x] init
- [ ] list features being used 
- [ ] what else?


caarlos0:
gonna merge as is, can keep adding more stuff later.

PR #5807, opened 2025-06-05
fix(dep): update gitlab, fix use of removed function

caarlos0:
apparently this is what we're supposed to use now - looks like it does the same thing under the hood

closes #5778 
see: https://gitlab.com/gitlab-org/api/client-go/-/releases/v0.129.0


mavogel:
Yes, just took a look at https://gitlab.com/gitlab-org/api/client-go/-/issues/2090 which explains the BC. Still pre-major, so it is fine and we are ready for GitLab v18

caarlos0:
Thanks @mavogel ❤️

PR #5844, opened 2025-06-22
fix(cask): add test and docs for custom URL

caarlos0:
based on suggestions by @sushichan044


internal/pipe/cask/testdata/github.rb
```diff
@@ -0,0 +1,17 @@
...
+    release = GitHub.get_release("goreleaser", "example", tag)
+    release["assets"].find { |asset| asset["name"] == name }["url"]
+  end
+
+  def self.token
+    require "utils/github"
+    @github_token = ENV["HOMEBREW_GITHUB_API_TOKEN"]
+    unless @github_token
+      @github_token = GitHub::API.credentials
+      raise CurlDownloadStrategyError, "Failed to retrieve token" if @github_token.nil? || @github_token.empty?
```
sushichan044: Sorry, I forgot to remove this code after pasting it from DownloadStrategy of a Formula.
```suggestion
      raise "Failed to retrieve token" if @github_token.nil? || @github_token.empty?
```

www/docs/customization/homebrew_casks.md
```diff
@@ -285,12 +285,46 @@ Your users can then `brew install foo@1.2` to keep using the previous version.
...
-## Limitations
+## Private GitHub Repositories
 
-- Only one `GOARM` build is allowed;
+The best way to support private repositories is to add by using a custom block,
+a custom template URL, and custom headers.
+
+Here's an example:
+
+```yaml title=".goreleaser.yaml"
```
sushichan044: We should probably add a warning about the use of an internal API.
```suggestion
Here's an example:

!!! warning

    Please note that this example uses an internal Homebrew API to retrieve the GitHub API token.

    Replace with your implementation as needed.

```yaml title=".goreleaser.yaml"
```

sushichan044 (changes requested):
Thanks for the great work 🙏 
I have two small revisions I'd like to request.

www/docs/customization/homebrew_casks.md
sushichan044: スクリーンショット 2025-06-23 9 34 41

caarlos0:
thanks for the review @sushichan044 🙏🏻

PR #5847, opened 2025-06-23
docs(cask): fix the incorrect Ruby code in the example

sushichan044:
Fixing the incorrect Ruby sample code I proposed for private cask releases in #5844.
https://github.com/goreleaser/goreleaser/pull/5844

to correct my mistake


caarlos0:
thanks!

PR #5900, opened 2025-07-11
docs: update required workflow permissions in attestations.md

mroth:
This updates the documentation regarding attestations to address two issues I hit when following the docs.

1. Adds `id-token: write` permissions to the Github Actions workflow.  Without this, the signature will fail.  See https://github.com/actions/attest-build-provenance?tab=readme-ov-file#usage for reference.
2. Show the goreleaser configuration file checksum filename modified to match the predictable filename in the actions workflow.


caarlos0:
good catch! thanks!🙏

PR #5915, opened 2025-07-22
ci: release loong64 archives

znley:
This commit could release loong64 archives.

I need to use gorelease to build binaries and then use these binaries to build a container image for the loong64 architecture.

https://github.com/goreleaser/goreleaser/pull/3277


caarlos0:
thanks

PR #5922, opened 2025-07-23
add .github/goreleaser.yaml

Efreak:
This commit adds `.github` directory as a search path for the config. 

This is for use with GitHub actions, it keeps the repo root cleaner. (Since you already need to have the .github directory for actions anyways)

Alternatives:
- add this logic to the GitHub action
- if goreleaser is not running interactively, check all first-level directories for an *unambiguous* goreleaser config (ie only use the config if it's the only one found).
  - Probably also add a parameter to skip this function in case there's a config but the user doesn't want to use any configs on this run.

I didn't bother to create an issue for this, as it seems simple enough to just add it.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -50,6 +50,8 @@ func loadConfigCheck(path string) (config.Project, string, error) {
 		".goreleaser.yaml",
 		"goreleaser.yml",
 		"goreleaser.yaml",
+		".github/goreleaser.yml",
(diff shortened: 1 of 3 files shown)


caarlos0:
i don't think that semantically it makes sense to put goreleaser config inside .github, at least not by default

you can in your goreleaser action call pass `args: release -f .github/goreleaser.yml` though, which seems a good enough compromise to me

PR #5955, opened 2025-08-05
ci: improve get-releases.sh, do not serve v0.* anymore

caarlos0:
This removes the `v0` from the `releases*.json`.

Not yet sure how much of a good idea this is, as the files aren't that big anyway:

###### on `main`:

```console
$ wc -c www/docs/static/releases*.json
    5125 www/docs/static/releases-pro.json
   19928 www/docs/static/releases.json
   25053 total
```

###### this branch:

```console
$ wc -c www/docs/static/releases*.json
    3925 www/docs/static/releases-pro.json
    3523 www/docs/static/releases.json
    7448 total
```

A blind attempt of improving https://github.com/orgs/goreleaser/discussions/5954


caarlos0:
I think we can leave this as is for a little longer.

PR #5986, opened 2025-08-15
fix(brew): use && instead of and

caarlos0:
according to some users, linuxbrew might complain for the `and` syntax: https://github.com/orgs/goreleaser/discussions/5947


mrfuxi:
Looks great, I can't wait to use it!
Thank you so much for the change.

PR #5989, opened 2025-08-15
feat: add makeself self-extracting archive format support

faisal-smarthub:
Add "makeself" archive format with configurable install scripts, compression options, LSM support, and custom extensions.

Features include inline/file-based install scripts, custom labels, multiple compression formats, and comprehensive error handling.

Add makself archive support to goreleaser

.This is an itch I wanted to scratch because we were developing a tool that we wanted to distribute using makeself and the only way to integrate it into the release process was to add a pre-release hook, which was still not the right solution since it needed to untar the archives and re-package it as makeself and we would still end up publishing both the tarballs and the .run makself archives. This fixes that and provides us a clean pipeline to publish self extracting archives for platforms that support makeself.

Makeself is documented [here](https://makeself.io/)

### DIFF
--- a/internal/pipe/archive/archive.go
+++ b/internal/pipe/archive/archive.go
@@ -178,7 +178,26 @@ func create(ctx *context.Context, arch config.Archive, binaries []*artifact.Arti
 	if err != nil {
 		return err
 	}
-	archivePath := filepath.Join(ctx.Config.Dist, folder+"."+format)
+	var archivePath string
+	if format == "makeself" {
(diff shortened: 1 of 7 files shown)


faisal-smarthub:
I was in two minds about this myself. But from a purely archive perspective it seemed simpler to add to the existing archive code. I will refactor this to a new pipeline.

thanks,
Faisal
________________________________

caarlos0:
cool, let me know if I can help in any way!

PR #5992, opened 2025-08-18
feat(nfpm): add `Prefix` template variable for termux

caarlos0:
It'll be empty in all formats but `termux.deb`, this way users can properly put the files in the right places.

Not sure about the name of the variable, though.


caarlos0:
I think this is a tad weird, probably worth just documenting it without adding a new feature.

PR #5994, opened 2025-08-19
feat(makeself): move from archive format to standalone pipeline

faisal-smarthub:
Add makeself support to goreleaser to create self-extracting archives/installers.

This is to scratch an itch I had with the current goreleaser implementation that I was using where I needed to create makeself archives but the current workaround using pre-release hooks was inelegent and inefficient. I wanted to add first class support for makeself to goreleaser.

Documentation about makeself available [here](https://makeself.io/)

### DIFF
--- a/internal/pipe/makeself/makeself.go
+++ b/internal/pipe/makeself/makeself.go
@@ -1,0 +1,528 @@
+// Package makeself implements the Pipe interface providing makeself self-extracting archive support.
+package makeself
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/caarlos0/log"
+	"github.com/goreleaser/goreleaser/v2/internal/archivefiles"
+	"github.com/goreleaser/goreleaser/v2/internal/artifact"
+	"github.com/goreleaser/goreleaser/v2/internal/ids"
+	"github.com/goreleaser/goreleaser/v2/internal/pipe"
+	"github.com/goreleaser/goreleaser/v2/internal/semerrgroup"
+	"github.com/goreleaser/goreleaser/v2/internal/skips"
+	"github.com/goreleaser/goreleaser/v2/internal/tmpl"
+	"github.com/goreleaser/goreleaser/v2/pkg/config"
+	"github.com/goreleaser/goreleaser/v2/pkg/context"
+)
+
+const (
+	defaultNameTemplate = `{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ with .Arm }}v{{ . }}{{ end }}{{ with .Mips }}_{{ . }}{{ end }}{{ if not (eq .Amd64 "v1") }}{{ .Amd64 }}{{ end }}`
+	extraFiles          = "Files"
+)
+
+// Pipe for makeself packaging.
+type Pipe struct{}
+
+func (Pipe) String() string { return "makeself packages" }
+func (Pipe) Skip(ctx *context.Context) bool {
+	return skips.Any(ctx, skips.Makeself) || len(ctx.Config.Makeselfs) == 0
+}
+
+// Default sets the pipe defaults.
+func (Pipe) Default(ctx *context.Context) error {
+	ids := ids.New("makeselfs")
+	for i := range ctx.Config.Makeselfs {
+		makeself := &ctx.Config.Makeselfs[i]
+		if makeself.ID == "" {
+			makeself.ID = "default"
+		}
+		if makeself.NameTemplate == "" {
+			makeself.NameTemplate = defaultNameTemplate
+		}
+		if makeself.Extension == "" {
+			makeself.Extension = ".run"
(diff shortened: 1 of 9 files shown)


www/docs/customization/makeself.md
```diff
@@ -0,0 +1,320 @@
...
+    # Templates: allowed.
+    meta: false
+
+    # Disable this makeself package.
+    # Templates: allowed.
+    disable: "{{ .Env.SKIP_MAKESELF }}"
+
+    # Deprecated: use 'ids' instead.
+    builds:
+      - default
```
caarlos0, line 140: we can leave this out

www/docs/customization/makeself.md
```diff
@@ -0,0 +1,320 @@
...
+      Begin4
+      Title: {{ .ProjectName }}
+      Version: {{ .Version }}
+      Description: {{ .ProjectName }} self-extracting installer
+      Author: Your Name
+      Maintained-by: your-email@example.com
+      Primary-site: https://github.com/youruser/yourproject
+      Platforms: Linux
+      Copying-policy: MIT
+      End
```
caarlos0, line 87: are there more options than these? if not, we could probably generate the LSM file from yaml, e.g. accepting options like `description`, `author`, etc (like casks, winget,...)

www/docs/customization/makeself.md
```diff
@@ -0,0 +1,320 @@
...
+      #!/bin/bash
+      echo "Installing {{ .ProjectName }}..."
+      chmod +x {{ .Binary }}
+      cp {{ .Binary }} /usr/local/bin/
+      echo "Installation complete!"
+
+    # Path to install script file within the archive.
+    # Alternative to install_script for external script files.
+    # Templates: allowed.
+    install_script_file: "install.sh"
```
caarlos0, line 65: I _think_ we should have only one of these... maybe simply `install_script` and always asking for a file?

realistically, i think these scripts will usually get rather big, and most people will want them outside of yaml? thoughts?

internal/pipe/makeself/makeself.go
```diff
@@ -0,0 +1,528 @@
...
+	// Check if disabled
+	if makeselfCfg.Disable != "" {
+		disable, err := template.Apply(makeselfCfg.Disable)
+		if err != nil {
+			return err
+		}
+		if disable == "true" {
+			return nil // Return nil instead of skip error for disabled packages
+		}
+	}
```
caarlos0, line 142: you could use `template.Bool`

internal/pipe/makeself/makeself.go
```diff
@@ -0,0 +1,528 @@
...
+	var err error
+
+	// For meta packages without binaries, use a simpler name template if the default is being used
+	if makeselfCfg.Meta && len(binaries) == 0 && makeselfCfg.NameTemplate == defaultNameTemplate {
+		// Use a meta-package friendly name template
+		metaTemplate := `{{ .ProjectName }}_{{ .Version }}_meta`
+		packageName, err = template.Apply(metaTemplate)
+	} else {
+		packageName, err = template.Apply(makeselfCfg.NameTemplate)
+	}
```
caarlos0, line 154: this could be handled in `Default`, if `Meta` use the simpler name template.
if `Meta` is not set, and there's no found binaries, it should fail instead I think 🤔

internal/pipe/makeself/makeself.go
```diff
@@ -0,0 +1,528 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Apply templates to all configuration fields
+	label := makeselfCfg.Label
+	if label != "" {
```
caarlos0, line 161: don't need to check if label != "" (and the other fields below as well)... can simply template it - if I recall `Apply` already checks it

internal/pipe/makeself/makeself.go
```diff
@@ -0,0 +1,528 @@
...
+		label, err = template.Apply(label)
+		if err != nil {
+			return fmt.Errorf("failed to apply template to label: %w", err)
+		}
+	}
```
caarlos0, line 166:
```suggestion
	label, err := template.Apply(makeselfCfg.Label)
	if err != nil {
		return fmt.Errorf("failed to apply template to label: %w", err)
	}
```

www/docs/customization/makeself.md
faisal-smarthub: The mandatory ones are Title, Version, Entered-date, Description, Author and Primary-site.  There are a few more, but I don't think we need to police the content. If the user puts in things that are non-standard, that is their prerogative.

www/docs/customization/makeself.md
faisal-smarthub: Makes sense. Install files tend to be biggish.

faisal-smarthub:
Please feel free to take over and bring it to the a usable state to go into the mainstream code. I am currently testing my forked version on my own codebase.

caarlos0:
Continuing it in #6023 

will close this one

thanks!

PR #6051, opened 2025-09-04
docs: Fixed typo in dockers_v2. Should be images for string list.

ZAlexanderV:


caarlos0:
Good catch! Thanks 🙏🙏

PR #6083, opened 2025-09-15
docs: fix typo in signs config section

stanhu:
The comment said to use `binaries_sign`, but the actual value should be `binary_signs`.

...

...

...


caarlos0:
ah, good catch! thanks!

PR #6094, opened 2025-09-17
docs: fix inconsistency about symlink in nfpm.md

kaitokimuraofficial:
This PR fixs inconsistency between comment and actual configuration in nfpm symlink example.


caarlos0:
Thanks!

PR #6138, opened 2025-10-02
feat: do not create a docker manifest if image contains one platform

tomMoulard:
When a docker image is built using dockers_v2, and a single platform is selected, the produced image contains a useless manifest.

Imagine my configuration looks like that:

```yaml
dockers_v2:
  - ids: [mybuild]
    platforms:
      - linux/arm64
    images:
      - "myuser/my-image"
    tags:
      - latest
      - "{{ .ShortCommit }}"

```

The produced image will not use `docker buildx` to be built, and will simply use `docker build`.
Doing so  does not burden the docker registry with a meaningless docker image manifest.

### notes

This PR does not (currently) provide tests for this feature. Do not hesistate to ask me to add tests, change a few things, or close the PR is there is no need for it.

Thanks a lot for your awesome work on this great project !


caarlos0:
BTW: I haven't merged this yet because I'm wondering if it is really needed... 

Using buildx, we have the benefit auto-generating SBOMs with `--attest`, and that doesn't seem to always work with plain `docker build` (at least I get a `Attestation is not supported for the docker driver` error)... 🤔

tomMoulard:
For sure, the issue from me is that this dockers_v2 thing breaks my usage of goreleaser...
My current setup builds and push a docker image to a private ecr on aws to be used by a lambda. The lambda fails to start when there is a docker manifest (this might be a configuration issue on my side tho).

I wouldn't add a new flag just for this use case, and thats why I did my PR like this.

What do you think would be the better way to use docker build (instead of buildx) for a single image?

caarlos0:
I'm not sure about this. On one hand, this is, for sure, not sufficient. It'll fail if you have SBOM enabled, or if you pass other `--attest` flags. 

On the other, I get your point. 

I'll have to think about this for a bit longer.

caarlos0:
[this](https://stackoverflow.com/questions/75521775/buildx-docker-image-claims-to-be-a-manifest-list) also recommends setting `--provenance=false` (even though I think it is false by default?).

Anyway, I tried it here:

```yaml
dockers_v2:
  - images:
      - "ghcr.io/caarlos0/test-dockers-v2"
    tags:
      - "v{{ .Version }}"
      - "{{ if .IsNightly }}nightly{{ end }}"
      - "{{ if not .IsNightly }}latest{{ end }}"
    platforms:
      - linux/amd64
    sbom: false
    flags:
      - "--provenance=false"
    labels:
      "org.opencontainers.image.description": "Some description"
      "org.opencontainers.image.created": "{{.Date}}"
      "org.opencontainers.image.name": "{{.ProjectName}}"
      "org.opencontainers.image.revision": "{{.FullCommit}}"
      "org.opencontainers.image.version": "{{.Version}}"
      "org.opencontainers.image.source": "{{.GitURL}}"
```

and seems like it works:

```
❯ docker image inspect ghcr.io/caarlos0/test-dockers-v2:v1.14.43

[
    {
        "Id": "sha256:d4e4a9330013c635d465f6e4082b3ab48d02d04f601cf0e00c0f33fe302408d2",
        "RepoTags": [
            "ghcr.io/caarlos0/test-dockers-v2:v1.14.43"
        ],
        "RepoDigests": [
            "ghcr.io/caarlos0/test-dockers-v2@sha256:3401754aada7e8f0d9eaad01c4018a71a837341de4c9e2e7399d72e976e99d1d"
        ],
        "Parent": "",
        "Comment": "buildkit.dockerfile.v0",
        "Created": "2025-11-29T12:07:28.45177019Z",
        "DockerVersion": "",
        "Author": "",
        "Architecture": "amd64",
        "Os": "linux",
        "Size": 9399732,
        "GraphDriver": {
            "Data": {
                "LowerDir": "/var/lib/docker/overlay2/4207c781d1537ab5995de97d23ccfa35dc622698acd80d998553821df2460167/diff",
                "MergedDir": "/var/lib/docker/overlay2/fc0609c70382fad251b7468071109d83b564f535f9260440abce4bf19746c1ce/merged",
                "UpperDir": "/var/lib/docker/overlay2/fc0609c70382fad251b7468071109d83b564f535f9260440abce4bf19746c1ce/diff",
                "WorkDir": "/var/lib/docker/overlay2/fc0609c70382fad251b7468071109d83b564f535f9260440abce4bf19746c1ce/work"
            },
            "Name": "overlay2"
        },
        "RootFS": {
            "Type": "layers",
            "Layers": [
                "sha256:256f393e029fa2063d8c93720da36a74a032bed3355a2bc3e313ad12f8bde9d1",
                "sha256:06b9f3f7cc16cb691945a52ff3f972f90d6e148dd5f3d95a7611acfe1ba23f21"
            ]
        },
        "Metadata": {
            "LastTagTime": "0001-01-01T00:00:00Z"
        },
        "Config": {
            "Cmd": null,
            "Entrypoint": [
                "/usr/bin/mybin"
            ],
            "Env": [
                "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
            ],
            "Labels": {
                "org.opencontainers.image.created": "2025-11-29T12:07:26Z",
                "org.opencontainers.image.description": "Some description",
                "org.opencontainers.image.name": "test",
                "org.opencontainers.image.revision": "3f0522d62f1fe0ed508912b15306ff8dcda6e548",
                "org.opencontainers.image.source": "git@github.com:caarlos0-graveyard/test.git",
                "org.opencontainers.image.version": "1.14.43"
            },
            "OnBuild": null,
            "User": "",
            "Volumes": null,
            "WorkingDir": "/"
        }
    }
]
``` 

So maybe this can be only a documentation note or something like that 🤔

caarlos0:
added it to docs, which seems to make the most sense to me.

this PR would break many things, I want to keep it using docker buildx as it seamlessly supports other backends as well.

thanks for raising this, though!

caarlos0:
forgot to comment here, but ended up just documenting it so people can keep using buildx, which is compatible with other backends as well.

thanks for the PR and for bringing this up, though <3

tomMoulard:
no worries, it's fine as it is !

Keep up the good work :shipit:

PR #6163, opened 2025-10-13
docs: add default value for wrap directory

FelicianoTech:
Some of the keys for "archives" list defaults, some don't. This one didn't and I assume it was 'true' given the example. The default is actually 'false'. I figured I'd document it for future me and others.


caarlos0:
good catch, thanks!

PR #6174, opened 2025-10-14
refactor(homebrew): replace Binary with Binaries slice in HomebrewCask

Nurtic-Vibe:
This will align goreleaser to specification from https://docs.brew.sh/Cask-Cookbook#at-least-one-artifact-stanza-is-also-required where multiple `binary` items are permitted.


caarlos0:
this is a good catch! thank you!

PR #6184, opened 2025-10-16
Add Flyte to the list of users

eapolinario:
Add Flyte to the list of projects that rely on goreleaser (as can be evidenced by https://github.com/flyteorg/flyte/blob/master/.github/workflows/create_release.yml#L181C19-L187)

Flyte has been relying on goreleaser since the beginning of the project and it's using the pro version for the last 2 years without any issues.

N/A


caarlos0:
NICE!

thanks 🙏🏻

PR #6198, opened 2025-10-19
Allow GOARM softfloat and hardfloat

MDr164:
Go added the ability to add an optional softfloat/hardfloat suffix to the `GOARM` variable when building to control what assembly is emitted for variants with and without a FPU. This change should allow the usage of those variants. Maybe it's worth using regex but because the amount of variants is rather limited and my regex skills are lacking I decided to opt for written out strings.


caarlos0:
is this supported? I didn't find this in docs anywhere...

MDr164:
It is but looks like this was never documented that well: https://github.com/golang/go/issues/61588

MDr164:
> is this supported? I didn't find this in docs anywhere...

https://go.dev/doc/go1.22#arm it was added in 1.22 and mentioned in the release notes. I'm relying on it for some of my release targets so it would be nice if goreleaser could support these targets directly. Would anything else needed for a support or is modifying the `targets(_test).go` file(s) sufficient?

caarlos0:
hmm i missed that i think.

ok, so we should probably support this. That said, we'll need more changes. basically, need to check every reference of `ByGoarm` and see.

MDr164:
As mentioned it is rather underdocumented upstream so I'm not surprised you missed it. I'll skim through the code base then and update this PR :+1:

caarlos0:
thank you!

MDr164:
There are multiple other occurrences of GOARM but I think given that softfloat/hardfloat are an optional feature it wouldn't make sense to update it everywhere (e.g. locations like DefaultGOARM and such). I think I caught all relevant locations now.

internal/builders/golang/targets.go
```diff
@@ -357,7 +357,7 @@ var (
 
 	validGoamd64   = []string{"v1", "v2", "v3", "v4"}
 	validGo386     = []string{"sse2", "softfloat"}
-	validGoarm     = []string{"5", "6", "7"}
+	validGoarm     = []string{"5", "6", "7", "5,softfloat", "5,hardfloat", "6,softfloat", "6,hardfloat", "7,softfloat", "7,hardfloat"}
```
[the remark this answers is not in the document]
MDr164: I used neither the AUR, NFPM or Nix release pipelines before but attempted to resolve this issue to my best knowledge. Please validate if the fix is adequate.

internal/builders/golang/build_test.go
```diff
@@ -346,6 +394,37 @@ func TestWithDefaults(t *testing.T) {
...
+			targets: []string{
+				"linux_arm_5",
+				"linux_arm_6",
+				"linux_arm_7",
+				"linux_arm_5,softfloat",
+				"linux_arm_5,hardfloat",
+				"linux_arm_6,softfloat",
+				"linux_arm_6,hardfloat",
+				"linux_arm_7,softfloat",
+				"linux_arm_7,hardfloat",
```
caarlos0: should probably replace the `,` with `_` like we do with `mips`

caarlos0 (commented):
thanks for this!
looking good overall!

I think we might need to change `ByGoarm` in `artifacts`, I would probably do it so it respects `,softfloat`/`,hardfloat` if informed in the filter, and ignore it otherwise.

also, the actual target name should probably replace `,` with `_` to match the others 🤔

internal/builders/golang/build_test.go
MDr164: Sounds good, will update the PR. Can you trigger a CI run once so I can check if there are other spots I might have missed?

caarlos0: sure! done

MDr164: I changed the usage to _ everywhere. I wasn't entirely sure if we also want _ in the goreleaser config so I added the changes in another commit. Let me know which one you would prefer for the config format. I have no strong opinion there.

caarlos0:
just fwiw i'm still on the fence on how to best do this - its full of conflicts now, but don't worry, when i reach a decision on how, i'll do it and add you as co-commiter so you get the green squares hehe

MDr164:
> just fwiw i'm still on the fence on how to best do this - its full of conflicts now, but don't worry, when i reach a decision on how, i'll do it and add you as co-commiter so you get the green squares hehe

I can rebase this PR myself in it's current state and make sure that it at least builds fine again if that helps you.

Issue #6196, opened 2025-10-18
Add a Discourse Announcer


FelicianoTech:
I would like to announce releases via GoReleaser to a [Discourse](https://discourse.org/) forum. This request has been [mentioned & greenlit](https://github.com/orgs/goreleaser/discussions/6192) on GitHub Discussions. 

This Issue is to track the task and to provide notes and a place for discussion if needed.

---

Discourse API docs: https://docs.discourse.org/
I believe the relevant endpoint would be: https://docs.discourse.org/#tag/Topics/operation/createTopicPostPM

I'm not currently aware of an existing Discourse Go SDK but I would want to avoid pulling in another dependency anyway considering we only need to work with one end-point.

I'll get started on a PR shortly. Anyone have any thoughts, let me know.

FelicianoTech:
Got a little busy. The PR will be opened later today.

PR #6199, opened 2025-10-21
feat: add Discourse announcer

FelicianoTech:
This PR adds a new Announcer for [Discourse](https://discourse.org/)-based forums. This is a bare minimum PR. There's more I could do with error codes and with allowing threading. I will wait-and-see for those and if there's interest, I can add that in a future PR.

I will comment in-line on why I made certain decisions and/or ask questions.

Closes: #6196

### DIFF
--- a/internal/pipe/discourse/discourse.go
+++ b/internal/pipe/discourse/discourse.go
@@ -8,7 +8,6 @@ import (
 	"net/http"
 
 	"github.com/caarlos0/env/v11"
-	"github.com/caarlos0/log"
 	"github.com/goreleaser/goreleaser/v2/internal/tmpl"
 	"github.com/goreleaser/goreleaser/v2/pkg/context"
 )
@@ -92,11 +91,10 @@ func (p Pipe) Announce(ctx *context.Context) error {
 	}
 
 	req.Header.Set("Content-Type", "application/json")
-	req.Header.Set("User-Agent", "GoReleaser/v2")
+	req.Header.Set("User-Agent", "GoReleaser/v3")
 	req.Header.Set("Api-Username", ctx.Config.Announce.Discourse.Username)
 	req.Header.Set("Api-Key", cfg.APIKey)
 
-	log.Infof("posting: '%s'", msg)
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
 		return fmt.Errorf("%s: %w", p, err)
--- a/www/docs/customization/announce/discourse.md
+++ b/www/docs/customization/announce/discourse.md
@@ -61,7 +61,7 @@ announce:
 
 If you get the error message:
 
-```
+```text
 discourse: There was an error posting to Discourse. Check your config again. HTTP code: XXX
 ```
 
--- a/www/docs/static/schema.json
+++ b/www/docs/static/schema.json
@@ -548,6 +548,37 @@
 				"additionalProperties": false,
 				"type": "object"
 			},
+			"Discourse": {
+				"properties": {
+					"enabled": {
+						"oneOf": [
+							{
+								"type": "string"
+							},
+							{
+								"type": "boolean"
+							}
+						]
+					},
+					"server": {
+						"type": "string"
+					},
+					"title_template": {
+						"type": "string"
+					},
+					"message_template": {
+						"type": "string"
+					},
+					"username": {
+						"type": "string"
+					},
+					"category_id": {
+						"type": "integer"
+					}
+				},
+				"additionalProperties": false,
+				"type": "object"
+			},
 			"Build": {
 				"properties": {
 					"id": {


internal/pipe/discourse/discourse.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	return nil
+}
+
+func (p Pipe) Announce(ctx *context.Context) error {
+
+	title, err := tmpl.New(ctx).Apply(ctx.Config.Announce.Discourse.TitleTemplate)
+	if err != nil {
+		return fmt.Errorf("%s: %w", p, err)
```
FelicianoTech, line 52: Other announcers handle the error message like this: `return fmt.Errorf("bluesky: %w", err)`. This has lead to an issue where the Mattermost announcer, which must have started as a copy from the Teams announcer, left this line in: https://github.com/goreleaser/goreleaser/blob/main/internal/pipe/mattermost/mattermost.go#L63

This will give an incorrect error message to users. My question is:

1. Can I leave my method that I used here?
2. Can I make a new PR changing the other announcers to use my method?
3. Or should I undo what I did here and make it work like the other announcers?

internal/pipe/discourse/discourse.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, endpoint, bytes.NewReader(payload))
+	if err != nil {
+		return fmt.Errorf("%s: %w", p, err)
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("User-Agent", "GoReleaser/v2")
```
FelicianoTech: Not all announcers set the User-Agent. In the PR for adding the Bluesky announcer, they tried to use `ctx.Version` but Carlos mentioned that that will provide the software version, not the GoReleaser version. In the meantime, the Bluesky announcer and possibly others, are using using the "goreleaser/v1" string instead of v2.

Should I...
1. Create a PR to update/add this string in the other announcers?
2. Can someone (or I can try) add the GoVersion to the context so that it can be used later in the pipeline? Some software that uses UserAgents add the name, version, OS and arch. Typically that's browsers. I think for something like this, the name and version would suffice.

internal/pipe/discourse/discourse.go
```diff
@@ -0,0 +1,112 @@
...
+	req.Header.Set("Api-Username", ctx.Config.Announce.Discourse.Username)
+	req.Header.Set("Api-Key", cfg.APIKey)
+
+	log.Infof("posting: '%s'", msg)
```
FelicianoTech: Not all announcers have this line. Is this something that was phased out? Or should we be using it going forward?

www/docs/customization/announce/discourse.md
```diff
@@ -0,0 +1,74 @@
...
+discourse: There was an error posting to Discourse. Check your config again. HTTP code: XXX
+```
+
+then double check the Discourse section of your GoReleaser config.
+Make sure everything is correct.
+Here are some common error codes and what they **might** mean:
+
+- 404 - The server field is either incorrect or your forum is down/unreachable.
+- 403 - The API Key doesn't have the correct permission it needs or the `username` key in GoReleaser config doesn't match what's configured in Discourse.
+- 5XX - The Discourse forum is having a bad day and throwing back errors.
```
FelicianoTech, line 74: I've added more info in the doc page that most announcers have. Is this okay or should I remove some of it?

FelicianoTech:
I'm working on addressing golangci-lint now.

caarlos0:
Thanks! I'll be at GitHub universe this week, but will eventually put in the time into reviewing this!

Sorry in advance if it takes some time - if I don't interact in ~1 week feel free to ping me :D

FelicianoTech:
@caarlos0 friendly bump =)

www/docs/customization/announce/discourse.md
```diff
@@ -0,0 +1,74 @@
+# Discourse
```
caarlos0, line 1:
```suggestion
# Discourse

<!-- md:version v2.13-unreleased -->.

```

internal/pipe/discourse/discourse.go
caarlos0: > Can I make a new PR changing the other announcers to use my method?

yes this is probably the right choice

internal/pipe/discourse/discourse.go
caarlos0: I'll do the goreleaser version thing 🙏🏻

www/docs/customization/announce/discourse.md
caarlos0: let it in :D

internal/pipe/discourse/discourse.go
caarlos0: i don't think we have a pattern here, but I would say its probably better to leave it out as it also logs the success and isn't something that should take too long 🤔

internal/pipe/discourse/discourse.go
caarlos0: actually, for now, lets just fix them all to `/v3`

caarlos0:
Thank you!

PR #6210, opened 2025-10-26
fix(dcos): typo in nfpm.md from 'headers' to 'header'

jkroepke:
Fix docs, ref:

https://github.com/goreleaser/goreleaser/blob/23b03dad313594d9b49d7952ee29f71cf7dabe32/pkg/config/config.go#L696


caarlos0:
thank you!

PR #6211, opened 2025-10-26
Enhance CGO and cross-compiling documentation

tsukinoko-kun:
Add information about libc dependency and static linking for Linux builds.

I had difficulties setting up a CGO build with Goreleaser. I additionally wanted to solve the libc problem.

I had trouble setting this up for my project. Therefore I made a small test repository and started experimenting. I figured this could be helpful for other devs that run into the same issue.

For the interested ones: https://frankmayer.dev/blog/shipping_go_software_with_goreleaser_and_zig


caarlos0:
the example is nice, but you don't need split merge if you're using zig, you can instead template the builds.env field setting it according to Os/Arch. Did you try to do something like that?

This would also make a great addition to the examples in @goreleaser

tsukinoko-kun:
I was trying to fix my problem with libc and missed that using Zig as a C compiler could help a lot more than I realized. I will try this out. Thanks for the input @caarlos0 .

tsukinoko-kun:
You were right. I could simplify the setup and get rid of the split-and-merge.

But I just realized that there already is an example setup that does exactly that. https://github.com/goreleaser/example-zig-cgo

Maybe those examples should be somehow more prominent so that people find them and can use them instead of doing it themselves. My solution is a bit different, but I like the one in the existing example more.

caarlos0:
> But I just realized that there already is an example setup that does exactly that

right! i forgot about that myself



> Maybe those examples should be somehow more prominent so that people find them and can use them instead of doing it themselves. My solution is a bit different, but I like the one in the existing example more.

yes, we should probably have a page in the docs with all the examples, and which features each of them use 🤔

caarlos0:
I improved the CGO docs a bit in e11cef3bd4253a5a90a07ad2862c61c5e2a599dd


should be live at https://goreleaser.com/limitations/cgo/ in a few

PR #6233, opened 2025-11-04
feat: support ko loading base image from daemon

scav:
**What**
Allow users to use a locally cached image as the base image for ko. 

**Why**
The current default behaviour is trying to fetch the image remotely, but this is not always the expected behaviour. E.g ko will support this behaviour through `defaultBaseImage`

Note: this could be hidden behind a flag, and this may be a better approach from what this PR does which is to prioritize local images.

### DIFF
--- a/internal/pipe/ko/ko.go
+++ b/internal/pipe/ko/ko.go
@@ -219,10 +219,8 @@ func (o *buildOptions) makeBuilder(ctx *context.Context) (*build.Caching, error)
 				return ref, cached.(build.Result), nil
 			}
 			localImage, err := daemon.Image(ref)
-			if err != nil {
-				if !errdefs.IsNotFound(err) {
-					return nil, nil, err
-				}
+			if err != nil && !errdefs.IsNotFound(err) {
+				return nil, nil, err
 			}
 			if localImage != nil {
 				baseImages.Store(o.baseImage, localImage)


internal/pipe/ko/ko.go
```diff
@@ -215,6 +217,17 @@ func (o *buildOptions) makeBuilder(ctx *context.Context) (*build.Caching, error)
...
+				localImage, err := daemon.Image(ref)
+				if err != nil {
+					if !errdefs.IsNotFound(err) {
+						return nil, nil, err
+					}
+				}
+				if localImage != nil {
+					baseImages.Store(o.baseImage, localImage)
+					return ref, localImage, err
+				}
```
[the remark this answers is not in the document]
scav: Just to add to this:

The docker daemon supports holding multi arch builds the same way a repository does and just as there is no guarantee that a base image from a remote repo is multi arch we cannot guarantee it here either.

However, on an `arm64` machine, when running build `docker buildx build --platform linux/amd64 [...]` and setting this up in .goreleaser, it works just fine.
```yaml
    platforms:
      - linux/amd64
```

caarlos0:
looks good, thank you!

internal/pipe/ko/ko_test.go
```diff
@@ -621,7 +622,12 @@ func TestPublishLocalBaseImage(t *testing.T) {
 					Tags:         []string{"latest", "{{.Tag}}"},
 					BaseImage:    "local-base:latest",
 					SBOM:         "none",
-					Platforms:    []string{"linux/arm64"},
+					Platforms: []string{fmt.Sprintf("%s/%s", func() string {
+						if runtime.GOOS != "linux" {
+							return "linux"
+						}
+						return runtime.GOOS
```
caarlos0: isn't it always `"linux/"+runtime.GOOS`?

scav: Absolutely! I have no idea why I did that. Probably age and attempted multi tasking. It has been simplified now.

www/docs/customization/ko.md
```diff
@@ -38,6 +38,8 @@ kos:
 
     # Base image to publish to use.
     #
+    # Local images will take priority over fetching remote images
```
caarlos0:
```suggestion
    # Local images will take priority over fetching remote images (since v2.13-unreleased).
```

internal/pipe/ko/ko.go
```diff
@@ -215,6 +217,17 @@ func (o *buildOptions) makeBuilder(ctx *context.Context) (*build.Caching, error)
 
 			if cached, found := baseImages.Load(o.baseImage); found {
 				return ref, cached.(build.Result), nil
+			} else {
```
caarlos0: lint error here, can drop the else

internal/pipe/ko/ko.go
caarlos0:
```suggestion
			}
			localImage, err := daemon.Image(ref)
			if err != nil {
				if !errdefs.IsNotFound(err) {
					return nil, nil, err
				}
			}
			if localImage != nil {
				baseImages.Store(o.baseImage, localImage)
				return ref, localImage, err
```

internal/pipe/ko/ko.go
```diff
@@ -216,6 +218,16 @@ func (o *buildOptions) makeBuilder(ctx *context.Context) (*build.Caching, error)
 			if cached, found := baseImages.Load(o.baseImage); found {
 				return ref, cached.(build.Result), nil
 			}
+			localImage, err := daemon.Image(ref)
+			if err != nil {
+				if !errdefs.IsNotFound(err) {
+					return nil, nil, err
+				}
+			}
```
caarlos0:
```suggestion
			if err != nil && !errdefs.IsNotFound(err) {
				return nil, nil, err
			}
```

caarlos0:
thanks so much 🙏

PR #6246, opened 2025-11-09
fix: dynamically use announcer names in errors

FelicianoTech:
This PR uses this setup in announcer error messages. It also fixes an error where the Mattermost announcer mistakenly refers to itself as Teams. Creating this PR was brought up in #6199.


caarlos0:
thanks!

PR #6252, opened 2025-11-10
fix(deps): update charm.land, use forked (clean) quill

caarlos0:
- uses charm.land/lipgloss/v2 
- updates log library 
- updates fang
- replaces quill with a our fork, which removes the TUI, leaving only the things we actually use


caarlos0:
it seems like it'll probably be easier to copy the parts of quill we actually use inside our codebase. So I'll close this for now, eventually I'll make the time to do this.

PR #6300, opened 2025-12-18
feat(mcp): always use current schema url

tjhop:
Found this during publication of
https://github.com/tjhop/prometheus-mcp-server/releases/tag/v0.13.0.

Previously, goreleaser was hard coding the MCP schema registry URL. The
`2025-10-17` schema is no longer supported, resulting in publication
errors like the following:

```
 ⨯ release failed after 2m48s
    error=
    │ 1 error occurred:
    │ 	* mcp registry: got status code 400: {"title":"Bad Request","status":400,"detail":"Failed to publish server","errors":[{"message":"schema version https://static.modelcontextprotocol.io/schemas/2025-10-17/server.schema.json is not supported. Please use schema version 2025-12-11"}]}
Error: The process '/opt/hostedtoolcache/goreleaser-action/2.13.1/x64/goreleaser' failed with exit code 1
```

The registry package already defines the current/latest schema version,
as well as the fully assembled URL, so let's use that.

Note: This would require bumping the version of the registry package and
putting out a fix/patch release to get the MCP publication working
again. We're already in that boat with hard-coding, but now we can rely
on tools like dependabot/renovate to watch for package updates for us.
It also reduces the schema update process from `sed -i s/old/new/g` to a
simple package bump, rebuild, and publish.

Signed-off-by: TJ Hoplock <t.hoplock@gmail.com>


caarlos0 (approved):
good catch!

tjhop:
Last CI publication to the registry went through without issue. Thanks for getting this out so quick! :heart:

PR #6310, opened 2025-12-20
fix(deps): use quill fork

caarlos0:
this removes the dependencies on bubbletea, and old versions of lipgloss


wagoodman:
@caarlos0 would it be more helpful to have quill make a separate go module in the `cmd` package, thus isolating entries in the go.sum away from lib users? That way there would be no go.sum entries for things only used by the cmd package (such as bubbletea and libgloss).

The log package [defaults](https://github.com/anchore/quill/blob/main/internal/log/log.go#L13) to a [discard logger](https://github.com/anchore/go-logger/blob/main/adapter/discard/logger.go) and doesn't pull in any actual logger (such as logrus). Though logrus would be in the go.sum, as long as there is no package using logrus in your code it would not be pulled into the built binary.

Anyway, we were thinking about adding these nested go modules if they would be helpful for you.

caarlos0:
@wagoodman yes that would be awesome! in the fork i stripped down everything that wasn't "necessary" from a lib standpoint, but something in the middle that works for both cases would be amazing as well! 

PS: i was going to ping you all eventually about it hehe 
PPS: happy to help in this front in any way i can 🙏

PR #6322, opened 2025-12-27
feat: extract co-authors

### LINKED ISSUE #6292: Mention in logs should consider `Co-authored-by` as well
In logs, commits only include the commit author for now, in form of `(@user)`.

When one or more `Co-authored-by` lines are included in the commit, these users should be mentioned as well: `(@user and @user)` or `(@user, @user and @user)`.

caarlos0:
This is a first pass at trying to extract the co-authors from the changelog.

Current state:

- it is able to extract name + email from commit messages in GitHub, Git, Gitea, and Gitlab 
- it might be able to find a co-author username on GitHub (e.g. if they user their `@users.noreply.github.com` email OR if their email is public in their profile)
- it doesn't try to find the username on gitea and gitlab (yet)

closes #6292

### DIFF
--- a/internal/changelog/changelog.go
+++ b/internal/changelog/changelog.go
@@ -1,3 +1,4 @@
+// Package changelog holds the basic changelog struct and shared logic.
 package changelog
 
 import (


internal/client/github.go
```diff
@@ -140,6 +149,28 @@ func (c *githubClient) Changelog(ctx *context.Context, repo Repo, prev, current
 	return log, nil
 }
 
+func (c *githubClient) authorsLookup(ctx *context.Context, authors []Author) []Author {
+	cache := map[string]string{}
+	for i := range authors {
+		author := &authors[i]
+		if before, ok := strings.CutSuffix(author.Email, "@users.noreply.github.com"); ok {
```
andreynering, line 166: Double-check before proceeding, but IIRC the email style can change a bit per user. AFAIK old users take only the username, but for new users it also take the user ID (in case the user changes the username later).

- Classic: `andreynering@users.noreply.github.com`
- New: `andreynering+1234567890@users.noreply.github.com`

Stripping out anything after `+` should get we covered.

caarlos0: good catch!

PR #6326, opened 2025-12-30
feat: artifacts and metadata JSON schema output

scop:
Add ability to output JSON schema for artifacts.json and metadata.json.

See discussion in https://github.com/orgs/goreleaser/discussions/6295

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,6 +8,8 @@
 
 www/docs/static/schema.json linguist-generated=true
 www/docs/static/schema-pro.json linguist-generated=true
+www/docs/static/schema-artifacts.json linguist-generated=true
+www/docs/static/schema-metadata.json linguist-generated=true
 www/docs/static/releases.json linguist-generated=true
 www/docs/static/releases-pro.json linguist-generated=true
 
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -103,17 +103,27 @@ tasks:
     desc: Generate JSONSchema
     cmds:
       - go run . schema -o ./www/docs/static/schema.json
+      - go run . schema artifacts -o ./www/docs/static/schema-artifacts.json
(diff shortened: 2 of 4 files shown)


scop:
This is WIP/draft/call for initial comments. At least docs are missing, probably other things as well. But the basic generation does work.

caarlos0:
hey! thanks for looking into this!

the main point for this is to use it with tools like `jq`, I'm assuming?

if so, maybe the jsonschema is not as useful as a docs page somewhere?

what do you think?

scop:
Yes, docs is the most important thing for this, I agree.

Then again having a JSON schema done like this is useful in the sense that it is "integrated" with the code and less subject to bitrot than if it was a separate, manually maintained document.

To me, JSON schema serves as documentation just fine. But I appreciate others might prefer some other form of documentation, but that could then again be (automatically) generated from the JSON schema.

scop:
Also, regarding the artifact struct and schema: the `Extra` map contains some useful information about artifacts, but that being just a string to any map, it isn't naturally captured to the schema in any useful sense and it's harder to figure out the possible data in it from browsing the code.

Maybe consider converting at least some of those Extra things into regular artifact fields? For backwards compatibility, it would make sense to keep them in also in `extra` in JSON for some time though.

caarlos0:
yeah, that part will be hard to do automatically.

I think that we can just document manually for now. We don't need to document everything, just the common ones:

- all the fields in the artifact struct
- all the extra fields defined in artifacts.go (which are the ones used in multiple places)
- maybe add a note that there might be other fields, and also add a note into the artifacts package to update the docs on change. 

Let me come up with that.

caarlos0:
see https://github.com/goreleaser/goreleaser/pull/6341

scop:
Awesome, thanks!

Issue #3759, opened 2023-02-06
Support building and pushing images through `buildkit` without a docker daemon


valerian-roche:
### Is your feature request related to a problem? Please describe.

In some restricted environments, machines executing build code (e.g. CI) may not be running a local docker daemon and rely on buildx to handle docker images build on remote buildkit executors.
In this context the current use of `--load` in the `docker buildx build` command (https://github.com/goreleaser/goreleaser/blob/main/internal/pipe/docker/api_docker.go#L76) is not working as this requires a docker daemon to push the image to. The use of `--push` instead allows to completely build and push images without a docker daemon. It is also not an issue for manifests as those do not use the docker daemon at all.

### Describe the solution you'd like

When using buildx builder, allow to push the image directly from the remote builder rather than the daemon. 
As this can impact current workflows (as the machine pushing the image might then be different from the current one), a specific flag might be needed to not break existing use cases. 
In the general case I would expect this change to improve user experience as it would avoid the transfer of the image to/from the client machine. When the remote executor is running in a higher bandwidth environment (e.g. public cloud), this might greatly improve the build speed. 

### Describe alternatives you've considered

The main alternatives I considered:
 - use flags in a way allowing to push through the build command: this nearly works, but `docker buildx build` does not override `load` with `push` but errors if both are set. As the `load` flag is hardcoded in goreleaser, this cannot be avoided and will require the docker daemon to be available for the client
 - configure the client to use a full-fledge docker daemon. This would likely work, but in constrained environment this is potentially out of user control. A major advantage of buildkit is to allow the build of oci images without the need for a privileged docker daemon (e.g. by running buildkit executors on a kubernetes cluster)
 - bypass the docker build part in goreleaser, and only use it to build go. This would work, but is somewhat defeating the goal of goreleaser

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Additional context

I checked on the code side if there was an easy way to do this change. It is non-trivial given the currently provided interfaces, as the `imager` interface assumes distinct `build` and `push` phases. 
It can be done by running the `build` command without the `load` flag if requested and then running the `push` command with the same build flags and the `--push -q` flags, but this leads to strict coupling between the two phases that breaks the context. 
If the imager object was not common to all goroutines, some context could be passed but this again requires a significant rewrite of the interface

caarlos0:
this would probably have to be implemented more or less like we implemented the Ko integration.

[...]

PR #6348, opened 2026-01-21
feat(dockers/v2): allow to skip load image in snapshot for daemonless clients

BaptisteLalanne:
# Context

Related to #3759 - While docker v2 works fine with `goreleaser release` on daemonless environment and remote buildkit drivers (ex: Kubernetes driver), snapshot builds fails.  

GoReleaser forces the --load option on the snapshot mode. The --load option requires a local Docker daemon.

# Changes

Implements automatic Docker daemon detection for snapshot builds:                                                                                                                                                                 
                                                                                                                                                                                                                                     
- GoReleaser now automatically detects if a Docker daemon is accessible by running `docker info`                                                                                                                                  
- When no daemon is detected the `--load` flag is automatically skipped                                                                                                    
   - Logs a clear message: `"snapshot build: skipping --load option because cannot connect to docker daemon"`                                                                                                                        
   - No configuration needed 


This enables daemonless CI environment to run `goreleaser release --snapshot` with docker_v2 and make sure the image can be built. The `goreleaser release` already works without a daemon. There is also the thin benefit of having the build for the snapshot that is equivalent to the release one.

### DIFF
--- a/internal/pipe/changelog/changelog_test.go
+++ b/internal/pipe/changelog/changelog_test.go
@@ -1264,7 +1264,7 @@ func TestFormatEntry(t *testing.T) {
 				{Username: "bob"},
 				{Username: "alice"}, // duplicate
 				{Username: "charlie"},
-				{Username: "bob"},   // duplicate
+				{Username: "bob"}, // duplicate
 			},
 		}
 		result, err := formatEntry(ctx, entry)


caarlos0:
i wonder if we could detect that, and do it automatically when there's no docker daemon 🤔

caarlos0:
I guess we could `docker buildx inspect` and see the driver? 🤔

BaptisteLalanne:
> i wonder if we could detect that, and do it automatically when there's no docker daemon 🤔

🎯  sure, I like it a lot

BaptisteLalanne:
@caarlos0 I went with an automatic switch as suggested, I think that makes a lot of sense

internal/pipe/docker/v2/docker.go
```diff
@@ -98,16 +98,28 @@ func (p Snapshot) Run(ctx *context.Context) error {
...
+		if !canLoad {
+			log.WithField("id", d.ID).
+				Info("snapshot build: skipping --load option because cannot connect to docker daemon")
 			g.Go(func() error {
-				// buildx won't allow us to `--load` a manifest, so we create
-				// one image per platform, adding it to the tags.
-				d := ctx.Config.DockersV2[i]
-				d.Platforms = []string{plat}
-				return buildImage(ctx, d, "--load")
+				return buildImage(ctx, d, []string{}...)
```
caarlos0, line 111: wouldn't this also push the manifest?

I _think_ we should create the separated images when its a snapshot, because AFAIK you can't really create a manifest without pushing it 🤔

BaptisteLalanne: > wouldn't this also push the manifest?

In snapshot case we wouldn't even create a manifest but just do the platform builds and that's it. Without the `--load` (equivalent `--output=type=docker`) we will default to `--output=type=cacheonly`
https://docs.docker.com/build/exporters/#cache-only-export

It won't push nor create the manifest
https://github.com/docker/buildx/blob/master/build/opt.go#L399-L406

https://github.com/docker/buildx/blob/44945d71ff077ce7fc142fbdee6acec8d9acb630/build/build.go#L272


`docker buildx build --platform linux/amd64,linux/arm64 -t test-multiplatform:latest .`
```
#0 building with "ci" instance using kubernetes driver
#1 [internal] load build definition from Dockerfile
#1 transferring dockerfile: 200B done
#1 DONE 0.0s
#2 [internal] load build definition from Dockerfile
#2 transferring dockerfile: 200B done
#2 DONE 0.0s
#3 [linux/arm64 internal] load metadata for registry.xx.xx/docker:27.3.1@sha256:e3175bd12c87162e984992671a307a06ef51edca6a72f61fe543980344d9468d
#3 DONE 0.0s
#4 [linux/amd64 internal] load metadata for registry.xx.xx/docker:27.3.1@sha256:e3175bd12c87162e984992671a307a06ef51edca6a72f61fe543980344d9468d
#4 DONE 0.0s
#5 [internal] load .dockerignore
#5 transferring context: 2B done
#5 DONE 0.0s
#6 [internal] load .dockerignore
#6 transferring context: 2B done
#6 DONE 0.0s
#7 [linux/arm64 1/2] FROM registry.xx.xx/docker:27.3.1@sha256:e3175bd12c87162e984992671a307a06ef51edca6a72f61fe543980344d9468d
#7 resolve registry.xx.xx/docker:27.3.1@sha256:e3175bd12c87162e984992671a307a06ef51edca6a72f61fe543980344d9468d done
#7 DONE 0.0s
#8 [linux/arm64 2/2] RUN echo "Hello from $(uname -m)"
#8 CACHED
#9 [linux/amd64 1/2] FROM registry.xx.xx/docker:27.3.1@sha256:e3175bd12c87162e984992671a307a06ef51edca6a72f61fe543980344d9468d
#9 resolve registry.xx.xx/docker:27.3.1@sha256:e3175bd12c87162e984992671a307a06ef51edca6a72f61fe543980344d9468d done
#9 extracting sha256:5d67b1ae10e433f1defd83b179460272526a8d858241f29fa5b5a429e9207998
#9 extracting sha256:5d67b1ae10e433f1defd83b179460272526a8d858241f29fa5b5a429e9207998 1.3s done
#9 DONE 1.3s
#9 [linux/amd64 1/2] FROM registry.xx.xx/docker:27.3.1@sha256:e3175bd12c87162e984992671a307a06ef51edca6a72f61fe543980344d9468d
#9 extracting sha256:58e4fbfd5d507b220677fa06b834a49503c50d22b0daeedbdea450413beb6d71 done
#9 extracting sha256:136d802b80b3351fd2975b2ba725199d43cdb64d40b93896ca0041ed83bb99e4 done
#9 DONE 1.3s
#10 [linux/amd64 2/2] RUN echo "Hello from $(uname -m)"
#10 0.045 ++ uname -m
#10 0.046 + echo 'Hello from x86_64'
#10 0.046 Hello from x86_64
#10 DONE 0.1s
WARNING: No output specified with kubernetes driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
```

caarlos0: ah, right! cool! I think this should be ready to merge then

caarlos0:
Thank you! Assuming CI passes, let's get this in 🙏

PR #6364, opened 2026-01-29
Fix/6363

### LINKED ISSUE #6363: Generated nix expressions may miss autoPatchelfHook in parameter list
### What happened?

When GoReleaser generates nix expressions for dynamically linked binaries, `autoPatchelfHook` is included in `nativeBuildInputs`, but not in the parameter list at the top of the nix file. This leads to failures during nix evaluation because `autoPatchelfHook` is undefined. The golden files (e.g. internal/pipe/nix/testdata/TestDynamicallyLinked.golden) show this issue as the generated nix output requires `autoPatchelfHook` to be passed as a function argument.

### Expected behavior

If `autoPatchelfHook` is in `nativeBuildInputs`, it should also be present in the parameter list at the top of the nix expression (next to e.g. installShellFiles) so the generated file evaluates correctly.

### Context / References

- Code position responsible: [internal/pipe/nix/nix.go](https://github.com/goreleaser/goreleaser/blob/main/internal/pipe/nix/nix.go) lines 293ff, logic for dynamically linked binaries
- Related downstream issue: [charmbracelet/nur#50](https://github.com/charmbracelet/nur/issues/50) — this is breaking downstream consumers as `autoPatchelfHook` is undefined.
- Example output: [internal/pipe/nix/testdata/TestDynamicallyLinked.golden](https://github.com/goreleaser/goreleaser/blob/main/internal/pipe/nix/testdata/TestDynamicallyLinked.golden)

### Minimal reproducible example

- Build any GoReleaser project that results in a dynamically linked binary and check the generated nix expression

### Additional details

The fix is likely to add `autoPatchelfHook` to the dependencies list in addition to the inputs slice ([suggested code change](https://github.com/goreleaser/goreleaser/blob/main/internal/pipe/nix/nix.go#L293)): 

```go
if !dynamicallyLinked && artifact.ExtraOr(*arch, artifact.ExtranDynLink, false) {
    inputs = append(inputs, "autoPatchelfHook")
    dependencies = append(dependencies, "autoPatchelfHook") // <-- missing
    dynamicallyLinked = true
}
```

---

Downstream issue: https://github.com/charmbracelet/nur/issues/50

malikwirin:
## Summary

This PR fixes an issue where generated Nix expressions for dynamically linked binaries would fail during evaluation because `autoPatchelfHook` was missing from the function parameter list.

## Problem

When GoReleaser generates Nix expressions for dynamically linked binaries, `autoPatchelfHook` is added to `nativeBuildInputs` but was not being added to the `dependencies` list. This meant it wouldn't appear in the parameter list at the top of the generated `.nix` file, causing Nix evaluation to fail with an undefined variable error.

## Solution

Added `autoPatchelfHook` to the `dependencies` slice alongside the `inputs` slice when a dynamically linked binary is detected:

```go
if !dynamicallyLinked && artifact.ExtraOr(*arch, artifact.ExtranDynLink, false) {
    inputs = append(inputs, "autoPatchelfHook")
    dependencies = append(dependencies, "autoPatchelfHook") // added
    dynamicallyLinked = true
}
```

## Changes

- `internal/pipe/nix/nix.go`: Add `autoPatchelfHook` to `dependencies` when processing dynamically linked binaries
- `internal/pipe/nix/testdata/TestDynamicallyLinked.golden`: Update golden file to include `autoPatchelfHook` in the parameter list

## Testing

- Ran `go test -run TestDynamicallyLinked -update` to regenerate the golden file

Resolves #6363

### DIFF
--- a/internal/pipe/nix/testdata/TestDynamicallyLinked.golden
+++ b/internal/pipe/nix/testdata/TestDynamicallyLinked.golden
@@ -6,6 +6,7 @@
   installShellFiles,
   stdenvNoCC,
   stdenv,
+  autoPatchelfHook,
 }:
 let
   inherit (stdenvNoCC.hostPlatform) system;


caarlos0:
thank you!

PR #6370, opened 2026-01-30
chore(winget): update manifest schema to latest (1.12.0)

vedantmgoyal9:
Use the latest schema `1.12.0` available on winget-pkgs.

The new schema adds some new fields and is fully backwards compatible with the previous version, but they do not affect goreleaser.


caarlos0:
thanks @vedantmgoyal9

PR #6376, opened 2026-02-04
fix(nix): autoPatchelfHook on Linux only

caarlos0:
refs https://github.com/charmbracelet/nur/issues/50

### DIFF
--- a/internal/pipe/nix/nix.go
+++ b/internal/pipe/nix/nix.go
@@ -291,8 +291,6 @@ func preparePkg(
 			dependencies = append(dependencies, "unzip")
 		}
 		if !dynamicallyLinked && artifact.ExtraOr(*arch, artifact.ExtranDynLink, false) {
-			inputs = append(inputs, "autoPatchelfHook")
-			dependencies = append(dependencies, "autoPatchelfHook")
 			dynamicallyLinked = true
 		}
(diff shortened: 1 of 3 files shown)


malikwirin:
I would exclude on Darwin instead becsuse on *bsd it would need patching as well I believe

caarlos0:
apparently there's an `isElf` as well? https://github.com/NixOS/nixpkgs/blob/6f656b00df45feee6bf8b6b13bba2ca3f62eb820/pkgs/development/compilers/rust/cargo.nix#L92

caarlos0:
don't know what we should do in this part then though:
```nix
  buildInputs = lib.optionals stdenvNoCC.isLinux [
    stdenv.cc.cc.lib
  ];
```

malikwirin:
> don't know what we should do in this part then though:
> 
> ```nix
>   buildInputs = lib.optionals stdenvNoCC.isLinux [
>     stdenv.cc.cc.lib
>   ];
> ```

Negating `isDarwin`?
I might look into it tomorrow

caarlos0:
oh, on the other hand - we only generate nix for darwin and linux, so this doesn't really matter right now. 

i think we can merge as is

thanks for taking a look, though!

PR #6388, opened 2026-02-11
sec: redact secrets from command outputs

caarlos0:
introduces a `redact.Writer` which we can use in `exec.Command` so it automatically redacts things that look like secrets from the output.

also updated the usage of `gerrors` so we have nicer outputs when things go awry, e.g.:

CleanShot 2026-02-11 at 23 01 03@2x


caarlos0:
closed in favor of #6396 and #6397

PR #6390, opened 2026-02-11
docs: deb compression related improvements

scop:


www/docs/customization/nfpm.md
```diff
@@ -460,6 +460,9 @@ nfpms:
       breaks:
         - some-package
 
+      # Data compression algorithm (gzip (default), xz, zstd or none).
```
scop, line 463: _Data_ compression instead of just compression in the comment, because control.tar can use another compression method (and I may have a look into adding support for changing it in nfpm). 

Edit:  maybe I won't bother looking into adding support for other control compression methods as the archive is typically so small that there's very little to be gained sizewise by switching let's say to zstd from gzip. Then again size might not be the only consideration, search for "uniform-compression" at https://manpages.debian.org/trixie/dpkg/dpkg-deb.1.en.html. If anything, maybe mimicking that uniform-compression behavior in nfpm would make sense.

caarlos0, line 463:
```suggestion
      # Data compression algorithm (gzip (default), xz, zstd or none).
      # 
      # <!-- md:inline_version v2.14-unreleased -->.
```

caarlos0:
Thanks!

PR #6432, opened 2026-03-07
chore(ci): switch artifact attestations gen to actions/attest

scop:
Ref https://github.com/actions/attest-build-provenance#usage

> As of version 4, actions/attest-build-provenance is simply a wrapper
> on top of actions/attest.


caarlos0:
I think we might have places in docs to fix this as well

caarlos0:
Thanks!

PR #6448, opened 2026-03-12
feat: flatpak

### LINKED ISSUE #2927: [Feature Request] add flatpak package building support

Some linux distributions use Flatpaks by default or don't have snap installed.

It would be great if goreleaser supported generating flatpak packages. 
Similar to what is currently done for snaps.

N/A 
Haven't considered any alternatives

### Search

### Code of Conduct

### Additional context

https://docs.flatpak.org/en/latest/

caarlos0:
closes #2927

This will build `.flatpak`s for the linux binaries!

It obviously only works on Linux and requires `flatpak` to be installed, but it works.

To work on this in macOS, I created a wrapper script and shell out to a VM using OrbStack.

I also tested that it works with:

```bash
orb -m flatpak-vm flatpak install --user -y --noninteractive --bundle /path/to/goreleaser_arm64.flatpak
orb -m flatpak-vm flatpak run --user com.goreleaser.GoReleaser --version
orb -m flatpak-vm flatpak info --user com.goreleaser.GoReleaser
```

And everything seems to work as expected!

### Flathub

I think that Flathub should probably be another implementation, more akin to winget, as it's basically generating the manifest and creating a PR for it – that is if we do choose to support it, and if so, we should do it at a later time in another PR.

---

let me know your thoughts on the subject.


daegalus (commented):
This looks good to me. And this method is a good first step.

PR #6462, opened 2026-03-17
fix: filterOut returns excluded tag when multiple ignore_tags are set

### LINKED ISSUE #6446: filterOut only checks first exclude pattern when multiple ignore_tags are configured
### Description

`filterOut` in `internal/pipe/git/git.go` has a logic bug when multiple `ignore_tags` are configured. It returns a tag as soon as it doesn't match **any one** exclude entry, instead of checking that it doesn't match **all** of them.

### Current behavior

```go
for _, tag := range tags {
    for _, exl := range exclude {
        if exl != tag {
            return tag  // BUG: returns immediately on first non-match
        }
    }
}
```

With `tags=["v0.0.3"]` and `exclude=["v0.0.3", "v0.0.2"]`:
1. Checks `v0.0.3 != v0.0.3` → false, continues
2. Checks `v0.0.3 != v0.0.2` → true → returns `v0.0.3`

The tag `v0.0.3` is returned even though it's in the exclude list.

### Expected behavior

`v0.0.3` should be excluded since it matches one of the entries. The fix should check that a tag doesn't match **any** exclude entry before returning it:

```go
for _, tag := range tags {
    excluded := false
    for _, exl := range exclude {
        if exl == tag {
            excluded = true
            break
        }
    }
    if !excluded {
        return tag
    }
}
```

### Impact

Only affects users with **multiple** `ignore_tags` entries where a tag matches one but not all patterns.

abhay1999:
## Problem

`filterOut` in `internal/pipe/git/git.go` has a logic bug when multiple `ignore_tags` are configured.

The inner loop returns a tag as soon as it does **not** match any single exclude entry, without checking the remaining entries. This means a tag that matches one entry but not another is incorrectly returned.

Example with `tags=["v0.0.3"]` and `exclude=["v0.0.3", "v0.0.2"]`:
1. Checks `v0.0.3 != v0.0.3` → false, continues
2. Checks `v0.0.3 != v0.0.2` → true → **returns `v0.0.3`** (wrong — it should be excluded)

## Fix

Track whether any exclude entry matched, and only return the tag when none did.

```go
for _, tag := range tags {
    excluded := false
    for _, exl := range exclude {
        if exl == tag {
            excluded = true
            break
        }
    }
    if !excluded {
        return tag
    }
}
```

## Tests

Added `TestFilterOut` with cases covering:
- No excludes → returns first tag
- Single exclude match
- Multiple excludes where the tag matches one entry (the bug case)
- Multiple excludes where the tag matches none
- All tags excluded → empty string
- Empty tag list → empty string

Fixes #6446

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,7 @@ A big **thank you** to all current, past, and future sponsors!
   <a href="https://github.com/encoredev" target="_blank" rel="noopener sponsored"><img src="https://avatars.githubusercontent.com/u/50438175?v=4" alt="Encore" width="64" height="64" style="border-radius: 8px; margin: 8px;"></a>
   <a href="https://about.gitea.com/" target="_blank" rel="noopener sponsored"><img src="https://images.opencollective.com/gitea/bf35c2f/logo/96.png" alt="Gitea" width="64" height="64" style="border-radius: 8px; margin: 8px;"></a>
   <a href="https://github.com/comet-ml" target="_blank" rel="noopener sponsored"><img src="https://avatars.githubusercontent.com/u/31487821?v=4" alt="Comet" width="64" height="64" style="border-radius: 8px; margin: 8px;"></a>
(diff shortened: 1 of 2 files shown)


caarlos0:
good find - thank you!

Issue #5450, opened 2025-01-13
standardize targets


caarlos0:
Go doesn't have a "target" per se, so we do it manually on our side.
In some places, it's all parts joined together, and sometimes we have an underline between each part.

Other builders use `-` to do it.

This might be a breaking change, but I *think* we should probably change our internal targets for Go to also split by `-`.

This would allow us to do things link `strings.Split(target, "-")` and then check if any of its parts is `linux`, for example.

The breaking changes this would introduce:

- manual targets in `builds.targets` (currently split by `_`, could make both work for the time being)
- what else?

caarlos0:
this will be done in v3 - https://github.com/orgs/goreleaser/discussions/6491

PR #6467, opened 2026-03-20
refactor: standardize internal target strings to use '-' separator

omnicoder9:
Change Go builder internal target strings from underscore-separated (linux_amd64_v1) to hyphen-separated (linux-amd64-v1), matching the convention already used by Rust, Zig, Bun, and Deno builders. Add backward compatibility by normalizing underscores to hyphens in fixTarget(), so existing configs with the old format continue to work. Update the universal binary target from darwin_all to darwin-all.
Go was the only builder using underscores as the target separator, which made it impossible to write generic cross-builder target handling code like strings.Split(target, "-"). Any feature that needs to inspect target components such as unified filtering, partial builds, or cross-builder matching had to special-case Go targets or use fragile heuristics. With this change, every target in the entire codebase can be parsed the same way regardless of which builder produced it.
BREAKING CHANGE: Go build targets now use hyphens (linux-amd64-v1) instead of underscores (linux_amd64_v1). This affects the .Target template variable, dist directory names, and builds.targets config values. The old underscore format is still accepted in config files and is silently normalized.

issue: https://github.com/goreleaser/goreleaser/issues/5450


caarlos0:
unfortunately this is a breaking change, as some people have scripts that depend on the `./dist/` structure. 

this is something we'll do on v3. I'm planning to launch v3 sometime later this year.

caarlos0:
closing for now, as as mentioned, this will be on v3.

see https://github.com/orgs/goreleaser/discussions/6491

caarlos0:
in the mean time, thanks for the pr!

Issue #4565, opened 2024-01-22
Integrating godownloader into goreleaser


kishaningithub:
### Is your feature request related to a problem? Please describe.

I would like to have the ability to publish an install script as part of the deployment process

### Describe the solution you'd like

goreleaser creates shell scripts like godownloader and publishes them.

### Describe alternatives you've considered

Hand written shell scripts

### Search

- [X] I did search for other open and closed issues before opening this

### Supporter

- [ ] I am a [sponsor](https://github.com/sponsors/caarlos0/) or a [Pro](https://goreleaser.com/pro) customer

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

caarlos0:
this is something I've been wanting to do for a while (since godownloader was created, actually).


but yeah, I need to do it eventually... good to have an issue for it as well


thanks!

kishaningithub:
@caarlos0 Let know if you need any help from the community on this.

andreynering:
We still rely heavily on godownloader for Task, despite being deprecated. Having this integrated officially would be awesome.

PR #6475, opened 2026-03-22
feat: install scripts

caarlos0:
made with Claude

tested to some degree 

not sure about any of this yet


caarlos0:
I don't think this makes sense. The mechanics are weird, you're not generating this script on every release, but generally only once.

Rather not do this right now.

Maybe we revive godownloader? I don't know, at this point it seems to me its easy enough to generate it with a LLM once and proceed with life.

Anyway, closing for now.

caarlos0:
https://github.com/goreleaser/goreleaser/issues/4565

PR #6498, opened 2026-03-28
fix(snapcraft): correct channel template args in fmt.Errorf

cuiweixie:
## What

Fixes incorrect arguments to `fmt.Errorf` in `processChannelsTemplates`: the format string uses `%s` for the channel template and `%w` for the error, but `err` was passed twice, so the template string was never included in the message.

## Why

Improves error messages when channel template execution fails (shows which template failed).


caarlos0:
good catch! thanks

PR #6499, opened 2026-03-28
fix(archive): use current binary for ExtraReplaces in skip

cuiweixie:
## What

In `skip()`, when emitting uploadable binary artifacts for skipped archives, `ExtraReplaces` was read from `binaries[0]` inside the loop over all binaries.

## Why

Each emitted artifact should reflect the **current** binary’s metadata so multiple binaries get the correct replaces data when they differ.


caarlos0:
good catch! thank you :D

PR #6502, opened 2026-03-29
test(cmd): cover replace+gomod.proxy snapshot vs release behavior

Iheanacho-ai:
This test verifies that a `go.mod` replace directive is permitted when running `release --snapshot`, but causes `release` to fail when the `--snapshot` flag is not used.

Supports the CI behaviour discussed in #2355.

To run test:

```
go test ./cmd/... -run TestReleaseGoModProxyReplaceNotCaughtBySnapshotAlone -count=1 -v ⁠
```

### DIFF
--- a/cmd/release_test.go
+++ b/cmd/release_test.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"os/exec"
 	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


caarlos0:
hi, thanks for the pr!

i believe this is already tested in `gomod_proxy_test.go`.

caarlos0:
> i believe this is already tested in gomod_proxy_test.go.

confirmed

PR #6511, opened 2026-03-30
test: require to use assert and require only

alexandear:
Refactor tests to use `assert` or `require` instead of `t.Error` and `t.Fatal`.

Extend `forbidigo.forbid` patterns to prevent future regressions.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -37,6 +37,9 @@ linters:
     forbidigo:
       forbid:
         - pattern: ioutil\.*
+        - pattern: ^testing\.T\.Errorf?$
+        - pattern: ^testing\.T\.Fatalf?$
+      analyze-types: true
     gocritic:
       disabled-checks:
         - appendAssign
--- a/internal/client/github_test.go
+++ b/internal/client/github_test.go
@@ -514,7 +514,7 @@ func TestGitHubOpenPullRequestCrossRepo(t *testing.T) {
 			return
 		}
 
-		t.Error("unhandled request: " + r.URL.Path)
+		assert.Fail(t, "unhandled request: "+r.URL.Path)
 	}))
 	t.Cleanup(srv.Close)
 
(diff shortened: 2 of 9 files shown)


caarlos0:
I don't see the point of switching these to testify - could you elaborate on why this is better? 

only reason i can think of is to "use testify all the way", but i'm not sure that's better either

alexandear:
> I don't see the point of switching these to testify - could you elaborate on why this is better?
> 
> only reason i can think of is to "use testify all the way", but i'm not sure that's better either

The main point is consistency and avoiding mixing different approaches. The goreleaser project already uses the `testify` library extensively.

The number of `assert` + `require` statements is 4962. The number of `t.Fail` + `t.Fatal` + `t.Error` is 221. From this we can conclude that `testify` is prevailing approach.

### How I counted

Use [rg](https://github.com/burntsushi/ripgrep):

```console
❯ rg --stats "assert\." -g "*\_test\.go" -q

180 matches
180 matched lines
14 files contained matches
165 files searched
0 bytes printed
1326584 bytes searched
0.030779 seconds spent searching
0.011137 seconds total

❯ rg --stats "require\." -g "*\_test\.go" -q

4782 matches
4782 matched lines
162 files contained matches
165 files searched
0 bytes printed
1326584 bytes searched
0.037167 seconds spent searching
0.024622 seconds total

❯ rg --stats "t.Fail" -g "*\_test\.go" -q

3 matches
3 matched lines
2 files contained matches
165 files searched
0 bytes printed
1326584 bytes searched
0.002091 seconds spent searching
0.010646 seconds total

❯ rg --stats "t.Fatal" -g "*\_test\.go" -q

18 matches
18 matched lines
6 files contained matches
165 files searched
0 bytes printed
1326584 bytes searched
0.000996 seconds spent searching
0.028437 seconds total

❯ rg --stats "t.Error" -g "*\_test\.go" -q

200 matches
200 matched lines
63 files contained matches
165 files searched
0 bytes printed
1326584 bytes searched
0.001623 seconds spent searching
0.021781 seconds total
```

caarlos0:
yes, i'm aware of that, but honestly this doesn't seem to be worth it

thanks for the pr, though.

PR #6513, opened 2026-03-31
docs: Fix signature verification of checksum file

nicholasdille:
The fix correctly mentions the necessary files for signature verification.

After the switch to the sigstore bundle, the commands for signature verification were updated but the text still mentioned the old signature and certificate files.


caarlos0:
good catch - thanks!

PR #6518, opened 2026-04-01
feat(docker): check for buildx plugin in dependencies

SiddharthShrimali:
### Summary
Address TODO: Added an explicit dependency check for the `docker-buildx` plugin in the Docker pipe.

### Description
This PR implements a "fail-fast" mechanism by adding `docker-buildx` to the list of required dependencies during the validation phase. 

Currently, GoReleaser checks for the `docker` binary but assumes `buildx` is present if configured. This change ensures that if a user has Docker installed but is missing the buildx plugin, the release process stops immediately with a clear error message rather than failing deep into the build cycle.

As noted in the codebase's existing **TODOs**, we should verify that the environment has all required plugins before execution. This improves the developer experience by providing immediate feedback on environment requirements.

### Changes
* Updated `internal/pipe/docker/docker.go` to include `docker-buildx` in the `Dependencies` return slice when `use: buildx` is detected.
* Improved error reporting by ensuring the binary check covers the specific plugin.

### Technical & Testing Note
- **Local Testing:** I observed some test failures locally due to environment-specific issues on macOS (Go toolchain version mismatch `1.26.1` vs `1.26.0` and Docker socket access). 
- **Verification:** I stashed my changes and confirmed that these failures persist on the `main` branch as well. I am relying on the GitHub Actions CI pipeline to provide a clean-room verification of the logic.

### DIFF
--- a/internal/pipe/docker/docker.go
+++ b/internal/pipe/docker/docker.go
@@ -48,20 +48,26 @@ func (Pipe) Skip(ctx *context.Context) bool {
 	return len(ctx.Config.Dockers) == 0 || skips.Any(ctx, skips.Docker)
 }
 
-func (Pipe) Dependencies(ctx *context.Context) []string {
-	var cmds []string
+func (Pipe) Dependencies(ctx *context.Context) []func() (string, error) {
+	var checks []func() (string, error)
 	for _, s := range ctx.Config.Dockers {
 		switch s.Use {
 		case useDocker, useBuildx:
-			cmds = append(cmds, "docker")
+			checks = append(checks, func() (string, error) {
+				_, err := exec.LookPath("docker")
+				return "docker", err
+			})
+
 			if s.Use == useBuildx {
-				if err := checkBuildx(); err != nil {
-					log.Warn(logext.Warning(err.Error()))
-				}
+				checks = append(checks, func() (string, error) {
+					return "docker buildx", checkBuildx()
+				})
 			}
+
+			return checks
 		}
 	}
-	return cmds
+	return checks
 }
 
 // Default sets the pipe defaults.
@@ -423,12 +429,8 @@ func isRetriablePush(err error) bool {
 }
 
 func checkBuildx() error {
-	if _, err := exec.LookPath("docker"); err != nil {
-		return errors.New("docker not found in PATH")
-	}
-	cmd := exec.Command("docker", "buildx", "version")
-	if err := cmd.Run(); err != nil {
-		return errors.New("docker buildx is not installed")
+	if err := exec.Command("docker", "buildx", "version").Run(); err != nil {
+		return fmt.Errorf("docker buildx plugin is not installed or not working: %w", err)
 	}
 	return nil
 }
--- a/cmd/healthcheck.go
+++ b/cmd/healthcheck.go
@@ -1,11 +1,8 @@
 package cmd
 
 import (
-	stdctx "context"
 	"errors"
 	"io"
-	"os/exec"
-	"strings"
 	"sync"
 
 	"charm.land/lipgloss/v2"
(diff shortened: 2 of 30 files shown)


internal/pipe/docker/docker.go
caarlos0: we'll need to refactor this interface I think, something that allows to return a `[]func() (string, error)` or something like that. 

the warning in l59 would be helpful but would not exit 1 🤔

SiddharthShrimali: Thanks for the review @caarlos0.

The refactor makes sense, since shifting from returning a slice of strings to a slice of function checks allows much more granular validation irrespective of whether it is a binary or not, solving the issue where a dependency isn't just a binary in the `PATH`.

We can also ensure we trigger a **hard failure** during the validation phase instead of just a warning, so that the build stops early.

I will start working on this.. 
A couple of questions: 
* Should I apply this signature change to the `Pipe` interface **globally** across all pipes, or would you prefer I implement it here first as a poc?
* For the check result, should it always return an `error` to trigger an `exit 1`, or do we want to support a "warning-only" mode for certain checks?

caarlos0: nice!

i would change all pipes, as this is an internal interface, it should be fine.

we can return errors, we can use `internal/gerrors` if we need to customize exit codes, but the behavior of `goreleaser hc` right now is: check all, show the ones who failed, and then exit 0 or exit 1. I _think_ we can keep it that way 🤔

caarlos0: BTW, i just remembered: I have this implemented in the pro codebase. Let me copy it over from there in another PR.

caarlos0: it is here as well, see `internal/pipe/docker/v2`, func `Healthcheck`

SiddharthShrimali: ## Update: Refactor of Healthcheck Dependency Architecture

I have completed the core logic refactor for the healthcheck system. 
the `Builder` and `Pipe` interfaces, along with all internal pipes (Docker, Sign, Nix, SBOM, etc.), now implement the requested functional signature: 

`Dependencies(*context.Context) []func() (string, error)`

### Key Changes
* **Docker Buildx:** implemented a `checkBuildx()` helper to verify the plugin specifically when `use: buildx` is set in the configuration.
* **Decoupled Logic:** healthcheck now returns "check functions" instead of simple strings. This allows the system to verify tool health (like running `docker buildx version`) only when the healthcheck command is actually executed, avoiding side effects during the initial dependency gathering.
* **Global Architecture:** updated `pkg/build`, `pkg/healthcheck`, and all `internal/pipe` implementations.

### Current status of the project:
* **Build:** the project compiles successfully with `go build ./...`.
* **Testing:** there are currently several unit test failures in the CI (e.g., `docker_test.go`, `sign_test.go`). these are **assertion mismatches** because the existing tests expect `[]string` but are receiving the new `[]func`.

### I had a few questions:
Before I manually update every individual test assertion to execute these functions and verify the returned strings, I wanted to check:
1. Do you have a preferred helper or pattern you'd like me to use for testing these function slices?
2. Is the current architecture correct before I finalize the test suite?

caarlos0: this comment looks awfully ai-generated.

also see my previous comments

caarlos0: https://github.com/goreleaser/goreleaser/pull/6526

SiddharthShrimali: > this comment looks awfully ai-generated.
> 
Yes, sorry for that, i did use ai to help me summarize my changes, so that i could convey them in a better way. But, the refactor was manual on your feedback. I am still getting used to the project's discussions.. I'll keep it more direct moving forward.

I see that you have opened #6526 to address this, so do i close this one to make it easier for the project?

caarlos0: yes, thanks 🙏

PR #6556, opened 2026-04-12
fix(homebrew_cask): sha256 before url in per-arch blocks

kkd16:
Follow-up to #6466, which only reordered the top-level stanzas in cask.rb. brew style's Cask/StanzaOrder also requires sha256 before url inside on_intel/on_arm blocks.

see https://docs.brew.sh/Cask-Cookbook#stanza-order


caarlos0:
Thanks!

caarlos0 (approved):

Issue #6482, opened 2026-03-23
Support specifying libc versions for targets passed into zigbuild


andriygm:
creating from https://discord.com/channels/890434333251362866/890434334803247126/1485652797255979149.

for the rust builder, we should support specifying libc versions for targets passed into zigbuild https://github.com/rust-cross/cargo-zigbuild?tab=readme-ov-file#specify-glibc-version e.g. appending .`2.17` to generated target to compile against libc `2.17`? currently, can implement a bit hackily using a wrapper to manually inject this:
```bash
#!/bin/bash
args=()
for arg in "$@"; do
  args+=("${arg/unknown-linux-gnu/unknown-linux-gnu.2.17}")
done
exec cargo-zigbuild "${args[@]}"
```
where goreleaser builder gets configured with:
```yaml
builds:
  - builder: rust
    tool: .github/scripts/cargo-zigbuild-glibc228.sh
```
could implement as a specific libc parameter but maybe more generally allowing customization of the template used to generate the rust target via the OS/arch pair?

caarlos0:
yes we should probably support versioned targets!

caarlos0:
do you know if this is documented anywhere? I didn't find it... 

Generally I think we could make `rust/targets.go` accept these, but I would like to know what other things I might need to think about when doing it

andriygm:
@caarlos0 the github link above should take you to the cargo-zigbuild readme, will paste here for reference:
>By default --target for *-gnu will have Zig implicitly build for a default version of glibc that varies based on the release of Zig ([v12 to v14 releases default to glibc 2.28](https://github.com/ziglang/zig/blob/0.14.1/lib/std/Target.zig#L473)).

>To build for a specific minimum glibc version, add that version as a suffix to the --target value. For example, to compile with --target aarch64-unknown-linux-gnu for glibc 2.17:

>```cargo zigbuild --target aarch64-unknown-linux-gnu.2.17```

caarlos0:
oh, i'm absolutely blind, sorry 😅

caarlos0:
see https://github.com/goreleaser/goreleaser/pull/6492

mozzieongit:
Hi,
it seems to me that it only works if I specify the target using the TARGET environment variable (`TARGET=x86_64-unknown-linux-gnu.2.38 goreleaser build --verbose --snapshot --clean --single-target`) with the "normal" target in the goreleaser config file:

```yml
# default config file generated by goreleaser init
# ...
[...]

PR #6557, opened 2026-04-14
Allow custom glibc version in config file's targets

mozzieongit:
This change will strip the custom glibc version target (as introduced in https://github.com/annihilatorrrr/goreleaser/commit/6f074fe80b75b2db5009eb8b3e6d92e8412b8f43) for the `rustup target add` command, as well as add documentation for specifying a custom glibc version.

The custom glibc version target could not yet be used within the goreleaser config file's `targets` directive. With this change it is possible to specify the glibc version as part of the target in the config file.

[Original Issue about custom glibc versions](https://github.com/goreleaser/goreleaser/issues/6482)
[Further context for this PR](https://github.com/goreleaser/goreleaser/issues/6482#issuecomment-4243044999)


caarlos0:
Thanks!

PR #6560, opened 2026-04-15
fix(build): set per-binary artifact IDs for ./... builds

caarlos0:
When main=./... resolves to multiple mains and no explicit ID is set, each artifact now gets the binary name as its ID instead of sharing the build-level default. Errors if user sets an explicit ID with a multi-main ellipsis path.

see: https://github.com/goreleaser/goreleaser/discussions/6559

### DIFF
--- a/internal/builders/golang/build.go
+++ b/internal/builders/golang/build.go
@@ -582,6 +582,10 @@ func checkBuildElipsis(
 		return nil, nil, err
 	}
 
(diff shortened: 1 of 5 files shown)


till:
Will my `all` hack still work?

caarlos0:
nope, now each will have an id

but if you want all binaries, you can simply not set any ids 🙏🏻

PR #6566, opened 2026-04-16
fix(docker): install syft, cosign, and buildx from upstream releases

kevinburkesegment:
Alpine-packaged binaries bundle google.golang.org/grpc versions older than the 1.79.3 fix for CVE-2026-33186. Pin and download the tools directly from their GitHub releases instead.

- syft 1.42.4 (grpc 1.79.3)
- cosign 3.0.6 (grpc 1.80.0)
- docker-buildx 0.33.0 (grpc 1.79.3)

CVE-2026-33186 allows an attacker to bypass 'deny' rules set for a GRPC server. While it's possible this is not used or exploitable there's a lot of code in here and it seems safest to update to non-vulnerable software versions retrieved directly from the source. https://nvd.nist.gov/vuln/detail/CVE-2026-33186

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,13 @@
+# Pull syft, cosign, and docker-buildx from their upstream images so we
+# control the dependency versions (Alpine packages bundle older grpc, see
+# CVE-2026-33186) and Dependabot can keep these tags up to date.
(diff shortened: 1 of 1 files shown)


caarlos0:
we should probably copy from their own docker images instead, so dependabot can auto update things 🤔

caarlos0:
see #6576

caarlos0:
thanks, merged #6576

PR #6570, opened 2026-04-17
docs(templates): note git fields unavailable in top-level env block

### LINKED ISSUE #6500: Git-based template variables evaluate to empty or confusing values in global `env` field
### Description

I have noticed that in my build process the global environment variables are loaded BEFORE GoReleaser retrieves the Git state.
This causes Git-based template variables like `.FullCommit`, `.CommitDate`, `.CommitTimestamp`, and so on, to evaluate to empty or confusing values when used in the global `env` field.
Either they are empty, or for example in the case of `.CommitDate` and `.CommitTimestamp` they evaluate to Go's zero-time representation (`0001-01-01T00:00:00Z` and `-62135596800`).
This is super confusing in my opinion and can definitely lead to unforeseen bugs and mistakes while writing a GoReleaser config, especially since the docs mention the common fields in the table are "always available" at https://goreleaser.com/customization/general/templates/#common-fields

### How I came across this problem

I was trying to set a global [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/docs/source-date-epoch/) variable for my whole build process, especially for Docker BuildKit to do a more reproducible build, as instructed here: https://docs.docker.com/build/ci/github-actions/reproducible-builds/

```yaml
env:
  - SOURCE_DATE_EPOCH={{ if index .Env "SOURCE_DATE_EPOCH"  }}{{ .Env.SOURCE_DATE_EPOCH }}{{ else }}{{ .CommitTimestamp }}{{ end }}
```

But then I noticed that for some reason it always evaluated to `-62135596800`, and I spent like 30mins trying to debug the issue...

### How to reproduce

I think it is pretty easy to reproduce based on the description above, but nonetheless I have created a reproduction repository here: https://github.com/jonasgeiler/goreleaser-git-env-bug

When you run `goreleaser build --clean --snapshot --single-target --verbose`, it will print the weird/confusing evaluated values in the global before-hooks step:
```
  • running before hooks
    • running                                        hook=sh -c 'echo COMMIT_HASH=$COMMIT_HASH'
    • running                                        cmd=[sh -c echo COMMIT_HASH=$COMMIT_HASH] dir=
      │ COMMIT_HASH=
    • running                                        hook=sh -c 'echo COMMIT_TIMESTAMP=$COMMIT_TIMESTAMP'
    • running                                        cmd=[sh -c echo COMMIT_TIMESTAMP=$COMMIT_TIMESTAMP] dir=
      │ COMMIT_TIMESTAMP=-62135596800
    • running                                        hook=sh -c 'echo COMMIT_DATE=$COMMIT_DATE'
    • running                                        cmd=[sh -c echo COMMIT_DATE=$COMMIT_DATE] dir=
      │ COMMIT_DATE=0001-01-01T00:00:00Z
    • running                                        hook=sh -c 'echo PROJECT_NAME=$PROJECT_NAME'
    • running                                        cmd=[sh -c echo PROJECT_NAME=$PROJECT_NAME] dir=
      │ PROJECT_NAME=goreleaser-git-env-bug
    • running                                        hook=sh -c 'echo FOO=$FOO'
    • running                                        cmd=[sh -c echo FOO=$FOO] dir=
      │ FOO=bar
```

### What I expect to happen

It is difficult to say what SHOULD happen instead.
I think there are multiple options:
- An error is thrown when using those template variables.
- You make them work in the global `env` field just as they work everywhere else (aka. load environment variables AFTER retrieving Git state).
- You don't make changes to the code, but give a warning in the docs about Git-based template variables in the global `env` field, and how they don't work there.

### Environment

- GoReleaser version output:
  ```
  GitVersion:    2.14.3
  GitCommit:     7cc45095d4e0bebd44c8d2bbf97823856405cfd7
  GitTreeState:  false
  BuildDate:     2026-03-09T13:56:16Z
  BuiltBy:       goreleaser
  GoVersion:     go1.26.1
  Compiler:      gc
  ModuleSum:     h1:vWFc2clxPb1DUkafYVdr93Vx27Mb5VPDRghY8AfWZFo=
  Platform:      linux/amd64
  ```
- OS: `EndeavourOS` (based on Arch Linux)
- Architecture: `x86_64` / `amd64`

SAY-5:
Fixes #6500.

## Problem

The templates reference says the common fields are "always available", but the top-level `env:` block is evaluated before goreleaser collects the repository git state. Fields that depend on git (`.FullCommit`, `.Commit`, `.ShortCommit`, `.CommitDate`, `.CommitTimestamp`, `.GitURL`, `.GitTreeState`, `.IsGitClean`, `.IsGitDirty`, `.Branch`, `.Tag`, `.PreviousTag`, `.Summary`, `.TagSubject`, `.TagContents`, `.TagBody`) therefore evaluate to empty strings, or to Go zero-time values like `0001-01-01T00:00:00Z` / `-62135596800` for the date/timestamp fields.

The reporter ran into this while setting up a reproducible-build `SOURCE_DATE_EPOCH` and ended up debugging a bogus `-62135596800` value for 30 minutes.

## Fix

Add an explicit note at the top of "Common Fields" documenting the carve-out, listing the affected fields, and pointing readers at `before.hooks` (or an exported CI env var) as the right way to produce something like `SOURCE_DATE_EPOCH` from the commit timestamp. No behaviour change.

This is the least-disruptive of the three options the reporter enumerated (error / reorder / doc warning). Reordering the env resolution relative to git state collection would work but affects other template paths; the doc note is a narrow, backport-safe fix that stops future users from hitting the same pit.

Signed off per DCO.

### DIFF
--- a/www/content/customization/general/templates.md
+++ b/www/content/customization/general/templates.md
@@ -11,7 +11,19 @@ support templating.
 
 ## Common Fields
 
-In fields that support templates, these fields are always available:
+In fields that support templates, these fields are always available,
+**with one exception**: the top-level `env` block is resolved before the
+git state is collected, so `.FullCommit`, `.Commit`, `.ShortCommit`,
+`.CommitDate`, `.CommitTimestamp`, `.GitURL`, `.GitTreeState`,
+`.IsGitClean`, `.IsGitDirty`, `.Branch`, `.Tag`, `.PreviousTag`, `.Summary`,
+`.TagSubject`, `.TagContents`, and `.TagBody` all evaluate to their
+zero values there (empty strings, or `0001-01-01T00:00:00Z` /
+`-62135596800` for the date/timestamp fields). Use these fields in
+hooks, builds, archives, etc. instead. For reproducible-build use cases
+in particular, the common `SOURCE_DATE_EPOCH={{ .CommitTimestamp }}` trick
+will not work in the top-level `env:` block and should be set in a
+`before.hooks` step or exported from the CI system instead
+(see [goreleaser/goreleaser#6500](https://github.com/goreleaser/goreleaser/issues/6500)).
 
 | Key                | Description                                                                                                |
 | ------------------ | ---------------------------------------------------------------------------------------------------------- |


caarlos0:
thanks @SAY-5  - ended up going with a different wording, but added you as co-author.

have a great week :)

PR #6572, opened 2026-04-17
chore(deps): swap acr credhelper to remove deprecated go-autorest

gaganhr94:
If applied, this commit will replace chrismellard/docker-credential-acr-env with gaganhr94/docker-credential-acr to drop the deprecated github.com/Azure/go-autorest dependency from the ACR credential helper used by the ko pipe.

### Why is this change being made? 
chrismellard/docker-credential-acr-env still depends on github.com/Azure/go-autorest, which has been deprecated by Microsoft in favor of the Azure SDK for Go. The repo is not maintained as well. This PR migrates the ACR credential helper fully onto the current Azure SDK for Go, eliminating the deprecated transitive dependency while preserving identical behavior for ACR authentication.


gaganhr94:
@caarlos0 can you please review this PR when you get a chance. Thanks!

caarlos0:
Hey, thanks for the PR, but I think for now I rather stay with the same lib ko itself is using.

gaganhr94:
Hi @caarlos0, 

I understand the preference to continue with it, but I wanted to share some concrete findings:

**As of now, the package has 18 vulnerabilities. 1 Critical, 7 High, 9 Medium**

The package which was replaced hasn't seen active maintenance — its last commit was over two years ago, and there's no indication these vulnerabilities will be addressed. As a result of which, there's a risk of all the downstream projects also being affected. 

These vulnerabilities may propagate to every downstream project that depends on goreleaser. I believe this is important for the security posture of goreleaser and its downstream consumers.

Thanks for your time.

PR #6136, opened 2025-10-01
feat(builders): add nodejs

vedantmgoyal9:

PR #6579, opened 2026-04-18
feat(builders): add node sea support

caarlos0:
based on @vedantmgoyal9's work on #6136, but without a rust binary dependency.


vedantmgoyal9:
this is crazy work 🤯 i think we can sign macos binaries with an ad-hoc signature using quill, so they can run on m-series macs. will try to test and give feedback before end of the month 🫡

PR #6580, opened 2026-04-20
Fix grouping of different arm version when creating archives of rust builds

mozzieongit:
Archiving multiple rust arm builds would fail because armv7 and arm would be grouped into a single archive and fail with the following error:

```
  • archives
    • group linuxarm64musl has 1 binaries
    • group linuxarmgnueabihf has 2 binaries
...
  ⨯ release failed after 2s                          error=failed to add: 'dist/dnst-cross_armv7-unknown-linux-gnueabihf/dnst' -> 'dnst': add dnst: file already exists
```

This was because rust builds didn't store the arm version in the `Goarm` field. This PR fixes that.

The relevant config snippet was this:
```yml
builds:
  - builder: rust
    id: dnst-cross
    flags:
      - --release
      - --features=static-openssl
    targets:
      - aarch64-unknown-linux-musl
      - arm-unknown-linux-gnueabihf
      - armv7-unknown-linux-gnueabihf

archives:
  - formats: [tar.gz]
    format_overrides:
      - goos: windows
        formats: [zip]
    allow_different_binary_count: true # added to skip the error of having 2 binaries vs 1 binary in linuxarm64musl
```

With this fix, it successfully separates the binaries into different archives:

```
    • group linuxarmgnueabihf7 has 1 binaries
    • group linuxarm64musl has 1 binaries
    • group linuxarmgnueabihf6 has 1 binaries
```

For the test added in 978108fb037f51c8dd68038d0188ae6becc51767 I just copied the existing build test for aarch64 and changed the expected artifact values at the bottom.

### DIFF
--- a/internal/builders/rust/build.go
+++ b/internal/builders/rust/build.go
@@ -75,7 +75,7 @@ func (b *Builder) Parse(target string) (api.Target, error) {
 		Target: target,
 		Os:     parts[2],
 		Vendor: parts[1],
-		Arch:   convertToGoarch(parts[0]),
+		Arch:   parts[0],
 	}
 
 	if len(parts) > 3 {
--- a/internal/builders/rust/build.go
+++ b/internal/builders/rust/build.go
@@ -161,7 +161,7 @@ func (b *Builder) Build(ctx *context.Context, build config.Build, options api.Op
 		Path:   options.Path,
 		Name:   options.Name,
 		Goos:   t.Os,
-		Goarch: convertToGoarch(t.Arch),
+		Goarch: t.Arch,
 		Target: t.Target,
 		Extra: map[string]any{
 			artifact.ExtraBinary:  strings.TrimSuffix(filepath.Base(options.Path), options.Ext),
--- a/internal/builders/rust/build.go
+++ b/internal/builders/rust/build.go
@@ -162,6 +162,7 @@ func (b *Builder) Build(ctx *context.Context, build config.Build, options api.Op
 		Name:   options.Name,
 		Goos:   t.Os,
 		Goarch: t.Arch,
+		Goarm:  t.Arm,
 		Target: t.Target,
 		Extra: map[string]any{
 			artifact.ExtraBinary:  strings.TrimSuffix(filepath.Base(options.Path), options.Ext),
@@ -173,9 +174,6 @@ func (b *Builder) Build(ctx *context.Context, build config.Build, options api.Op
 	}
 	if t.Libc != "" {
 		a.Extra[keyLibc] = t.Libc
-	}
-	if t.Arm != "" {
-		a.Goarm = t.Arm
 	}
 
 	env := []string{}
--- a/internal/builders/rust/build.go
+++ b/internal/builders/rust/build.go
@@ -78,6 +78,10 @@ func (b *Builder) Parse(target string) (api.Target, error) {
 		Arch:   convertToGoarch(parts[0]),
 	}
 
+	if arm, ok := getArmVersion(parts[0]); ok {
+		t.Arm = arm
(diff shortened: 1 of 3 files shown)


internal/builders/rust/targets.go
```diff
@@ -123,3 +123,17 @@ func defaultTargets() []string {
 		"aarch64-apple-darwin",
 	}
 }
+
+// Get ARM version for `(artifact.Artifact).Goarm`
+func getArmVersion(s string) (string, bool) {
+	// Values sourced from https://go.dev/wiki/GoArm in combination with
+	// https://doc.rust-lang.org/nightly/rustc/platform-support.html
+	ss, ok := map[string]string{
+		"arm":     "6",
```
mozzieongit, line 134:
```suggestion
```

I wonder if this line should be removed. 

With this PR, for the value `arm` (aka arm v6) the resulting archive would have the name `dnst_0.2.0..._linux_armv6.tar.gz`. The version number `v6` is included. For `armv7` that is what we want, but usually "`armv6`" is just called `arm` in packages and archives.

Or should this better be reflected in the default file name template? That if the template would print `armv6` it's just `arm`?

mozzieongit: Thinking a bit more about it, it should probably be a template string change. Either people do it themselves when they want to keep the `arm` without `v6` behaviour, or set it by default.

In my config right now I changed the with-Arm part in the file name templates (for archives and packages) to:
```
      {{ if not (eq .Arm "6") }}{{ with .Arm }}v{{ . }}{{ end }}{{ end }}
```

internal/builders/rust/build.go
```diff
@@ -75,7 +75,7 @@ func (b *Builder) Parse(target string) (api.Target, error) {
 		Target: target,
 		Os:     parts[2],
 		Vendor: parts[1],
-		Arch:   convertToGoarch(parts[0]),
+		Arch:   parts[0],
```
caarlos0: i think this should remain as it was

mozzieongit: Should then the `Target struct` instead get an `Arm` field to retain that information for the `Artifact struct` in the `Build` function?

internal/builders/rust/build.go
mozzieongit: And if the other `convertToGoarch` is staying, this one is redundant, right?
Because it takes in the t.Arch, which has been created by `convertToGoarch` in the first place in `Parse()`.

internal/builders/rust/build.go
```diff
@@ -157,7 +161,7 @@ func (b *Builder) Build(ctx *context.Context, build config.Build, options api.Op
 		Path:   options.Path,
 		Name:   options.Name,
 		Goos:   t.Os,
-		Goarch: convertToGoarch(t.Arch),
+		Goarch: t.Arch,
```
caarlos0, line 164:
```suggestion
		Goarch: t.Arch,
		Goarm:  t.Arm,
```

internal/builders/rust/build.go
```diff
@@ -170,6 +174,9 @@ func (b *Builder) Build(ctx *context.Context, build config.Build, options api.Op
 	if t.Libc != "" {
 		a.Extra[keyLibc] = t.Libc
 	}
+	if t.Arm != "" {
+		a.Goarm = t.Arm
+	}
```
caarlos0:
```suggestion
```

per comment in line 164, this can be removed

internal/builders/rust/build.go
caarlos0: ah ok, right, so we can simplify it, yes! thanks

caarlos0:
thanks! merge to main in 03735a4

PR #6584, opened 2026-04-21
fix: wrong goreleaser URL 

ldez:
This PR fixes invalid URL, ex: https://github.com/orgs/goreleaser/discussions/6005

Those URLs lead to a 404.

Luckily, I think nobody can claim the name `orgs` :wink:


caarlos0 (approved):

caarlos0:
thanks! it's due a settings change on the org, didn't realize it would break these links

thanks for the fix 🙏

PR #6634, opened 2026-05-22
feat(dockers_v2): pre/post hooks

### LINKED ISSUE #6621: Add before and after hooks to the Docker section.
By having before and after hooks in the Docker section, we gain:

- parity with most of the other sections (build, release, global).
- the ability to perform docker specific actions before or after the docker section, such as extracting a base image digest so it can be included in a label/annotation or running a vulnerability scan over the recently created image so that i can then be included in the release files.

This was discussed privately with @caarlos0 over a chain of emails so that is why there was no Github Discussion.

caarlos0:
adds pre/post hooks to dockers_v2

closes #6621

### DIFF
--- a/www/content/customization/package/dockers_v2.md
+++ b/www/content/customization/package/dockers_v2.md
@@ -139,6 +139,7 @@ dockers_v2:
     # Available extra template fields:
     #   - .Dockerfile: the resolved path to the Dockerfile.
     #   - .Images:     the compiled `image:tag` list for this build.
+    #   - .ContextDir: the temporary build context directory.
     #   - .Digest:     the resulting image digest (post hook only).
     #
     # {{< g_inline_version "v2.16-unreleased" >}}
@@ -146,7 +147,7 @@ dockers_v2:
       pre:
         - cmd: ./scripts/before-docker.sh
           # Working directory for the command.
-          dir: ./
+          dir: "{{ .ContextDir }}"
           # Extra env vars to inject into the hook.
           env:
             - DOCKERFILE={{ .Dockerfile }}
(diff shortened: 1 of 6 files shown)


www/content/customization/package/dockers_v2.md
```diff
@@ -132,6 +132,30 @@ dockers_v2:
...
+    #   - .Dockerfile: the resolved path to the Dockerfile.
+    #   - .Images:     the compiled `image:tag` list for this build.
+    #   - .Digest:     the resulting image digest (post hook only).
+    #
+    # {{< g_inline_version "v2.16-unreleased" >}}
+    hooks:
+      pre:
+        - cmd: ./scripts/before-docker.sh
+          # Working directory for the command.
+          dir: ./
```
martin-viggiano: Is this relative to the docker build context location or from the dist folder?

caarlos0: repository root - but might be worth it making it relative to build context? 🤔 

OR - we pass a `.ContextDir` template key

martin-viggiano: I was thinking about running it in the build context directory for the use-case where you generate a file that needs to be included in the image but that can be achieved by running the command from somewhere relative to the repository root and using the extra_files section to pick it up.

I also think that running relative to the repository root makes it easier to understand for end-users.

caarlos0: that would be inconsistent with the other places where there's a hook

added `ContextDir` , so you can `dir: '{{.ContextDir}}'` which seems like a fair compromise

PR #6644, opened 2026-06-01
fix: return error when go.mod is unreadable in CheckGoModPipe

SebTardif:
Return an error when `os.ReadFile` fails with a non-`ErrNotExist` error in `CheckGoModPipe.Run`, instead of silently swallowing the error.

`CheckGoModPipe.Run` validates that `go.mod` does not contain `replace` directives when `gomod.proxy` is enabled. When `os.ReadFile` fails (e.g., permission denied, I/O error), the error is logged but `nil` is returned, silently skipping the validation. This means a user whose `go.mod` is unreadable proceeds to the proxy build step, which fails later with a confusing error instead of a clear early message.

The fix distinguishes between `os.ErrNotExist` (expected for non-Go builds, skip gracefully) and other errors (return to caller). This matches the pattern already used by `copyGoSum` in the same file.

**Bug origin:** Introduced in #4398 (2023-11-03).

- #4398 (original PR that introduced the behavior)

AI was used in authoring this PR (Grok, assisted by user).

### DIFF
--- a/internal/pipe/gomod/gomod_proxy.go
+++ b/internal/pipe/gomod/gomod_proxy.go
@@ -42,8 +42,10 @@ func (CheckGoModPipe) Run(ctx *context.Context) error {
 		path := filepath.Join(build.UnproxiedDir, "go.mod")
 		mod, err := os.ReadFile(path)
 		if err != nil {
-			log.Errorf("could not check %q", path)
-			return nil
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):

caarlos0:
thanks!

Issue #3299, opened 2022-08-09
GitLab self-hosted 13.4.3 release link is a 404 link


laojianzi:
### What happened?

gitlab Self-hosted 13.4.3

release link `direct_asset_url` is a 404 link

### How can we reproduce this?

```yaml
project_name: test
gitlab_urls:
  api: https://domain/api/v4/
  download: https://domain
  skip_tls_verify: false
builds:
  - id: test
    binary: test
    env: [CGO_ENABLED=0]
    goos:
      - linux
      - windows
      - darwin
    goarch:
      - amd64
      - arm64
snapshot:
  name_template: '{{ incpatch .Version }}'
archives:
  - name_template: "{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}"
    files:
      - none*
release:
  gitlab:
    owner: repo
    name: test
  disable: false
```

### goreleaser version

```bash
goreleaser version 1.10.3
commit: 315935aedc9908b078ae525b33ddb42b2fafcc54
built by: homebrew
goos: darwin
goarch: amd64
```

### GoReleaser Check

- [X] goreleaser check shows no errors

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Additional context

If you edit the release in gitlab and save it without changing any data

`direct_asset_url` will become the same as `url` and will be a downloadable link

caarlos0:
> If you edit the release in gitlab and save it without changing any data
> 
> `direct_asset_url` will become the same as `url` and will be a downloadable link

this seems to indicate a gitlab issue, doesn't it?

laojianzi:
Sorry, I'm not sure if this is a Gitlab issue

What I know: 
- the repository is public
- my user rights are developer
- when I open this release page I get the download link as 
https://domain/repo/test/-/releases/v0.0.1/test_0.0.1_linux_arm64.tar.gz and this link will return 404 page

caarlos0:
which is the URL that works? You also mentioned in the original post something about the url changing when editing and saving the release... does the URL change then?

also, did this started to happen after a goreleaser or gitlab update?

laojianzi:
> which is the URL that works? You also mentioned in the original post something about the url changing when editing and saving the release... does the URL change then?

Editing process
**Before:**

[...]

PR #6645, opened 2026-06-02
fix(gitlab): add direct asset URL compatibility option

laojianzi:
## Summary

This adds an opt-in GitLab compatibility mode for release asset links through
`gitlab_urls.use_direct_asset_url`.

When enabled, GoReleaser sends the uploaded artifact URL as
`direct_asset_url` when creating GitLab release links, instead of asking GitLab
to derive the direct asset URL from `filepath` or `direct_asset_path`.

## Motivation

Some older GitLab Community Edition instances, including GitLab CE 13.4.3, may
accept release links successfully but generate broken direct asset URLs when the
direct URL is derived from an asset path. In that case, release links can appear
valid while the generated direct asset URLs return 404.

This addresses the compatibility case described in
#3299 while keeping the existing default behavior
unchanged.

## Changes

- Added `gitlab_urls.use_direct_asset_url` to the project configuration.
- Added a GitLab release-link creation path that sends `direct_asset_url`
  directly in the request body.
- Kept the existing `filepath` / `direct_asset_path` behavior as the default.
- Updated the JSON schema with the new option.
- Documented when and why to enable the compatibility option.
- Added tests for config loading and GitLab upload request behavior.

## Validation

```bash
go test ./internal/client ./pkg/config
```

### DIFF
--- a/internal/client/gitlab.go
+++ b/internal/client/gitlab.go
@@ -602,23 +602,7 @@ func (c *gitlabClient) Upload(
 			WithField("url", baseLinkURL).
 			Debug("uploaded file")
 
-		name := artifact.Name
-		filename := "/" + name
-		opt := &gitlab.CreateReleaseLinkOptions{
-			Name: &name,
-			URL:  &linkURL,
-		}
-		if c.isV17OrLater {
-			opt.DirectAssetPath = &filename
-		} else {
-			opt.FilePath = &filename
-		}
-
-		releaseLink, resp, err := c.client.ReleaseLinks.CreateReleaseLink(
-			projectID,
-			releaseID,
-			opt,
-		)
+		releaseLink, resp, err := c.createReleaseLink(ctx, projectID, releaseID, artifact.Name, linkURL)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


caarlos0:
hey! thanks for the PR!

if we know which versions have this issue, we could check the server version and do the 'right' thing automatically... what do you think?

caarlos0:
on a second thought, gitlab 13.4.3 if from 2020 - is this fixed on newer gitlab versions? 

if so, IMHO, not really worth it doing this.

laojianzi:
> on a second thought, gitlab 13.4.3 if from 2020 - is this fixed on newer gitlab versions?仔细想想，如果是 2020 年的 GitLab 13.4.3 版本，那么这个问题在新版本的 GitLab 中是否已经修复了？
> 
> if so, IMHO, not really worth it doing this.如果真是这样，恕我直言，我觉得这么做不太值得。

Your decision is correct. I can resolve individual issues by using a fork. Thank you for your reply. I’ll close this PR now.

PR #4721, opened 2024-03-22
feat(winget): sync fork before opening PR

caarlos0:
closes https://github.com/goreleaser/goreleaser/issues/4720

TODO: review, clean up, implement for nix et al too

maybe merge forksyncer and pullrequestopener interfaces

Issue #4720, opened 2024-03-22
Winget: base branch not updated before committing


Trenly:
### What happened?

When GoReleaser is used for winget releases, the base branch is not updated. This leads to the branches used for PRs being `1 commit ahead, 28150 commits behind`. This causes automated tooling at winget to fail occasionally when PRs are being reviewed by moderators, as the PR checkout fails due to files being overwritten by checkout.

I expect that either -
1) The base branch is brought up to latest before the commits are added 
\- or -
2) The new branch is created based on a detached head from microsoft/winget-pkgs@master

This way, the branches from PRs will always be `1 commit ahead, 0 commits behind` at the time of creation

### How can we reproduce this?

* https://github.com/microsoft/winget-pkgs/pull/145974
* https://github.com/microsoft/winget-pkgs/pull/145973

### goreleaser version

```bash
¯\_(ツ)_/¯
```

### GoReleaser Check

- [ ] goreleaser check shows no errors

### Search

- [X] I did search for other open and closed issues before opening this

### Supporter

- [ ] I am a [sponsor](https://github.com/sponsors/caarlos0/) or a [Pro](https://goreleaser.com/pro) customer

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

caarlos0:
hey! right, I think this wasn't available in go-github or something, but just created the pr:

https://github.com/goreleaser/goreleaser/pull/4721

PR #6646, opened 2026-06-02
fix(client): skip merge-upstream when target repo is not a fork

jamessawle:
## Problem

When a publisher (brew, cask, nix, scoop, krew, winget) is configured with `repository.pull_request.enabled: true` and the target repository is **not a fork**, every release logs a misleading error:

```
homebrew formula
  • could not sync fork  error=POST /repos/<owner>/<repo>/merge-upstream: 422 Validation Failed
```

The release and the PR are still created successfully — the warning is cosmetic — but it's noisy, conditions readers to ignore "errors" from the publisher step, and obscures the genuine 422s that fork users actually need to act on (merge conflicts, branch protection, divergent history, etc.).

The root cause is that `SyncFork` calls GitHub's `merge-upstream` endpoint unconditionally. That endpoint is documented to only work on actual forks and returns 422 for everything else. The sync step was added in #4720 / #4721 for the winget flow, where contributors really do fork `microsoft/winget-pkgs`. It's correct for that case but fires for every `pull_request`-enabled publisher today — including same-owner Homebrew taps, which are the common Homebrew setup.

## Change

Before calling `merge-upstream`, query `Repositories.Get` on the target (head) repo and check `fork`:

- `fork == false` → log at debug and return early. No wasted API call, no scary warning.
- `fork == true` → proceed exactly as today.
- Precheck call itself errors → fall through and attempt the sync anyway. This preserves current behavior on transient API issues so the change can't regress fork users.

This is a single change in `SyncFork`; every pipe (`brew`, `cask`, `krew`, `nix`, `scoop`, `winget`) routes through it, so no per-publisher edits are needed.

### Why a precheck instead of swallowing the 422?

422 is GitHub's catch-all for `merge-upstream` and is also returned for genuine failures (merge conflicts surface here in some cases, branch protection rules, etc.). Silently dropping all 422s would hide real, actionable failures from fork users — exactly the people the sync step was added for. A precheck distinguishes "this call is structurally impossible" from "this call failed for a fixable reason", which is the actual semantic difference that matters.

### Why no config knob?

Behavior is detectable from the GitHub API. A `fork:` or `sync_fork:` field would add a config surface that users would have to set correctly to get the same result. Auto-detection is zero-config and correct by construction.

### Cost

One additional `GET /repos/:owner/:name` per release per publisher. Runs once per release, well inside rate limits. Not worth caching.

## Tests

`internal/client/github_test.go`:
- Existing `TestGitHubSyncFork` happy-path / error / no-base-branch tests updated to mock the new `GET /repos/headowner/headname` response.
- New `skips merge-upstream when target is not a fork` — asserts `merge-upstream` is never called when precheck reports `fork: false`.
- New `falls through when fork precheck errors` — asserts `merge-upstream` is still attempted when the precheck call returns 500, preserving current behavior on API hiccups.
- `TestGitHubSyncForkDefaultBranchError` updated for the new precheck.

All tests in `./internal/client/...` pass locally. `gofumpt -l` and `go vet` clean.

---

*Drafted with [Claude Code](https://claude.com/claude-code) assistance.*

### DIFF
--- a/internal/client/github.go
+++ b/internal/client/github.go
@@ -371,10 +371,10 @@ func (c *githubClient) SyncFork(ctx *context.Context, head, base Repo) error {
 	case err != nil:
 		log.WithField("repo", head.String()).
 			WithError(err).
-			Debug("could not check if target is a fork; attempting sync anyway")
+			Warn("could not check if target is a fork; attempting sync anyway")
 	case !fork:
 		log.WithField("repo", head.String()).
-			Debug("target is not a fork; skipping merge-upstream")
+			Info("target is not a fork; skipping merge-upstream")
 		return nil
 	}
 
--- a/internal/client/github.go
+++ b/internal/client/github.go
@@ -226,6 +226,18 @@ func (c *githubClient) authorsLookup(authors []Author) []Author {
 	return authors
 }
 
+// isFork reports whether the given repository is a fork.
+func (c *githubClient) isFork(ctx *context.Context, repo Repo) (bool, error) {
+	c.checkRateLimit(ctx)
+	p, _, err := githubDo(ctx, func() (*github.Repository, *github.Response, error) {
+		return c.client.Repositories.Get(ctx, repo.Owner, repo.Name)
+	})
+	if err != nil {
+		return false, err
+	}
+	return p.GetFork(), nil
+}
+
 // getDefaultBranch returns the default branch of a github repo
 func (c *githubClient) getDefaultBranch(ctx *context.Context, repo Repo) (string, error) {
 	c.checkRateLimit(ctx)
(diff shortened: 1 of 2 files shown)


internal/client/github.go
```diff
@@ -353,6 +365,19 @@ func (c *githubClient) OpenPullRequest(
 }
 
 func (c *githubClient) SyncFork(ctx *context.Context, head, base Repo) error {
+	// merge-upstream returns 422 for non-forks; skip the call when we can tell.
+	fork, err := c.isFork(ctx, head)
+	switch {
+	case err != nil:
+		log.WithField("repo", head.String()).
+			WithError(err).
+			Debug("could not check if target is a fork; attempting sync anyway")
```
caarlos0: I think this could be a warn

internal/client/github.go
```diff
@@ -353,6 +365,19 @@ func (c *githubClient) OpenPullRequest(
...
+	case !fork:
+		log.WithField("repo", head.String()).
+			Debug("target is not a fork; skipping merge-upstream")
```
caarlos0: and this probably an info

caarlos0 (commented):
hey! thanks for the PR!

i have a few cosmetic suggestions, lgtm otherwise

internal/client/github.go
caarlos0:
```suggestion
			Warn("could not check if target is a fork; attempting sync anyway")
```

internal/client/github.go
caarlos0:
```suggestion
			Info("target is not a fork; skipping merge-upstream")
```

caarlos0:
merged, thank you!

PR #6647, opened 2026-06-02
feat(nfpm): support msix packages

### LINKED ISSUE #6519: support msix
nfpm v0.46 added support for msix: https://github.com/goreleaser/nfpm/releases/tag/v2.46.0

we could probably support it here.

it would require some refactor, though, as currently the nfpm pipe only runs for
linux packages.

documentation around that also is mostly linux oriented.

that said, its probably worth it investigating.

umaidshahid:
Add support for building Windows `.msix` packages via the nfpm pipe.

nfpm [v2.46](https://github.com/goreleaser/nfpm/releases/tag/v2.46.0) added an
msix packager. This wires it into GoReleaser's `nfpms` pipe so a `msix` format
can be requested alongside the existing Linux formats.

What's included:

- New `msix` format and `msix:` config block (publisher, identity, properties,
  applications, dependencies, capabilities, signature).
- The nfpm pipe previously only fed `linux`/`ios`/`android`/`aix` binaries to
  nfpm. `windows` is now included in the artifact filter, with a per-format
  guard that skips Windows binaries for the Linux formats and skips non-Windows
  binaries for `msix` — mirroring the existing aix/android/ios skip pattern. A
  single `nfpms` entry can therefore list both Linux formats and `msix`; each
  binary lands only in the package matching its platform.
- For `msix`, `bindir` is forced to the package root (`/`) so the binary lands
  at the root of the package, matching the manifest's root-level `Executable`
  (users no longer need to hand-align `bindir` with each application's
  `executable`).
- Templating for the relevant msix string fields (publisher, logo, display
  names, PFX path).
- MSIX signing passphrase lookup via `NFPM_<ID>_MSIX_PASSPHRASE` (reuses the
  existing passphrase resolver).
- `msix` is flagged as experimental: a one-time runtime warning plus the
  `g_experimental` docs callout, matching the other experimental features.
- Regenerated `www/static/schema.json`.
- Tests and documentation (incl. an "A note about MSIX" section explaining how
  it differs from the Linux formats).

nfpm gained msix support, but GoReleaser's nfpm pipe was hardwired to Linux
(artifact filtering, format list, and docs were all Linux-oriented), so the new
packager couldn't be used. This makes it available without a separate pipe, by
generalizing the existing per-format skip logic the pipe already uses for
aix/android/ios.

The produced msix artifact uses a dedicated `artifact.MSIX` type (with
`Goos: windows`) rather than reusing `artifact.LinuxPackage`, since a msix is a
Windows package, not a Linux one. It is added to the release uploadable types so
the downstream release/sign/checksum consumers keep working, while no longer
being treated as a Linux package (e.g. it won't be fed into Linux docker
images).

Claude Code has been used.

Closes #6519
nfpm msix release: https://github.com/goreleaser/nfpm/releases/tag/v2.46.0


caarlos0:
Reviewed the change — implementation logic looks solid (Windows/Linux routing guard, signature clearing on skip-sign, passphrase env lookup, the mappers, format registration, and schema all check out). One issue worth fixing:

### 🟡 `bindir` should be forced to `/` for msix

`bindir` is a Linux filesystem concept (`/usr/bin`) that does not map to MSIX — an MSIX package is a virtual filesystem rooted at the install location, and the executable belongs at the package root.

Today the binary is placed at `filepath.Join(bindir, name)` regardless of format (`nfpm.go:426`), and the nfpm msix packager only strips the leading `/`. So with the default `bindir: /usr/bin` the binary lands at `usr/bin/myapp.exe` inside the `.msix`, while the manifest declares `Executable="myapp.exe"` (root) — the declared path does not exist in the package and the installed app will not launch. It also forces users to hand-align `executable` with `bindir`.

The fix is to special-case the `msix` format in `create()` and force `fpm.Bindir = "/"` (like `termux` rewrites `fpm.Bindir` at `nfpm.go:273`). Then the binary lands at the package root, `executable: myapp.exe` becomes the natural default, and the bindir/executable confusion disappears — the docs no longer need to tell users to match the two by hand.

Everything else looks good to merge.

umaidshahid:
Alright. I'll look into it :))

caarlos0:
thanks @umaidshahid! 🙏

PR #6661, opened 2026-06-18
docs: fix "lets" -> "let's" in quick-start guide

s3onghyun:
Tiny grammar fix in the Quick Start guide: "Now, lets run a local-only release" should be "let's" (contraction of "let us"). It's the line right before the first command in the onboarding tutorial.


caarlos0:
thanks!

Issue #6527, opened 2026-04-03
release summary


caarlos0:
at the end of the release, print a summary like:

Published v1.2.3 to GitHub with 43 assets.
Opened PR to microsoft/winget-pkgs.
Pushed foo to AUR.
Updated foo in myorg/homebrew-tap.

and so on 

idea i just have, probably needs to evolve it further

robaa12:
Can i grab that issue ?

PR #6666, opened 2026-06-20
feat: release summary

Zybyte85:
Print a summary at the end of release showing published artifacts and targets, as in #6527.

Assisted-by: Claude (AI)

Example:
Screenshot 2026-06-20 at 11 59 59 AM

Please let me know if I missed anything that needs to be added, or if something needs to be changed.

### DIFF
--- a/cmd/release.go
+++ b/cmd/release.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/caarlos0/log"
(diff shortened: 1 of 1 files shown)


caarlos0:
sorry, this is not the way I was thinking in going about this one. that issue was mostly a note-to-self. i have a local branch with a different approach (its WIP)

will close this. thanks for the work either way 🙏🏻

PR #6668, opened 2026-06-22
fix(nfpm): produce valid arch for Termux packages

bltavares:
This is a workaround to https://github.com/goreleaser/nfpm/issues/1103
using the `info.Deb.Arch` field to pass forward the correct architecture
for Termux packages.

`goreleaser` `termux.deb` packages are emiting the wrong `Architecture:`
control field on since the update of architecture mapping for Debian
packages on [nfpm](https://github.com/goreleaser/nfpm/pull/1032).

Currently, `goreleaser` has a replacer to adjust the `.deb` value
according to Termux's expectation (`aarch64` in Termux vs `arm64` in
Debian), but the value used in `nfpm.Info{Arch}` is being normalized
upstream.

It used to work as the mapping did not cover `aarch64`, but now it's
normalizing to `arch64` causing errors in Termux:

```
$ dpkg -i opkssh*.termux.deb

pkg install ./opkssh_snapshot_arm64.deb.termux.deb
dpkg: error processing archive opkssh_snapshot_arm64.deb.termux.deb (--unpack):
 package architecture (arm64) does not match system (aarch64)
E: Sub-process /data/data/com.termux/files/usr/bin/dpkg returned an error code (1)
```

Wring tests for it has been a challenge, as the `Pipe` tests don't
expose the internal `nfpm.Info` struct, nor the `Artifacts` file have
the resulting information.

An end-to-end test would be to unpack the produced `.termux.deb` and
`.deb` files and check their control field, but this starts to feel like
testing the upstream `nfpm` code through the client. I could write this
test to avoid regressions tho if desired.


caarlos0:
cool, thank you!

could you also add add a test for this so we don't break it again 🙏

EDIT: nevermind, I just did it :)

PR #6679, opened 2026-06-29
fix(builders): reject empty target string

sueun-dev:
The rust, zig, deno, and bun builders load their valid-target lists from embedded `.txt` files and parse them with `strings.Split(string(bts), "\n")`. Those files end with a trailing newline, so the split leaves an empty string at the end of `allTargets` (and zig's `errTargets`). That makes an empty target validate as if it were a real one: `isValid("")` returns true for rust/deno/bun, and zig's `checkTarget("")` returns `targetBroken` instead of `targetInvalid`.

As a result, an empty entry in a build's `targets` list (for example a stray empty item in the YAML list) is silently accepted instead of failing with the usual `invalid target` error.

This drops empty entries when building the lists, so an empty target is rejected like any other unknown one. I added an `empty target` case to each builder's `TestWithDefaults`; it fails before the change and passes after.

AI disclosure: I used AI assistance while writing this change, and I've reviewed and understand all of it.

### DIFF
--- a/internal/builders/bun/build_test.go
+++ b/internal/builders/bun/build_test.go
@@ -109,6 +109,13 @@ func TestWithDefaults(t *testing.T) {
 		})
 		require.Error(t, err)
 	})
+
+	t.Run("empty target", func(t *testing.T) {
+		_, err := Default.WithDefaults(config.Build{
+			Targets: []string{""},
+		})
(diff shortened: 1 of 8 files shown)


caarlos0:
good catch, thank you

PR #6696, opened 2026-07-06
ci: use nightly goreleaser in the release job

### LINKED ISSUE #6692: release: use latest nightly goreleaser in the release job
### Problem

The `release` job in `.github/workflows/release.yml` pins the goreleaser binary via `goreleaser-action` with:

```yaml
- uses: goreleaser/goreleaser-action@... # v7.2.3
  with:
    distribution: goreleaser-pro
    version: "~> 2"
```

`version: "~> 2"` resolves to the latest **stable** release. When we tag a new minor (e.g. `v2.17.0`) whose `.goreleaser.yaml` uses config fields introduced in that same release, the action downloads the *previous* stable binary (e.g. `v2.16.0`), which can't parse the new config, and the release fails:

```
failed to load configuration: yaml: unmarshal errors:
  line 401: field verify not found in type config.Project
  line 440: field token not found in type config.Repo
```

### Proposal

Make the release job use the latest **nightly** build instead, so it always runs a binary new enough to parse the config it's releasing:

```yaml
- uses: goreleaser/goreleaser-action@... 
  with:
    distribution: goreleaser-pro
    version: nightly
```

(or otherwise ensure the release runs a binary >= the version being tagged.)

### Context

Hit while releasing `v2.17.0`. Worked around manually for this release; filing so we fix the workflow properly.

youdie006:
Closes #6692.

Changes the `release` job's `goreleaser-action` step from `version: "~> 2"` to `version: nightly`, so the release binary is always new enough to parse config fields introduced in the same release being tagged (as described in the issue). `distribution: goreleaser-pro` is unchanged, and only the release job is affected.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -149,7 +149,7 @@ jobs:
         if: ${{ startsWith(github.ref, 'refs/tags/v') }} # only on tags
         with:
           distribution: goreleaser-pro
-          version: "~> 2"
+          version: nightly
           args: release --clean --timeout 60m
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


caarlos0:
thanks!

PR #6698, opened 2026-07-07
fix(builder/go): parse GORISCV64 and GOPPC64 for riscv64 and ppc64le targets

upuddu:
Correctly parse the microarchitecture level for `riscv64` and `ppc64le` build targets, so `GORISCV64` and `GOPPC64` are populated in the build environment.

The Go builder's target parser matched on `riscv` and `ppc64`, but Go's actual GOARCH values are `riscv64` and `ppc64le` (the latter alongside `ppc64`). So for a target like `linux_riscv64_rva22u64` or `linux_ppc64le_power9` the microarch level (`extra`) was dropped and the build ran with an empty `GORISCV64`/`GOPPC64`. The `case "riscv"` branch was effectively dead code since no GOARCH is ever literally `riscv`.

This widens the `ppc64` case to also cover `ppc64le` and corrects `riscv` to `riscv64`.

GOARCH-specific env var reference: https://go.dev/doc/install/source#environment (GORISCV64, GOPPC64).

Added the missing rows to `TestParse` in `internal/builders/golang/build_test.go` covering `linux_ppc64_power8`, `linux_ppc64le_power9`, `linux_riscv64_rva20u64`, and `linux_riscv64_rva22u64`. Verified with `go test ./internal/builders/golang/...` and `gofmt`.

### DIFF
--- a/internal/builders/golang/build.go
+++ b/internal/builders/golang/build.go
@@ -96,9 +96,9 @@ func (b *Builder) Parse(target string) (api.Target, error) {
 			t.Abi = abi
 		case "mips", "mipsle", "mips64", "mips64le":
 			t.Gomips = extra
-		case "ppc64":
+		case "ppc64", "ppc64le":
 			t.Goppc64 = extra
-		case "riscv":
+		case "riscv64":
 			t.Goriscv64 = extra
 		}
 	}
--- a/internal/builders/golang/build.go
+++ b/internal/builders/golang/build.go
@@ -98,7 +98,7 @@ func (b *Builder) Parse(target string) (api.Target, error) {
 			t.Gomips = extra
 		case "ppc64", "ppc64le":
 			t.Goppc64 = extra
-		case "riscv64":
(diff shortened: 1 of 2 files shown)


internal/builders/golang/build.go
```diff
@@ -96,9 +96,9 @@ func (b *Builder) Parse(target string) (api.Target, error) {
 			t.Abi = abi
 		case "mips", "mipsle", "mips64", "mips64le":
 			t.Gomips = extra
-		case "ppc64":
+		case "ppc64", "ppc64le":
 			t.Goppc64 = extra
-		case "riscv":
+		case "riscv64":
```
caarlos0: should probably be `case "riscv", "riscv64":`?

upuddu:
Good call - done, matches the `ppc64`/`ppc64le` line right above.

caarlos0 (approved):

caarlos0:
thanks!

PR #6706, opened 2026-07-12
fix: migrate from deprecated s3/manager to s3/transfermanager

joeyberkovitz:
There was a corresponding change in gocloud (https://github.com/google/go-cloud/pull/3717) which was merged in v0.46.0 which goreleaser is already using

That change has migrated goreleaser to effectively using s3/transfermanager which is breaking goreleaser S3 uploads if an ACL is being used because those depend on the no longer used s3/manager struct

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/agnivade/levenshtein v1.2.1
 	github.com/atc0005/go-teams-notify/v2 v2.14.0
 	github.com/avast/retry-go/v4 v4.7.0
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):
thank you!

joeyberkovitz:
sorry, didn't have the commit signed, will need a second approval

caarlos0 (approved):

joeyberkovitz:
looks like it wanted a rebase before merging + re-sign 😢

caarlos0:
> looks like it wanted a rebase before merging + re-sign 😢

weird, shouldn't have, I think :| 

anyway, merged! thanks for the PR

PR #6709, opened 2026-07-15
feat: add iru custom apps publisher

wimwenigerkind:
Add a new `iru` publisher that uploads artifacts as Custom Apps to iru.com endpoint management. It follows the documented three-step flow (pre-signed S3 upload, then create or update of the library item) and supports updating an existing Custom App via `library_item_id`.

We deploy internal macOS CLI tools to our managed Macs via Iru. Publishing them manually on every release is error prone. This makes it a regular GoReleaser publish step. Since we already use GoReleaser to publish our Docker images, this keeps the whole release in a single tool and a single `.goreleaser.yaml`, instead of maintaining a custom upload script next to it.

[Iru Endpoint Management API Documentation](https://api-docs.iru.com)

### DIFF
--- a/www/content/customization/publish/iru.md
+++ b/www/content/customization/publish/iru.md
@@ -1,0 +1,135 @@
+---
+title: "Iru Custom Apps"
+linkTitle: Iru
+weight: 165
+---
+
+{{< g_version "v2.18" >}}
+
+GoReleaser can publish artifacts as _Custom Apps_ to
+[iru.com](https://www.iru.com) (formerly Kandji) endpoint management, making
+them available for deployment to your managed Macs.
+
+## How it works
+
+Publishing follows the three-step flow of the
+[Iru API](https://api-docs.iru.com): GoReleaser requests pre-signed S3 upload
+details, uploads the file to S3, and then creates (or updates) the Custom App
+library item.
+
+Custom Apps are a macOS-only library item: the supported install types
+(`package`, `zip`, `image`) correspond to `.pkg`, `.zip`, and `.dmg` files,
+and the pre/post install scripts run on the target Mac.
+
+Prerequisites:
+
+- An [Iru API token](https://support.kandji.io/api), read from the
+  `$IRU_API_TOKEN` environment variable by default. Permissions are granted
+  per endpoint, so the token needs the following Library permissions:
+  - `Upload Custom App`: always required.
+  - `Create Custom App`: required when creating a new Custom App on each
+    release (no `library_item_id` set).
+  - `Update Custom App`: required when `library_item_id` is set.
+
+The `iru` section specifies how the Custom App should be created:
+
+```yaml {filename=".goreleaser.yaml"}
+iru:
+  # Your Iru API base URL.
+  # You can find it in Settings > Access, e.g.
+  # US: https://SubDomain.api.kandji.io
+  # EU: https://SubDomain.api.eu.kandji.io
+  #
+  # Required.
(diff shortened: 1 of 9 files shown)


www/content/customization/publish/iru.md
```diff
@@ -0,0 +1,135 @@
+---
+title: "Iru Custom Apps"
+linkTitle: Iru
+weight: 165
+---
+
+{{< g_version "v2.18" >}}
```
caarlos0:
```suggestion
{{< g_version "v2.18-unreleased" >}}
```

caarlos0 (commented):
lgtm, one minor comment, + lint issues

thanks! and sorry for the delay reviewing this

caarlos0:
I found three correctness issues:

1. **High — Existing-app updates overwrite dashboard-managed settings** (`internal/pipe/iru/iru.go:45-47,245-248`). With `library_item_id` set, omitted `name`, `install_type`, and `install_enforcement` are defaulted and always included in the PATCH. An installer-only update can rename the app or reset `image`/`zip` and `continuously_enforce` to `package`/`install_once`. Please build separate create/update payloads and PATCH only `file_key` plus explicitly configured fields.

2. **High — The default artifact selection publishes incompatible files** (`internal/pipe/iru/iru.go:83-90`). With no `ids`, a normal cross-platform release selects every uploadable archive, binary, and file, then publishes all of them under one default name and `install_type: package`. Linux/Windows binaries and `.tar.gz`/`.zip` artifacts can therefore be uploaded as macOS packages. Please require explicit compatible artifacts or validate/filter each artifact against the selected install type before uploading.

3. **Medium — API companion rules are validated only after uploading** (`internal/pipe/iru/iru.go:112-122`). `no_enforcement` without enabled Self Service/category and a non-empty `audit_script` with `install_once` both pass local validation, but Iru rejects them after the S3 upload. Please add both inverse checks to `validate` so invalid configurations fail before external side effects.

caarlos0:
Thanks for the updates. The default-leaking PATCH behavior and original companion checks are fixed, but I found three remaining issues:

1. **Medium — Artifact validation still accepts non-macOS files** (`internal/pipe/iru/iru.go:160-172`). A Windows artifact named `*.zip` passes when `install_type: zip` because only the suffix is checked. Updates with `install_type` omitted skip validation entirely, even for unsupported `.tar.gz` files. Please always reject known non-`darwin` artifacts and, when the remote type is unknown, still require one of `.pkg`, `.zip`, or `.dmg`.

2. **Medium — Invalid Self Service recommendations upload before failing or being ignored** (`internal/pipe/iru/iru.go:126-146,328-335`). On create, `self_service_recommended: true` with Self Service omitted/false passes validation and is sent after the S3 upload; Iru rejects it or strips the recommendation. Please require enabled Self Service and a category for recommendations on create, while preserving update behavior when remote state is unknown.

3. **Medium — Updates cannot clear scripts or other string fields** (`pkg/config/config.go:1687-1692`, `internal/pipe/iru/iru.go:317-326`). Plain strings cannot distinguish omitted from explicitly empty, and empty values are always dropped from PATCH. For example, `audit_script: ""` cannot clear an existing script when moving away from continuous enforcement. Please use pointer/presence-aware fields and send explicitly configured empty strings.

caarlos0:
Thank you!

caarlos0 (approved):

wimwenigerkind:
Thanks for the review, much appreciated!

PR #6715, opened 2026-07-21
feat(apkbuild): add Alpine Linux package publishing

### LINKED ISSUE #4708: Add support for APKBUILD (Alpine Linux)

I'd like GoReleaser to have explicit support for apk's APKBUILD format, used by Alpine Linux.

APKBUILD is very similar to Arch's PKGBUILD. The only difference (that I'm aware of) is that APKBUILD scripts must be pure `sh`, while PKGBUILD are `bash`.

It would be basically the same as the current `aurs` feature, but with a different name.

Use `aurs` to push to alpine linux repos and somehow automate renaming PKGBUILD to APKBUILD?

### Search

### Supporter

### Code of Conduct

### Additional context

_No response_

davicbtoliveira:
## What

  Add first-class support for generating and publishing Alpine Linux `APKBUILD` files from
  prebuilt Linux release artifacts.

  The implementation:

  - Adds the new `apkbuilds` configuration.
  - Generates architecture-specific sources and SHA-512 checksums.
  - Supports `amd64`, `386`, `arm64`, `arm/v6`, `arm/v7`, `ppc64le`, `s390x`, and `riscv64`.
  - Supports optional Git publishing, following the existing AUR workflow.
  - Supports artifact ID and GOAMD64 filtering, templates, skips, and custom package
  instructions.
  - Updates the JSON schema and documentation.
  - Adds unit, integration, shell syntax, and Git publishing tests.

  ## Why

  GoReleaser supports generating and publishing Arch Linux `PKGBUILD` files, but does not
  offer an equivalent workflow for Alpine Linux `APKBUILD` recipes.

  This implementation follows the existing binary AUR integration: it packages prebuilt
  release artifacts and optionally publishes the generated recipe to a configured Git
  repository.

  It does not build from source, generate `.apk` packages, or automatically create an Alpine
  `aports` merge request.

## Relevant Resources

  - Closes #4708
  - APKBUILD reference: https://wiki.alpinelinux.org/wiki/APKBUILD_Reference
  - Creating an Alpine package: https://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package

  ## AI disclosure

  AI assistance was used during planning, implementation, testing, and documentation.

 I reviewed and understand all changes. I also manually ran the relevant build, test, formatting, schema, and lint
  checks described in `CONTRIBUTING.md`.

  Co-authored-by: GPT-5.6-sol


internal/pipe/apkbuild/apkbuild.go
```diff
@@ -0,0 +1,397 @@
...
+			pkg.CommitMessageTemplate = defaultCommitMsg
+		}
+		if pkg.Name == "" {
+			pkg.Name = ctx.Config.ProjectName
+		}
+		if pkg.Rel == "" {
+			pkg.Rel = "0"
+		}
+		if pkg.Goamd64 == "" {
+			pkg.Goamd64 = "v1"
```
caarlos0, line 68: does alpine support amd64!=v1?

www/content/customization/publish/apkbuilds.md
```diff
@@ -0,0 +1,153 @@
+---
+title: "Alpine Linux APKBUILDs"
+linkTitle: APKBUILD
+weight: 125
+---
+
```
caarlos0, line 6: missing the `g_version` shortcode call

caarlos0 (commented):
some minor comments, but lgtm overall

thanks for the PR!

internal/pipe/apkbuild/apkbuild.go
davicbtoliveira: Good point. APKBUILD only models this as x86_64, not GOAMD64 microarchitecture levels, so exposing goamd64 here is misleading. I removed the option and the pipeline now always selects the baseline v1 artifact.

www/content/customization/publish/apkbuilds.md
davicbtoliveira: added it, thanks!

caarlos0:
I took another correctness pass over the full branch and found a few issues worth addressing before merge:

1. **Multi-architecture wrapped archives break** (`internal/pipe/apkbuild/apkbuild.go:145`): the default `package()` hardcodes `WrappedIn` from only `archives[0]`. With `wrap_in_directory: true`, each architecture has a different folder, so every other architecture installs from a nonexistent path.
2. **Duplicate package configs publish to the wrong repository** (`internal/pipe/apkbuild/apkbuild.go:155`, `:170`, `:358`): entries sharing `name` overwrite the same dist file and artifact ID, then publishing uses only the first entry's repository/directory. Later repositories are never updated.
3. **Publishing always targets `master`** (`internal/pipe/apkbuild/apkbuild.go:375`): a repository whose default branch is `main` gets a new `master` branch rather than updating its default branch.
4. **Common SemVer prereleases produce invalid Alpine versions** (`internal/pipe/apkbuild/apkbuild.go:279`): `1.2.3-beta.1` becomes `1.2.3_beta.1`, which Alpine rejects; the valid form is `1.2.3_beta1`.
5. **Binary paths containing directories generate broken installs** (`internal/pipe/apkbuild/apkbuild.go:180`, `:185`): `binary: bin/foo` installs to `/usr/bin/bin/foo`; with `strip_binary_directory: true`, the generated source path is also nonexistent.
6. **Some supported GoReleaser archive formats are not unpacked by Alpine** (`internal/pipe/apkbuild/apkbuild.go:119`, `:322`): `gz`, `xz`, `txz`, and `tzst` artifacts are accepted, but Alpine's `default_unpack` does not handle those suffixes, so the default package command cannot find the binary.
7. **Required APKBUILD metadata is not validated** (`internal/pipe/apkbuild/tmpl.go:40`): empty `description`, `homepage`, or `license` generates successfully, but `abuild` rejects the result as missing `pkgdesc`, `url`, or `license`.
8. **Multiple maintainers make the APKBUILD invalid** (`internal/pipe/apkbuild/tmpl.go:30`): the slice emits multiple `# Maintainer:` lines, while `abuild` rejects more than one.

davicbtoliveira:
> I took another correctness pass over the full branch and found a few issues worth addressing before merge:
> 
> 1. **Multi-architecture wrapped archives break** (`internal/pipe/apkbuild/apkbuild.go:145`): the default `package()` hardcodes `WrappedIn` from only `archives[0]`. With `wrap_in_directory: true`, each architecture has a different folder, so every other architecture installs from a nonexistent path.
> 2. **Duplicate package configs publish to the wrong repository** (`internal/pipe/apkbuild/apkbuild.go:155`, `:170`, `:358`): entries sharing `name` overwrite the same dist file and artifact ID, then publishing uses only the first entry's repository/directory. Later repositories are never updated.
> 3. **Publishing always targets `master`** (`internal/pipe/apkbuild/apkbuild.go:375`): a repository whose default branch is `main` gets a new `master` branch rather than updating its default branch.
> 4. **Common SemVer prereleases produce invalid Alpine versions** (`internal/pipe/apkbuild/apkbuild.go:279`): `1.2.3-beta.1` becomes `1.2.3_beta.1`, which Alpine rejects; the valid form is `1.2.3_beta1`.
> 5. **Binary paths containing directories generate broken installs** (`internal/pipe/apkbuild/apkbuild.go:180`, `:185`): `binary: bin/foo` installs to `/usr/bin/bin/foo`; with `strip_binary_directory: true`, the generated source path is also nonexistent.
> 6. **Some supported GoReleaser archive formats are not unpacked by Alpine** (`internal/pipe/apkbuild/apkbuild.go:119`, `:322`): `gz`, `xz`, `txz`, and `tzst` artifacts are accepted, but Alpine's `default_unpack` does not handle those suffixes, so the default package command cannot find the binary.
> 7. **Required APKBUILD metadata is not validated** (`internal/pipe/apkbuild/tmpl.go:40`): empty `description`, `homepage`, or `license` generates successfully, but `abuild` rejects the result as missing `pkgdesc`, `url`, or `license`.
> 8. **Multiple maintainers make the APKBUILD invalid** (`internal/pipe/apkbuild/tmpl.go:30`): the slice emits multiple `# Maintainer:` lines, while `abuild` rejects more than one.

  Thanks for the thorough correctness pass, I addressed all eight points:

  1. Default package instructions are now generated per Alpine architecture, so each archive uses its own WrappedIn directory.
  2. Configurations sharing the same package name now produce distinct artifacts and are published independently to their respective repositories and directories.
  3. Git publishing now follows the repository's default branch instead of forcing master.
  4. SemVer prereleases are normalized to Alpine-compatible versions, e.g. 1.2.3-beta.1 becomes 1.2.3_beta1.
  5. Binary destinations now use the basename, and archive metadata records the actual path after strip_binary_directory is applied.
  6. APKBUILD generation now filters archives to the formats handled by Alpine's default_unpack: tar, tgz, tar.gz, tar.xz, and zip.
  7. description, homepage, and license are now validated at runtime and marked as required in the JSON schema.
  8. Multiple maintainers are rejected at runtime and limited to one entry in the schema.

caarlos0:
Thanks for addressing the previous review. I pulled `0a75b6a` and re-reviewed the full updated diff. Seven of the eight earlier issues are fixed; the version conversion is only partially fixed. I found four remaining correctness issues:

1. **SemVer conversion can still emit invalid or colliding Alpine versions** (`internal/pipe/apkbuild/apkbuild.go:302`): valid versions such as `1.2.3-preview.1`, `1.2.3-rc-1`, `1.2.3+build.1`, and the default snapshot version still produce invalid `pkgver` values. Removing all prerelease dots also maps distinct versions such as `beta.1.0` and `beta.10` to the same `pkgver`.
2. **Multiple formats for one archive ID cannot be selected** (`internal/pipe/apkbuild/apkbuild.go:343`): `formats: [tar.gz, zip]` produces two same-ID artifacts per architecture and fails with “use ids to select one,” but IDs cannot distinguish those formats. This needs a format selector or a deterministic supported-format choice.
3. **Whitespace-only custom package scripts create empty packages** (`internal/pipe/apkbuild/apkbuild.go:151`, `:362`; `internal/pipe/apkbuild/tmpl.go:73`): the warning treats whitespace as empty, but generation does not, so a templated package block that renders to whitespace yields an empty `package()` instead of inferred install commands.
4. **Reused repository checkouts retain the first config's commit author/signing settings** (`internal/pipe/apkbuild/apkbuild.go:416`, `internal/client/git.go:84`): when multiple APKBUILD entries target the same `git_url`, Git config is applied only for the first clone, so later entries ignore their own `commit_author` and signing configuration.

davicbtoliveira:
> Thanks for addressing the previous review. I pulled `0a75b6a` and re-reviewed the full updated diff. Seven of the eight earlier issues are fixed; the version conversion is only partially fixed. I found four remaining correctness issues:
> 
> 1. **SemVer conversion can still emit invalid or colliding Alpine versions** (`internal/pipe/apkbuild/apkbuild.go:302`): valid versions such as `1.2.3-preview.1`, `1.2.3-rc-1`, `1.2.3+build.1`, and the default snapshot version still produce invalid `pkgver` values. Removing all prerelease dots also maps distinct versions such as `beta.1.0` and `beta.10` to the same `pkgver`.
> 2. **Multiple formats for one archive ID cannot be selected** (`internal/pipe/apkbuild/apkbuild.go:343`): `formats: [tar.gz, zip]` produces two same-ID artifacts per architecture and fails with “use ids to select one,” but IDs cannot distinguish those formats. This needs a format selector or a deterministic supported-format choice.
> 3. **Whitespace-only custom package scripts create empty packages** (`internal/pipe/apkbuild/apkbuild.go:151`, `:362`; `internal/pipe/apkbuild/tmpl.go:73`): the warning treats whitespace as empty, but generation does not, so a templated package block that renders to whitespace yields an empty `package()` instead of inferred install commands.
> 4. **Reused repository checkouts retain the first config's commit author/signing settings** (`internal/pipe/apkbuild/apkbuild.go:416`, `internal/client/git.go:84`): when multiple APKBUILD entries target the same `git_url`, Git config is applied only for the first clone, so later entries ignore their own `commit_author` and signing configuration.

Thanks for the review, Carlos!

I'll work on these points asap

davicbtoliveira:
@caarlos0 
Thanks for the thorough follow-up review! I addressed all four remaining issues:
  
1. SemVer conversion now handles prereleases, build metadata, and snapshot versions using Alpine-compatible suffixes. Distinct versions such as beta.1.0 and beta.10 no longer collide. The reported cases were also validated with Alpine’s apk version.
2. When the same archive ID produces multiple supported formats for an architecture, the APKBUILD pipe now selects one deterministically, following this priority: tar.gz, tgz, tar.xz, tar, then zip. Different IDs or actual duplicate artifacts still produce an error.
3. Custom package scripts are trimmed after template evaluation. Scripts that render only whitespace now fall back to the inferred installation commands.
4. Reused Git checkouts now reapply the commit author and signing configuration before every commit. Optional signing settings from previous configurations are cleared to prevent them from leaking into later APKBUILD entries.

I added regression tests for all four cases, including reused checkout author/signing behavior. The affected package tests, race tests, and go vet pass.

caarlos0:
hi! could you fix the lint issues? ty

davicbtoliveira:
Hi, Carlos!!

I've made the fixes :D

internal/client/git.go
```diff
@@ -131,6 +103,28 @@ func (g *gitClient) CreateFiles(
...
+	if commitAuthor.Signing.Program != "" {
+		gitCmds = append(gitCmds, []string{"config", "--local", "gpg.program", commitAuthor.Signing.Program})
+	}
+	if commitAuthor.Signing.Format != "" && commitAuthor.Signing.Format != "openpgp" {
+		gitCmds = append(gitCmds, []string{"config", "--local", "gpg.format", commitAuthor.Signing.Format})
+	}
+	if err := runGitCmds(ctx, cwd, env, gitCmds); err != nil {
+		return fmt.Errorf("git: failed to setup local repository: %w", err)
+	}
+
```
caarlos0: why was this moved?

it was previously checking id signing was enabled, it isn't anymore...

www/content/customization/publish/apkbuilds.md
```diff
@@ -0,0 +1,161 @@
...
+{{< g_version "v2.18" >}}
```
caarlos0:
```suggestion
{{< g_version "v2.19-unreleased" >}}
```

caarlos0 (dismissed):
Thanks for this — it is a careful piece of work. The pipe follows the `aur` structure closely, the `abuild` semantics are mostly right, and the tests are better than the ones we have for `aur`: they run the generated file through `sh` with `CARCH` set and a stubbed `install`, and the publish tests clone a real bare repository. I checked the generated output against the `abuild` and `apk-tools` sources rather than the wiki, and most of it holds up.

I have one correctness defect that I want fixed before merge, plus a scope change I want made. The rest is smaller.

---

## 1. `pkgver` inverts the apk upgrade order for `-rc1` / `-beta1` / `-alpha1` tags

`internal/pipe/apkbuild/apkbuild.go:361`, `:394`, `:397`

`toAPKPrerelease` matches a candidate only when the word is exact, or is followed by `.` or `-`:

```go
if value != candidate.name && !strings.HasPrefix(value, candidate.name+".") && !strings.HasPrefix(value, candidate.name+"-") {
    continue
}
```

A pre-release word with no separator (`rc1`, `beta1`, `alpha1`) matches nothing and falls through to `"_pre" + toAPKVersionNumber(value)`. That function writes `1` and then **three digits for each character**, so the size of the number follows the length of the word, not the release stage:

| tag | `pkgver` | digits |
| --- | --- | --- |
| `v1.0.0-alpha1` | `1.0.0_pre1097108112104097049` | 19 |
| `v1.0.0-beta1` | `1.0.0_pre1098101116097049` | 16 |
| `v1.0.0-rc1` | `1.0.0_pre1114099049` | 10 |

apk compares that suffix as a number (`apk_blob_pull_uint`, `apk-tools/src/blob.c:418`, `uint64_t`). So apk sees `alpha1 > beta1 > rc1`. A user on `alpha1` never gets `beta1` or `rc1`, and `apk upgrade` does nothing. The separated forms are correct: `-rc.1` gives `1.2.3_rc1`.

Two more effects of the same encoding. A value above 20 digits wraps `uint64`:

- our default snapshot template `{{ .Version }}-SNAPSHOT-{{ .ShortCommit }}` (`internal/pipe/snapshot/snapshot.go:23`) gives a 49-digit number;
- `+build.1` metadata gives 22 digits.

**Suggested fix:** in the loop at `:361`, also accept a digit directly after the candidate word. Use `strings.CutPrefix`, then `continue` only when the remainder starts with a character that is not `.`, `-`, or a digit. Then `rc1` gives `_rc1` and `beta1` gives `_beta1`.

**Tests:** `TestToAPKVersion` (`apkbuild_test.go:65-88`) covers only the separated forms, so this is not caught today. Please add `"1.2.3-rc1": "1.2.3_rc1"`, `"1.2.3-beta1": "1.2.3_beta1"`, `"1.2.3-alpha1": "1.2.3_alpha1"`, and an order assertion that `alpha1 < beta1 < rc1 < 1.2.3`. The regex test proves only that the string is valid, not that the order is correct.

---

## 2. Please split the two shared-code fixes into their own pull requests

Both changes are correct, and both repair real defects. Neither is needed for Alpine support, and both change behaviour for users who do not use `apkbuilds`.

**`internal/pipe/archive/archive.go:246`** — `bins = append(bins, dst)`. This differs from `binary.Name` only when `strip_binary_directory: true`, and it changes `ExtraBinaries`, which is also the documented `.Binaries` template field. It therefore changes generated output in `brew` (`brew.go:402`), `nix` (`nix.go:531`), `scoop` (`scoop.go:400`), `aur` (`aur.go:157`), `krew` (`krew.go:246`), `winget` (`winget.go:557`), and `cask` (`cask.go:459`). The new value is the correct one — the old one pointed at a path that is not in the archive — but no test in any of those packages exercises `strip_binary_directory`, and today it ships as a one-line edit to a docs table (`artifacts.md:87`).

**`internal/client/git.go:93-125`, `:262-274`** — moving the `git config` block out of the fresh-clone branch, plus `unsetGitConfigs`. This fixes a reused checkout keeping the first caller's identity and signing configuration, and it affects all nine `CreateFiles` callers. I checked it and found no regression: `commit.gpgSign` gives the same value, `init.defaultBranch` has no effect after `clone` in either order, and `checkout -b` needs no identity. It is a good fix. But it is not needed here — the `apkbuild` checkout directory is `<sha256>-` and `aur`'s is `<sha256>-master`, so they never collide, and the new test uses one client with two authors rather than this pipe.

Please send them as `fix(archive):` and `fix(client):`, each with its own changelog line. For the archive one, please add a `strip_binary_directory` case to at least `brew_test.go` and `nix_test.go`.

---

## 3. `formats: ["tar.gz", "binary"]` stops the release, and `ids` cannot help

`internal/pipe/apkbuild/apkbuild.go:206`, `:440`

A single archive block can emit both formats (`archive.go:129-148`: `binary` goes to `skip()`, the others go to `create()`). That gives an `UploadableArchive` and an `UploadableBinary` with the **same ID** for the same platform. `selectArtifacts` removes duplicates by format only when both artifacts are archives; for mixed types it appends, and `dataFor` then fails with:

> multiple artifacts found for Alpine architecture x86_64; use ids to select one

`ids` cannot fix this, because both artifacts have the same ID. `ContinueOnError` applies only to the publish phase (`publish.go:99`), so this stops the full run after the builds, archives, and signing are done.

**Suggested fix:** in `selectArtifacts`, prefer the archive over the binary for the same ID and architecture, with the priority path that is already there.

**Test:** one archive ID that emits both `tar.gz` and `binary`, and an assertion that the `tar.gz` source is selected.

---

## 4. `goamd64` has no configuration field

`internal/pipe/apkbuild/apkbuild.go:113` pins `artifact.ByGoamd64("v1")`. `aur` (`aur.go:68`), `nix` (`nix.go:85`), and `brew` (`brew.go:76`) all have a `Goamd64` field with the default `v1`. A project that builds `goamd64: [v3]` gets the hard error `no linux binaries or archives found` and has no way to configure around it. The restriction is not in the documentation, and the PR description says GOAMD64 filtering is supported.

Please add `Goamd64 string` to `config.APKBuild`, default it to `"v1"` in `Default`, and use it at `:113`.

---

## 5. Smaller items

- **`options` drops `!check` without telling the user.** `apkbuild.go:62` sets the default only when the list is empty, so `options: ["!strip"]` removes `!check`. The docs present `!check` as the default "because this integration packages prebuilt binaries". `abuild` then gives a warning, and stops with an error where `REQUIRE_CHECK` is set (`abuild.in:360`). Please either always append `!check`, or document the replacement.
- **The `sh` test has no Windows guard.** `apkbuild_test.go:247`, `:252`, `:263` are the only `exec.Command("sh", ...)` calls under `internal/`, and `task test` runs on `windows-latest` without a guard (`build.yml:48`, `:125`). Please use `testlib.SkipIfWindows` or `testlib.CheckPath` (`internal/testlib/path.go:12`, `:37`), and `exec.CommandContext(t.Context(), ...)` as `git_test.go:434` does.
- **The publish is missing from the release summary.** `doPublish` (`apkbuild.go:526`) has no `summary.Appendf`. Compare `aur.go:438`, `aursources.go:390`, and `nix.go:417`.
- **Documentation version.** `apkbuilds.md:6` says `v2.18`, but `v2.18.0` is tagged and the nightlies are `v2.19.0-*`. Please use `v2.19`.
- **Templated fields.** The same file marks only `description`, but `doRun` (`apkbuild.go:97-105`) also applies templates to `name`, `homepage`, `license`, and `skip_upload`.
- **Indentation of `package()`.** `fixLines` (`apkbuild.go:279`) writes one tab, but the per-architecture branch puts the label at two tabs and `;;` at three (`tmpl.go:75-84`). The shell accepts it, but this file gets committed into a package repository and people read it.

---

## What I checked and found correct

I traced the generated file against the `abuild` sources, not the wiki:

- `sha512sums="$_sha512  $_source"` passes `default_validate`, which counts two fields for each source. The two-space separator is right.
- `source="$_source::$_url"` works with `filename_from_uri` and `is_remote`.
- All five accepted formats are in `default_unpack`, and busybox supplies `unzip` and `unxz` on Alpine (`main/busybox/busyboxconfig`: `CONFIG_UNZIP=y`), so no extra `makedepends` is necessary.
- An extension-less binary source still appears in `$srcdir`, because `default_fetch` makes a symbolic link.
- No `builddir` is necessary: `default_prepare` returns 0 when there are no patches, and `build()` is a no-op.
- The source name `pkgname-pkgver-arch.ext` avoids the "needs to be renamed" rule, and the one-maintainer limit agrees with `abuild`.
- The output is deterministic. The three maps are used for lookups only, and every slice that is written out is sorted (`apkbuild.go:142`, `:419-422`).
- `shellQuote` is correct and is applied to each field that the user controls.
- `artifact.APKBuild` is appended last before `lastMarker`, so the integers in `dist/artifacts.json` do not move, and it is correctly excluded from `isUploadable`.
- The tests are behavioural. There are no golden files that were made from the implementation itself.

---

Disclosure: this review was prepared with AI assistance. I read and agree with each point above.

internal/pipe/apkbuild/apkbuild.go
```diff
@@ -0,0 +1,530 @@
...
+		suffix string
+	}{
+		{name: "preview", suffix: "pre"},
+		{name: "snapshot", suffix: "git"},
+		{name: "alpha", suffix: "alpha"},
+		{name: "beta", suffix: "beta"},
+		{name: "pre", suffix: "pre"},
+		{name: "rc", suffix: "rc"},
+	} {
+		if value != candidate.name && !strings.HasPrefix(value, candidate.name+".") && !strings.HasPrefix(value, candidate.name+"-") {
```
caarlos0: **This inverts the apk upgrade order. Please fix before merge.**

This condition matches a candidate only when the word is exact, or is followed by `.` or `-`. A pre-release word with no separator (`rc1`, `beta1`, `alpha1`) matches nothing, and falls through to `"_pre" + toAPKVersionNumber(value)` at line 394.

`toAPKVersionNumber` writes three digits for each character, so the size of the number follows the **length of the word**, not the release stage:

| tag | `pkgver` | digits |
| --- | --- | --- |
| `v1.0.0-alpha1` | `1.0.0_pre1097108112104097049` | 19 |
| `v1.0.0-beta1` | `1.0.0_pre1098101116097049` | 16 |
| `v1.0.0-rc1` | `1.0.0_pre1114099049` | 10 |

apk compares that suffix as a number (`apk_blob_pull_uint`, `apk-tools/src/blob.c`, `uint64_t`). So apk reads `alpha1 > beta1 > rc1`. A user on `alpha1` never gets `beta1` or `rc1`, and `apk upgrade` does nothing.

The separated forms are correct, which hides the problem: `-rc.1` gives `1.2.3_rc1`.

**Smallest fix:** also accept a digit directly after the candidate word. Use `strings.CutPrefix`, then `continue` only when the remainder starts with a character that is not `.`, `-`, or a digit. Then `rc1` gives `_rc1`, and `beta1` gives `_beta1`.

internal/pipe/apkbuild/apkbuild.go
```diff
@@ -0,0 +1,530 @@
...
+		}
+		if len(parts) > 0 {
+			result += "_p" + strings.Join(parts, "_p")
+		}
+		return result
+	}
+	return "_pre" + toAPKVersionNumber(value)
+}
+
+func toAPKVersionNumber(value string) string {
```
caarlos0, line 425: Related to the order defect above: this encoding also overflows.

`apk_blob_pull_uint` reads into a `uint64_t`, which holds 20 digits. This function writes `1` plus three digits for each character, so a value of more than six characters wraps:

- our default snapshot template `{{ .Version }}-SNAPSHOT-{{ .ShortCommit }}` (`internal/pipe/snapshot/snapshot.go:23`) gives a **49-digit** number;
- `+build.1` metadata gives 22 digits.

If you keep this fallback for words that are not known, please limit its length.

internal/pipe/apkbuild/apkbuild_test.go
```diff
@@ -0,0 +1,687 @@
...
+		require.Equal(t, "armhf", toAPKArch("arm", ""))
+	})
+	t.Run("experimental default GOARM", func(t *testing.T) {
+		t.Setenv("GORELEASER_EXPERIMENTAL", "defaultgoarm")
+		require.Equal(t, "armv7", toAPKArch("arm", ""))
+	})
+}
+
+func TestToAPKVersion(t *testing.T) {
+	for input, expected := range map[string]string{
```
caarlos0, line 66: These cases all use a separator (`beta.1`, `rc.2`, `preview.1`, `rc-1`), so the defect at `apkbuild.go:361` is not caught.

Please add:

```go
"1.2.3-rc1":    "1.2.3_rc1",
"1.2.3-beta1":  "1.2.3_beta1",
"1.2.3-alpha1": "1.2.3_alpha1",
```

Please also add an order assertion, not only a validity assertion. The regex test below proves that the string is valid; it does not prove that the order is correct. Something of this form:

```go
require.Less(t, toAPKVersion("1.2.3-alpha1"), toAPKVersion("1.2.3-beta1"))
```

An assertion that prints the two values on failure is worth more here than the fix that it ships with.

internal/pipe/apkbuild/apkbuild.go
```diff
@@ -0,0 +1,530 @@
...
+		arch := toAPKArch(art.Goarch, art.Goarm)
+		index, ok := selected[arch]
+		if !ok {
+			selected[arch] = len(result)
+			result = append(result, art)
+			continue
+		}
+
+		current := result[index]
+		if current.Type != artifact.UploadableArchive || art.Type != artifact.UploadableArchive ||
```
caarlos0: `formats: ["tar.gz", "binary"]` stops the release, and `ids` cannot correct it.

One archive block can emit both formats (`archive.go:129-148`: `binary` goes to `skip()`, the others go to `create()`). That gives an `UploadableArchive` and an `UploadableBinary` with the **same ID**, for the same platform.

This condition removes duplicates by format only when both artifacts are archives. For mixed types it appends, and `dataFor` then fails at line 440. `ContinueOnError` applies only to the publish phase (`publish.go:99`), so this stops the full run after the builds, archives, and signing are complete.

**Smallest fix:** prefer the archive over the binary for the same ID and architecture, with the priority path that is already here.

**Test:** one archive ID that emits both `tar.gz` and `binary`, and an assertion that the `tar.gz` source is selected.

internal/pipe/apkbuild/apkbuild.go
```diff
@@ -0,0 +1,530 @@
...
+	}
+
+	seen := make(map[string]struct{}, len(artifacts))
+	for _, art := range artifacts {
+		arch := toAPKArch(art.Goarch, art.Goarm)
+		if arch == "" {
+			continue
+		}
+		if _, ok := seen[arch]; ok {
+			return result, fmt.Errorf("multiple artifacts found for Alpine architecture %s; use ids to select one", arch)
```
caarlos0, line 464: This message sends the user to a control that cannot help.

When the two artifacts come from one archive block with `formats: ["tar.gz", "binary"]` (see my note at line 206), their IDs are identical, so `ids` cannot select one of them. The user must change the `archives` configuration, but the message does not say so.

If you keep the error for the cases that `ids` can in fact correct, please make the mixed-type case select an artifact instead of failing.

internal/pipe/apkbuild/apkbuild.go
```diff
@@ -0,0 +1,530 @@
...
+	); err != nil {
+		return err
+	}
+
+	filters := []artifact.Filter{
+		artifact.ByGoos("linux"),
+		artifact.Or(
+			artifact.And(
+				artifact.ByGoarch("amd64"),
+				artifact.ByGoamd64("v1"),
```
caarlos0: `goamd64` is pinned here, and `config.APKBuild` has no field for it.

`aur` (`aur.go:68`), `nix` (`nix.go:85`), and `brew` (`brew.go:76`) all have a `Goamd64` field with the default `v1`. A project that builds `goamd64: [v3]` gets the hard error `no linux binaries or archives found`, and has no way to configure around it. The restriction is not in the documentation, and the PR description says that GOAMD64 filtering is supported.

Please add `Goamd64 string` to `config.APKBuild`, default it to `"v1"` in `Default`, and use it here.

internal/pipe/apkbuild/apkbuild.go
```diff
@@ -0,0 +1,530 @@
...
+		if pkg.CommitMessageTemplate == "" {
+			pkg.CommitMessageTemplate = defaultCommitMsg
+		}
+		if pkg.Name == "" {
+			pkg.Name = ctx.Config.ProjectName
+		}
+		if pkg.Rel == "" {
+			pkg.Rel = "0"
+		}
+		if len(pkg.Options) == 0 {
```
caarlos0: This removes `!check` without telling the user.

The default applies only when the list is empty. A user who writes `options: ["!strip"]` loses `!check`, which the documentation presents as the default "because this integration packages prebuilt binaries".

`abuild` then gives a warning from `default_check`, and stops with an error where `REQUIRE_CHECK` is set (`abuild.in:360`), which is the aports policy.

Please either always append `!check`, or record the replacement in the documentation.

internal/pipe/apkbuild/apkbuild.go
```diff
@@ -0,0 +1,530 @@
...
+			SSHCommand: cfg.GitSSHCommand,
+		},
+		Name: fmt.Sprintf("%x", sha256.Sum256([]byte(cfg.GitURL))),
+	})
+
+	content, err := os.ReadFile(file.Path)
+	if err != nil {
+		return err
+	}
+	return uploader.CreateFiles(ctx, author, repo, message, []client.RepoFile{{
```
caarlos0: Please add a `summary.Appendf` call after this, as the other publish pipes do:

- `aur.go:438` - `Pushed AUR package ...`
- `aursources.go:390`
- `nix.go:417`

Without it, the APKBUILD push is missing from the release summary.

internal/pipe/archive/archive.go
```diff
@@ -243,7 +243,7 @@ func create(ctx *context.Context, arch config.Archive, binaries []*artifact.Arti
 		}); err != nil {
 			return fmt.Errorf("failed to add: '%s' -> '%s': %w", binary.Path, dst, err)
 		}
-		bins = append(bins, binary.Name)
+		bins = append(bins, dst)
```
caarlos0, line 246: **Please move this to its own pull request.**

This change is correct. `dst` is the true path in the archive, and `binary.Name` was not. But it differs from `binary.Name` only when `strip_binary_directory: true`, and it changes `ExtraBinaries`, which is also the documented `.Binaries` template field.

It therefore changes generated output for users who do not use `apkbuilds`, in seven pipes: `brew.go:402`, `nix.go:531`, `scoop.go:400`, `aur.go:157`, `krew.go:246`, `winget.go:557`, and `cask.go:459`.

No test in any of those packages exercises `strip_binary_directory`, and today the change ships as a one-line edit to a documentation table (`artifacts.md:87`).

Please send it as `fix(archive):` with its own changelog line, and add a `strip_binary_directory` case to at least `brew_test.go` and `nix_test.go`.

internal/client/git.go
```diff
@@ -131,6 +103,28 @@ func (g *gitClient) CreateFiles(
 		}
 	}
 
+	if err := unsetGitConfigs(ctx, cwd, env, "user.signingKey", "gpg.program", "gpg.format"); err != nil {
```
caarlos0: **Please move this to its own pull request.**

Moving the `git config` block out of the fresh-clone branch, and adding `unsetGitConfigs`, repairs a real defect: a reused checkout kept the identity and signing configuration of the first caller. It is a good fix, and I found no regression for the nine `CreateFiles` callers. `commit.gpgSign` gives the same value, `init.defaultBranch` has no effect after `clone` in either order, and `checkout -b` needs no identity.

But this feature does not need it. The `apkbuild` checkout directory is `<sha256>-` and `aur`'s is `<sha256>-master`, so they never collide. The new test uses one client with two authors, not this pipe.

Please send it as `fix(client):` with its own changelog line. It repairs `aur`, `nix`, `brew`, `krew`, `scoop`, and `winget`, so it deserves its own review.

internal/pipe/apkbuild/apkbuild_test.go
```diff
@@ -0,0 +1,687 @@
...
+	require.FileExists(t, file)
+	content, err := os.ReadFile(file)
+	require.NoError(t, err)
+	require.Contains(t, string(content), `pkgdesc='It'"'"'s "$HOME"'`)
+	require.Contains(t, string(content), `arch='aarch64 armhf armv7 ppc64le riscv64 s390x x86 x86_64'`)
+	require.Contains(t, string(content), `options='!check' # prebuilt binaries`)
+	require.Contains(t, string(content), `install -Dm755 "$srcdir/wrapped-x86_64/foo" "$pkgdir/usr/bin/foo"`)
+	require.NotContains(t, string(content), "ignored_v3")
+
+	cmd := exec.Command("sh", "-n", file)
```
caarlos0: These are the only `exec.Command("sh", ...)` calls under `internal/`, and `task test` runs on `windows-latest` with no guard (`.github/workflows/build.yml:48`, `:125`).

Please add `testlib.SkipIfWindows(t, "APKBUILD is POSIX sh")` or `testlib.CheckPath(t, "sh")` (`internal/testlib/path.go:12`, `:37`). There are about 40 uses of that helper already.

Please also use `exec.CommandContext(t.Context(), ...)`, as `internal/client/git_test.go:434` does.

www/content/customization/publish/apkbuilds.md
caarlos0: Please use `v2.19`. `v2.18.0` is tagged, and the nightlies are `v2.19.0-*`, so this feature lands in v2.19.

www/content/customization/publish/apkbuilds.md
```diff
@@ -0,0 +1,161 @@
...
+    # Artifact IDs to filter for.
+    # Empty means all IDs (no filter).
+    # There must be at most one artifact per Alpine architecture.
+    ids:
+      - foo
+      - bar
+
+    # Your app's homepage.
+    # Required.
+    homepage: "https://example.com/"
```
caarlos0, line 45: `doRun` (`apkbuild.go:97-105`) applies templates to `name`, `homepage`, `license`, and `skip_upload` as well, but only `description` is marked below.

Please add `# Templates: allowed.` to those four fields.

internal/pipe/apkbuild/tmpl.go
```diff
@@ -0,0 +1,87 @@
...
+
+package() {
+{{- if .Package }}
+{{ fixLines .Package }}
+{{- else }}
+	case "$CARCH" in
+{{- range .ReleasePackages }}
+		{{ .Arch }})
+{{ fixLines .Package }}
+			;;
```
caarlos0, line 80: The body of the per-architecture branch is misindented. `fixLines` (`apkbuild.go:291`) writes exactly one tab, but this branch puts the label at two tabs and `;;` at three:

```sh
	case "$CARCH" in
		x86_64)
	install -Dm755 "$srcdir/wrapped-x86_64/foo" "$pkgdir/usr/bin/foo"
			;;
	esac
```

The shell accepts it, and `sh -n` passes, but this file gets committed into a package repository and people read it. The `{{ if .Package }}` branch above is correct.

Please give `fixLines` a depth, or add an `indent` function. A test that asserts on the rendered `package()` block, rather than `Contains`, would hold this.

caarlos0 (changes requested):
I read the complete change. Thanks - this is careful work. The pipe follows the `aur` structure, and the tests are better than the ones we have for `aur`: they run the generated file through `sh` with `CARCH` set and a stubbed `install`, and the publish tests clone a real bare repository.

I checked the generated file against the `abuild` and `apk-tools` sources, not the wiki. Most of it holds up. I have put each request on the line it applies to.

**One defect must be fixed:** the `pkgver` conversion inverts the apk upgrade order for `-rc1`, `-beta1`, and `-alpha1` tags.

**Two changes should move to their own pull requests:** `internal/pipe/archive/archive.go` and `internal/client/git.go`. Both are correct, and both repair defects in other pipes, but neither is necessary for Alpine support.

### What I examined and found correct

- `sha512sums="$_sha512  $_source"` satisfies `default_validate`, which counts two fields for each source. The two-space separator is right.
- `source="$_source::$_url"` works with `filename_from_uri` and `is_remote`.
- All five accepted formats are in `default_unpack`. busybox supplies `unzip` and `unxz` on Alpine (`main/busybox/busyboxconfig`: `CONFIG_UNZIP=y`), so no added `makedepends` is necessary.
- A binary source with no extension still appears in `$srcdir`, because `default_fetch` makes a symbolic link.
- No `builddir` is necessary. `default_prepare` returns 0 when there are no patches, and `build()` is a no-op.
- The source name `pkgname-pkgver-arch.ext` avoids the "needs to be renamed" rule, and the one-maintainer limit agrees with `abuild`.
- The output is deterministic. The three maps are used for lookups only, and every slice that is written out is sorted.
- `shellQuote` is correct, and is applied to each field that the user controls.
- `artifact.APKBuild` is appended last before `lastMarker`, so the integers in `dist/artifacts.json` do not move.
- The tests are behavioural. There are no golden files that were made from the implementation itself.

This review replaces my earlier one, which put all of this in a single comment. I have dismissed that one.

Disclosure: this review was prepared with AI assistance. I read and agree with each point.

internal/client/git.go
davicbtoliveira: It was moved outside the fresh-clone block so that an existing/reused checkout is reconfigured for the current caller. Previously, user.name, user.email, and signing settings were only applied when
the repository was first cloned, so later calls could retain the first caller’s author and signing configuration.

Signing is still conditional: commit.gpgSign is set from commitAuthor.Signing.Enabled, while user.signingKey, gpg.program, and gpg.format are only set when provided. Those optional values are unset first to prevent stale settings from a previous call.

That said, this is a shared Git-client fix unrelated to APKBUILD support, so I’ll move it into a separate fix(client) branch/PR as requested.

internal/pipe/apkbuild/apkbuild.go
davicbtoliveira: Addressed. selectArtifacts now prefers an UploadableArchive over an UploadableBinary when they have the same ID and Alpine architecture. Therefore, an archive configuration with formats: ["tar.gz", "binary"] selects the archive and no longer reaches this error. I added a regression test asserting that the tar.gz source is used.

davicbtoliveira:
Thanks for the thorough review, @caarlos0! I addressed all requested changes:

 - Fixed APK prerelease conversion for alpha1, beta1, and rc1.
- Added ordering tests using apk version --test, including ordering before the final release.
- Bounded fallback version numbers to uint64.
- Prefer archives over binaries sharing the same ID and architecture.
- Added configurable goamd64, defaulting to v1.
- Ensured !check is always included.
- Made shell tests portable and switched them to exec.CommandContext.
- Added APKBUILD publishing to the release summary.
- Updated the documentation version and template annotations.
- Fixed the generated package() indentation.

I also extracted the unrelated shared fixes:

 - Archive binary paths: [fix(archive): report stripped binary paths](https://github.com/goreleaser/goreleaser/pull/6820)
 - Reused Git checkout configuration: [fix(client): reconfigure reused Git checkouts](https://github.com/goreleaser/goreleaser/pull/6821)
 
The full test suite pass.

PR #6717, opened 2026-07-23
feat: allow PR creation to use a different auth token

emily-curry:
This change adds support for providing a token to the `PullRequest` configuration, which will be used for opening the PR _only_. This utilizes the same patterns used for creating a client with a different auth token in `RepoRef`.

My use case is that I have a GitHub org which has classic PATs disallowed at the org level, so we can only use app installations or fine-grained tokens. This works for almost everything goreleaser does, except for opening PRs into public repos we don't control such as [microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs), this can only be accomplished using a classic PAT. If we use the fine-grained token to write to our fork, and use the PAT to open the PR from the fork to the upstream, we can keep our "allow classic PAT" setting disabled at the org level. I tested this scenario with a local build, where the [fork](https://github.com/DopplerTest/winget-pkgs) was in an org with these restrictive settings ([generated PR](https://github.com/emily-curry/winget-pkgs/pull/1)).

I looked into adding tests, but `client.NewIfToken` won't create another mock client, so it'll try to create a real PR if I set the token property. There weren't any existing tests that I saw which tested the branching behavior if `RepoRef.Token` was set, and I didn't want to change the shared test utilities as that might make this more difficult to review, but I can look into it more if necessary. Thanks for all your work on this project!

### DIFF
--- a/www/content/includes/repository.md
+++ b/www/content/includes/repository.md
@@ -48,6 +48,12 @@
         # Whether to open the PR as a draft or not.
         draft: true
 
+        # An optional token that can be provided for opening the pull request.
+        #
+        # Templates: allowed.
+        # {{< g_inline_version "v2.18-unreleased" >}}
+        token: "{{ .Env.GITHUB_PERSONAL_AUTH_TOKEN }}"
+
         # If the pull request template has checkboxes, enabling this will
         # check all of them.
         #
--- a/internal/pipe/brew/brew.go
+++ b/internal/pipe/brew/brew.go
@@ -188,7 +188,11 @@ func doPublish(ctx *context.Context, formula *artifact.Artifact, cl client.Clien
 	}
 
 	log.Info("brews.pull_request enabled, creating a PR")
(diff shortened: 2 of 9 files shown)


caarlos0:
thanks!

www/content/includes/repository.md
```diff
@@ -48,6 +48,11 @@
         # Whether to open the PR as a draft or not.
         draft: true
 
+        # An optional token that can be provided for opening the pull request.
+        #
+        # Templates: allowed.
```
caarlos0, line 53:
```suggestion
        # Templates: allowed.
        # {{< g_inline_version "v2.18-unreleased" >}}
```

emily-curry: Applied in [this](https://github.com/goreleaser/goreleaser/compare/cf5af5100a8d19daa3f076e155db2c492e3c525b..ac697c0bb93bef6e01597378ad1664761063a6f8) force push.

caarlos0 (approved):

PR #6725, opened 2026-07-24
feat: support Gentoo Overlays as an output target

arran4:
If applied, this commit will.... Allow for goreleaser to publish Gentoo Ebuilds to gentoo overlays via Git and optionally PRs.

This is an AI supported build, mostly with Jules, heavily reviewed each commit, and I understand how it works.

I have attempted to make it reasonably featured, there are still of a lot of potential improvements from different use cases that will need to be discovered.

It supports:
* simple but supports most of the install functions: https://devmanual.gentoo.org/function-reference/install-functions/
* `category/package/files` storage (for extra files)
* Optional purning of older ebuilds (highly recommended)
* Generation of the manifest files (and adhering to `metadata/layout.conf` 's specification as to which checksums to use if any otherwise sane defaults
* Upserting `category/package/metadata.xml`
* Reasonably LOC efficient ebuilds
* Useflags for installed extra files

It doesn't support:
* Building from source (but it supports the idea of it, it's clear these are `-bin` repos) The reason why is the gentoo standard go install system is a bit immature

I am trying to reduce my reliance on my [own overlay](https://github.com/arran4/arrans_overlay) which heavily uses [generated](https://github.com/arran4/arrans_overlay_workflow_builder/) code checks.

I have a github action verifying the install via a docker contianer:

https://github.com/arran4/goreleaser-gentoo-smoke-overlay/actions/runs/31479941333/job/93742337812?pr=29

The config file look as follows:

https://github.com/arran4/goreleaser-gentoo-smoke/blob/main/.goreleaser.yaml#L35-L90

Smoke test github action run:
https://github.com/arran4/goreleaser-gentoo-smoke/actions/runs/31479785100/job/93741840394

And output:
https://github.com/arran4/goreleaser-gentoo-smoke-overlay/pull/29

Where I will need help
* gitea
* gitlab
* git

I needed to add a directory listing api to this to get it to work IDK how it will work with the other ones. I tried to make it conditional as when you have fat manifests (ie thin-manifest is off) it contains a complete listing of the files including the manifests. Ideally we should be able to support thin manifest in the other git hosts.

I believe that this is ready for review.


caarlos0:
Hi, thanks for the PR!

Due to low usage of gentoo out there, i'm not sure if this is something I want to merge in (and maintain).


Thanks for the PR either way.

arran4:
Ah okay. I was going to continue working on it.. Is it worth while finishing it?

arran4:
@caarlos0 would it help I strip it back to just deploying a basic ebuild, manifest, and the cache? Which would be mvp. However it will need to read the manifest to append info it.

PR #6752, opened 2026-08-08
fix(cask): emit Casks that pass brew style

### LINKED ISSUE #6747: Generated Homebrew Cask fails current brew style stanza checks
### What happened?

GoReleaser v2.17.1 successfully generates a multi-platform binary Cask, but the generated Ruby fails the current `brew style` checks. The file is generated by GoReleaser and has no user-controlled template for stanza ordering/grouping.

### Reproduction

Configuration excerpt:

```yaml
builds:
  - id: cast
    goos: [linux, windows, darwin]
    goarch: [amd64, arm64]
archives:
  - id: cast
    formats: [tar.gz]
homebrew_casks:
  - name: cast
    ids: [cast]
    binaries: [cast]
    directory: Casks
```

Run:

```console
goreleaser release --snapshot --clean --skip=publish,notarize
brew style dist/homebrew/Casks/cast.rb
```

Result with Homebrew 6.0.15-73-g7d75aaa:

```text
Cask/StanzaOrder: on_intel stanza out of order
Cask/StanzaOrder: on_arm stanza out of order
Cask/StanzaGrouping: stanzas within the same group should have no lines between them
Layout/EmptyLinesAroundBlockBody: Extra empty line detected at block body end
1 file inspected, 6 offenses detected, 6 offenses autocorrectable
```

The generated Cask uses nested `on_macos` / `on_linux` blocks with `on_intel` before `on_arm`, blank lines between same-group platform stanzas, and an extra blank line before the final `end`.

### Expected behavior

A Cask emitted by the latest GoReleaser should pass the current Homebrew style rules without post-processing.

### Environment

- GoReleaser: v2.17.1
- Homebrew: 6.0.15-73-g7d75aaa
- macOS arm64

Downstream reproduction: https://github.com/neosapience/cast/pull/32

r0h1tb:
Generated Casks now pass `brew style` for two rules they previously always
tripped — output the user has no template control over.

**Why:** `brew style` failed on every Cask GoReleaser emits.

`on_arm` must precede `on_intel`. Homebrew's canonical order is

```ruby
ON_SYSTEM_METHODS_STANZA_ORDER = [:arm, :intel, ...]
```

(`rubocops/cask/constants/stanza.rb`). Packages were sorted by comparing the
arch strings, and `"amd64" < "arm64"`, so `on_intel` always came first and
`Cask/StanzaOrder` complained. Now sorted by an explicit stanza rank.

`Layout/EmptyLinesAroundBlockBody` rejects a blank line before a block's `end`,
and the template left one — with no caveats, `{{ with .Caveats -}}` still
emitted its leading newlines. The `on_macos` and `on_linux` blocks were also
separated by a blank line, though every `on_*` method belongs to a single
`Cask/StanzaGrouping` group.

## Measured, not assumed

`brew style --cask` over all 27 generated fixtures, before and after, on
Homebrew 6.0.9:

| cop | before | after |
|---|---:|---:|
| `Layout/EmptyLinesAroundBlockBody` | **10** | **0** |
| `Cask/StanzaGrouping` | 52 | 52 |
| `Layout/FirstArrayElementIndentation` | 72 | 72 |
| `Layout/HashAlignment` | 71 | 71 |
| `Layout/ArgumentAlignment` | 37 | 37 |
| `Cask/HomepageUrlStyling` | 21 | 21 |
| `Cask/ArrayAlphabetization` | 16 | 16 |
| `Sorbet/BlockMethodDefinition` | 4 | 4 |

Nothing else moves. Goldens regenerated with
`go test ./internal/pipe/cask/ -update`; reverting the source with the goldens
kept fails the suite showing `on_intel` first and the trailing blank line, so
the fixtures pin the change rather than restate it.

`go test ./internal/pipe/cask/... ./internal/pipe/brew/...` pass, `go build ./...`,
`go vet` and `gofmt -l` clean.

## Two things I want to flag rather than let you find

**`Cask/StanzaOrder` does not fire on my Homebrew 6.0.9** for `on_*` nested
inside `on_macos`/`on_linux` — the reporter saw it on 6.0.15. So the ordering
half is verified against the documented constant and the regenerated fixtures,
not against a local `brew style` run. If you would rather see that half proven
against 6.0.15 before merging, say so and I will get on that version.

**The remaining 52 `Cask/StanzaGrouping` offences are pre-existing and left
alone.** They concern blank lines between other stanza groups and need a wider
pass over the template than a bug fix should carry. Happy to follow up.

## AI disclosure

Per CONTRIBUTING.md: **AI was used.** This change was written with Claude Code,
and the commit carries an `Assisted-by:` marker. The before/after `brew style`
numbers, the test runs and the revert check above are real runs on my machine,
not generated text.

In the interest of your "agents pretending to be a human" rule: I am a human
contributor using an assistant, this is not an unattended bot posting on random
repositories, and I will respond to review. Close it if the disclosure still
puts it outside what you want.

Fixes #6747

### DIFF
--- a/internal/pipe/cask/cask.go
+++ b/internal/pipe/cask/cask.go
@@ -465,7 +465,31 @@ func dataFor(ctx *context.Context, cfg config.HomebrewCask, cl client.ReleaseURL
 	return result, nil
 }
 
+// archStanzaRank orders packages so the generated `on_*` blocks come out in
+// Homebrew's canonical stanza order, which puts arm before intel:
+//
+//	ON_SYSTEM_METHODS_STANZA_ORDER = [:arm, :intel, ...]
+//
+// (Homebrew/rubocops/cask/constants/stanza.rb). Sorting the arch strings
+// alphabetically put amd64 first, so every generated Cask emitted `on_intel`
+// before `on_arm` and tripped Cask/StanzaOrder.
+func archStanzaRank(arch string) int {
+	switch arch {
+	case "all":
+		return 0
+	case "arm64":
+		return 1
+	case "amd64":
+		return 2
+	default:
+		return 3
+	}
+}
+
 func compareByArch(a, b releasePackage) int {
+	if c := cmp.Compare(archStanzaRank(a.Arch), archStanzaRank(b.Arch)); c != 0 {
+		return c
+	}
 	return cmp.Compare(a.Arch, b.Arch)
 }
 
--- a/internal/pipe/cask/templates/cask.rb
+++ b/internal/pipe/cask/templates/cask.rb
@@ -13,13 +13,12 @@ cask "{{ .Name }}" do
   on_macos do
   {{- include "macos_packages" . | indent 2 }}
   end
-  {{ end }}
-
-  {{ if .LinuxPackages }}
+  {{- end }}
+  {{- if .LinuxPackages }}
   on_linux do
   {{- include "linux_packages" . | indent 2 }}
   end
-  {{ end }}
+  {{- end }}
 
   name "{{ .Name }}"
   desc "{{ .Description }}"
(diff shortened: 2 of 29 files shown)


caarlos0 (approved):
thank you!

PR #6755, opened 2026-08-09
ci: disable Zig cache on Windows

### LINKED ISSUE #6754: ci: Windows tests fail permanently once the shared setup-zig cache is corrupted
## Summary

The Windows `test` job fails intermittently — and then **permanently** — with linker errors pointing at missing files inside the shared `.zig-cache`:

```
--- FAIL: TestBuild (14.48s)
    internal/builders/zig/build_test.go:148
        exit status 2: error: lld-link: could not open
        'D:\a\goreleaser\goreleaser\.zig-cache\o\7a0e4a9a703ea62c018fa090e09494ee\compiler_rt.lib':
        No such file or directory

--- FAIL: TestBuild (16.54s)
    internal/builders/rust/build_test.go:164
        exit status 101: error: linking with `zigcc-aarch64-unknown-linux-gnu...bat` failed
        ld.lld: error: cannot open ...\.zig-cache\o\440b11de9981c96f2cd928d4a17ac08d\Scrt1.o: No such file or directory
        ld.lld: error: cannot open ...\.zig-cache\o\65a43333793b394990036c8f85ebe5ab\libc.so.6: No such file or directory
        ld.lld: error: cannot open ...\.zig-cache\o\d461fcf302559f0461370016dc0435a4\libcompiler_rt.a: No such file or directory
        (... and ~8 more)
```

Example: https://github.com/goreleaser/goreleaser/actions/runs/31270375643/job/93135210685

This is a Zig cache manifest **hit** whose output directory is missing — i.e. a corrupt cache, not a compile error.

## Not the runner image

Initial suspicion was the `windows-latest` image. It isn't:

| | passing run `31075257398` | failing run `31270375643` |
|---|---|---|
| Image | `win25-vs2026/20260728.188` | `win25-vs2026/20260803.193` |
| Zig | 0.16.0 | 0.16.0 |

The delta between those two images is only .NET SDK patch bumps (8.0.128→8.0.129 etc.). Nothing was removed and nothing extra needs installing.

## Root cause

`mlugg/setup-zig` sets **both** `ZIG_LOCAL_CACHE_DIR` and `ZIG_GLOBAL_CACHE_DIR` to `$GITHUB_WORKSPACE/.zig-cache`:

```js
function getZigCachePath() {
  return path.join(process.env['GITHUB_WORKSPACE'] ?? process.cwd(), '.zig-cache');
}
```

and persists that directory via `actions/cache` using a **prefix restore-key**
(`setup-zig-cache-v2-test-zig-x86_64-windows-0.16.0--`). So the directory is:

1. shared between `internal/builders/zig` and `internal/builders/rust` (cargo-zigbuild), which run **concurrently** under `go test ./...`;
2. shared between local and global Zig cache namespaces;
3. carried forward across runs *and branches*.

Because of (3), once the cache is corrupted, **every subsequent Windows run restores the corruption and fails deterministically**, then re-saves it. The cache blob sizes made this visible — Windows entries split cleanly into two populations:

- **~33.3 MB** → runs that pass (`33334176`, `33331088`, `33233301`, `33397111`)
- **~30.4 MB** → runs that fail (`30431920`, `30436604`, `30436337`, `30483009`)

Run `31235473562` (2026-08-08 02:49) restored a healthy 33.2 MB cache, corrupted it during the run, and saved `30431920`. Everything after that inherited it.

Linux is unaffected by the same setup, which is consistent with Windows file-locking/deletion semantics.

## Immediate mitigation (already applied)

Deleted all 17 `setup-zig-cache-v2-*-windows-*` entries so Windows starts cold. Verified on job `93138132470`:

```
Attempting restore of Zig cache with prefix 'setup-zig-cache-v2-test-zig-x86_64-windows-0.16.0--'
Cache miss: leaving Zig cache directory ... unpopulated
...
ok  github.com/goreleaser/goreleaser/v2/internal/builders/rust  37.533s
ok  github.com/goreleaser/goreleaser/v2/internal/builders/zig   32.057s
```

## Why it will come back

This is only a mitigation — nothing stops the cache being re-poisoned.

A likely seeding mechanism showed up while verifying: a Windows job **cancelled 3 minutes into `task test`** (right in the zig/rust builder tests) still ran `setup-zig`'s post step and published the half-written cache:

```
2026-08-08T18:18:56Z ##[group]Run task test RACE=false COVER=false
2026-08-08T18:22:07Z ##[error]The operation was canceled.
2026-08-08T18:22:11Z Saving Zig cache with key 'setup-zig-cache-v2-...-31270375643-2'
```

Killed `zig`/`zig cc` processes can leave a committed manifest whose output dir was never finished. `post.js` saves unconditionally, so that state becomes the newest cache and the next run picks it up via the prefix key. Now that `ci: merge queue` has landed, cancellations are more frequent.

The concurrent access in (1) is the other candidate — `builders/rust` and `builders/zig` overlap on Windows against one mutable cache dir.

## Possible fixes

1. **Split local vs global** — keep a shared, cached `ZIG_GLOBAL_CACHE_DIR` (where the expensive `compiler_rt`/libc stubs live) but give each test its own `ZIG_LOCAL_CACHE_DIR`. Keeps the speed benefit, removes the collision.
2. **Isolate per test** — `t.Setenv` both cache dirs to the test's temp dir in the rust and zig build tests. Simple, but discards all cache reuse.
3. **Don't cache on Windows** — `use-cache: false` on `setup-zig` for the Windows matrix leg. One line, slower job.
4. **Upstream** — `setup-zig` arguably shouldn't save the cache when the job was cancelled or failed.

## Workaround for anyone hitting this now

Purge the caches and re-run:

```console
gh cache list --repo <owner>/<repo> --key setup-zig-cache-v2 --json id,key \
  --jq '.[]|select(.key|test("windows"))|.id' \
  | xargs -n1 gh cache delete --repo <owner>/<repo>
```

alloutflo:
Disable persisted `setup-zig` caching only for the Windows test matrix leg. Linux keeps the existing cache behavior.

The pinned action's `use-cache` input gates both its restore path and its post-job `saveCache` call. A cancelled or failed Windows job can therefore use a fresh, run-local cache while it is executing, but can no longer publish a partially written cache or restore one into later runs.

Windows jobs have repeatedly restored a corrupt shared Zig cache whose manifest points to missing compiler/runtime outputs. Clearing the cache makes a cold run pass, but the action's unconditional post step can poison the prefix-restored cache again after cancellation. Disabling persistence on the affected matrix leg removes that cross-run failure mode at the cost of a cold Windows Zig cache; other platforms retain the speed benefit.

Fixes #6754.

## Verification

- `actionlint 1.7.12 .github/workflows/build.yml`
- `actionlint 1.7.12` across all repository workflows
- Verified the pinned `mlugg/setup-zig` action defines `use-cache` and uses it to guard both restore and post-job save paths
- `git diff --check`

The PR-triggered Windows job provides the real cold-cache integration check.

- #6754
- Pinned action input: https://github.com/mlugg/setup-zig/blob/d1434d08867e3ee9daa34448df10607b98908d29/action.yml

AI disclosure: I used an AI coding assistant to inspect the issue evidence and pinned action behavior, make this focused workflow change, and run the checks above. I reviewed and understand the resulting workflow semantics before submitting.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -106,6 +106,10 @@ jobs:
         with:
           node-version: "25"
       - uses: mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29 # v2.2.1
+        with:
+          # Cancelled Windows jobs can persist a partially written Zig cache,
+          # which poisons subsequent runs restored through the prefix key.
+          use-cache: ${{ matrix.os != 'windows-latest' }}
       - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
       - uses: denoland/setup-deno@22d081ff2d3a40755e97629de92e3bcbfa7cf2ed # v2.0.5
       - uses: cargo-bins/cargo-binstall@e00d2c94cc0067b77737821097a62d91c0301baa # v1.21.1
--- a/.gitignore
+++ b/.gitignore
@@ -21,4 +21,3 @@ internal/pipe/ko/testdata/app/testapp-linux-amd64.spdx.json
 .env
 tmp/
 sponsors.json
-.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ internal/pipe/ko/testdata/app/testapp-linux-amd64.spdx.json
 .env
 tmp/
 sponsors.json
+.DS_Store


caarlos0:
this will make CI slower, IMHO not a proper fix.

PR #6756, opened 2026-08-09
docs: use formats plural syntax

FelicianoTech:
I believe this bottom part of the archives doc was never updated to use the plural syntax for formats.


caarlos0:
thank you!

PR #6758, opened 2026-08-10
fix(notarize): cap macOS notarization timeout at 20m

caarlos0:
Cap the macOS notarization `timeout` at `20m`, failing configuration validation with an actionable error when it is exceeded, and document the new maximum.

Quill derives the App Store Connect authentication token lifetime from the configured notarization timeout:

```go
func (c *NotarizeConfig) WithStatusConfig(cfg notary.StatusConfig) *NotarizeConfig {
	c.StatusConfig = cfg
	c.TokenConfig.TokenLifetime = cfg.Timeout + (2 * time.Minute)
	return c
}
```

Apple [rejects tokens with a lifetime greater than 20 minutes](https://developer.apple.com/documentation/appstoreconnectapi/generating-tokens-for-api-requests), so raising `timeout` past that made every submission fail with:

```
⨯ release failed after 25s   error=dist/foo_darwin_arm64_v8.0/foo: unable to start submission: http status="401 Unauthorized"
```

Nothing in that error points at the timeout, so the natural reaction is to go hunting for malformed keys and certificates instead.

The check runs during defaulting, so the problem now surfaces on `goreleaser check` — before a release starts — with a message that says what to do:

```
⨯ configuration is invalid: sign & notarize macOS binaries: timeout must be at most 20m0s, got 1h0m0s: Apple rejects authentication tokens with longer lifetimes
```

### Changes

- `internal/pipe/notary/macos.go`: new `maxNotarizeTimeout` constant; `Default` errors when the configured timeout exceeds it.
- `internal/pipe/notary/macos_test.go`: `20m` is accepted, `21m` is rejected.
- `www/content/customization/sign/notarize.md`: documents `Maximum: 20m` and why it exists.

### Note

Because quill adds 2 minutes, an allowed `20m` timeout still yields a 22-minute token. That is above Apple's documented limit on paper, but is the value the reporter confirmed working in practice — dropping the cap to `18m` would make it strictly in-spec if this ever proves to be a problem.

Closes #6757

### DIFF
--- a/internal/pipe/notary/macos.go
+++ b/internal/pipe/notary/macos.go
@@ -18,6 +18,12 @@ import (
 	"github.com/goreleaser/quill/quill/pki/load"
 )
 
+// maxNotarizeTimeout is the maximum allowed notarization timeout.
+//
+// Apple's App Store Connect API rejects tokens with a lifetime greater than 20
+// minutes, and the token lifetime is derived from this timeout.
+const maxNotarizeTimeout = 20 * time.Minute
+
 type MacOS struct{}
 
 func (MacOS) String() string { return "sign & notarize macOS binaries" }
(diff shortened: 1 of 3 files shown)


robobario:
checking directly against their API https://gist.github.com/robobario/b11f935b7202bedfa638e9fb3d513c3d (listing submissions, but hopefully comparable to submit)

it fails reliably with exp `now+3602` and works with `now+3600` (and fails sometimes at now+3601). So 60 minutes might be their undocumented limit and my config of 1h+ the 2 minutes pushed it over.

caarlos0:
@robobario I think some endpoints might allow higher timeouts, but it seems like the notarizing one is limited at 20 min, unless I'm understanding this wrong. 🤔

Issue #5487, opened 2025-01-26
config missing from NFPMDebScripts


Daniel15:
### What happened?

From nfpm docs:
```yaml
  # Custom deb special files.
  scripts:
    # Deb rules script.
    rules: foo.sh

    # Deb templates file, when using debconf.
    templates: templates

    # Deb config maintainer script for asking questions when using debconf.
    config: config
```

However, only `rules` and `templates` work with Goreleaser's nfpm config. `config` [is missing](https://github.com/goreleaser/goreleaser/blob/894cf38964e5f495ab31a0419a80bf2c0fec4b75/pkg/config/config.go#L875). Trying to use `config` results in:
```
  ⨯ release failed after 0s                 
    error=
    │ yaml: unmarshal errors:
    │   line 91: field config not found in type config.NFPMDebScripts

```

### How can we reproduce this?

Use `nfpms` -> `deb` -> `scripts` -> `config` field

### goreleaser version

```bash
____       ____      _
 / ___| ___ |  _ \ ___| | ___  __ _ ___  ___ _ __
| |  _ / _ \| |_) / _ \ |/ _ \/ _` / __|/ _ \ '__|
| |_| | (_) |  _ <  __/ |  __/ (_| \__ \  __/ |
 \____|\___/|_| \_\___|_|\___|\__,_|___/\___|_|
goreleaser: Release engineering, simplified.
https://goreleaser.com

GitVersion:    2.6.1
GitCommit:     b6bb03ddc28d6de71a37012107cda26af53cc116
GitTreeState:  false
BuildDate:     2025-01-22T11:11:07Z
BuiltBy:       goreleaser
GoVersion:     go1.23.5
Compiler:      gc
ModuleSum:     h1:wszzGB1PkrkbtjJIKIEdJv5EX6VCFpqYYlP6kH8dQXE=
Platform:      linux/amd64
```

### GoReleaser Check

- [x] goreleaser check shows no errors

### Search

- [x] I did search for other open and closed issues before opening this

### Supporter

- [ ] I am a [sponsor](https://github.com/sponsors/caarlos0/) or a [Pro](https://goreleaser.com/pro) customer

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

caarlos0:
Good catch! Fixed on main!

PR #6759, opened 2026-08-10
Use correct script for debconf

Daniel15:
Putting the debconf questions in the postinst script was a workaround for a bug in goreleaser (#5487). That bug was fixed long ago, so the questions can now be asked in their proper place.


caarlos0:
thank you!

PR #6771, opened 2026-08-15
fix(winget): fall back to the default description in additional locales

VXNCXNX:
Per CONTRIBUTING: AI was used here. Claude Opus 5 via Claude Code, and the commit carries an `Assisted-by` marker. I have read and understand the change.

## What's broken

An additional winget locale that does not set `description` renders it empty, instead of inheriting the default locale's value.

The docs say the opposite, in `www/content/customization/publish/winget.md`:

> Any field that is not set falls back to the default locale manifest value

## The fix

In `doAdditionalLocales`, all fifteen locale fields use `cmp.Or(aloc.X, winget.X)` except one:

```go
ShortDescription:    cmp.Or(aloc.ShortDescription, winget.ShortDescription),
Description:         strings.ReplaceAll(aloc.Description, "\t", "  "),
ReleaseNotes:        cmp.Or(aloc.ReleaseNotes, winget.ReleaseNotes),
```

`Description` now takes the same fallback, with the tab replacement kept on the outside so the inherited value gets the same treatment the default manifest gives it at line 267.

## Verification

New table case `additional-locale-description-fallback` in `winget_test.go`, with a pt-BR locale that sets its own short description and leaves `description` unset. Its golden shows the inherited value:

```
ShortDescription: descricao curta
Description: default description
```

Goldens were generated with `-update`, not hand-written, and regenerating the whole `TestRunPipe` set touches no existing golden. With the one-line change reverted the new test fails on that exact missing line. `go test ./internal/pipe/winget/...` passes.


caarlos0 (approved):

caarlos0:
Thank you!

PR #6773, opened 2026-08-15
fix(changelog): a commit matching two include filters is listed twice

VXNCXNX:
A commit that matches more than one `changelog.filters.include` pattern is listed once per matching pattern.

```yaml
changelog:
  use: git
  filters:
    include:
      - '^feat:'
      - 'parser'
```

Against a range containing one commit, `feat: add parser`:

```
before:
## Changelog
* f2ddf71 feat: add parser
* f2ddf71 feat: add parser

after:
## Changelog
* f2ddf71 feat: add parser
```

Overlapping includes are easy to write without noticing, since each pattern is individually reasonable. A conventional-commit prefix plus a keyword or a scope is enough.

## Cause

`filterEntries` looped over the patterns and appended the full match set of each:

```go
for _, filter := range filters.Include {
	r, err := regexp.Compile(filter)
	...
	newEntries = append(newEntries, keep(r, entries)...)
}
```

So the result is a concatenation of per-pattern matches rather than the subset of entries matching any pattern.

## The fix

Compile the patterns first, then walk the entries once and keep an entry if any pattern matches. `keep` becomes `keepAny` and takes the compiled set; it had no other callers.

## One visible side effect, please look at this

Output is now in commit order rather than grouped by which filter matched.

With disjoint filters whose order disagrees with commit order, this reorders the changelog. Two commits, `feat: alpha` older and `docs: beta` newer, with `include: ['^feat:', '^docs:']`:

```
before:  * feat: alpha
         * docs: beta

after:   * docs: beta
         * feat: alpha
```

I believe commit order is correct, because it is what the unfiltered path and the `exclude` path already produce, so today `include` is the only filter that reorders. But it is a user-visible change to generated release notes and it is your call, not mine. If you would rather preserve grouping-by-filter, dedup while keeping the old append order is a small change and I am happy to send that instead.

## Testing

`TestChangelogIncludeOverlapping` in `internal/pipe/changelog/changelog_test.go`, following the shape of the neighbouring `TestChangelogInclude`. Two commits, two overlapping includes, asserting the matching one appears exactly once and the non-matching one not at all.

Reverting the source while keeping the test:

```
--- FAIL: TestChangelogIncludeOverlapping
    expected: 1
    actual  : 2
```

`gofumpt -l` clean, `go vet` clean, `golangci-lint run --tests --config ./.golangci.yaml` reports 0 issues. The full `go test ./internal/... ./pkg/... ./cmd/...` has one pre-existing failure, `internal/builders/node TestBuild`, which needs Node >= v25.5.0 and fails identically on a clean tree here.

*AI disclosure, per the AI usage guidelines in CONTRIBUTING: I used an AI assistant (Claude Code) on this, and the commit carries an `Assisted-by` trailer. I built goreleaser from main and produced every before and after above by running the two binaries against real git repositories, including the reordering case. I understand the change and take ownership of it.*

### DIFF
--- a/internal/pipe/changelog/changelog.go
+++ b/internal/pipe/changelog/changelog.go
@@ -310,15 +310,16 @@ func formatEntries(ctx *context.Context, entries []Item) ([]string, error) {
 func filterEntries(ctx *context.Context, entries []Item) ([]Item, error) {
 	filters := ctx.Config.Changelog.Filters
 	if len(filters.Include) > 0 {
-		var newEntries []Item
+		res := make([]*regexp.Regexp, 0, len(filters.Include))
 		for _, filter := range filters.Include {
 			r, err := regexp.Compile(filter)
 			if err != nil {
 				return entries, err
 			}
-			newEntries = append(newEntries, keep(r, entries)...)
+			res = append(res, r)
 		}
-		return newEntries, nil
+		// an entry matching several includes must still be listed once
+		return keepAny(res, entries), nil
 	}
 	for _, filter := range filters.Exclude {
 		r, err := regexp.Compile(filter)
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):
Nice catch, and thanks for the clear write-up. Ordering side effect is fine: nothing documented the old per-filter grouping, and this makes include consistent with exclude. `sort` and `groups` remain the ordering knobs.

PR #6774, opened 2026-08-16
fix(winget): count arm64 in the duplicate-archive check

VXNCXNX:
The winget duplicate-archive guard only counts `386` and `amd64`, so two `windows/arm64` archives slip through and produce a manifest with two `arm64` installers.

```go
switch archive.Goarch {
case "386":
    i386count++
case "amd64":
    amd64Count++
}
...
if i386count > 1 || amd64Count > 1 {
    return Installer{}, errMultipleArchives
}
```

`arm64` has been in the artifact filter since `0f5b0583` ("feat(winget): support arm64"), which added `artifact.ByGoarch("arm64")` at line 206 but did not add it to these counters. So the arch reaches `installer.Installers` and is never counted.

Calling `makeInstaller` with two `windows/arm64` zips:

```
before:  err = <nil>
         Installers:
         - Architecture: arm64
           ...
         - Architecture: arm64
           ...

after:   err = found multiple archives for the same platform, please consider filtering by id
         Installers: []
```

A winget manifest may only carry one installer per architecture, so the `before` output is invalid and fails validation downstream rather than at build time.

## The fix

Replace the two fixed counters with a `map[string]int` keyed by `archive.Goarch`, so every architecture that reaches the installer list is counted. Adding a new arch to the filter can no longer silently skip the check.

## Verification

`TestMakeInstallerMultipleArchivesSameArch` with a subtest per arch (`amd64`, `386`, `arm64`).

Reverting only the guard to the old two-arch form fails it, and fails it *only* on arm64:

```
--- FAIL: TestMakeInstallerMultipleArchivesSameArch
    --- PASS: TestMakeInstallerMultipleArchivesSameArch/amd64
    --- PASS: TestMakeInstallerMultipleArchivesSameArch/386
    --- FAIL: TestMakeInstallerMultipleArchivesSameArch/arm64
        Error: Expected error with "found multiple archives for the same platform, please consider filtering by id" in chain but got nil.
```

The two passing subtests are the point: they pin the existing behaviour while arm64 isolates the gap.

`go build ./...`, `go vet`, `gofmt -l` are clean, and `internal/pipe/winget`, `internal/pipe/archive` and `internal/artifact` all pass.

One note: CONTRIBUTING asks for `task ci`, which I did not run in full because it pulls in Docker, snapcraft and cosign suites. The winget package is pure functions over config structs, so the targeted runs above are the meaningful signal for this change.

*Disclosure: written with AI assistance (Claude Code). I confirmed arm64 is in the artifact filter, produced the before/after by calling makeInstaller directly, and ran the mutation check myself.*

### DIFF
--- a/internal/pipe/winget/winget.go
+++ b/internal/pipe/winget/winget.go
@@ -604,9 +604,8 @@ func makeInstaller(ctx *context.Context, winget config.Winget, archives []*artif
 			installer.Commands = []string{cmd}
 		}
 		installer.Installers = append(installer.Installers, item)
-		// count every supported architecture: a manifest may only have one
-		// installer per architecture.
-		archCounts[archive.Goarch]++
+		// a manifest may only have one installer per architecture.
+		archCounts[item.Architecture]++
 	}
 
 	if binaryCount > 0 && zipCount > 0 {
--- a/internal/pipe/winget/winget.go
+++ b/internal/pipe/winget/winget.go
@@ -571,7 +571,8 @@ func makeInstaller(ctx *context.Context, winget config.Winget, archives []*artif
 		},
 	}
 
-	var amd64Count, i386count, zipCount, binaryCount int
+	var zipCount, binaryCount int
+	archCounts := map[string]int{}
 	for _, archive := range archives {
 		sha256, err := archive.Checksum("sha256")
 		if err != nil {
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):
Verified the bug is reachable: `arm64` entered the artifact filter in 0f5b0583 but was never added to the duplicate-archive counter, so two windows/arm64 archives produced a manifest with two arm64 installers instead of the error. Checked out the pre-fix code and confirmed the new test fails only on the arm64 subtest.

Pushed a small follow-up keying the count on `item.Architecture` (the manifest field the invariant is about) rather than `archive.Goarch`, plus dropping a few test inputs `makeInstaller` never reads.

Thanks @VXNCXNX!

PR #6775, opened 2026-08-16
fix(krew): apply the documented goarm default

VXNCXNX:
The krew docs give `goarm` a `Default: 6`, but the pipe never applies it.

With `goarm` unset and a build producing only 32-bit arm archives, the whole
release fails:

```
before:
  • krew plugin manifest
  ⨯ release failed after 1s      error=no archives found
$ ls dist/krew
ls: cannot access 'dist/krew': No such file or directory

after:
  • krew plugin manifest
    • writing                    manifest=dist/krew/krewdemo.yaml
      uri: .../krewdemo_0.0.0-SNAPSHOT-none_linux_armv6.tar.gz
      selector: matchLabels: {os: linux, arch: arm}
```

In a mixed build the failure is quieter: `artifact.ByGoarm("")` matches nothing,
so the armv6 archive is simply missing from the manifest.

## The fix

Three lines in `Default()`, the same ones brew already has at `brew.go:72`:

```go
if krew.Goarm == "" {
    krew.Goarm = experimental.DefaultGOARM()
}
```

Going through `experimental.DefaultGOARM()` rather than a literal `"6"` keeps it
consistent with brew, including under the `defaultgoarm` experiment.

## Verification

`TestDefault` gains the goarm assertion, and `TestRunPipeDefaultGoarm` builds an
arm-only project end to end.

Removing the three lines fails both, one on the default and one on the symptom:

```
--- FAIL: TestDefault
    Error: Not equal: expected: "6"  actual  : ""
--- FAIL: TestRunPipeDefaultGoarm
    Error: Received unexpected error: no archives found
```

No golden file changed, because every existing krew test either sets `goarm`
explicitly or uses non-arm artifacts.

`go test ./internal/... ./pkg/...` is ok everywhere except
`internal/builders/node`, whose `TestBuild` needs Node >= v25.5.0 and this host
has v24.19.0. I confirmed by stashing that it fails identically on unmodified
`main`. `go build`, `go vet`, `gofmt`, `gofumpt` clean, and
`golangci-lint run internal/pipe/krew/...` reports 0 issues.

*Disclosure: written with AI assistance (Claude Code). I produced the before and after by running binaries built from each tree against a real arm-only project, and ran the mutation check myself.*

### DIFF
--- a/internal/pipe/krew/krew.go
+++ b/internal/pipe/krew/krew.go
@@ -18,6 +18,7 @@ import (
 	"github.com/goreleaser/goreleaser/v2/internal/artifact"
 	"github.com/goreleaser/goreleaser/v2/internal/client"
 	"github.com/goreleaser/goreleaser/v2/internal/commitauthor"
+	"github.com/goreleaser/goreleaser/v2/internal/experimental"
 	"github.com/goreleaser/goreleaser/v2/internal/pipe"
 	"github.com/goreleaser/goreleaser/v2/internal/tmpl"
 	"github.com/goreleaser/goreleaser/v2/internal/yaml"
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):

caarlos0:
thank you!

PR #6777, opened 2026-08-16
fix(build): node windows targets get no .exe extension

VXNCXNX:
Windows binaries from the `node` builder never get the `.exe` extension.

`/tmp/nodedemo`, targets `[win-x64, linux-x64]`, with a pre-hook echoing
`{{ .Name }}` and `{{ .Ext }}`:

```
before:
  │ ARTIFACT NAME=demo EXT=[]
  │ ARTIFACT NAME=demo EXT=[]

after:
  │ ARTIFACT NAME=demo EXT=[]
  │ ARTIFACT NAME=demo.exe EXT=[.exe]
```

So `dist/demo_win-x64/demo` is written instead of `demo.exe`, and any hook or
template reading `.Name` or `.Ext` sees the wrong value.

## Cause

`extFor` detects Windows by looking for the literal string `windows`:

```go
if strings.Contains(target, "windows") {
    return ".exe"
}
```

The node builder does not use that spelling. It names targets after
`nodejs.org/dist`, where Windows is `win-x64` and `win-arm64`
(`internal/builders/node/targets.go`). The same builder already recognises them
that way at `download.go:138`:

```go
isWin := strings.HasPrefix(target, "win-")
```

The `c-shared` and `c-archive` branches of `extFor` have the same miss, so those
targets lose `.dll` and `.lib` too.

## The fix

One predicate, `isWindowsTarget`, accepting both spellings, used at the three
call sites. It reuses the node builder's own `win-` prefix rule rather than
inventing a second convention.

I deliberately left the ext selection in `extFor` rather than pushing it into
each builder. There is an existing TODO suggesting that refactor, but it is a
much larger change than this bug needs.

## Verification

`TestExtWindowsNodeTargets` covers the three output modes for `win-x64` and
`win-arm64`, and `TestBuildOptionsForNodeWindowsTarget` checks the value that
actually reaches the artifact.

Replacing the new prefix check with `false` fails both:

```
--- FAIL: TestExtWindowsNodeTargets
    Error: Not equal: expected: ".exe"  actual  : ""
--- FAIL: TestBuildOptionsForNodeWindowsTarget
    Error: Not equal: expected: ".exe"  actual  : ""
```

The existing `windows` cases pass under that mutation, which is what pins the
behaviour the change must not alter.

`internal/pipe/build`, `internal/pipe/env`, `internal/pipe/before`,
`internal/pipe/snapshot` and `internal/pipe/partial` all pass. `go build`,
`go vet`, `gofmt -l` clean, `golangci-lint run ./internal/pipe/build/...` reports
0 issues.

`internal/builders/node` `TestBuild` fails here, but it needs Node >= v25.5.0 and
this host has v24.19.0. I confirmed by stashing that it fails identically on
unmodified `main`. That does mean the node builder's own suite could not
exercise this end to end; the before and after above come from running the real
binaries instead.

*Disclosure: written with AI assistance (Claude Code). I produced the before and after by running binaries built from each tree against a real node project, and ran the mutation check myself.*

### DIFF
--- a/internal/pipe/build/build.go
+++ b/internal/pipe/build/build.go
@@ -262,7 +262,7 @@ func extFor(target string, build config.BuildDetails) string {
 		if strings.Contains(target, "darwin") {
 			return ".dylib"
 		}
-		if strings.Contains(target, "windows") {
+		if isWindowsTarget(target) {
 			return ".dll"
 		}
 		if strings.Contains(target, "wasm") {
@@ -270,7 +270,7 @@ func extFor(target string, build config.BuildDetails) string {
 		}
 		return ".so"
 	case "c-archive":
-		if strings.Contains(target, "windows") {
+		if isWindowsTarget(target) {
 			return ".lib"
 		}
 		return ".a"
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):

caarlos0:
Good catch, thanks!

PR #6782, opened 2026-08-17
fix(nfpm): overrides ignore package_name, epoch, release and prerelease

VXNCXNX:
## What's broken

`nfpms.overrides.<format>` silently ignores `package_name`, `epoch`, `release`, `prerelease` and `version_metadata`. `create()` builds a merged `overridden` but then reads those five off the un-merged `fpm`.

All five are fields of `NFPMOverridables`, and `www/static/schema.json` accepts them under `overrides`, so they are written without any warning and dropped.

## Repro

A config with top level `epoch: 1`, `release: 1`, `prerelease: alpha1`, `version_metadata: top`, plus an `overrides.rpm` block changing all five:

```
before:
  creating  package=nfpmdemo  format=rpm  file=dist/nfpmdemo_..._amd64.rpm
  rpm lead: nfpmdemo-0.0.0~alpha1+top-1

after:
  creating  package=nfpmdemo-rpm  format=rpm  file=dist/nfpmdemo-rpm_..._amd64.rpm
  rpm lead: nfpmdemo-rpm-0.0.0~beta1+git-9
```

The deb package, which has no override block, is unaffected either way.

## The fix

Read those five from `overridden` instead of `fpm`, in the three places they are used.

## Verification

`TestOverridesVersionFields` configures all five at the top level and overrides them for rpm only, then asserts on both output names. Reverting the three spots fails it, and the deb entry stays correct under the mutation, which isolates the override path.

The test needs no external binary, since nfpm is a library, so it runs rather than skipping.

`go test ./internal/pipe/nfpm/` passes. `go test ./internal/...` has one failure, `builders/node TestBuild`, which needs Node >= v25.5.0 and fails identically on a clean `main` here. `golangci-lint run ./internal/pipe/nfpm/...` reports nothing, `gofmt` and `go vet` clean.

I did not run `task ci` in full, since it pulls Docker, snapcraft and cosign.

### DIFF
--- a/internal/pipe/nfpm/nfpm.go
+++ b/internal/pipe/nfpm/nfpm.go
@@ -308,7 +308,7 @@ func create(ctx *context.Context, fpm config.NFPM, format string, artifacts []*a
 		return err
 	}
 
-	packageName, err := tmpl.New(ctx).Apply(fpm.PackageName)
+	packageName, err := tmpl.New(ctx).Apply(overridden.PackageName)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):
thanks!

PR #6788, opened 2026-08-17
fix(nix): a skipped nix entry stops the ones after it

VXNCXNX:
`runAll` returns on the first error, so a nix entry that gets skipped (no `repository.name`) aborts every nix entry after it in the same release. With two entries where only the first lacks a repository name, the second never gets written:

```
before   • pipe skipped or partially skipped   reason=repository name is not set
         dist/nix: (no files)

after    • writing   nixpkg=dist/nix/pkgs/second/default.nix
         • pipe skipped or partially skipped   reason=repository name is not set
```

It now collects skips in a `pipe.SkipMemento` and evaluates them at the end, which is what cask (#6785), brew (#6783), krew (#6787), flatpak (#6781) and snapcraft (#6784) already do. `publishAll` in this same file was already written that way, only `runAll` was missed.

Test added in `nix_test.go`; reverting the change fails it with "second nix package should have run". It runs without any external binary, using the package's existing `alwaysZeroHasher`. `go test ./internal/pipe/nix/...` passes, gofmt and vet clean.

AI disclosure: written with Claude Code. I ran the release binary before and after and checked the mutation myself.

### DIFF
--- a/internal/pipe/nix/nix.go
+++ b/internal/pipe/nix/nix.go
@@ -114,13 +114,20 @@ func (p Pipe) Publish(ctx *context.Context) error {
 }
 
 func (p Pipe) runAll(ctx *context.Context, cli client.ReleaseURLTemplater) error {
+	// even if one of them is skipped, we still go through all of them, and
+	// return the skips all at once in the end.
+	skips := pipe.SkipMemento{}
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):

caarlos0:
thank you!

PR #6790, opened 2026-08-17
fix(sign): artifacts: none masks real signing failures

VXNCXNX:
In `signs`, `binary_signs` and `docker_signs`, an entry with `artifacts: none` returns a skip into a plain errgroup, so that skip becomes the group's result and hides real signing failures from the other entries in the same list. The release then exits 0 with unsigned artifacts.

With `signs: [{artifacts: none}, {artifacts: checksum, cmd: sh, args: ["-c","exit 7"]}]`:

```
before   • signing artifacts
           • pipe skipped or partially skipped   reason=artifact signing is disabled
         • release succeeded after 0s                                        exit=0

after    • signing artifacts
         x release failed   error=exit status 7  message=could not sign artifact  cmd=sh   exit=1
```

The three groups are now wrapped in `semerrgroup.NewSkipAware`, which sbom, blob, ko, notary and docker already use. In `signs`, `ctx.Artifacts.Refresh()` was also being skipped in that case, so signatures produced by the other entries were not registered; it now runs, and the skip is still returned afterwards.

One test per pipe, using `cmd: false` so no cosign, syft or gon is needed. Reverting the three wrappers fails all three with `"artifact signing is disabled" does not contain "exit status 1"`. sign, sbom and notary suites pass, gofmt and vet clean.

AI disclosure: written with Claude Code. I ran the release binary before and after and checked the mutations myself.

### DIFF
--- a/internal/pipe/sign/sign.go
+++ b/internal/pipe/sign/sign.go
@@ -84,7 +84,9 @@ func (Pipe) Default(ctx *context.Context) error {
 
 // Run executes the Pipe.
 func (Pipe) Run(ctx *context.Context) error {
-	g := semerrgroup.New(ctx.Parallelism)
+	// skip-aware so that a config with `artifacts: none` doesn't mask the
+	// errors of the other configs, and doesn't prevent the refresh below.
+	g := semerrgroup.NewSkipAware(semerrgroup.New(ctx.Parallelism))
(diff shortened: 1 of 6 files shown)


caarlos0 (approved):

caarlos0:
thanks

PR #6791, opened 2026-08-17
fix(aur): expand description templates before escaping quotes

VXNCXNX:
Both `aur.md` and `aursources.md` document `description` as `Templates: allowed`, but neither pipe expands it before the PKGBUILD is built, so `quoteField` escapes the raw `{{ ... }}` string instead of the final value. When the template result contains an apostrophe, the generated PKGBUILD is not valid shell.

With `description: "{{ .Env.DESC }}"` and `DESC=Let's go`:

```
before   pkgdesc='Let's go'
after    pkgdesc="Let's go"

$ bash -n PKGBUILD
before   exit=2   unexpected EOF while looking for matching `''
after    exit=0
```

The fix adds `&aur.Description` to the existing `tmpl.ApplyAll` in `doRun`, in both pipes. It has to happen there rather than later, because the whole file is templated again afterwards and by then the quoting decision has already been made.

The hardcoded version of this case is already covered by `simple-quote-inside-description`; only the templated one was missing. One test per pipe, both fail on revert with `does not contain "pkgdesc=\"Let's go\""`. aur and aursources suites pass, gofmt and vet clean. No docs or schema change, the docs were already correct.

AI disclosure: written with Claude Code. I ran the release binary before and after, checked the mutation, and validated the output with `bash -n` myself.

### DIFF
--- a/internal/pipe/aur/aur.go
+++ b/internal/pipe/aur/aur.go
@@ -104,9 +104,7 @@ func doRun(ctx *context.Context, aur config.AUR, cl client.ReleaseURLTemplater) 
 		&aur.Name,
 		&aur.Directory,
 		&aur.SkipUpload,
-		// needs to be evaluated here so `quoteField` sees the final value:
-		// the whole file is templated again later, and by then it is too late
-		// to escape quotes coming from the template result.
+		// must be expanded here so `quoteField` sees the final value.
 		&aur.Description,
 	); err != nil {
 		return err
@@ -290,7 +288,7 @@ func applyTemplate(ctx *context.Context, tpl string, data templateData) (string,
 func toPkgBuildArray(ss []string) string {
 	result := make([]string, 0, len(ss))
 	for _, s := range ss {
-		result = append(result, fmt.Sprintf("'%s'", s))
+		result = append(result, quoteField(s))
 	}
 	return strings.Join(result, " ")
 }
--- a/internal/pipe/aur/aur.go
+++ b/internal/pipe/aur/aur.go
@@ -104,6 +104,10 @@ func doRun(ctx *context.Context, aur config.AUR, cl client.ReleaseURLTemplater) 
 		&aur.Name,
 		&aur.Directory,
 		&aur.SkipUpload,
+		// needs to be evaluated here so `quoteField` sees the final value:
+		// the whole file is templated again later, and by then it is too late
+		// to escape quotes coming from the template result.
+		&aur.Description,
 	); err != nil {
 		return err
 	}
--- a/internal/pipe/aur/aur_test.go
+++ b/internal/pipe/aur/aur_test.go
@@ -927,6 +927,9 @@ func TestRunPipeTemplatedDescriptionWithQuotes(t *testing.T) {
 			ProjectName: "foo",
 			AURs: []config.AUR{{
 				Description: "{{ .Env.DESC }}",
+				Homepage:    "https://example.com/~o'brien",
+				License:     "Nobody's",
+				Provides:    []string{"fo'o", "bar"},
 			}},
 			Env: []string{`DESC=Let's go`},
 		},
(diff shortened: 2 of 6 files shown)


caarlos0:
Reviewed this locally (checked out the branch, ran the suites, reverted the source hunks to confirm the tests actually catch the bug). The report and the fix both hold up.

**Bug confirmed.** `dataFor` passes `cfg.Description` into `templateData.Desc` verbatim, so `quoteField` decides the quoting against the raw `{{ ... }}` string, which contains no apostrophe, and picks `'...'`. The file-level `tmpl.Apply` in `applyTemplate` then expands the template inside the already-chosen single quotes. Reverting `aur.go`/`aursources.go` and keeping the tests reproduces it exactly:

```
pkgdesc='Let's go'    # does not contain "pkgdesc=\"Let's go\""
```

Both new tests fail on revert and pass with the fix. `aur`, `aursources` and `nix` suites are green, gofmt and vet clean.

**Fix is the right shape.** Expanding in `doRun`'s existing `ApplyAll` is the only place it can go — after that point the quoting decision is already baked in. It also matches what the sibling pipes already do: `nix.go:152` has `&nix.Description` and `winget.go:132` has `&winget.Description` in exactly the same position. `aur`/`aursources` were the outliers.

**One behavior change worth naming.** The description now goes through two template passes: the new one in `doRun`, then the whole-file pass in `applyTemplate`. So a description whose *result* contains `{{ ... }}` used to pass through literally and now errors. With `DESC=hi {{ brace }}` and `description: "{{ .Env.DESC }}"`:

```
main   pkgdesc='hi {{ brace }}'
PR     template: failed to apply ...: function "brace" not defined
```

I don't consider this a blocker: `aur.Name` has always had this property, nix and winget already behave this way, and a hardcoded `description` containing braces already failed on `main` (the PR just fails earlier with a tighter error message). Just flagging it so it isn't a surprise later.

**Out of scope, for a follow-up:** the same template single-quotes `url='{{ .Homepage }}'` and `license=('{{ .License }}')` without escaping, and `toPkgBuildArray` does `'%s'` for provides/conflicts/depends. Same class of problem, much less likely to bite. Not something this PR needs to take on.

Nits, take or leave:

- The three-line comment is duplicated verbatim in both files and is longer than it needs to be. One line would carry it: `// must be expanded here so quoteField sees the final value.`
- The two tests are near-identical copies of existing setup boilerplate in each package. Folding them into the surrounding table tests would be tidier, but the standalone form is easier to read, so no strong opinion.

Nice catch, and thanks for the disclosure and for actually validating the output with `bash -n` instead of just trusting the diff. LGTM.

caarlos0:
Pushed a follow-up commit to this branch (4a82d69) with the review items:

- Collapsed the duplicated three-line comment to one line in both pipes.
- Took the out-of-scope item myself since it is the same bug: `url=` and `license=` were hardcoded to single quotes in both templates, and `toPkgBuildArray` did `'%s'` for provides/conflicts/depends/optdepends/backup/makedepends. All of them now go through `quoteField`.

Baseline on `main` with an apostrophe in those fields:

```
url='https://example.com/~o'brien'
license=('Nobody's')
provides=('fo'o' 'bar')

$ bash -n PKGBUILD
line 9: syntax error near unexpected token `)'
```

After:

```
url="https://example.com/~o'brien"
license=("Nobody's")
provides=("fo'o" 'bar')

$ bash -n PKGBUILD
exit 0
```

Output is byte-identical for values without quotes, so no golden file churn: every existing `aur` and `aursources` test passes untouched. I extended your two tests to assert the new cases rather than adding more boilerplate.

One thing to be aware of, not introduced here: `quoteField`'s both-quotes branch is lossy, it rewrites `"` to `'`. So `both'and"quotes` comes out as `"both'and'quotes"`. Valid shell, but not the input. That was already true for `pkgdesc` and now applies to the array fields too. Worth a proper fix (`'\''` escaping) at some point, but not in this PR.

gofmt, vet and both suites clean on my side. Thanks again for the catch.

caarlos0 (approved):

caarlos0:
thank you!

PR #6792, opened 2026-08-17
fix(chocolatey): error on multiple archives for the same platform

VXNCXNX:
chocolatey has no duplicate-archive guard, so two Windows archives with the same GOARCH both get written into `chocolateyinstall.ps1`, producing a hashtable with duplicate `url64bit` and `checksum64` keys. PowerShell fails on that at runtime, and goreleaser reports success.

Calling `doRun` with a client and a server zip, both windows/amd64:

```
before   err: <nil>
         $packageArgs = @{
             url64bit   = '.../client_1.0.0_windows_amd64.zip'
             checksum64 = '5e2bf5...'
             url64bit   = '.../server_1.0.0_windows_amd64.zip'   <- duplicate key
             checksum64 = '5e2bf5...'
         }

after    err: found multiple archives for the same platform, please consider filtering by id
         no chocolateyinstall.ps1 written
```

winget and scoop both guard this already, chocolatey was the one that did not. This is also the exact case scoop's error page tells users to fix with `ids`, so I added two lines about it under `ids` in the docs.

I used a plain error rather than `pipe.Skip`, matching this pipe's existing `errNoWindowsArchive`, though winget uses a skip for its equivalent. Say the word if you would rather have it skip.

Test added covering 386 and amd64; removing the guard fails both. chocolatey and scoop suites pass, gofmt and vet clean. No `choco` binary needed, the pipe's `cmder` is faked.

AI disclosure: written with Claude Code. I ran the before/after and checked the mutation myself.

### DIFF
--- a/internal/pipe/chocolatey/chocolatey.go
+++ b/internal/pipe/chocolatey/chocolatey.go
@@ -18,7 +18,12 @@ import (
 	"github.com/goreleaser/goreleaser/v2/pkg/context"
 )
 
-var errNoWindowsArchive = errors.New("chocolatey requires at least one windows archive")
+var (
+	errNoWindowsArchive = errors.New("chocolatey requires at least one windows archive")
+	// a chocolateyinstall.ps1 may only have one url/checksum pair per
+	// architecture.
+	errMultipleArchives = errors.New("found multiple archives for the same platform, please consider filtering by id")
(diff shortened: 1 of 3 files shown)


caarlos0 (approved):
thanks!

PR #6793, opened 2026-08-17
fix(healthcheck): register upx and makeself dependency checkers

VXNCXNX:
`goreleaser healthcheck` reports a clean bill of health for configs using `upx` or `makeselfs`, even though both docs pages say the binary has to be in `$PATH`. Neither pipe is registered in `healthcheck.DependencyCheckers`, so the command that exists to catch a missing tool stays silent about these two.

Same config, neither binary installed:

```
before   • ✓ git
         • ✓ go
         • done!                                      exit 0

after    • ✓ git
         • ✓ go
         • ⚠ makeself - not present in path
         • ⚠ upx - not present in path
         x command failed   error=one or more checks failed    exit 1
```

`upx.Pipe` already implemented `Dependencies()`, it was just never registered, so that was dead code. makeself had none and now has one. A config with neither section is unaffected, which I checked separately.

The new test in `pkg/healthcheck` asserts the behaviour rather than a hardcoded list: it walks the registered checkers, skips the ones whose `Skip` returns true for the config, and requires the tools to show up. srpm, makeself, upx and healthcheck suites pass, gofmt and vet clean.

This touches `pkg/healthcheck` as well as the two pipes, since that is where the registry lives.

AI disclosure: written with Claude Code. I ran healthcheck before and after and checked the mutation myself.

### DIFF
--- a/internal/pipe/makeself/makeself.go
+++ b/internal/pipe/makeself/makeself.go
@@ -38,6 +38,9 @@ func (Pipe) Skip(ctx *context.Context) bool {
 	return skips.Any(ctx, skips.Makeself) || len(ctx.Config.Makeselfs) == 0
 }
 
+// Dependencies returns the binaries this pipe needs.
+func (Pipe) Dependencies(*context.Context) []string { return []string{"makeself"} }
+
 // Default sets the pipe defaults.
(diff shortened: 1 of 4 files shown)


caarlos0 (approved):
thanks!

PR #6795, opened 2026-08-18
fix(mcp): mcp.disable is documented but never read

VXNCXNX:
`mcp.disable` is documented, present in `config.MCPDetails`, and accepted by `schema.json`, but `internal/pipe/mcp/mcp.go` never reads it. Setting it has no effect and the publish runs anyway.

From `www/content/customization/publish/mcp.md`:

> Setting this will prevent GoReleaser from actually trying to publish the server manifest - instead, the manifest file will be stored in the dist directory only [...] If set to auto, the manifest will not be published in case there is an indicator for prerelease in the tag

The field is currently only ever written to, in the `mcp.github` deprecation migration.

Before, with `mcp.disable: true`:

```
• mcp registry
  • mcp is experimental and subject to change
⨯ release failed after 0s
  │ * mcp registry: could not get token: failed to get anonymous token (status 404)
```

After:

```
• mcp registry
  • pipe skipped or partially skipped   reason=mcp.disable is set
• release succeeded after 0s
```

Both documented forms are handled, `true` and `auto` on a prerelease, using the same two-line shape as `brew.skip_upload`, `docker_manifest.skip_push` and the other seven pipes that implement `auto`.

One deliberate change beyond the skip: `warnExperimental()` now runs after the disable check, so a disabled MCP config no longer prints the experimental warning. Happy to move it back if you would rather it always warn.

AI-assisted: written with Claude Code, reviewed and tested by me before sending.

### DIFF
--- a/internal/pipe/mcp/mcp.go
+++ b/internal/pipe/mcp/mcp.go
@@ -9,10 +9,12 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"strings"
 
 	"github.com/caarlos0/log"
 	"github.com/goreleaser/goreleaser/v2/internal/deprecate"
 	"github.com/goreleaser/goreleaser/v2/internal/logext"
+	"github.com/goreleaser/goreleaser/v2/internal/pipe"
 	"github.com/goreleaser/goreleaser/v2/internal/retryx"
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):
thx

PR #6797, opened 2026-08-18
fix(winget): a skipped winget entry stops the ones after it

VXNCXNX:
`winget.runAll` returns on the first error, including `pipe.Skip`, so a winget entry that skips takes every later entry with it and their manifests are never generated. An entry skips when it is missing `license`, `publisher`, `short_description` or `repository.name`, or when `package_identifier` is invalid.

With two entries where only the first is missing a license:

```
before:
  • winget
    • pipe skipped or partially skipped    reason=winget.license is required
  (no dist/winget/*.yaml produced at all)

after:
  • winget
    • writing   path=dist/winget/manifests/f/Foo/valid/1.0.0/Foo.valid.yaml
    • writing   path=.../Foo.valid.installer.yaml
    • writing   path=.../Foo.valid.locale.en-US.yaml
    • pipe skipped or partially skipped    reason=winget.license is required
```

The fix is `pipe.SkipMemento`, the same change already merged for brew (#6783), snapcraft (#6784), cask (#6785), upload (#6786), krew (#6787) and nix (#6788). `publishAll` in this same file already does it this way, `runAll` was the last place in winget that did not.

AI-assisted: written with Claude Code, reviewed and tested by me before sending.

### DIFF
--- a/internal/pipe/winget/winget.go
+++ b/internal/pipe/winget/winget.go
@@ -103,13 +103,20 @@ func (p Pipe) Publish(ctx *context.Context) error {
 }
 
 func (p Pipe) runAll(ctx *context.Context, cli client.ReleaseURLTemplater) error {
+	// even if one of them is skipped, we still go through all of them, and
+	// return the skips all at once in the end.
+	skips := pipe.SkipMemento{}
 	for _, winget := range ctx.Config.Winget {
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):

caarlos0:
thanks!

PR #6800, opened 2026-08-20
fix(dockers/v2): annotation scopes

### LINKED ISSUE #6799: `dockers_v2`: `annotations` scope prefix (`manifest:`/`index:`) is accepted but ignored — everything lands on the image index
## Summary

`dockers_v2.annotations` keys support Docker Buildx's `[type:]key=value` scope-prefix
syntax (e.g. `manifest:org.opencontainers.image.revision=...`,
`index,manifest:...`) — the prefix is parsed and stripped from the resulting
annotation key correctly — but the requested scope is not honoured. Every
annotation ends up on the top-level OCI image **index** only, never on the
per-platform image **manifests**, regardless of what scope prefix is used.

This matters because any consumer that resolves a multi-platform tag down to a
specific platform's manifest (e.g. `docker pull`, `regctl image manifest`,
buildx's own `imagetools inspect <ref>@<platform-digest>`, or internal tooling
that walks straight to the platform manifest for a given architecture) will
never see these annotations. For us specifically, this silently broke
attribution of container builds to DORA/lead-time metrics, since our
attribution proxy resolves the `linux/amd64` manifest directly and never reads
the index's annotations.

## Versions tested

Confirmed on two GoReleaser releases, three minor versions apart — behaviour
is identical on both, so this isn't a recent regression that's already fixed
on `main`:

- GoReleaser `2.14.3`
- GoReleaser `2.17.1`

Environment:

```
Docker:  29.7.2 (client + server)
Buildx:  v0.36.0-desktop.1
OS:      macOS (Darwin 25.6.0, arm64)
```

## Minimal repro config

```yaml
version: 2

builds:
  - id: api
    builder: go
    goos: [linux]
    goarch: [amd64, arm64]

dockers_v2:
  - images:
      - 'ghcr.io/example/example'
    tags:
      - 'latest'
    dockerfile: Dockerfile
    platforms:
      - linux/amd64
      - linux/arm64
    annotations:
      'index,manifest:org.opencontainers.image.revision': '{{.FullCommit}}'
      'index,manifest:org.opencontainers.image.description': 'Example service'
```

## Steps to reproduce

```bash
goreleaser release --clean --auto-snapshot --skip=publish
# "snapshot build: will not push any images" — GoReleaser tags/loads each
# platform image locally under its fully-qualified name instead of pushing.

docker save ghcr.io/example/example:latest-amd64 -o out.tar
mkdir extract && tar -xf out.tar -C extract
INNER=$(jq -r '.manifests[0].digest' extract/index.json | cut -d: -f2)
jq . "extract/blobs/sha256/$INNER"
```

## Expected

With `manifest:` (or `index,manifest:`) requested, the per-platform manifest
entry (the one with `"platform": {"architecture": "amd64", "os": "linux"}`)
should carry its own `.annotations` block containing the requested keys.

## Actual

The per-platform manifest entry has **no** `.annotations` field at all. Every
requested key — regardless of whether it was prefixed `manifest:`,
`index:`, or `index,manifest:` — appears only once, on the outer index's
`.annotations`:

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:255bdd7a80686d364508388d9c3c5893a8dcbe48f7af986cf8e45fb1137caa95",
      "size": 4009,
      "platform": { "architecture": "amd64", "os": "linux" }
      // <-- no "annotations" key here, despite "manifest:" scope requested
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:4ac20b55ac87d70301f143f69153f1f20dae85b36b2dff128b80bc67959dd526",
      "size": 837,
      "annotations": {
        "vnd.docker.reference.digest": "sha256:255bdd7a...",
        "vnd.docker.reference.type": "attestation-manifest"
      },
      "platform": { "architecture": "unknown", "os": "unknown" }
    }
  ],
  "annotations": {
    "org.opencontainers.image.revision": "5610b7ec6715a0466ce28be9677f2c1030acb43c",
    "org.opencontainers.image.description": "Example service"
  }
}
```

The requested keys arrive clean (no leftover `index,manifest:` prefix text),
confirming the prefix *is* parsed — it's just not routed anywhere except the
index.

Tested with `manifest:` alone, `index:` alone, and `index,manifest:` combined
— all three produce the identical index-only result. There appears to be no
way, via `dockers_v2.annotations`, to get an annotation onto a per-platform
manifest.

## Why this is surprising

- Docker's own `buildx build --annotation` docs state the *unprefixed*
  default is manifest-level (["By default, the image manifest is
  annotated."](https://docs.docker.com/reference/cli/docker/buildx/build/#annotation)) —
  so even the default here is inverted relative to plain buildx.
- `dockers_v2`'s own docs describe it as running a single
  `docker buildx build --push` per release, which per Docker's docs should
  support the full `[type:]key=value` annotation syntax natively — suggesting
  GoReleaser's `annotations` field isn't passing the prefix through to the
  underlying `--annotation` flag as given.

## Impact / workaround attempted

We also tried working around this at the registry level via
`docker buildx imagetools create --annotation manifest:...` after the fact,
but that command [explicitly only supports `index:` and
`manifest-descriptor:` scopes](https://docs.docker.com/reference/cli/docker/buildx/imagetools/create/)
— it cannot write manifest-level annotations either. So there is currently no
way to get GoReleaser-built multi-platform images annotated at the
per-platform-manifest level, short of bypassing `dockers_v2` for a raw
`docker buildx build --push --annotation manifest:...` step.

CraigAstillRVU:
If applied, this commit will...

...honor buildx annotation scope prefixes in `dockers_v2.annotations`, so
`manifest:`, `index,manifest:`, the `-descriptor` variants and
platform-qualified scopes reach the platform manifests instead of being mangled
into the image index.

On multi-platform builds every annotation was rewritten to
`index:<whatever the user wrote>`, with only a literal `index:` stripped first.
buildx cuts an annotation key at its **first** colon, so any other prefix wasn't
ignored — it was absorbed into the key:

| Config key | Flag emitted (before) | Result (before) |
| --- | --- | --- |
| `foo` | `index:foo=v` | index annotation `foo` ✅ |
| `index:foo` | `index:foo=v` | index annotation `foo` ✅ |
| `manifest:foo` | `index:manifest:foo=v` | index key **`manifest:foo`** ❌ |
| `index,manifest:foo` | `index:index,manifest:foo=v` | index key **`index,manifest:foo`** ❌ |
| `manifest[linux/amd64]:foo` | `index:manifest[linux/amd64]:foo=v` | index key **`manifest[linux/amd64]:foo`** ❌ |

The fix defaults to `index:` only when the key carries no scope of its own:

```go
if len(d.Platforms) > 1 {
	for i := 1; i < len(annotationFlags); i += 2 {
		if !hasAnnotationScope(annotationFlags[i]) {
			annotationFlags[i] = "index:" + annotationFlags[i]
		}
	}
}
```

Deliberately **not** changed: the `index:` default itself. It comes from #6053
so that multi-arch annotations are visible to
`docker buildx imagetools inspect`, and changing it would alter manifest content
for every existing user. Unscoped and `index:`-scoped keys emit byte-identical
flags to before, which the existing `TestMakeArgs` assertions pin.

A scope is only recognized when every comma-separated entry is one of buildx's
four types (`index`, `manifest`, `index-descriptor`, `manifest-descriptor`),
optionally platform-qualified. Anything else (e.g. `bogus:foo`) is treated as
part of the key and still gets the `index:` default, rather than reaching buildx
as an `unknown annotation type`.

**Tests**

- `TestMakeArgsAnnotationScopes` — table over all four scope types, comma-separated
  lists, platform-qualified scopes, unknown scopes, colon-in-value,
  single-platform, plus the unscoped and `index:` cases as regression guards.
- `TestPublish` — annotates with `index,manifest:` and asserts the key lands on
  the index *and* on each per-platform manifest (`linux/amd64`, `linux/arm64`)
  against a real registry. Verified to fail without the fix.

**Docs**: new "Annotation scopes" section in `dockers_v2.md`, including the note
that the unscoped default here is `index:` — the inverse of plain
`docker buildx build`.

Annotations scoped to the platform manifests were unreachable: whatever prefix
you wrote, the annotation landed only on the OCI image index, under a corrupted
key.

Anything that resolves a single-platform manifest directly (`docker pull`,
`crane manifest`, and registry-scanning tooling generally) therefore sees no
annotations at all. In our case a proxy resolves the `linux/amd64` manifest, so
`org.opencontainers.image.revision` never arrived and commit-to-deploy
attribution broke — with no error anywhere to point at, since the flag was
accepted and the release succeeded.

One clarification on the issue as I filed it: I wrote that the prefix "is parsed
and stripped from the resulting annotation key correctly". That's wrong. The keys
are silently **corrupted**. The updated integration test fails on `main` with:

```
expected: map[string]string{"org.opencontainers.image.description":"My multi-arch image", "org.opencontainers.image.revision":"abc123"}
actual  : map[string]string{"index,manifest:org.opencontainers.image.revision":"abc123", "org.opencontainers.image.description":"My multi-arch image"}
```

Confirmed independently with a real multi-platform OCI export on buildx v0.36.0:
passing prefixes through unmangled routes them correctly (`index,manifest:` →
index + both platform manifests, `manifest:` → platform manifests only), so
buildx needs no help beyond being handed the flag as written.

- Closes #6799
- #6053 — the PR that introduced the `index:` default, preserved here
- [buildx `ParseAnnotations`](https://github.com/docker/buildx/blob/master/util/buildflags/export.go) — cuts the key at the first colon
- [Docker docs: Annotations → Specify annotation level](https://docs.docker.com/build/metadata/annotations/#specify-annotation-level)
- [OCI image spec: annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md)

### DIFF
--- a/internal/pipe/docker/v2/docker.go
+++ b/internal/pipe/docker/v2/docker.go
@@ -371,7 +371,9 @@ func makeArgs(ctx *context.Context, d config.DockerV2, extraArgs []string) (dock
 	}
 	if len(d.Platforms) > 1 {
 		for i := 1; i < len(annotationFlags); i += 2 {
-			annotationFlags[i] = "index:" + strings.TrimPrefix(annotationFlags[i], "index:")
+			if !hasAnnotationScope(annotationFlags[i]) {
+				annotationFlags[i] = "index:" + annotationFlags[i]
+			}
 		}
 	}
 
@@ -574,6 +576,18 @@ func parsePlatform(p string) platform {
 		result.arm = strings.TrimPrefix(parts[2], "v")
 	}
 	return result
+}
+
+// hasAnnotationScope reports whether a `key=value` annotation already carries a
+// buildx scope prefix.
+//
+// buildx parses annotations as `[type:]key=value`, taking everything before the
+// first colon of the key as the scope list, so a colon there means the user
+// picked the scopes themselves. Invalid scopes are passed through as-is for
+// buildx to report.
+func hasAnnotationScope(annotation string) bool {
+	key, _, _ := strings.Cut(annotation, "=")
+	return strings.Contains(key, ":")
 }
 
 // tplMapFlags templates all keys and values in the given map, returning a
--- a/internal/pipe/docker/v2/docker.go
+++ b/internal/pipe/docker/v2/docker.go
@@ -578,16 +578,34 @@ func parsePlatform(p string) platform {
 	return result
 }
 
+// annotationScopes are the annotation types buildx accepts, optionally
+// qualified with a platform, e.g. `manifest[linux/amd64]`.
+var annotationScopes = []string{
+	"index",
+	"manifest",
+	"index-descriptor",
+	"manifest-descriptor",
+}
+
 // hasAnnotationScope reports whether a `key=value` annotation already carries a
 // buildx scope prefix.
 //
-// buildx parses annotations as `[type:]key=value`, taking everything before the
-// first colon of the key as the scope list, so a colon there means the user
-// picked the scopes themselves. Invalid scopes are passed through as-is for
-// buildx to report.
+// buildx parses annotations as `[types:]key=value`, in which types is a
+// comma-separated list of [annotationScopes]. Keys with any other prefix are
+// scoped by us, as buildx would reject them.
 func hasAnnotationScope(annotation string) bool {
 	key, _, _ := strings.Cut(annotation, "=")
-	return strings.Contains(key, ":")
+	scopes, _, ok := strings.Cut(key, ":")
+	if !ok {
+		return false
+	}
+	for scope := range strings.SplitSeq(scopes, ",") {
+		scope, _, _ = strings.Cut(scope, "[")
+		if !slices.Contains(annotationScopes, scope) {
+			return false
+		}
+	}
+	return true
 }
 
 // tplMapFlags templates all keys and values in the given map, returning a
--- a/www/content/customization/package/dockers_v2.md
+++ b/www/content/customization/package/dockers_v2.md
@@ -123,6 +123,8 @@ dockers_v2:
 
     # Annotations to be added to the image.
     #
+    # Keys may carry a scope prefix, see "Annotation scopes" below.
+    #
     # Items with empty keys or values will be ignored.
     #
     # Templates: allowed.
@@ -137,6 +139,9 @@ dockers_v2:
       # You can also use `.BaseImage` and `.BaseImageDigest`. {{< g_inline_version "v2.16" >}}
       "org.opencontainers.image.base.name": "{{.BaseImage}}"
       "org.opencontainers.image.base.digest": "{{.BaseImageDigest}}"
+
+      # Keys may be scoped to where the annotation lands. {{< g_inline_version "v2.18" >}}
+      "index,manifest:org.opencontainers.image.licenses": "MIT"
 
     # Platforms to build.
     #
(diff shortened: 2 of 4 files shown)


internal/pipe/docker/v2/docker.go
```diff
@@ -576,6 +578,18 @@ func parsePlatform(p string) platform {
...
+// hasAnnotationScope reports whether a `key=value` annotation already carries a
+// buildx scope prefix.
+//
+// buildx parses annotations as `[type:]key=value`, taking everything before the
+// first colon of the key as the scope list, so a colon there means the user
+// picked the scopes themselves. Invalid scopes are passed through as-is for
+// buildx to report.
+func hasAnnotationScope(annotation string) bool {
+	key, _, _ := strings.Cut(annotation, "=")
+	return strings.Contains(key, ":")
```
caarlos0: should we make this more strict, e.g. match `index:` `manifest:` and `manifest[*]:` specifically?

www/content/customization/package/dockers_v2.md
```diff
@@ -138,6 +140,9 @@ dockers_v2:
       "org.opencontainers.image.base.name": "{{.BaseImage}}"
       "org.opencontainers.image.base.digest": "{{.BaseImageDigest}}"
 
+      # Keys may be scoped to where the annotation lands. {{< g_inline_version "v2.18" >}}
```
caarlos0:
```suggestion
      # Keys may be scoped to where the annotation lands. {{< g_inline_version "v2.18-unreleased" >}}
```

caarlos0 (changes requested):
requested some changes, looks good overall though, thanks!

caarlos0 (approved):

caarlos0:
Made some changes, hope it's ok!

Thanks for the PR 🙏🏻

PR #6801, opened 2026-08-20
fix(config): type retry durations as strings in schema

skatkov:
## Problem

According to schema.json `retry.delay` and `retry.max_delay` should have an integer type value. `time.Duration` is internally an `int64`, so schema reflection otherwise emits "type": "integer"

Screenshot:
Screenshot From 2026-08-21 02-15-30

But expected values are `10s` or `5m` e.g. `time.Duration` should be a string-like value.

## Proposed solution

Describe `retry.delay` and `retry.max_delay` as strings in the generated JSON schema.


caarlos0 (approved):
thanks

PR #6808, opened 2026-08-23
test(dockers/v2): cover debug helpers

0jaspahwa:
Add test coverage for the docker v2 debug helpers.

- `isFileNotFoundError`, `fileNotFoundDetails`, and `buildTree` had no test coverage. This adds table-driven tests for all three, matching the style in `heuristics_test.go`.

- Package coverage goes from **76.2%** to **79.9%**.

- **Note on `buildTree` (91.7%):** the uncovered branch is the error return from the recursive call, which needs a nested directory that fails `os.ReadDir`. That isn't reproducible on Windows — `os.Chmod` only toggles the read-only bit, so I left it rather than adding a platform-specific test.

### DIFF
--- a/internal/pipe/docker/v2/debug_test.go
+++ b/internal/pipe/docker/v2/debug_test.go
@@ -1,0 +1,66 @@
+package docker
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestIsFileNotFoundError(t *testing.T) {
+	tests := []struct {
+		name     string
+		out      string
+		expected bool
+	}{
+		{
+			name:     "empty output",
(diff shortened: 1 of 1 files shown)


0jaspahwa:
The `ruleguard` failure isn't from this PR — it's also the only failing job on `main` right now (e.g. run #11993 on `060260a`, everything else green).

```
ruleguard: internal error: package "regexp" without types was imported from "github.com/goreleaser/goreleaser/v2/internal/changelog"
```

An internal error in `internal/changelog`, which this change doesn't touch. Seems to have started around the Go 1.27 bump (#6802).

All required checks pass here.

caarlos0:
yeah, ruleguard needs some fixing for go 1.27 it seems

PR #6816, opened 2026-08-29
fix(aursources): clarify error and document source.enabled requirement

### LINKED ISSUE #6815: Undocumented: aur_sources fails if source isn't enabled
`aur_sources` requires `source.enabled = true` to be set, however this isn't documented and the error message is a little misleading: `no linux archives found`. 

The same constraint applies to `srpm` but there it is documented.

0jaspahwa:
Closes #6815.

`aur_sources` filters for `artifact.UploadableSourceArchive` like `srpm` does, but the error said `no linux archives found`, which doesn't point at the actual cause. The docs page didn't mention the requirement either, while `srpm.md` does.

- error now mentions `source.enabled`
- added a note to the docs, mirroring the one in `srpm.md`

Happy to change the wording or drop the docs part if you'd rather handle it differently.

### DIFF
--- a/internal/pipe/aursources/aursources.go
+++ b/internal/pipe/aursources/aursources.go
@@ -30,7 +30,7 @@ const (
 	defaultCommitMsg = "Update to {{ .Tag }}"
 )
 
-var ErrNoArchivesFound = errors.New("no linux archives found")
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):

caarlos0:
thank you!

PR #6820, opened 2026-08-29
fix(archive): report stripped binary paths

davicbtoliveira:
Report the actual paths written into archives through the `.Binaries`
  artifact template field.

  When `strip_binary_directory: true` is enabled, archive creation writes a
  binary such as `bin/foo` as `foo`. The artifact metadata previously continued
  to report `bin/foo`, even though that path did not exist inside the archive.

  This changes `ExtraBinaries` to use the final archive destination path.

  Tests cover the resulting metadata and its use by the Archive, Homebrew, and
  Nix pipes. The artifact documentation now describes `.Binaries` as paths
  inside the archive.

  

  Consumers such as Homebrew and Nix use `ExtraBinaries` when generating their
  default installation instructions. Reporting the original binary name can
  produce installation commands referencing paths that are not present when
  `strip_binary_directory` is enabled.

  This was discovered while reviewing #6715 and extracted into a separate PR
  because it affects existing archive consumers independently of APKBUILD
  support.

  

  Related to #6715.

  AI assistance disclosure: Codex assisted with implementation and review. I
  reviewed and understand the resulting changes.


internal/pipe/brew/brew_test.go
```diff
@@ -1402,6 +1402,21 @@ func TestInstalls(t *testing.T) {
 		}, install)
 	})
 
+	t.Run("from archive with stripped binary directory", func(t *testing.T) {
```
caarlos0: This subtest cannot fail if `internal/pipe/archive/archive.go:246` is reverted, so it does not protect the fix.

`installs()` (`internal/pipe/brew/brew.go:396-405`) reads `ExtraBinaries` from the literal built here. The archive pipe is not on the call path, so the input `[]string{"foo"}` is the same before and after the change. The test only shows that `installs` echoes its input.

It is also a strict subset of `"from archives"` directly above (line 1387), which already asserts `bin.install "foo"` for `[]string{"foo", "bar"}`.

Smallest fix: remove it. `internal/pipe/archive/archive_test.go:231-233` is the regression test. If you want cross-pipe coverage, build the artifact with `archive.Pipe{}.Run` and `StripBinaryDirectory: true` over a build named `bin/foo`, then give the produced artifact to `installs`. That is the only shape that pins the contract this PR restores.

internal/pipe/nix/nix_test.go
```diff
@@ -757,6 +757,24 @@ func TestBinInstallFormats(t *testing.T) {
 	})
 }
 
+func TestInstallsStrippedBinaryDirectory(t *testing.T) {
```
caarlos0: Same problem as `internal/pipe/brew/brew_test.go:1405`: nothing here is stripped, and this test cannot fail if `internal/pipe/archive/archive.go:246` is reverted. `installs()` (`internal/pipe/nix/nix.go:513-540`) never sees the archive configuration, so the hand-built `[]string{"foo"}` is constant across the change. The name `TestInstallsStrippedBinaryDirectory` claims coverage that does not exist.

The asserted output is also already covered: `cp -vr ./foo $out/bin/foo` appears in 21 golden files under `internal/pipe/nix/testdata/`.

Smallest fix: remove it. If you want to keep it because it is the first direct unit test of nix `installs()`, rename it to what it tests (the default install guess for an archive) so it does not imply strip coverage.

www/content/customization/general/artifacts.md
```diff
@@ -84,7 +84,7 @@ The most common fields are:
 | ------------------- | ---------- | ---------------------------------------------------------- |
 | `ID`                | `string`   | The artifact ID from the configuration                     |
 | `Binary`            | `string`   | The binary name (for archives with a single binary)        |
-| `Binaries`          | `[]string` | List of binary names (for archives with multiple binaries) |
+| `Binaries`          | `[]string` | List of binary paths inside an archive                     |
```
caarlos0: "inside an archive" is not accurate when `wrap_in_directory` is set.

`EnhancedArchive.Add` (`internal/pipe/archive/archive.go:385`) stores the file at `Join(wrap, dst)`, while `ExtraBinaries` holds only `dst`. That is why every consumer re-joins it with `WrappedIn`: `scoop.go:403`, `aur.go:158`, `winget.go:559`, and the `rename` stanzas in `cask/templates/linux_packages.rb:15-18` and `macos_packages.rb:24-27`. A user who reads this row and uses the value as an archive-root path gets a missing file whenever `wrap_in_directory` is on.

```suggestion
| `Binaries`          | `[]string` | List of binary paths, relative to `WrappedIn`              |
```

caarlos0 (commented):
The fix at `internal/pipe/archive/archive.go:246` is correct and minimal. `dst` is the exact `Destination` handed to `a.Add`, so the reported value can no longer disagree with what the archive holds, and the change is a no-op when `strip_binary_directory` is false.

I traced all seven `ExtraBinaries` consumers (`brew.go:402`, `nix.go:531`, `aur.go:157`, `scoop.go:400`, `winget.go:557`, `krew.go:246`, `cask.go:459`). None becomes newly wrong, and all become correct under strip. Two of them use the same string both as the in-archive source path and as the installed name (`nix.go:544`, `aur.go:159`), so before this change both halves were wrong. The top-level `binary` stanza in `cask/templates/cask.rb:39` reads `config.HomebrewCask.Binaries`, not the artifact extra, so it is unaffected.

`internal/pipe/archive/archive_test.go:231-233` is a true regression test: with line 246 reverted it gets `bin/mybin` instead of `mybin`. No other test or golden file asserts `ExtraBinaries` under `strip_binary_directory`.

Three non-blocking remarks inline. Two are about tests that cannot detect the bug they are named after; one is about the documentation line.

Out of scope, do not fix here: `aur.go:159` and `nix.go:544` still build a nested install target (`/usr/bin/bin/foo`) when the binary name has a directory and `strip_binary_directory` is false. That defect exists today and is independent.

Verification gap: at the requester's instruction I did not build, lint, or run tests. Every claim above comes from static tracing of the source; I did not execute the new tests.

This review comes from a bot.

caarlos0 (approved):

caarlos0:
thanks

PR #6821, opened 2026-08-29
fix(client): reconfigure reused Git checkouts

davicbtoliveira:
Apply the current caller's commit-author and signing configuration whenever a
  Git checkout is used, including when the checkout already exists.

  Optional signing settings from a previous caller are cleared before applying
  the current configuration. This prevents a reused checkout from retaining the
  first caller's identity, signing key, GPG program, or signing format.

  Regression tests verify that:

  - The second caller's name and email are used in a reused checkout.
  - Signing configuration from a later caller is applied.

  

  Git configuration was previously applied only immediately after cloning.
  Subsequent calls that reused the checkout inherited its existing author and
  signing settings, potentially creating commits with the wrong identity or
  configuration.

  This was discovered while reviewing #6715 and extracted into a separate PR
  because the shared Git client is used by several existing publishing pipes.

  

  Related to #6715.

  AI assistance disclosure: Codex assisted with implementation and review. I
  reviewed and understand the resulting changes.


internal/client/git.go
```diff
@@ -131,6 +104,28 @@ func (g *gitClient) CreateFiles(
...
+	}
+	if commitAuthor.Signing.Key != "" {
+		gitCmds = append(gitCmds, []string{"config", "--local", "user.signingKey", commitAuthor.Signing.Key})
+	}
+	if commitAuthor.Signing.Program != "" {
+		gitCmds = append(gitCmds, []string{"config", "--local", "gpg.program", commitAuthor.Signing.Program})
+	}
+	if commitAuthor.Signing.Format != "" && commitAuthor.Signing.Format != "openpgp" {
+		gitCmds = append(gitCmds, []string{"config", "--local", "gpg.format", commitAuthor.Signing.Format})
+	}
```
caarlos0: These three keys are now written even when `Signing.Enabled` is false. The base code wrote them only inside the `if commitAuthor.Signing.Enabled` branch.

This is not observable today, because `commit.gpgSign=false` (line 114) makes git ignore `user.signingKey`, `gpg.program` and `gpg.format`. But it is an unnecessary change of meaning in a block that is otherwise a pure move, and it puts a signing key into `.git/config` for a user who disabled signing.

Smallest fix: keep the guard, so that lines 116-124 stay inside `if commitAuthor.Signing.Enabled { ... }`. No new test is needed.

caarlos0 (commented):
The fix is correct and the premise holds. `cwd` is `dist/git/<repo.Name>-<branch>` (`internal/client/git.go:80-82`), and eight pipes share this client (`brew.go:166`, `cask.go:179`, `scoop.go:241`, `nix.go:413`, `winget.go:447`, `krew.go:329`, `aur.go:407`, `aursources.go:359`), each with its own `commit_author`. Two entries that push to the same tap did reuse the first caller's identity and signing settings. Moving the config block after the branch checkout is inert: `git checkout` and `git checkout -b` read none of these keys, and `init.defaultBranch` was already written after `git clone` before this change, so it was inert then too. I confirmed that both new subtests fail on the base commit.

I have two comments inline. Neither blocks the fix.

On the `--get` probe in `unsetGitConfigs`: I checked this and it is right, so no change is needed. With git 2.55, `config --local --get <key>` exits 1 for a missing key and 0 otherwise (including for multi-valued keys), so the probe never skips a necessary unset. Calling `--unset-all` directly is not a good alternative: it exits 5 with empty stderr for a missing key, and `git.RunWithEnv` (`internal/git/git.go:43`) would turn that into an error with a blank message on the normal first-call path.

Verification gaps, stated openly:
- I did not run the test suite or the build. I was told to assume build, lint and fmt are OK, and the machine load average was 20.34, above the local threshold.
- I did not verify the `signing` subtest on Windows. CI runs `task test` on `windows-latest`, and the identical `/usr/bin/gpg` plus `ErrorContains(err, "gpg")` pattern at `git_test.go:325` and `git_test.go:350` is green today, so it must be safe.

This review is automated.

internal/client/git_test.go
```diff
@@ -421,6 +422,49 @@ func TestGitClientWithSigning(t *testing.T) {
...
+		err := cli.CreateFile(ctx, config.CommitAuthor{
+			Name:  "Second",
+			Email: "second@example.com",
+			Signing: config.CommitSigning{
+				Enabled: true,
+				Key:     "missing-key",
+				Program: "/usr/bin/gpg",
+			},
+		}, repo, []byte("second"), "file.txt", "second")
+		require.ErrorContains(t, err, "gpg")
```
caarlos0, line 466: The new `unsetGitConfigs` helper is never exercised. In both subtests the *first* caller sets no `Signing.Key`, `Program` or `Format`, so the probe at `git.go:265` always fails, the unset branch never runs, and the test still passes if you delete the call at `git.go:107`.

The leak that the unset exists for is different from the one you test, and a user can trigger it: entry A sets `commit_author.signing.key`, entry B sets `signing.enabled: true` with no `key` because it depends on the key that git selects from the e-mail. Both push to the same tap. Without the unset, the local `user.signingKey` of entry A stays in the reused checkout and git signs the commit of entry B with the key of entry A. `commit.gpgSign` is rewritten on each call, so the return value alone cannot show this.

Smallest test: call 1 with `Signing{Enabled: true, Key: "first-key", Program: "/usr/bin/gpg"}` (the commit fails, but git writes the local config first, so use `require.Error`), call 2 with no signing, then read the checkout config directly and require it to be empty:

```go
cwd := filepath.Join(ctx.Config.Dist, "git", repo.Name+"-")
out, err := exec.CommandContext(t.Context(), "git", "-C", cwd, "config", "--local", "--get", "user.signingKey").CombinedOutput()
require.Error(t, err) // exit 1: the key is not set
require.Empty(t, strings.TrimSpace(string(out)))
```

The same assertion on the local config is also a better check for this subtest than `ErrorContains(err, "gpg")` at line 464, which proves the fix only through a downstream gpg failure. Adding `require.Equal(t, "true", ...)` for `commit.gpgSign` here would test the contract directly.

internal/client/git.go
```diff
@@ -265,6 +262,19 @@ func runGitCmds(ctx *context.Context, cwd string, env []string, cmds [][]string)
 	return nil
 }
 
+func unsetGitConfigs(ctx *context.Context, cwd string, env []string, keys ...string) error {
```
caarlos0, line 271: Consider dropping this helper and passing the configuration as `-c` flags on the commit command instead.

Only `git commit` reads `user.name`, `user.email`, `commit.gpgSign`, `user.signingKey`, `gpg.program` and `gpg.format`. `clone`, `checkout`, `add` and `push` ignore them. If those values travel as `-c key=value` on the commit, they become per-call data instead of persistent checkout state, and no reset is necessary: this function, the probe-then-unset dance, and the `strconv` import all disappear.

I verified that `-c` persists nothing:

```
$ git -c user.name=Ephemeral -c user.email=e@e.com commit --allow-empty -m viaflags
last author:            Ephemeral <e@e.com>
persisted user.name:    [FromWorktree1]   # unchanged
```

`git.RunWithEnv` already prepends `-c log.showSignature=false`, so the mechanism is present.

The get-then-unset probe here is a reasonable workaround as written, because `git config --unset-all` exits 5 on a missing key and `RunWithEnv` discards the exit status. But it exists only to undo state that does not need to be stored in the first place.

One aside while you are here: `init.defaultBranch` (line 113) is written **after** the clone, and nothing in this path runs `git init`, so no command ever reads it. It can go with the rest.

Test needed: the existing `TestGitClientReconfiguresReusedCheckout` covers this design too, unchanged — which is a good sign for the swap.

internal/client/git.go
```diff
@@ -131,6 +104,30 @@ func (g *gitClient) CreateFiles(
 		}
 	}
 
+	if err := unsetGitConfigs(ctx, cwd, env, "user.signingKey", "gpg.program", "gpg.format"); err != nil {
```
caarlos0, line 103: This reset makes checkout reuse safe for the commit identity, but reuse has a second and worse failure that it does not address. Flagging it here because this is the line that makes reuse a supported path; the defect itself is on the unchanged line 81 and predates this PR (it arrived with 08851dce6, #4712).

`name := repo.Name + "-" + g.branch` (line 81) is not derived from the git URL, and `url` is never used again after the clone — the push at line 154 goes to whatever `origin` the existing checkout already has. So two publishers with the same `repository.name` and branch but different `git.url` share one checkout, and the second one pushes into the first one's repository.

Reproduced with plain git:

```
--- repoA.git (pipe 1's remote) ---
a2e7cbe two      <- pipe 2's commit landed here
e5a48f9 one
--- repoB.git (pipe 2's intended target) ---
fatal: unknown revision 'master'
```

A user reaches this with, for example, brew and cask both named `homebrew-tap` under different owners. `RepoFromRef` copies the user-supplied name verbatim, and when the name is empty `nameFromURL` takes only the last path segment, so two different hosts or owners collide just as easily. `aur.go:414` and `aursources.go:366` already hash the git URL by hand to dodge exactly this.

Smallest fix: key the directory on the templated URL plus branch (a SHA-256 of `url + "\x00" + branch`), so reuse can only ever happen for the same remote. The AUR and aursources workarounds can then be deleted. `dist` is wiped per run (`internal/pipe/dist/dist.go:23`), so there is no on-disk compatibility cost.

This is out of scope for this PR — please do it separately. Note that it does **not** remove the need for this reset: two configs on the same URL with different `commit_author` values still share a checkout.

Test needed: two bare repositories, `CreateFile` twice with the same `Repo.Name` and branch but different URLs, and assert each bare repository receives only its own file.

internal/client/git_test.go
```diff
@@ -421,6 +423,66 @@ func TestGitClientWithSigning(t *testing.T) {
...
+		require.NoError(t, err, string(out))
+		require.Equal(t, "true", strings.TrimSpace(string(out)))
+
+		require.NoError(t, cli.CreateFile(ctx, config.CommitAuthor{
+			Name: "Third", Email: "third@example.com",
+		}, repo, []byte("third"), "file.txt", "third"))
+
+		out, err = exec.CommandContext(t.Context(), "git", "-C", cwd, "config", "--local", "--get", "user.signingKey").CombinedOutput()
+		require.Error(t, err)
+		require.Empty(t, strings.TrimSpace(string(out)))
```
caarlos0, line 478: The clearing of `gpg.program` and `gpg.format` is new behaviour with no coverage, so both can regress silently.

`gpg.format` is never exercised at all: the second caller (line 460) sets `Key` and `Program` but no `Format`, so `gpg.format` is never written and its removal is never tested. `gpg.program` **is** written, but nothing asserts it is gone afterwards. Because the third caller has signing disabled, a stale `gpg.program` does not affect the commit — git only reads it when signing. So dropping either key from the `unsetGitConfigs` call on line 107 leaves this test green.

Smallest fix: set `Format: "ssh"` on the second caller, and after the third call assert that `gpg.program` and `gpg.format` are absent, next to the `user.signingKey` check here.

While editing these lines: the bare `require.Error(t, err)` on line 477 aborts before line 478 runs, so a regression reports "An error is expected but got nil" and hides the key that was actually retained. Pass the value: `require.Error(t, err, strings.TrimSpace(string(out)))`.

One more on line 466: `require.ErrorContains(t, err, "gpg")` passes on a machine without gpg (`cannot run /usr/bin/gpg`) and on one with an empty keyring, and it would also pass if only `commit.gpgSign=true` were applied while the key and program were dropped. Asserting the actual `user.signingKey` and `gpg.program` values immediately after the failed second call would make the subtest prove what it claims.

caarlos0 (commented):
Automated review (GitHub Copilot CLI agent), run by @caarlos0.

**The fix is correct.** Every local key this client can write is now overwritten or cleared on each call, and moving the config after the checkout block is safe: `clone`, `checkout` and `checkout -b` read neither the commit identity nor the signing keys. `TestGitClientReconfiguresReusedCheckout` passes on this branch (both subtests). At the merge base the config block sits inside the `os.Stat` clone branch, so the second caller keeps `user.name=First` and the `commit author` subtest must fail there.

My concern is the shape, not the behaviour. This change writes per-caller state into a shared checkout and then repairs it before each use. Two comments below propose a smaller design and flag the deeper defect the reuse model still has; a third flags missing coverage for the new clearing logic.

**Verification gaps, disclosed:**

- Build, lint and gofmt were not run. The requester stated they pass.
- `TestGitClient/full`, `/with_new_branch` and `/no_repo_name` already fail on this machine with `fatal: cannot use bare repository ... (safe.bareRepository is 'explicit')`, through `internal/testlib/git.go:156`, which runs `git -C <bare-repo>`. The new test uses the same pattern and fails identically. This is pre-existing and is **not** a defect of this PR. Switching testlib to `git --git-dir=` would fix all of them at once, in its own PR.
- `TestGitClientWithSigning/commit_signing_with_ssh_format` fails here with `could not create temporary file: Operation not permitted`, an artefact of the review sandbox, unrelated to this change.
- Windows behaviour and a system-level `gpg.format` override were not checked.

Two findings reported by the assisting agents were rejected: the `safe.bareRepository` test failure (pre-existing, affects the existing suite equally) and the error classification inside `unsetGitConfigs` (no user-observable impact beyond a less precise error message).

caarlos0:
Superseded by #6852, which is merged. It fixes the same leak by passing the author and signing options as `git -c` flags on the commit, so no per-caller state is written to the shared checkout, and it includes the regression test from this PR.

Thanks a lot for finding the problem and for the tests!

PR #6824, opened 2026-08-30
feat(nix): drop the nix-hash dependency

caarlos0:
Remove GoReleaser's dependency on Nix being installed.

`nix-hash --type sha256 --flat --base32` is simply the sha256 of the file
rendered in Nix's own base32 alphabet, so GoReleaser now does it directly
instead of shelling out.

Users had to have Nix installed to publish a Nix package: the pipe listed
`nix-hash` as a dependency and skipped the whole thing when it was missing,
which is a surprising failure mode for a package manager you may not otherwise
use.

**How it is verified.** The repository already contained a reference vector:
`internal/pipe/nix/testdata/file.bin` with the expected
`1n7yy95h81rziah4ppi64kr6fphwxjiq8cl70fpfrqvr0ml1xbcl`, produced by `nix-hash`
itself. The new implementation reproduces it exactly, and the all-zeroes vector
matches the `zeroHash` constant already used by the tests.

A third vector, the sha256 of `goreleaser`, was added because those two only
ever emit 27 of the 32 characters in the alphabet: with `2` and `3` exchanged
in `nixBase32Alphabet` both of them still passed. With the third one, a typo
anywhere in the alphabet fails the test.

I did try a differential test that shelled out to real `nix-hash` and compared,
but `testlib.CheckPath` does not skip on CI — `InPath` returns true there on
purpose so a missing tool fails loudly rather than skipping silently — so that
test cannot survive removing Nix from CI. The reference vectors cover it.

`TestRunSkipNoNix` covered the skip that no longer exists, so it goes too. The
pipe no longer declares any dependency, so it is also removed from
`healthcheck.DependencyCheckers`.

**CI.** With hashing done in Go, nothing in CI needs Nix any more, so the
installation is gone from `build.yml`, `release.yml` and `nightly-oss.yml`. The
two formatters the tests exercise are fetched as pinned release binaries by
#6828, which landed first.

Docs updated at `www/content/customization/publish/nix.md`, marked
`v2.19-unreleased`.

AI disclosure: I used an AI coding assistant to implement and verify this
change against the existing reference vector.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -120,10 +120,6 @@ jobs:
         run: |
           flatpak remote-add --if-not-exists --user flathub https://dl.flathub.org/repo/flathub.flatpakrepo
           flatpak install -y --user --noninteractive flathub org.freedesktop.Platform/x86_64/24.08 org.freedesktop.Sdk/x86_64/24.08 org.freedesktop.Platform/aarch64/24.08 org.freedesktop.Sdk/aarch64/24.08
-      - uses: cachix/install-nix-action@13d8dd58da0234aa297dedd986986ccb8e7f3e24 # v31.11.1
-        if: matrix.os == 'ubuntu-latest'
-        with:
-          github_access_token: ${{ secrets.GITHUB_TOKEN }}
       - name: setup-nix-formatters
         # `nix profile add nixpkgs#...` resolves and builds from nixpkgs, which
         # took 18s. Both projects publish static linux binaries, so fetching
--- a/internal/pipe/nix/nix.go
+++ b/internal/pipe/nix/nix.go
@@ -5,8 +5,10 @@ import (
 	"bufio"
 	"bytes"
 	"cmp"
+	"crypto/sha256"
 	"errors"
 	"fmt"
+	"io"
 	"maps"
 	"os"
 	"os/exec"
@@ -64,7 +66,7 @@ type Pipe struct {
 
 func (Pipe) String() string                           { return "nixpkgs" }
 func (Pipe) ContinueOnError() bool                    { return true }
-func (Pipe) Dependencies(_ *context.Context) []string { return []string{nixHashBin} }
+func (Pipe) Dependencies(_ *context.Context) []string { return nil }
 
 func (p Pipe) Skip(ctx *context.Context) bool {
 	return skips.Any(ctx, skips.Nix) || len(ctx.Config.Nix) == 0
@@ -94,9 +96,6 @@ func (Pipe) Default(ctx *context.Context) error {
 }
 
 func (p Pipe) Run(ctx *context.Context) error {
-	if !p.hasher.Available() {
-		return pipe.Skipf("%s is not available", nixHashBin)
-	}
 	cli, err := client.NewReleaseClient(ctx)
 	if err != nil {
 		return err
@@ -599,33 +598,45 @@ func depNames(deps []config.NixDependency) []string {
 
 type fileHasher interface {
 	Hash(ctx stdctx.Context, name string) (string, error)
-	Available() bool
 }
 
-const nixHashBin = "nix-hash"
+// nixBase32Alphabet is nix's own base32 alphabet: it drops e, o, u and t.
+const nixBase32Alphabet = "0123456789abcdfghijklmnpqrsvwxyz"
 
-var realHasher fileHasher = nixHasher{bin: nixHashBin}
-
-type nixHasher struct{ bin string }
-
-func (p nixHasher) Available() bool {
-	_, err := exec.LookPath(p.bin)
-	return err == nil
-}
-
-func (p nixHasher) Hash(ctx stdctx.Context, name string) (string, error) {
-	// $ nix-hash --type sha256 --flat --base32 <(echo test)
-	out, err := exec.CommandContext(
-		ctx,
-		p.bin,
-		"--type", "sha256",
-		"--flat",
-		"--base32",
-		name,
-	).Output()
-	outStr := strings.TrimSpace(string(out))
-	if err != nil {
-		return "", fmt.Errorf("could not hash file: %s: %w: %s", name, err, outStr)
+// nixBase32 renders a hash the way nix does, reading the bit groups from the
+// end backwards. See printHash32 in nix's libutil/hash.cc.
+func nixBase32(h []byte) string {
+	n := (len(h)*8-1)/5 + 1
+	out := make([]byte, n)
+	for k := range n {
+		b := (n - 1 - k) * 5
+		i, j := b/8, b%8
+		c := h[i] >> j
+		if i+1 < len(h) {
+			c |= h[i+1] << (8 - j)
+		}
+		out[k] = nixBase32Alphabet[c&0x1f]
 	}
-	return outStr, nil
+	return string(out)
+}
+
+var realHasher fileHasher = goHasher{}
+
+// goHasher reproduces `nix-hash --type sha256 --flat --base32`, which is
+// simply the sha256 of the file rendered in nix's base32. Doing it here means
+// users do not need nix installed to publish a nix package, and saves a
+// process per artifact.
+type goHasher struct{}
+
+func (goHasher) Hash(_ stdctx.Context, name string) (string, error) {
+	f, err := os.Open(name)
+	if err != nil {
+		return "", fmt.Errorf("could not hash file: %s: %w", name, err)
+	}
+	defer f.Close()
+	h := sha256.New()
+	if _, err := io.Copy(h, f); err != nil {
+		return "", fmt.Errorf("could not hash file: %s: %w", name, err)
+	}
+	return nixBase32(h.Sum(nil)), nil
 }
--- a/internal/pipe/nix/nix.go
+++ b/internal/pipe/nix/nix.go
@@ -18,8 +18,6 @@ import (
 	"strings"
 	"text/template"
 
-	stdctx "context"
-
 	"github.com/caarlos0/log"
 	"github.com/goreleaser/goreleaser/v2/internal/artifact"
 	"github.com/goreleaser/goreleaser/v2/internal/client"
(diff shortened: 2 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -124,12 +124,21 @@ jobs:
...
+        # format tests compare against golden files, so floating with whatever
+        # nixpkgs currently ships could break them at any time.
         if: matrix.os == 'ubuntu-latest'
+        run: |
+          mkdir -p "$HOME/.local/bin"
+          curl -fsSL -o "$HOME/.local/bin/alejandra" \
+            https://github.com/kamadorueda/alejandra/releases/download/4.0.0/alejandra-x86_64-unknown-linux-musl
+          curl -fsSL -o "$HOME/.local/bin/nixfmt" \
+            https://github.com/NixOS/nixfmt/releases/download/v1.4.0/nixfmt
+          chmod +x "$HOME/.local/bin/alejandra" "$HOME/.local/bin/nixfmt"
```
caarlos0, line 141: These two binaries are fetched by tag and then executed in CI with no integrity check. They are the only unverified binary downloads in the workflows: every other tool comes from an action pinned to a full commit SHA, or from apt/snap.

**Trigger.** A GitHub release asset is not immutable — the upstream owner can delete and re-upload at the same tag, and CI then runs a different binary with no signal. Dependabot covered `cachix/install-nix-action` before this change, but it cannot read a URL inside a `run:` block, so these two versions will also age invisibly. Blast radius is limited: `permissions: contents: read` on this job means code execution in the job and cache poisoning, not repository write.

**Smallest fix** — pin the content, not only the tag:

```sh
echo "a23b9d47cba945805c6169541046de890e94e07a5aa416c86dee15bca2da6216  $HOME/.local/bin/alejandra" | sha256sum -c -
echo "62488394d5233283096466350487ed46470366f57db4bec824a87ecbacce960a  $HOME/.local/bin/nixfmt" | sha256sum -c -
```

I measured both checksums from the exact URLs in this diff, so they can be pasted as-is.

**Test.** CI itself: the step fails if either asset changes.

This is worth doing because these binaries are load-bearing, not optional. `testlib.InPath` returns `true` unconditionally when `CI=true`, so `CheckPath(t, "alejandra")` and `CheckPath(t, "nixfmt")` do **not** skip in CI — the format tests really execute what this step downloads.

internal/pipe/nix/nix_test.go
```diff
@@ -39,39 +38,31 @@ func TestSkip(t *testing.T) {
...
-			require.False(t, nixHasher{fakeNixHashBin}.Available())
-		})
-		t.Run("valid", func(t *testing.T) {
-			testlib.CheckPath(t, "nix-hash")
-			testlib.SkipIfWindows(t, "nix doesn't work on windows")
-			require.True(t, realHasher.Available())
-		})
+
+	t.Run("all zeroes", func(t *testing.T) {
+		require.Equal(t, zeroHash, nixBase32(make([]byte, sha256.Size)))
```
caarlos0, line 55: `nixBase32Alphabet` (nix.go:604) is a hand-typed 32-character constant introduced by this PR, and the two vectors in this test exercise only 27 of its characters. `2`, `3`, `d`, `g` and `s` are never produced, so nothing pins them.

**Trigger.** A transposition inside that unexercised set, in this constant or in a later refactor. I confirmed both tests still pass against a mutant alphabet with `2` and `3` swapped:

```
mutant alphabet "0132456789abcdfghijklmnpqrsvwxyz"
  mutant(file.bin) == reference vector? True
  mutant(zeroes)   == zeroHash?         True
```

**Impact.** A silently wrong `sha256-` in a published `default.nix`, not a crash. The reference vector is strong against every other mutation I tried (off-by-one in `n`, dropping `& 0x1f`, and guard changes all fail or panic) — the alphabet is the one axis it misses. The all-zeroes case only pins `alphabet[0] == '0'`.

**Smallest fix** — one more vector, no new testdata file, which brings the union to 32/32:

```go
t.Run("full alphabet", func(t *testing.T) {
	// printf 'goreleaser' > f; nix-hash --type sha256 --flat --base32 f
	sum := sha256.Sum256([]byte("goreleaser"))
	require.Equal(t, "19a48kbsv6pr8x1bnbbs61f80wilnc2dcgahb70xbmb73pr0s7sk", nixBase32(sum[:]))
})
```

I derived that value with my independent decoder, not from the code under review, and confirmed the code under review reproduces it and that it does catch the `2`/`3` mutant.

internal/pipe/nix/nix.go
```diff
@@ -64,7 +66,7 @@ type Pipe struct {
 
 func (Pipe) String() string                           { return "nixpkgs" }
 func (Pipe) ContinueOnError() bool                    { return true }
-func (Pipe) Dependencies(_ *context.Context) []string { return []string{nixHashBin} }
+func (Pipe) Dependencies(_ *context.Context) []string { return nil }
```
caarlos0: Non-blocking cleanup created by this line. With `Dependencies` returning `nil`, `nix.New()` in `pkg/healthcheck/healthcheck.go:62` is a registered `DependencyChecker` that can never produce a check — `cmd/healthcheck.go:60` ranges over `nil` and does nothing. Output is correct, so no user sees anything wrong; it is dead surface only.

If you want it gone: delete this method, drop `nix.New()` and its import from `DependencyCheckers`, and delete `TestDependencies`. The compiler proves it; no test needed.

Keeping it is also defensible if you expect the pipe to declare the configured `formatter` (`nixfmt`/`alejandra`) as a dependency later — those are real binaries the pipe execs and never declared. That would be a separate change.

caarlos0 (commented):
**Automated review — posted by an agent (GitHub Copilot CLI), not by a human.**

The change is correct. I verified the encoder against ground truth rather than by reading it: I decoded the committed vector `1n7yy95h81rziah4ppi64kr6fphwxjiq8cl70fpfrqvr0ml1xbcl` with an independent decoder and it returns exactly `shasum -a 256 internal/pipe/nix/testdata/file.bin` = `94ad1e68...4bf2fed8`. A differential run of `nixBase32` against an independent model (little-endian integer rendered base-32, most significant group first) found 0 mismatches over random inputs at 11 byte lengths, plus the boundary patterns; `n` is 52 for 32 bytes and the leading character is only ever `0` or `1`, so no stray high bits leak.

Two non-blocking findings are inline. Neither breaks a user today.

Also checked, no action needed:
- Dropping `install-nix-action` from `release.yml` and `nightly-oss.yml` is safe: this repo's own `.goreleaser.yaml` sets no `formatter`, so `format()` is never called during its release, and it only warns when it is.
- `p.hasher` cannot be nil on any path reaching `Hash`: every `Run`/`Publish` site uses `New()`, and the one `nix.Pipe{}` (`pkg/defaults/defaults.go`) only calls `Default`.
- The ignored `ctx` in `goHasher.Hash` is real but not worth fixing: it replaces killing an external process with an uninterruptible local sha256 read, which runs at ~1.7 GB/s. A `ctxReader` here would be defensive code with no symptom.
- `v2.19-unreleased` is right; `g_version.html` handles the suffix and the last tag is v2.18.0.
- The pipe now also runs on Windows, where `nix-hash` never existed. Nothing in it needs Unix, and `.gitattributes` (`* text eol=lf`) keeps `testdata/file.bin` byte-identical, so the vector holds there.

Verification gaps, stated plainly:
- I was asked not to build, lint or format locally, so I relied on CI. On the pre-merge head 9bc9467, `test (ubuntu-latest)` and `test (windows-latest)` both passed. The run for the current head 9984b6f is still in progress.
- `ruleguard / scan` fails here, but it also fails on the last 4 `main` runs with the same error. It is not from this change.
- No `nix-hash` binary was available to me, so ground truth came from decoding the committed vector back to the raw sha256, not from running `nix-hash`.

.github/workflows/build.yml
caarlos0: Addressed, and the step left this PR entirely.

The formatter downloads were extracted into #6828, which added the two `sha256sum -c` checks you asked for and merged as ae564f60a. That PR keeps `install-nix-action`, so it stands alone.

```
echo "a23b9d47cba945805c6169541046de890e94e07a5aa416c86dee15bca2da6216  $HOME/.local/bin/alejandra" | sha256sum -c -
echo "62488394d5233283096466350487ed46470366f57db4bec824a87ecbacce960a  $HOME/.local/bin/nixfmt" | sha256sum -c -
```

`test (ubuntu-latest)` passed on #6828, which proves the checksums match the assets and that the format tests still execute those binaries. In `build.yml`, this PR now only removes `install-nix-action`.

internal/pipe/nix/nix_test.go
caarlos0: Fixed in e2f0ae070.

Added a third vector, the sha256 of `goreleaser`, which brings the union to 32/32 characters:

```go
t.Run("full alphabet", func(t *testing.T) {
	// printf goreleaser > f && nix-hash --type sha256 --flat --base32 f
	sum := sha256.Sum256([]byte("goreleaser"))
	require.Equal(t, "19a48kbsv6pr8x1bnbbs61f80wilnc2dcgahb70xbmb73pr0s7sk", nixBase32(sum[:]))
})
```

Verified against the mutant you described: with `2` and `3` exchanged in `nixBase32Alphabet`, `reference_vector` and `all_zeroes` still pass, and only `full_alphabet` fails.

```
--- FAIL: TestHasher/full_alphabet
    expected: "19a48kbsv6pr8x1bnbbs61f80wilnc2dcgahb70xbmb73pr0s7sk"
    actual  : "19a48kbsv6pr8x1bnbbs61f80wilnc3dcgahb70xbmb72pr0s7sk"
```

internal/pipe/nix/nix.go
caarlos0: Removed in e2f0ae070.

Deleted the `Dependencies` method, the `nix.New()` entry and its import in `pkg/healthcheck/healthcheck.go`, and `TestDependencies`.

I did not keep it for a future `formatter` dependency: `format()` only warns when the binary is missing, so declaring it would make the healthcheck report a failure for something that is not fatal. If that becomes wanted, it is a separate change.

`go build ./...` and `go test ./internal/pipe/nix/ ./pkg/healthcheck/ ./cmd/` pass.

PR #6826, opened 2026-08-30
perf: skip needless docker probes; make the ko registry test hermetic

caarlos0:
Two findings from the `go test -json` timings captured on CI.

**`dockers_v2` probed docker even when every configuration was disabled**

`Snapshot.Run` and `Publish.Publish` ran `docker buildx inspect` and `docker
info` *before* evaluating each configuration's `disable` template. A fully
disabled configuration therefore paid two subprocesses and logged a misleading
`no docker daemon available` warning. They now evaluate `disable` first and
skip.

This is a small production win too: users with `disable: true` no longer shell
out to docker at all.

`TestDisable/disabled` went from **4.6s to 0.00s** on the windows runner, where
probing a docker that is not installed is slow.

**The ko test pointed at a single-label host name**

`fakerepo` is not guaranteed to fail to resolve: a wildcard search domain would
resolve it and quietly change what the test exercises. It also makes the
resolver walk the search domains first, which is why the subtest costs 4.6s on
CI.

`.invalid` never resolves (RFC 2606) and contains a dot, so it is tried as-is
first.

Continuing the pass over the slowest tests. For the record, the ones above these
in the list are genuinely toolchain-bound and I left them alone: zig, rust,
node, poetry and golang `TestBuild` run real compilers, `snapcraft.TestRunPipe`
really runs `snapcraft pack`, and the `docker`/`sign` tests really build and
sign images.

One I looked at and deliberately did not change: `docker.TestBuildCommand`
appears to cost 4.6s on windows, but that is the one-time
`dockerSupportsProvenanceFlags` probe (`sync.OnceValue` running
`docker build --help`) being attributed to whichever subtest happens to run
first. Production needs that probe.

I also measured the zig/rust builder tests, which are the two slowest overall.
They force a per-test `ZIG_GLOBAL_CACHE_DIR`, so every run recompiles zig's
stdlib. Sharing that cache is worth about **41%** on the zig test locally
(10.5s -> 6.2s), but the isolation is deliberate: the comment records that a
shared cache got corrupted when the zig and rust packages built in parallel
(#6754). That deserves its own PR with repeated-run validation, not a drive-by.

AI disclosure: I used an AI coding assistant to analyse the CI timings and make
these changes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -102,6 +102,15 @@ jobs:
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
           go-version-file: go.mod
+          # setup-go keys its cache on the hash of these files only, and saves
+          # it just once per key. license-check and golangci-lint resolve the
+          # same key and always finish first, so this job kept restoring a
+          # build cache without the -race/-coverpkg objects and recompiled
+          # everything (~2.5min). Taskfile.yml holds the test flags, so hashing
+          # it too gives this job a key of its own that it can populate.
+          cache-dependency-path: |
+            go.sum
+            Taskfile.yml
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           node-version: "25"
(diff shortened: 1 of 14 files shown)


caarlos0:
Opened by mistake: this worktree was on `perf/faster-tests` (upstream deleted after #6823 merged), so the push was a no-op and the PR was raised from the stale `ci` branch, which only contains the already-merged #6822 commits. Reopening from a fresh branch.

PR #6830, opened 2026-08-30
chore: measure per-test times in CI (do not merge)

caarlos0:
Temporary measurement branch: emits `go test -json` so per-test durations show up in the job logs. Not for merge.


caarlos0:
Measurement done: per-test durations for both jobs are in #6837 and the PRs it lists.

PR #6831, opened 2026-08-30
test(cmd): stop repeating work the tests already cover

caarlos0:
The cmd package was the slowest package on the windows job (40s) and the
fifth slowest on ubuntu (25s). Two things account for most of it.

`TestBuild` and `TestRelease` ran a full pipeline and asserted only that it
returned no error. The `dirty` case of `TestBuildAutoSnapshot` and
`TestReleaseAutoSnapshot` runs the same pipeline in the same snapshot mode --
`--auto-snapshot` on a dirty tree sets `ctx.Snapshot`, which applies exactly
the skips `--snapshot` applies -- and additionally asserts the snapshot name
reached the artifacts. The flags the two dropped tests passed move to the
auto-snapshot cases, so `--timeout`, `--parallelism` and `--deprecated` stay
covered, and four pipeline runs become two.

The config tests called `setup`, which creates a buildable Go module and a
git repository with four commits and two tags. None of them runs a pipeline;
they only need a config file to find. `TestConfigFlagNotSetButExists` paid
for that fixture six times. They now use a new `mktmp` helper, which saves 54
git processes -- expensive on windows, where the package spent 5.6s in that
one test.

Verified the remaining tests still fail for the right reason: forcing the
`options.autoSnapshot && git.CheckDirty(ctx) != nil` branch to false fails
`TestBuildAutoSnapshot/dirty` and `TestReleaseAutoSnapshot/dirty`, and
nothing else.

Measured locally, `go test ./cmd/`: 8.46s/9.20s before, 5.85s/6.37s after.

### DIFF
--- a/cmd/build_test.go
+++ b/cmd/build_test.go
@@ -12,18 +12,13 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestBuild(t *testing.T) {
-	setup(t)
-	cmd := newBuildCmd()
-	cmd.cmd.SetArgs([]string{"--snapshot", "--timeout=1m", "--parallelism=2", "--deprecated"})
-	require.NoError(t, cmd.cmd.Execute())
-}
-
 func TestBuildAutoSnapshot(t *testing.T) {
+	// --auto-snapshot on a clean tree does not imply --snapshot, so this also
+	// covers a build that runs the validation pipe.
 	t.Run("clean", func(t *testing.T) {
 		setup(t)
 		cmd := newBuildCmd()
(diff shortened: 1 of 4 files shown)


caarlos0:
Measured on CI. Package seconds from `go test`'s `ok <pkg> <time>` lines, against
four green `main` runs, because one run proves nothing on these runners.

| --- | --- | --- |
| ubuntu | 23.20 25.27 23.82 24.47 | **15.70** |
| windows | 45.89 40.29 31.52 43.59 | **23.80** |

Below the baseline minimum on both.

Issue #6832, opened 2026-08-30
gitlab: release.replace_existing_artifacts never takes effect


caarlos0:
While reviewing `TestGitLabURLsDownloadTemplate` I found that the replace path in `gitlabClient.Upload` is unreachable.

`internal/client/gitlab.go`:

```go
releaseLink, resp, err := c.client.ReleaseLinks.CreateReleaseLink(projectID, releaseID, opt)
if err != nil {
    // this status means the asset already exists
    if resp != nil && resp.StatusCode == http.StatusBadRequest && releaseLink != nil {
        ...delete and retry...
    }
    return retryx.HTTP(err, must(resp).Response)
}
```

`ReleaseLinks.CreateReleaseLink` is implemented with the generic `do[*ReleaseLink]` helper, which returns the zero value on error. So when GitLab answers `400 {"message":{"name":["has already been taken"]}}`, `releaseLink` is nil, the guard is false, and the upload fails instead of deleting and re-uploading the asset.

Reproduction: point `TestGitLabURLsDownloadTemplate` at a server that answers the first `assets/links` request with that 400 and set `release.replace_existing_artifacts: true`. Every subtest fails with `400 {message: {name: [has already been taken]}}`.

The test used to cover this. #6528 replaced the 400 with a 503 (`service unavailable`), which go-gitlab retries internally, so the regression went unnoticed. Removing the now-meaningless 503 in a separate PR.

A fix needs the existing link ID, which the failed create does not return, so it has to list the release links and match by name before deleting.

PR #6833, opened 2026-08-30
test(client): stop paying for a retry that tests nothing

caarlos0:
`TestGitLabURLsDownloadTemplate` made the first `assets/links` request of
each version return 503. That cost 0.8s per subtest, 6.6s for the test --
6.2s on ubuntu and 6.5s on windows, the second slowest test in the package on
both.

None of it exercised goreleaser. go-gitlab wraps its transport in
`retryablehttp` with `RetryMax: 5` and `RetryWaitMin: 100ms`, so the SDK
absorbs the 503 and sleeps on its own; `retryx` never sees a failure. The
`Retry{Attempts: 2}` and the twelve backoffs bought nothing.

The 503 arrived in #6528, which replaced a 400 `has already been taken`.
That one did exercise goreleaser: it drove the `ReplaceExistingArtifacts`
branch in `gitlabClient.Upload`. Restoring it now fails, because
`ReleaseLinks.CreateReleaseLink` returns a nil link on error, so the
`releaseLink != nil` guard is never true and the branch is unreachable. That
is a separate bug, reported in #6832; this commit removes the leftover
`ReplaceExistingArtifacts: true` rather than leave it implying coverage that
does not exist.

Measured locally: 6.60s -> 0.43s.

### DIFF
--- a/internal/client/gitlab_test.go
+++ b/internal/client/gitlab_test.go
@@ -9,7 +9,6 @@ import (
 	"os"
 	"strconv"
 	"strings"
-	"sync/atomic"
 	"testing"
 	"text/template"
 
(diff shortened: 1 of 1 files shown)


caarlos0:
Measured on CI, `internal/client` package seconds against four green `main` runs:

| --- | --- | --- |
| ubuntu | 11.54 11.13 11.44 11.38 | **4.51** |
| windows | 17.47 16.89 14.20 17.93 | **10.80** |

PR #6834, opened 2026-08-30
test(builders): drop TestBuildSpecificModes

caarlos0:
`TestBuildSpecificModes` was added with the poetry and uv builders in #5734.
It is `TestBuild` again: same `poetry new proj` / `uv init proj` fixture, same
two builds in wheel and sdist buildmode, same two expected artifacts, same
mod-timestamp check. The only differences are the build ids it asserts
(`wheel`/`sdist` instead of `proj-wheel`/`proj-sdist`) and that it calls
`Default.Build` twice by hand instead of looping over the two builds.

`TestBuild` already proves what the name promises: one build with
`buildmode: wheel` produces only the `.whl`, one with `buildmode: sdist`
produces only the `.tar.gz`.

The duplicate is not free. It runs `poetry new` and two more poetry builds:
4.65s on ubuntu and 7.49s on windows, the seventh slowest test on both jobs.


caarlos0:

| --- | --- | --- | --- |
| `poetry` | ubuntu | 8.83 9.08 9.58 9.40 | **5.13** |
| `poetry` | windows | 17.64 16.60 24.55 17.34 | **12.21** |
| `uv` | ubuntu | 1.39 1.46 1.29 1.38 | **1.25** |
| `uv` | windows | 0.95 0.77 0.48 0.85 | **0.55** |

`uv` is fast either way; poetry is where the duplicate cost real time.

PR #6835, opened 2026-08-30
test: stop cloning repositories from github

caarlos0:
Two tests cloned a repository over the network to set up a fixture they
could have written themselves.

`TestShallowClone` cloned goreleaser at v0.160.0. The pipe only looks for
`.git/shallow`, so any shallow clone proves the same thing: the test now
builds a two-commit repository and shallow clones it over `file://`, which
`--depth` needs because a plain local path clones everything. It also asserts
`.git/shallow` exists, so the test fails at the setup rather than silently
becoming a non-shallow clone check.

`TestRunPipeWithProxiedRepo` cloned goreleaser/test-mod into the working
directory. All it needed from it was a main package, because `Build` parses
`UnproxiedDir` to check `UnproxiedMain` has a main function -- dropping the
clone without a replacement fails with "build for foo does not contain a main
function". A `writeGoodMain` and a one-line go.mod say the same thing. The
proxied build still resolves test-mod through the module proxy, which is the
point of the test.

Both packages now pass with `GOPROXY=off`. Measured locally: `TestShallowClone`
1.48s -> 0.66s, `TestRunPipeWithProxiedRepo` 1.47s -> 0.28s. On CI they cost
3.33s and 7.40s on windows, 3.02s for the proxied one on ubuntu.

### DIFF
--- a/internal/builders/golang/build_test.go
+++ b/internal/builders/golang/build_test.go
@@ -1238,8 +1238,16 @@ func TestBuildTests(t *testing.T) {
 
 func TestRunPipeWithProxiedRepo(t *testing.T) {
 	folder := testlib.Mktmp(t)
-	out, err := exec.CommandContext(t.Context(), "git", "clone", "https://github.com/goreleaser/test-mod", "-b", "v0.1.1", "--depth=1", ".").CombinedOutput()
-	require.NoError(t, err, string(out))
+	// the unproxied dir only has to hold a main package: build.Build parses it
+	// to check that UnproxiedMain has a main function. Writing it beats
+	// cloning goreleaser/test-mod, which needs the network to say the same
+	// thing.
+	writeGoodMain(t, folder)
+	require.NoError(t, os.WriteFile(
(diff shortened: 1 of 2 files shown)


caarlos0:

| --- | --- | --- | --- |
| `internal/pipe/git` | ubuntu | 4.29 4.19 4.31 4.32 | **3.76** |
| `internal/builders/golang` | ubuntu | 6.12 5.99 5.76 6.37 | **4.48** |
| `internal/pipe/git` | windows | 35.84 30.59 21.78 28.77 | 31.39 |
| `internal/builders/golang` | windows | 15.54 14.76 9.75 13.74 | 15.27 |

Clear on ubuntu, inconclusive on windows: both packages swing by 14s there
between `main` runs, which is wider than what these two tests cost. The
per-test measurement from a `-json` run put them at 3.33s and 7.40s on
windows.

The hermeticity is the part that does not depend on runner noise: both
packages now pass with `GOPROXY=off`, and neither can fail because github is
rate limiting or unreachable.

PR #6836, opened 2026-08-30
test(sign): take the docker sign shell from testlib

caarlos0:
`TestDockerSignArtifacts` picks the shell for its fake signer with
`testlib.ShC`, and splits the result with `shellwords`.

#6842 landed the rest of this branch on `main`: the nine cases that no longer
run cosign, the non-empty assertion on every recorded signature and
certificate, and the certificate copy. `main` selects the shell inline:

    cmd := "sh"
    args := []string{"-c", "echo ${artifact}@${digest} > ${signature}"}
    if testlib.IsWindows() {
        cmd = "cmd.exe"
        args = []string{"/c", "echo ${artifact}@${digest} > ${signature}"}
    }

`testlib.ShC` already makes that choice for the rest of the test suite, so
this is the only place that repeats it. `shellwords.Parse` splits the string
back into a command and its arguments, the same way goreleaser splits a hook,
because `config.Sign` keeps `Cmd` and `Args` apart.

Behavior does not change. The signer still runs through a shell, because of
the `> ${signature}` redirect, and still runs `cmd.exe` on windows, where
`${signature}` is a native path whose backslashes `sh` reads as escapes.

### Verification

`test (ubuntu-latest)` and `test (windows-latest)` pass.


caarlos0:
Measured on CI, `internal/pipe/sign` package seconds against four green `main` runs:

| --- | --- | --- |
| ubuntu | 30.25 23.39 26.48 24.52 | **12.50** |
| windows | 28.68 22.31 23.49 20.85 | **11.76** |

Roughly half, on both jobs, with two cases still signing for real.

caarlos0:
Good call, though `Echo` itself does not fit: it has no shell in it, so `> ${signature}` would reach `echo` as a literal argument rather than redirecting.

`ShC` is the right decision, but it returns one command line and `config.Sign` keeps `Cmd` and `Args` apart. Splitting it with `strings.Fields`, the way `internal/shell` does with `Echo`, would break on the quoted argument.

So: added `testlib.ShCArgs`, which returns the same choice as the two pieces, and `ShC` now goes through it. Only one place in testlib names the shell, and the test no longer branches on `IsWindows` itself.

Verified the existing callers still pass: `internal/shell`, `internal/exec`, `internal/pipe/universalbinary`, `internal/pipe/build`.

caarlos0:
Variadic input would not do it, because the difference is in the output shape rather than the input.

`ShC` returns one string because `config.Hook.Cmd` is one string, which the pipes hand to `shellwords.Parse`. `config.Sign` keeps `Cmd` and `Args` apart and never parses, so it needs the pieces.

Turning `ShC` into the pieces pushes the join onto its six hook callers, and a plain join drops the quotes: `sh -c echo foo > bar` parses to `sh -c echo` plus three stray arguments. That leaves two helpers again, only swapped.

So I dropped `ShCArgs`. `ShC` is back to what it was on main, and the test splits it with the same `shellwords.Parse` the pipes use on a hook:

```go
signer, err := shellwords.Parse(testlib.ShC("echo ${artifact}@${digest} > ${signature}"))
require.NoError(t, err)
cmd, args := signer[0], signer[1:]
```

No new testlib surface, and the windows choice still lives in one place.

PR #6838, opened 2026-08-30
test: stop spawning docker to watch it fail slowly

caarlos0:
Two tests invoke the docker CLI only to observe a failure, and pay for it on
the windows job.

`TestCheckPath` ran `docker something-inalid` for the "tool is on PATH but
the command fails" case: 5.26s, the fifth slowest test in the cmd package on
windows. `go something-invalid` exercises the same branch, and does so
reliably -- `go` is always present when the tests run, whereas without docker
installed the LookPath branch answers first and the case proves nothing.

`TestIsDockerDaemonAvailableNoDaemon` pointed DOCKER_HOST at a missing unix
socket, which takes the windows docker CLI 5.46s to give up on.
`tcp://127.0.0.1:1` is refused immediately, and is honest about the platform:
a unix socket path is not what an unreachable daemon looks like on windows.

Measured locally: 5.26s -> 0.02s and 0.17s -> 0.16s (the unix socket case is
only slow on windows).

### DIFF
--- a/cmd/healthcheck_test.go
+++ b/cmd/healthcheck_test.go
@@ -36,7 +36,11 @@ func TestCheckPath(t *testing.T) {
 	checked := &sync.Map{}
 	require.NoError(t, checkPath(t.Context(), checked, "go"))
 	require.NoError(t, checkPath(t.Context(), checked, "git version"))
-	require.Error(t, checkPath(t.Context(), checked, "docker something-inalid"))
+	// `go` rather than `docker`: this case is about a tool that is on PATH but
+	// whose command fails, and docker is not guaranteed to be installed -- when
+	// it is not, the LookPath branch answers instead and the case proves
+	// nothing. It is also slow to refuse: 5.26s on the windows job.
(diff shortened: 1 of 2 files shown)


caarlos0:

| --- | --- | --- | --- |
| `cmd` | windows | 45.89 40.29 31.52 43.59 | 35.28 |
| `internal/pipe/docker/v2` | windows | 5.50 6.62 9.10 6.59 | 9.10 |
| `cmd` | ubuntu | 23.20 25.27 23.82 24.47 | 24.66 |
| `internal/pipe/docker/v2` | ubuntu | 11.69 11.09 11.59 12.93 | 11.99 |

Inconclusive at package level: both packages swing by more between `main` runs
than these two tests cost. The per-test numbers are the reason to take it --
from a `-json` run, `TestCheckPath` was 5.26s and
`TestIsDockerDaemonAvailableNoDaemon` 5.46s on windows, and neither needs
docker to say what it says.

`TestCheckPath` is also the more correct of the two afterwards: it was
depending on docker being installed to reach the branch it means to test.

PR #6842, opened 2026-08-30
fix(sign): warn when the signer writes no file, fail in v3

### LINKED ISSUE #6840: sign: a signature/certificate artifact is recorded even when the signer wrote no file
`signone` records a signature or certificate artifact based on the config being
set, not on the signer having produced the file:

```go
if cfg.Signature != "" {
    result = append(result, &artifact.Artifact{
        Type: artifact.Signature,
        Name: name,
        Path: env["signature"],
        ...
```

Nothing checks that `env["signature"]` exists. A signer that exits 0 without
writing the file gives the release an artifact pointing at nothing, which then
fails later at upload with a confusing error.

Found by adding a `require.FileExists` to `TestDockerSignArtifacts` (#6836).
Two cases fail with it:

1. `only_certificate` signs with `--key` and `--output-certificate`. cosign
   only issues a certificate when it signs keylessly through Fulcio, so with a
   key there is nothing to write. The case has been recording a `.pem`
   artifact whose file does not exist. It is excluded from the new assertion
   for now, with a pointer here.

2. On windows, the nine cases that sign with `sh -c 'echo ... > ${signature}'`
   were redirecting to `distfoo.sig`: `${signature}` is a native path,
   `dist\foo.sig`, and sh reads the backslash as an escape. `Artifacts.Add`
   runs the recorded path through `filepath.ToSlash`, so the artifact said
   `dist/foo.sig` and nothing noticed. Fixed in #6836 by using `cmd.exe` on
   windows, but goreleaser would have accepted the same thing from a user's
   signer.

A `stat` after the signer runs, failing with "the signer did not write
&lt;path&gt;", would catch both at the point where the cause is still obvious.

caarlos0:
Warn when a sign command does not write the file it was configured to write.

`signone` recorded a signature or certificate artifact from the config being
set, not from the signer having produced the file. The artifact then points at
nothing, and that only fails much later, while uploading, with an error that no
longer says which signer is at fault.

It now stats the file after the signer runs and warns with `the signer did not
write <path>: this will be an error in v3`. This covers `signs`,
`binary_signs`, and `docker_signs`, which share `signone`.

It is a warning, not an error, because `signature` has a default: a config that
signs the artifact in place, documented as supported since #2423, would fail
immediately. A `TODO(v3)` marks where to make it fail.

### It found real bugs in our own tests

1. `only certificate` signed with `--key` and `--output-certificate`. cosign
   only issues a certificate when it signs keylessly, so nothing was ever
   written. The case now copies a real certificate, and every recorded
   certificate must parse with `pem.Decode` plus `x509.ParseCertificate`.
2. the nine docker cases redirecting to `${signature}` also appended
   `&& cosign sign ... > ${signature}`. That second redirect truncated the echo
   with cosign's empty stdout, so every one of those signatures was empty,
   which the new non-empty assertion caught. Dropped; `no signature file` still
   runs cosign.
3. on windows those same cases wrote `distfoo.sig` next to `dist`, because
   `${signature}` is a native path and `sh` eats its backslashes. CI proved it
   once the test printed the listings, so they now use `cmd.exe`. This is
   #6840 case 2.
4. the two gpg certificate cases recorded a `.pem` that gpg cannot write. They
   now use a wrapper, and `testSign` asserts those files are in `dist`.

Docs: `sign.md` states the rule and that v3 will fail, and `binary_sign.md` and
`docker_sign.md` point at it.

Closes #6840

### Verification

`ok github.com/goreleaser/goreleaser/v2/internal/pipe/sign` on both
ubuntu-latest and windows-latest.

`lint` fails on `cmd/util_test.go:37 unparam`, which this branch does not
touch: it fails on `main` at edcfdec37 as well.


internal/pipe/sign/sign_docker_test.go
```diff
@@ -114,9 +114,12 @@ func TestDockerSignArtifacts(t *testing.T) {
...
 					Artifacts:   "all",
 					Stdin:       &password,
-					Cmd:         "cosign",
+					Cmd:         cmd,
 					Certificate: `{{ replace (replace (replace .Env.artifact "/" "-") ":" "-") "." "" }}.pem`,
-					Args:        []string{"sign", "--output-certificate=${certificate}", "--key=" + key, "--upload=false", "${artifact}@${digest}", "--yes"},
+					// cosign only issues a certificate when it signs
+					// keylessly, so `--output-certificate` would write nothing
+					// here: redirect its output to the certificate instead.
+					Args: []string{"-c", "cosign sign --key=" + key + " --upload=false ${artifact}@${digest} --yes > ${certificate}"},
```
caarlos0: **High — Windows CI fails on this case.**

Run [33328348825, job 99302370613](https://github.com/goreleaser/goreleaser/actions/runs/33328348825/job/99302370613) on `b449850`:

```
--- FAIL: TestDockerSignArtifacts/only_certificate (1.72s)
    Received unexpected error:
    the signer did not write dist\ghcrio-caarlos0-goreleaser-docker-manifest-actions-example-121-amd64.pem
```

`sh` exited 0, so the shell should have created the redirection target. Note that the other cases in this table use the same `sh` and the same `> ${signature}` shape with a `dist\...` path, and they **passed** in the same job. So this is not the generic backslash-eaten-by-`sh` problem from #6840, and swapping in `cmd.exe` is not obviously the fix.

That difference matters: until the mechanism is known, this is either a wrong test fixture **or** `checkWritten` returning a false failure for a file the signer really did write. The second would be a user-facing regression, and it is exactly the class of bug this PR exists to detect, so it should not be guessed at.

Smallest fix: replace the cosign-stdout redirect with a deterministic fake signer that writes the `.pem` itself, in the spirit of the new `testdata/gpg-and-certificate.sh`. That drops the shell redirect and the dependency on cosign stdout, and works on both platforms.

Test: whatever the fix, `TestDockerSignArtifacts/only_certificate` must pass on `windows-latest` before merge.

internal/pipe/sign/sign_docker_test.go
```diff
@@ -114,9 +114,12 @@ func TestDockerSignArtifacts(t *testing.T) {
 				{
 					Artifacts:   "all",
 					Stdin:       &password,
-					Cmd:         "cosign",
+					Cmd:         cmd,
```
caarlos0: **Medium — this case no longer tests a certificate.**

The case is named `only certificate`, but it now runs key-based `cosign sign` and redirects its **stdout** into a `.pem`. That output is not an issued certificate, and nothing parses it: the assertion at line 278 only compares artifact names, and production only stats the path. The case therefore passes whenever a shell can create a file, and would keep passing if certificate handling broke completely.

Your comment is right that `--output-certificate` writes nothing with `--key`. But the conclusion should be that this case cannot use cosign at all, not that cosign stdout becomes the certificate.

Smallest fix: use a fake signer that writes a fixed valid PEM (same approach as `testdata/gpg-and-certificate.sh`) and assert it parses with `pem.Decode` + `x509.ParseCertificate`. Otherwise rename the case so it stops claiming certificate coverage.

One fixture change can resolve this and the finding on lines 119-122.

www/content/customization/sign/sign.md
```diff
@@ -168,9 +168,14 @@ Please refer to [Docker Images Signing](/customization/sign/docker_sign/).
...
 
-You can sign with any command that either outputs a file or modify the file
+You can sign with any command that either writes a file or modifies the file
 being signed.
 
+If you point the command at `${signature}` — or at `${certificate}`, if you set
+one — that file must exist once the command finishes. GoReleaser checks it, and
+fails with `the signer did not write <path>` otherwise, as an artifact that
+points to nothing would only fail much later, while uploading it.
+
```
caarlos0: **Low — the rule is documented only for `signs`.**

`binary_signs` and `docker_signs` share `signone`, so they produce the same failure, but neither `binary_sign.md` nor `docker_sign.md` has a Limitations section or a link to this one. The docker page needs it most: `docker_signs` with `certificate` plus `--key` is exactly the config that turns from a late upload failure into an immediate `the signer did not write <path>`, and `docker_sign.md` documents `${certificate}` (line 95) without documenting the new requirement.

Smallest fix: add a one-line Limitations section to both pages pointing here.

caarlos0 (commented):
**Automated review (bot).** Changes requested — GitHub does not let me formally request changes on your own PR, so this is a comment.

The production change is one coherent concern and the new `TestSignerWroteNothing` subtests are decisive: both fail without `checkWritten`. `env["signature"]` and `env["certificate"]` are still the filesystem paths where they are checked — the `// re-execute template results` block only reassigns the local `name` and `cert` — so the check and the recorded `Path` agree. `docker_signs` defaults leave `Signature` and `Certificate` empty, so the check does not fire on a default docker config.

One blocker: **Windows CI fails on the single docker case this PR changed.** See the inline comment.

### Verification gaps

- `test (ubuntu-latest)` was **cancelled** by fail-fast, not failed. It never reached `internal/pipe/sign`, so there is no Linux result for `TestDockerSignArtifacts` on this branch.
- The `lint` failure (`cmd/util_test.go:37:27: setup - result 0 (string) is never used`) already fails on the merge base `5afe838`, and this PR does not touch that file. Not caused by this change.
- I did not run the suite locally, as instructed, and could not reproduce Windows behaviour on this host. The mechanism behind the Windows failure below is unresolved.

### Rejected suggestions

Both independent reviewers reported that `shell()` returning `cmd.exe /c` breaks `TestSignerWroteNothing/certificate`, because `;` is not a command separator in `cmd.exe`. **That is wrong.** The Windows job ran that test and it passed: `cmd.exe` ends the redirection filename token at `;`, so the `.sig` is created and the `.pem` is not, which is what the subtest asserts.

I also reject, as out of scope, that `os.Stat` accepts a stale file from an earlier run or a directory. Both are true, but both behave the same on `main`, so neither is a regression, and deleting the output before running the signer would add a risk this change has no evidence for.

internal/pipe/sign/sign_docker_test.go
caarlos0: Fixed in d8a1dcb95, by dropping the shell redirect from the fixture.

You are right that the backslash story does not explain it: I pulled the job log, and `only_certificate` was the only failure in the package, while the nine cases that redirect to `${signature}` with the same `sh` and the same `dist\...` path passed. I could not establish the mechanism from the log, so I did not guess at it.

I can rule out a false failure from `checkWritten`: the path it stats and the path it looks for in the command line are the same string, `env["certificate"]`, which is also the string the signer received. It fails only when the signer wrote somewhere other than where it was told to. Under any explanation of the windows behaviour, the file the signer was pointed at did not appear, which is the condition this PR reports.

The case now copies a certificate with `cp` (`cmd.exe /c copy` on windows), so there is no redirect and no dependency on cosign stdout. Watching windows CI before this is merged.

internal/pipe/sign/sign_docker_test.go
caarlos0: Agreed, and fixed in d8a1dcb95 by the same fixture change.

The case no longer uses cosign. It copies `testdata/cosign/certificate.pem`, a real self-signed certificate, to `${certificate}`, and `testWithArtifacts` now reads every recorded artifact and, for certificates, requires `pem.Decode` plus `x509.ParseCertificate` to succeed. Signatures are asserted non-empty as well.

So the case keeps its name honestly: it records certificates, and the recorded file is a certificate.

www/content/customization/sign/sign.md
caarlos0: Fixed in d8a1dcb95: both pages now have a Limitations section pointing at the one in `sign.md`, and `docker_sign.md` also states that cosign only issues a certificate when it signs keylessly, so `certificate` is of no use with `--key`.

internal/pipe/sign/sign_docker_test.go
caarlos0: Mechanism found, and it is the first of your two options: a wrong fixture, not a false failure.

I made the test print where the signer's files landed, and windows answered:

```
publish failed: the signer did not write dist\output_false_man1.sig
cwd D:\tmp\TestDockerSignArtifactssign_with_templated_output_false...\001:
  [.gitignore certificate.pem cosign.key cosign.pub dist distoutput_false_man1.sig]
dist: []
```

`sh` ate the backslash and wrote `distoutput_false_man1.sig` next to `dist`, which is exactly #6840 case 2. `checkWritten` reported it correctly.

Why it looked like one case at a time: the other cases appended `&& cosign sign ... > ${signature}`, and that second redirect truncated the echo with cosign's empty stdout. Ubuntu showed it once I asserted non-empty: `"output_false_man1.sig" is empty`. So the failure moved around with which case ran cosign, and the cosign tail also hid how consistently the redirect was misplacing files.

Fixes, in 4ead76e97 and c676399d9:
- the nine `${signature}` cases use `cmd.exe` on windows, and no longer redirect cosign's stdout into the signature. `no signature file` still runs cosign.
- `only certificate` copies a certificate, with no shell on unix at all.

Windows is green on f5507a460: `ok github.com/goreleaser/goreleaser/v2/internal/pipe/sign 5.877s` ([job](https://github.com/goreleaser/goreleaser/actions/runs/33332039796/job/99313757469)), ubuntu likewise.

PR #6843, opened 2026-08-30
fix(gitlab): own the retries, and honor replace_existing_artifacts

### LINKED ISSUE #6832: gitlab: release.replace_existing_artifacts never takes effect

caarlos0:
Hand the GitLab retries to `retryx`, and make `release.replace_existing_artifacts` take effect.

## Why

go-gitlab wraps its transport in `retryablehttp` with `RetryMax: 5` and `RetryWaitMin: 100ms`. The SDK absorbed every 429 and 5xx on its own budget and backoff, so `retryx` never saw a failure and the user's `retry` configuration did nothing for GitLab.

That also hid a dead branch. `release.replace_existing_artifacts` never took effect:

```go
releaseLink, resp, err := c.client.ReleaseLinks.CreateReleaseLink(projectID, releaseID, opt)
if err != nil {
    if resp != nil && resp.StatusCode == http.StatusBadRequest && releaseLink != nil {
```

`CreateReleaseLink` is implemented with the generic `do[*ReleaseLink]` helper, which returns the zero value on error. When GitLab answers `400 {"message":{"name":["has already been taken"]}}`, `releaseLink` is nil, the guard is false, and the upload fails instead of deleting and re-uploading the asset. #6528 replaced that 400 with a 503, which go-gitlab retried internally, so nobody noticed.

## What

- `newGitLab` adds `gitlab.WithoutRetries()`. `retryx` owns the retries, with the user's configuration.
- `gitlabError` replaces `retryx.HTTP` at every SDK call site. For a 429 it reads the `RateLimit-Reset` or `Retry-After` header into `HTTPError.RetryAfter`, which the retry layer honors. This keeps the rate-limit behavior the SDK gave, and mirrors `githubError`. The headers are read for a 429 only: `IsRetriable` treats any `RetryAfter > 0` as retriable, so reading them for every status would make each 4xx retriable.
- The version probe in `isV17` is retried too, since a transient 5xx there would silently pick `filepath` over `direct_asset_path`. It gets its own small budget (`versionRetry`: 3 attempts, 500ms, 2s max). The release budget defaults to 10 attempts with a 5m max delay, which would leave goreleaser looking wedged for ~25 minutes on a best-effort probe.
- `replaceReleaseLink` lists the release links, matches by name, deletes, and returns a retriable error. The failed create does not return the ID of the existing link, so the list is required. If the user did not allow replacement, or no link carries that name, the error is unrecoverable. Both failure paths join the create error, so the user still learns that the asset already exists.
- The documentation said the option is "Available only for GitHub". That was accidentally true while the branch was dead.

## Tests

- `TestGitLabUploadReleaseLinkExists` asserted an error where it meant to assert a replacement. It now covers replaces-it, replace-disabled, no-matching-link, and a failing list. Without the fix the first case fails with `400 has already been taken`; without the joined error the last one reports only `404 Not Found`.
- `TestGitLabUploadRetriesAreOurs` counts the requests. Without `WithoutRetries()` it sees 12 requests in 24.7s instead of the 2 attempts configured.
- `TestGitLabVersionProbeIsBounded` gives the release a 10-attempt/10s/5m budget and asserts the probe still stops after `versionRetry.Attempts`. Without its own budget it runs past a 30s test timeout.
- `TestGitLabRateLimitRetryAfter` and `TestGitLabErrorRateLimit` cover the header parsing, including a `Retry-After` HTTP date, which is unsupported on purpose because GitLab answers with delta-seconds.
- `TestGitLabURLsDownloadTemplate` no longer makes the first request fail. It tests URL templates only: 6.60s -> 0.04s.
- Five GitLab tests answered the version probe with an empty 200 body, which decodes to `io.EOF` and is retriable. `serveGitLabVersion` answers them properly, so the bounded probe costs them nothing: 5.8s -> 2.7s.

Closes #6832
Closes #6833


internal/client/gitlab.go
```diff
@@ -91,14 +127,16 @@ func newGitLab(ctx *context.Context, token string, opts ...gitlab.ClientOptionFu
...
+		isV17OrLater: isV17(ctx, client),
 	}, nil
 }
 
-func isV17(client *gitlab.Client) bool {
+func isV17(ctx *context.Context, client *gitlab.Client) bool {
 	v := os.Getenv("CI_SERVER_VERSION")
 	if v == "" {
-		gitlabVersion, _, err := client.Version.GetVersion(nil)
+		gitlabVersion, _, err := gitlabDo(ctx, func() (*gitlab.Version, *gitlab.Response, error) {
```
caarlos0: **The version probe now uses the full release retry budget, but it discards the error.**

**Trigger.** Run `goreleaser release` against GitLab from outside GitLab CI, so `CI_SERVER_VERSION` is empty. Make `/api/v4/version` answer 5xx, or make the host refuse the connection. `retryx.IsNetworkError` matches `connection refused`, and a 5xx is retriable.

**Impact.** With the defaults from `internal/pipe/defaults/defaults.go:40-42` (10 attempts, 10s delay, 5m max delay), `BackOffDelay` gives delays of 10, 20, 40, 80, 160, 300, 300, 300, and 300 seconds. That is about **25 minutes**. Before this change the SDK made 6 requests in about 12 seconds (`RetryMax: 5`, `LinearJitterBackoff(700ms, 900ms)`). goreleaser looks wedged for 25 minutes. Then it writes a warning, throws the error away, and continues as if GitLab is older than 17. The default `--timeout` is 1 hour (`cmd/release.go:82`), so the probe can use almost half of it. `client.New` does not cache (`internal/client/client.go:122-153`), and there are many construction sites, so the cost repeats.

The reason to retry the probe is correct: a transient 5xx must not silently select `filepath`. But a best-effort probe must not spend the budget of the whole release.

**Smallest fix.** Give the probe its own small budget, for example `retryx.Do(ctx, config.Retry{Attempts: 3, Delay: time.Second, MaxDelay: 5 * time.Second}, ...)`.

**Test.** A server that always answers 503 on `/api/v4/version`, with `Retry{Attempts: 10, Delay: 10 * time.Second}`. Assert that `newGitLab` returns inside a bounded window, and that `isV17OrLater` is false. `internal/retryx/retryx_test.go` already uses `testing/synctest` for fake time.

internal/client/gitlab.go
```diff
@@ -685,6 +710,72 @@ func (c *gitlabClient) Upload(
...
+// happens again.
+//
+// The failed creation does not return the ID of the existing link, so it has to
+// be found in the release link list first.
+func (c *gitlabClient) replaceReleaseLink(
+	ctx *context.Context,
+	projectID, releaseID, name string,
+	createErr error,
+) error {
+	if !ctx.Config.Release.ReplaceExistingArtifacts {
```
caarlos0, line 738: **This line makes the documentation false.**

`www/content/customization/publish/scm/_index.md:57-66` says:

```
# Whether to remove an artifact that already exists.
#
# Available only for GitHub.
# ... when the upload of an artifact fails with a 422 ...
```

I anchor this to the changed line, because this line is the code that the documentation describes. Before this pull request the sentence was accidentally true, because the GitLab branch was dead. Now the option works on GitLab, and the sentence is wrong.

**Impact.** A GitLab user reads that page, sees "Available only for GitHub", and does not set the flag. The repair of #6832 stays invisible.

**Smallest fix.** Remove "Available only for GitHub". Say that GitLab uses a 400 response, and that goreleaser lists the release links, deletes the link with the same name, and then uploads again.

**Test.** None. This is documentation.

internal/client/gitlab.go
```diff
@@ -685,6 +710,72 @@ func (c *gitlabClient) Upload(
...
+		return retryx.Unrecoverable(createErr)
+	}
+
+	link, err := c.getReleaseLinkByName(projectID, releaseID, name)
+	if err != nil {
+		return err
```
caarlos0: **A failure of the list hides the original create error.**

This `return err`, and the `gitlabError(err, resp)` for the delete below, both discard `createErr`.

**Trigger.** The create fails with `400 has already been taken`. Then `ListReleaseLinks` answers 403 or 404, for example because the token can write links but cannot list them.

**Impact.** The user sees only `404 Not Found`. He never learns that the asset already exists, and that `replace_existing_artifacts` is the subject. A 500 from the list is retriable, so the whole upload repeats, and the reported error is the 500.

**Smallest fix.** Join the two errors: `return errors.Join(createErr, err)` at both sites.

**Test.** A subtest where `GET .../assets/links` answers 404. Assert that the error text contains both `has already been taken` and the 404.

internal/client/gitlab.go
```diff
@@ -43,14 +45,44 @@ func gitlabDo[T any](ctx *context.Context, fn func() (T, *gitlab.Response, error
...
+			he.RetryAfter = rateLimitRetryAfter(r.Header)
+		}
+	}
+	return he
+}
+
+// rateLimitRetryAfter returns how long to wait before retrying a rate-limited
+// request, from either the RateLimit-Reset (unix timestamp) or the Retry-After
+// (seconds) header. It returns 0 when neither header gives a usable value.
+func rateLimitRetryAfter(header http.Header) time.Duration {
```
caarlos0, line 74: **No test covers this function.**

`grep` for `429`, `RetryAfter`, and `rateLimitRetryAfter` in `internal/client/*_test.go` finds nothing. This is a new function that parses raw headers, so it is more fragile than `githubError`, which uses typed SDK errors.

**Impact.** A wrong header name or a wrong unit is silent. The 429 stays retriable through `Status == 429`, so the only symptom is a slower release. Nothing fails, and nobody notices.

**Smallest fix.** None in the source. The function is correct: I confirmed that `Header.Get("RateLimit-Reset")` and the SDK's `Ratelimit-Reset` have the same canonical key, and that the SDK also reads the value as a unix timestamp (`client-go@v1.46.0/gitlab.go:775-778`).

**Test.** A table test on `rateLimitRetryAfter`: a `RateLimit-Reset` in the future; a `RateLimit-Reset` in the past together with `Retry-After: 30`; a `Retry-After` as an HTTP date; and no headers. Note that this function has no lower limit, but `githubError` uses `max(..., time.Second)`.

caarlos0 (commented):
Review by an AI agent (GitHub Copilot CLI), at the request of @caarlos0.

**Verdict: the change is correct, and the scope is clean. Two items to correct before you merge.**

I verified the premise in the SDK, and it is true. `do[T]` returns the zero value on error (`client-go@v1.46.0/request_handler.go:270-272`), so `releaseLink` was always nil on a 400. The old guard was dead. `WithoutRetries()` disables only `retryHTTPCheck` (`client-go@v1.46.0/gitlab.go:654`), so the client-side rate limiter stays active. `Ratelimit-Reset` is a unix timestamp in the SDK too (`client-go@v1.46.0/gitlab.go:775-778`), so `rateLimitRetryAfter` agrees with it. The retry budget does not decrease: `internal/pipe/defaults/defaults.go:40-42` gives 10 attempts, and the SDK gave 5.

The new tests are good. `TestGitLabUploadReleaseLinkExists/replaces_it` fails against the old code, because the old code returns a non-retriable 400. `TestGitLabUploadRetriesAreOurs` is deterministic: `Delay` is 0, so `BackOffDelay` gives 1ns.

The findings are in the inline comments.

**Verification gaps.** I did not build or test locally, because you told me to assume that the build, the lint, and the format are correct. Two independent agents ran `go test -run TestGitLab ./internal/client/ -count=1`, and both report `ok` in about 1.2s. Nobody tested against a real GitLab instance. The retry-budget numbers come from `defaults.go`, from retry-go's `BackOffDelay`, and from a measurement that replicates `retryx.opts`.

internal/client/gitlab.go
caarlos0: Confirmed and fixed in 1187debce.

`internal/pipe/defaults/defaults.go:40-42` is exactly as you say: 10 attempts, 10s delay, 5m max delay. The probe is best effort, so it now has its own budget:

```go
var versionRetry = config.Retry{
	Attempts: 3,
	Delay:    500 * time.Millisecond,
	MaxDelay: 2 * time.Second,
}
```

That is ~1.5s worst case, close to the ~4s the SDK spent before (`RetryMax: 5`, `LinearJitterBackoff(700ms, 900ms)`), and far from 25 minutes.

`TestGitLabVersionProbeIsBounded` gives the release a 10-attempt/10s/5m budget and asserts the probe still stops after `versionRetry.Attempts`. With `ctx.Config.Retry` restored it runs past a 30s test timeout; with the fix it takes 1.50s.

I did not use `synctest`: the bubble cannot contain the `httptest` server, so `synctest.Test` deadlocks on real network I/O. The test uses real time instead, which the small budget makes affordable.

I left the discarded error alone. Returning it would make `newGitLab` fail the whole release on a probe failure, which is a larger behavior change than this pull request should make, and the fallback to "older than v17" is the pre-existing intent.

One thing your report did not cover, which the fix exposed: five existing GitLab tests answered the probe with an empty 200 body. That decodes to `io.EOF`, which `retryx.IsNetworkError` matches, so the bounded probe made each of them 1.5s slower. `serveGitLabVersion` now answers them properly. The package went 5.8s -> 2.7s.

internal/client/gitlab.go
caarlos0: Correct, and fixed in 1187debce.

`www/content/customization/publish/scm/_index.md` now reads:

```
  # Whether to remove an artifact that already exists.
  #
  # This might be a bit expensive (rate-limiting speaking), so it is only done
  # when the upload of an artifact fails because it already exists in the
  # release: a 422 on GitHub, and a 400 on GitLab.
  # We then grab the list of artifacts from the release, and delete the file
  # that matches the one we are trying to upload.
  # GoReleaser will then retry its upload.
  #
  # Available for GitHub and GitLab.
```

I checked the other places that describe the option. `pkg/config/config.go:694` carries no comment, and no other page under `www/` mentions `replace_existing_artifacts`, so this was the only false sentence.

internal/client/gitlab.go
caarlos0: Confirmed and fixed in 1187debce. Both sites now use `errors.Join(createErr, err)`.

I checked that this keeps the retry classification. `errors.AsType` does a depth-first traversal of `Unwrap() []error` (`src/errors/wrap.go:187-199`), so `retryx.IsRetriable` still finds the `HTTPError` inside the join. A 500 from the list stays retriable, and a 404 stays terminal.

`TestGitLabUploadReleaseLinkExists` gets a "listing the links fails" case where `GET .../assets/links` answers 404, asserting the error contains both `has already been taken` and `404`. With `return err` restored it fails with exactly the symptom you described:

```
Error "404 Not Found" does not contain "has already been taken"
```

internal/client/gitlab.go
caarlos0: Agreed, and covered in 1187debce.

`TestGitLabRateLimitRetryAfter` is the table you asked for: a `RateLimit-Reset` in the future, a `RateLimit-Reset` in the past falling back to `Retry-After: 30`, a `Retry-After` alone, a `Retry-After` as an HTTP date, unparsable values, and no headers at all.

The HTTP-date case asserts 0, and says why: GitLab answers with delta-seconds, an HTTP date is valid per RFC 9110 but unsupported, and the retry layer then backs off on its own. This is the pre-existing behavior of go-gitlab too, which never reads `Retry-After`.

I added `TestGitLabErrorRateLimit` as well, since the header parsing is only reachable through `gitlabError`. It pins the part that a table on the parser alone would miss: the headers are read for a 429 only. A 400 that also carries `Retry-After` must keep `RetryAfter` zero, otherwise every 4xx would become retriable, because `IsRetriable` returns true for any `RetryAfter > 0`.

On the missing lower limit: I left it. `githubError` needs `max(..., time.Second)` because `RateLimitError.Rate.Reset` can already be in the past, which yields a negative duration. Here a past reset falls through to the `Retry-After` branch and then to 0, so the value is never negative, and `retry-go` applies `MaxDelay` on top. A sub-second wait only costs one extra request against a server that just told us its window is nearly over.

Issue #6839, opened 2026-08-30
internal/builders/node: TestBuild is 14% of the ubuntu test job, almost all race-detector tax


caarlos0:
Measured with `go test -json` on both CI jobs (run 33311791959):

| job | node.TestBuild | package | job total |
| --- | --- | --- | --- |
| ubuntu (`-race`) | 73.53s | 74.88s | 537s |
| windows (no `-race`) | 16.30s | 16.45s | 422s |

The 4.5x gap is the race detector, and it is doing no work: the test builds one
target, single goroutine. What it instruments is 136MB of byte pushing.

Measured locally (M-series, darwin-arm64 dist, v26 = 136MB extracted):

| step | plain | `-race` |
| --- | --- | --- |
| `ensureNode` (download + gzip inflate) | 2.26s | 14.52s |
| `signMachO` on a 53MB Mach-O | 0.46s | 6.14s |

At 136MB the signing figure scales to roughly 15s, so the two byte-heavy steps
account for about 40s of the 73s, before allowing for the runner being slower
than a laptop.

Both costs are inherent to using a real Node.js dist:

- `extractFromTarGz` inflates the whole 136MB binary out of the tarball.
- `signMachO` hashes the whole 136MB Mach-O.

Options, none of them free:

1. Leave it. The builder keeps its only end-to-end test on every job.
2. Point `TestBuild` at `win-x64`. That dist is a bare `node.exe`, so there is
   no gzip inflate, and no darwin means no quill: `ensureNode` under `-race`
   measured 6.02s instead of 14.52s. It stops covering the tar.gz path and the
   Mach-O signing path, which are the two most failure-prone parts, and
   `signMachO`'s doc comment describes a node-specific quirk (a placeholder
   `LC_CODE_SIGNATURE` at end-of-file) that a cross-built Go binary would not
   reproduce in a unit test.
3. Skip the darwin case when the race detector is on. Coverage stays on the
   windows job every run and ubuntu saves the whole 73s, but the ubuntu job
   stops covering the builder end to end, and a local `task test` would skip it
   too, since `RACE` defaults to true.

Filing rather than picking: options 2 and 3 both trade real coverage of the
newest builder for about 14% of one job.

Unrelated but noticed while measuring: `downloadHostBinary` creates a fresh
`os.MkdirTemp` per call and never removes it, so a six-target node release
leaves about 800MB behind and re-downloads all of it on the next run.

caarlos0:
[...]

PR #6844, opened 2026-08-30
perf(node): faster dist extraction, and stop leaking it into TMPDIR

caarlos0:
Two independent fixes to the node builder, both found while measuring #6839.

Make node dist extraction ~1.4x faster (1.8x under `-race`), and stop leaking
the downloaded dist into `TMPDIR`.

### 1. `perf(node): use klauspost's inflate for node dists`

Inflating the node dist is the single most expensive step of a node build, and
it is pure per-byte work in `compress/flate`. Measured on a 58MB `tar.gz`
holding a 146MB darwin-arm64 binary (M3 Pro, each figure a fresh process,
because TSan gets slower the longer one runs):

| impl | plain | `-race` |
| --- | --- | --- |
| `compress/gzip` | 0.76s | 19.6s |
| `klauspost/compress/gzip` | 0.55s | 10.9s |

`klauspost/compress` is already a direct dependency (`pkg/archive/tarzst`), and
`pkg/archive/gzip` already prefers `klauspost/pgzip`, so this adds no module.
`pgzip` measured worse than klauspost here (11.4s vs 8.0s in the same process):
its readahead goroutine adds synchronization the race detector has to track.

Larger buffers were measured and rejected. The `-race` cpuprofile blames
`syscall.rawsyscalln` for 62% of the run, under `flate.huffSym` ->
`bufio.ReadByte` -> `fill` -> `os.File.Read`, which reads like gzip's internal
4KB `bufio.Reader`. It is a misattribution — a 1MB or 4MB `bufio.Reader` around
the file, and a 1MB `bufio.Writer` around the destination, moved nothing:

```
BenchmarkExtract/current-12          13815140292 ns/op
BenchmarkExtract/read1M-12           13704658000 ns/op
BenchmarkExtract/read1M_write1M-12   13894214959 ns/op
BenchmarkExtract/read4M_write1M-12   14474876959 ns/op
```

### 2. `fix(node): stop leaking the downloaded node dist`

`downloadHostBinary` and `nodedist.Download` each created an `os.MkdirTemp`
that nothing ever removed. Per darwin target that is 58MB of archive plus a
146MB extracted binary, so a six-target release left roughly **1.2GB** behind
in `TMPDIR` and re-downloaded all of it on the next run.

- `nodedist.Download` holds a single file, so `os.CreateTemp` is enough.
- `downloadHostBinary` now removes the archive on every path, not just on the
  `tar.gz` success path.
- The extracted host binary has to outlive `downloadHostBinary`, since
  `node --build-sea` reads it, so ownership moves to `Build`, which already
  knows when the target is done with it.

Not covered by a test on purpose: a stranded temp dir is wasted disk, not a
broken build, and asserting on it needs `TMPDIR` scoping plus a prefix glob to
avoid tripping over what `node --build-sea` writes there itself. Not worth the
machinery.

### CI

`internal/builders/node` took **57.1s** on `test (ubuntu-latest)`, against the
74.88s recorded in #6839. Treat that as one sample, not a benchmark: GitHub
runners vary enough that the gap is wider than the ~9s the local measurements
predict.

Two failures on the way here came from `main`, not from this branch, and both
are fixed at the source rather than worked around: the `unparam` lint break
from #6831 (fixed in #6845) and a clock race in
`TestGitLabRateLimitRetryAfter` from #6843 (fixed in #6846). With both merged,
every check is green.

### Not in this PR

Profiling `quill.Sign` on the same 146MB Mach-O (2.58s plain, 28.5s under
`-race`) showed `debug/macho.parseSymtab` at 4.23s, with
`macho.(*File).refresh` called from both `macho.NewFile` and `macho.Patch` —
so node's symbol table is fully parsed more than once per signature, and quill
never looks at symbols to sign. That belongs in goreleaser/quill and would help
every macOS SEA build, not only the test.

Refs #6839, where the full measurements live. This does not close it — the
remaining cost is quill and the race detector itself
(`runtime._ExternalCode` is 66% of the signing profile), so the coverage
trade-off in that issue is still open.


internal/builders/node/build_test.go
```diff
@@ -96,6 +98,55 @@ func TestResolveVersionStringRejectsUnsupportedSEARelease(t *testing.T) {
 	require.ErrorContains(t, err, ">= v25.5.0")
 }
 
+func TestExtractFromTarGz(t *testing.T) {
+	const entry = "node-v25.5.0-linux-x64/bin/node"
+	payload := "#!/bin/sh\nexec node \"$@\"\n"
```
caarlos0: `writeTarGz` produces about 3KiB here: two 512-byte headers, two 512-byte padded bodies, and the 1KiB zero trailer. deflate's history window is 32KiB, so this fixture never crosses a window boundary or a multi-block boundary — the only places where a replacement inflate can plausibly diverge from the stdlib.

**Trigger:** revert the import to `compress/gzip`, or supply an inflate with broken window handling, and this test stays green.
**Impact:** the perf commit ships effectively uncovered. The test proves the two packages agree on a stream small enough that no interesting decoder state exists.
**Smallest fix:** make the fixture larger than the window and compressible:

```go
payload := strings.Repeat("#!/bin/sh\nexec node \"$@\"\n", 2000) // ~48KiB
```

Both subtests keep working unchanged; `require.Equal` already compares the full body.

internal/builders/node/build.go
```diff
@@ -147,7 +147,13 @@ func (b *Builder) Build(ctx *context.Context, build config.Build, options api.Op
 		return err
 	}
 
-	targetNode, err := ensureNode(ctx, build.Dir, target.Target)
+	targetNodeDir, err := os.MkdirTemp("", "goreleaser-node-*")
+	if err != nil {
+		return err
+	}
+	defer func() { _ = os.RemoveAll(targetNodeDir) }()
```
caarlos0, line 154: No test fails if this `defer`, or the one added at `internal/builders/node/download.go:153`, is deleted. `TestExtractFromTarGz` never reaches `Build`, `ensureNode`, or `nodedist.Download`, so the headline fix of this PR ships uncovered and can regress silently.

`os.TempDir()` reads `$TMPDIR` on every call, so a scoped test is cheap:

```go
destDir := t.TempDir()
tmpRoot := t.TempDir()
t.Setenv("TMPDIR", tmpRoot) // plus TMP and TEMP on windows
// stub server + nodedist.StubRelease / SetBaseURL for the archive
bin, err := downloadHostBinary(t.Context(), "v25.5.0", "linux-x64", destDir)
require.NoError(t, err)
require.FileExists(t, bin)
left, err := os.ReadDir(tmpRoot)
require.NoError(t, err)
require.Empty(t, left)
```

That fails on the pre-fix code with two leftover entries.

One blocker: `SetBaseURL`, `NewServer`, and `StubRelease` live in `internal/nodedist/testing_test.go`, whose doc comment says they are "including tests living in other packages". A `_test.go` file cannot export to another package, so that claim is false today. Moving them to `internal/nodedist/nodedisttest` unblocks the test and makes the comment true.

If that move is out of scope here, `TestBuild` already runs the whole path: point `TMPDIR` at a `t.TempDir()` there and assert that no `goreleaser-node*` entry survives `Default.Build`.

caarlos0 (commented):
**Comment — no blocking defect found.**

I traced every return path in `Build`, `ensureNode`, `downloadHostBinary`, and `nodedist.Download`. The archive and the extracted binary are removed on every path, and nothing is removed while it is still in use: `targetNode` is consumed only by `createSEAConfig` and by the synchronous `node --build-sea`. `Build` runs concurrently per target (`AllowConcurrentBuilds` returns true, and `internal/pipe/build/build.go` starts one goroutine per target), but `targetNodeDir` is a per-invocation `os.MkdirTemp`, so two targets can never share `bin`. On the windows path the deferred `os.Remove(archive)` runs after a successful `os.Rename`, gets ENOENT, and is discarded — it can never touch `bin`. `nodedist.Download` has exactly one caller in the tree, and that caller owns the file.

The gzip swap is behaviorally identical. `klauspost/compress/gzip` imports `compress/gzip` and aliases `ErrChecksum` and `ErrHeader`, keeps `multistream` on by default, and returns `io.ErrUnexpectedEOF` on truncation. Nothing in the tree compares gzip sentinels anyway, and `nodedist.Download` verifies the SHA-256 before extraction. Module and binary cost is zero: `klauspost/compress` is already a direct require, and `pkg/archive/gzip` already links klauspost's inflate through `pgzip`.

The two commits are separable and correctly scoped. All three comments below are about test coverage.

One accepted consequence, not a request: `dist/<target>/sea-config.json` survives the build with `executable` pointing at the directory this PR now removes. It pointed at a leaked temp directory before, so this only makes an already-unsupported manual re-run fail deterministically.

**Verification gaps, disclosed.** At the author's request I ran no build, test, lint, or fmt. The performance numbers were not reproduced — they need a 58MB download — and no benchmark is committed, so they are unverified rather than doubted; the mechanism is plausible and I found no dimension in which the swap regresses.

Reviewed by an automated reviewer (GitHub Copilot CLI).

internal/nodedist/download.go
```diff
@@ -15,8 +15,8 @@ import (
 // Download fetches https://nodejs.org/dist/<version>/<archiveName>
 // into a fresh temp file, verifies its SHA-256 against the embedded
 // release index, and returns the local file path. Each call hits the
-// network — there is no on-disk cache. The temp file lives under
-// os.TempDir() and is reaped by the OS on its usual schedule.
+// network — there is no on-disk cache. The file belongs to the
+// caller, who must remove it.
```
caarlos0, line 19: Nit. This new contract has two in-tree callers that do not honor it: `download_test.go:25` and `download_test.go:74` keep the returned path and never remove it, so `go test ./internal/nodedist/...` now leaves files in `TMPDIR`. One line each keeps the documented ownership honest:

```go
t.Cleanup(func() { _ = os.Remove(got) })
```

internal/builders/node/build_test.go
caarlos0: Agreed, fixed in 60efe0878.

The entry is now `strings.Repeat("#!/bin/sh\nexec node \"$@\"\n", 2000)` = 50,000 bytes, so it crosses the 32KiB window and spans more than one block. Both subtests are unchanged.

internal/nodedist/download.go
caarlos0: Fixed in 60efe0878, though not with `os.Remove`.

Both call sites now scope `os.TempDir()` to a `t.TempDir()` via a new `testlib.ScopeTempDir`, so the downloads die with the test and nothing is left to remove by hand. `TestDownload` reuses that scoping for a real assertion — see the other thread.

internal/builders/node/build.go
caarlos0: You are right that it shipped uncovered. Fixed in 60efe0878, taking your second option.

`testlib.ScopeTempDir(tb)` points `os.TempDir()` at a `t.TempDir()` (`TMPDIR`, plus `TMP`/`TEMP` for windows). Two assertions now sit on the fix:

- `TestBuild` scopes it and asserts no `goreleaser-node*` entry survives `Default.Build`. That covers both defers, since `downloadHostBinary` runs inside `Build`. On the pre-fix code this is the two leftover entries you predicted.
- `TestDownload` scopes it and asserts `nodedist.Download` leaves exactly one entry and that it is not a directory. **Verified it fails on the pre-fix code**: restoring the old `MkdirTemp`+`CreateTemp(dir, ...)` gives `download_test.go:39: Should be false`.

Two deliberate departures from your sketch:

1. I matched `goreleaser-node*` by prefix instead of asserting the scoped dir is empty. `node --build-sea` inherits `TMPDIR` and may write there itself, so `require.Empty` on the whole directory would be a flake waiting to happen on CI.
2. I did not move the helpers to `internal/nodedist/nodedisttest`. It is a bigger change than it looks: `nodedisttest` would have to import `nodedist`, so `download_test.go` could no longer be `package nodedist` without an import cycle, which drags in an `export_test.go` shim for `defaultRetry` and a new exported `SetBaseURL` hook on the production package. That is a lot of surface to buy coverage that `TestBuild` already gives, and #6839 is an argument for less machinery here, not more.

The false doc comment is gone either way — `SetBaseURL` no longer claims to serve other packages.

If you would rather have the hermetic `downloadHostBinary` test regardless, say so and I will do the move as its own PR.

caarlos0: Reversing my previous reply: the leftover assertions are gone again as of 0269cf385, and the `defer` here stays untested on purpose.

Carlos's call, and I agree with it. A stranded temp dir is wasted disk, not a broken build, so the assertion was not worth what it cost: a new `testlib.ScopeTempDir` helper, `TMPDIR`/`TMP`/`TEMP` scoping, and a prefix glob rather than a plain empty-dir check, because `node --build-sea` inherits `TMPDIR` and writes there itself. That is a lot of machinery guarding a disk-space regression.

So this thread stands as originally written — the fix is uncovered — but deliberately now rather than by oversight. The `nodedist` tests keep `t.Cleanup(func() { _ = os.Remove(got) })` per the other thread, which is the cheap half and still honors the new ownership contract.

The false `SetBaseURL` doc comment stays fixed, and the widened `extractFromTarGz` fixture is unaffected.

internal/builders/node/build_test.go
caarlos0: Reverting this one in a1259b321, and I think the original review was aiming at the wrong target — mine included, since I agreed without pushing back.

Crossing the 32KiB window exercises klauspost's decompressor, not our code. They have tests and fuzzing for that, and a single `strings.Repeat` string is a weak canary for it anyway: it compresses to almost nothing and touches a narrow corner of the LZ77/Huffman state space. If klauspost's inflate broke on a window boundary, this test is not what should catch it.

`extractFromTarGz` only does three things of its own: pick one entry out of several, write it to `dst`, and error when it is absent. The payload is small again, and the fixture is now an ordered slice rather than a map.

That last part fixed a real hole the big fixture had masked. Map iteration order is random in Go, so `writeTarGz` emitted the two entries in random order and the wanted one was first about half the time — meaning the skip loop went unexercised on those runs. It is now deterministically third of four.

Mutation-checked both branches:

- dropping the `h.Name != entry` guard fails `requested entry` with `Not equal`
- returning `nil` in place of the not-found error fails `missing entry` with `An error is expected but got nil`

caarlos0: Test dropped entirely in f89095e86; `build_test.go` is now identical to main.

Coverage on `TestBuild` alone showed it already reaches the entry skipping, the copy and the rename against a real dist. The only uncovered branch was the not-found error, which is not worth a fixture and a helper.

PR #6850, opened 2026-08-30
perf(git): ask git once

caarlos0:
`getInfo` asked git the same question several times.

- `git.IsRepo` ran twice in a row, because the snapshot branch and the error
  branch each called it.
- Three `git show HEAD --quiet` differed only in `--format`: `%ct`, `%h` and
  `%H`. One `%h%n%H%n%ct` gives all three.
- Three `git tag -l` for the same tag differed only in the format:
  `%(contents:subject)`, `%(contents)` and `%(contents:body)`. One format with
  the three separated by `%00` gives all three, and a tag message cannot
  contain NUL.

That is five fewer processes out of the fourteen a pipeline run spends here.
Process creation is expensive on windows, and every `goreleaser` invocation
pays it.

Counted with a `git` wrapper on PATH that logs each call:

    internal/pipe/git   556 -> 423 spawns
    cmd                 354 -> 319 spawns
    whole suite        1435 -> 1267 spawns

The behaviour is unchanged. Two error messages are gone, since the three tag
reads and the two commit reads are now one call each: "couldn't get tag
subject" and "couldn't get tag content body" fold into "couldn't get tag
contents", and "couldn't get commit date" folds into "couldn't get current
commit".

`TestSingleCommit` now pins what the merged formats could silently swap.
Verified by mutation: `%h%n%H` -> `%H%n%h` fails it, `%ct` -> `%at` fails it,
and reordering the tag format fails `TestAnnotatedTags` and
`TestAnnotatedTagsWithApostrophes`. None of the three failed anything before.

### DIFF
--- a/internal/git/config.go
+++ b/internal/git/config.go
@@ -14,11 +14,15 @@ import (
 
 // ExtractRepoFromConfig gets the repo name from the Git config.
 func ExtractRepoFromConfig(ctx context.Context) (result config.Repo, err error) {
-	if !IsRepo(ctx) {
-		return result, errors.New("current folder is not a git repository")
-	}
 	out, err := Clean(Run(ctx, "ls-remote", "--get-url"))
 	if err != nil {
+		// git says "No remote configured to list refs from" for both a
+		// repository without a remote and a directory that is not a
+		// repository at all, so it takes a second command to tell the user
+		// which one it is. Only the failing path pays for it.
+		if !IsRepo(ctx) {
+			return result, errors.New("current folder is not a git repository")
+		}
 		return result, errors.New("no remote configured to list refs from")
 	}
 	// This is a relative remote URL and requires some additional processing
--- a/internal/pipe/git/git.go
+++ b/internal/pipe/git/git.go
@@ -70,11 +70,11 @@ var fakeInfo = context.GitInfo{
 }
 
 func getInfo(ctx *context.Context) (context.GitInfo, error) {
-	if !git.IsRepo(ctx) && ctx.Snapshot {
-		log.Warn("accepting to run without a git repository because this is a snapshot")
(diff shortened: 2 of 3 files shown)


caarlos0:
CI measurement, package seconds against four green `main` runs:

| --- | --- | --- | --- |
| `internal/pipe/git` | windows | 24.46 31.36 30.28 31.81 | 24.49 |
| `cmd` | windows | 21.83 26.09 25.38 25.62 | 23.62 |

Inconclusive at n=1: the baseline minimum is 24.46, so landing at 24.49 is inside the spread rather than below it.

The deterministic metric is the process count, which does not depend on the runner. Counted with a `git` wrapper on PATH that logs every call:

```
internal/pipe/git   556 -> 423 spawns   (-24%)
cmd                 354 -> 319 spawns   (-10%)
whole suite        1435 -> 1267 spawns  (-12%)
```

Five of the fourteen processes a pipeline run spent in `getInfo` are gone. On windows each git spawn measures around 70ms, which is where the 24-32s spread for that package comes from in the first place.

caarlos0:
Added a second commit here: `ExtractRepoFromConfig` ran `git rev-parse --is-inside-work-tree` before every `git ls-remote --get-url`. The check only exists to pick between two error messages -- git says "No remote configured to list refs from" for both a repository without a remote and a directory that is not one -- so it now runs only when `ls-remote` fails. Same two messages, one process instead of two on the happy path.

Both messages are asserted in `internal/git`, `internal/pipe/milestone` and `internal/pipe/release`, and all three still pass.

`cmd` goes from 319 to 280 spawns with it, on top of the 354 it started at.

The combined effect of this branch and #6851, counted over `go test ./...`: **1757 -> 1393 git spawns, -21%**.

PR #6851, opened 2026-08-31
perf(client): configure the repository while cloning it

caarlos0:
The git upload client ran one `git config --local` per setting right after
cloning: user.name, user.email, init.defaultBranch and commit.gpgSign, plus
user.signingKey, gpg.program and gpg.format when signing is on. That is four
to seven processes, every time goreleaser publishes a brew formula, a cask, a
scoop manifest, a krew plugin, an AUR package, a nix derivation or a winget
manifest.

`git clone -c <key>=<value>` writes the setting straight into the new
repository's config, so the clone that already runs can carry all of them.

Counted with a `git` wrapper on PATH that logs each call, across
internal/pipe/aur and internal/pipe/aursources: 290 -> 202 spawns. The two
packages take 15-20s each on the windows job against 2-3s on ubuntu, which is
the shape of a package that only spends its time creating processes.

`TestGitClient/full` now reads the commit author back out of the bare
repository. Verified by mutation: dropping user.email or commit.gpgSign from
the list already failed a test, dropping user.name failed nothing, and now
fails `TestGitClient/full`.


caarlos0:
Stacked on #6850, so its two commits show up in this diff as well. Review or merge #6850 first; this branch is `internal/client/git.go` only.

caarlos0:
Suite-wide count, measured with a `git` wrapper on PATH that logs every call,
`go test ./... -count=1` on this branch against `origin/main`:

```
main         1757 git spawns
this branch  1393 git spawns   (-364, -21%)
```

Per package, same method:

| package | main | after |
| --- | --- | --- |
| `internal/pipe/git` | 556 | 423 |
| `cmd` | 354 | 280 |
| `internal/pipe/aur` + `aursources` | 290 | 202 |

The process count is the metric worth trusting here: it is exact and does not
move between runs. The package seconds on the windows job do not resolve a
change this size at n=1, because `internal/pipe/git` alone swings between
24.46s and 31.81s across four green `main` runs. At the ~70ms a git spawn
measures on that runner, 364 fewer of them is around 25s.

Three separate "ask git twice" defects, none of which changes behaviour:

1. `getInfo` called `git.IsRepo` twice in a row.
2. Three `git show HEAD --quiet` differing only in `--format`.
3. Three `git tag -l` for one tag differing only in the format.
4. `ExtractRepoFromConfig` ran `IsRepo` before every `ls-remote --get-url`,
   when it only needs it to choose between two error messages on failure.
5. The upload client ran four to seven `git config --local` after each clone,
   which `git clone -c` can carry.

Two coverage holes turned up in the code being touched, both now closed and
both verified by mutation:

- Nothing pinned `%h` to `ShortCommit` and `%H` to `FullCommit`, so swapping
  them passed. Nothing pinned `%ct` either, so reading the author date instead
  of the committer date passed.
- Nothing asserted the commit author the upload client configures, so dropping
  `user.name` passed.

caarlos0:
Superseded: after a merge with main this branch has no changes left.

- The per-commit `git -c` flags landed in #6852, so the clone no longer needs to write the author and signing options into the checkout config.
- The rest of the work landed in #6850.

Closing.

PR #6852, opened 2026-08-31
fix(client): apply the commit config per command

caarlos0:
Apply the commit author and signing options as `git -c` flags on the commit,
instead of writing them into the checkout with `git config --local`.

Several publishing pipes share one checkout under `dist/git/<name>-<branch>`.
Because the configuration lived in that checkout, a caller that reused it
inherited the identity and signing options of whoever cloned it first, and
could push a commit with the wrong author, signing key, GPG program, or
signature format.

Passing the values per command removes the shared state instead of repairing
it, so there is nothing to reset before the next caller and the leak cannot
come back.

Two smaller changes come with it:

- `init.defaultBranch` is dropped. It was written after the clone, and nothing
  in this path runs `git init`, so no command ever read it. Cloning an empty
  remote follows the branch the remote advertises.
- `gpg.format` is now passed even when it is `openpgp`. The old code skipped
  that value to avoid persisting a redundant setting; with `-c` there is
  nothing to persist, and passing it explicitly also overrides any ambient
  format.

Every remaining call site ran a single command, so `runGitCmds` collapses into
`runGitCmd`, with `runGitCmdWith` for the commit that carries the flags. The
flags stay out of the error message.

This follows a pattern the repository already uses: `internal/testlib/git.go`
builds `-c user.name=... -c commit.gpgSign=false` the same way.

#6821 fixed the same leak by clearing `user.signingKey`, `gpg.program` and
`gpg.format` and then rewriting the configuration on every call. That works,
but it keeps per-caller state in shared storage and repairs it afterwards.
Scoping the values to the command that reads them is smaller and cannot drift:
only `git commit` reads `user.name`, `user.email`, `commit.gpgSign`,
`user.signingKey`, `gpg.program` and `gpg.format`, so `clone`, `checkout`,
`add` and `push` never needed them.

This PR supersedes #6821 and contains its regression coverage.

**Tests**

- `TestCommitConfigFlags` covers the flag builder directly: disabled, disabled
  with options set, enabled bare, enabled with all options, and `openpgp`.
- `TestGitClientReconfiguresReusedCheckout` now asserts behaviour rather than
  persisted configuration. The signing case points `program` at a path under
  `t.TempDir()` that cannot exist, so it no longer depends on a GPG
  installation or on the keyring of whoever runs the tests. It asserts that
  git names that program, then that the next commit has the correct author and
  no `gpgsig` header.

Both cases fail against `main` with the expected diagnostics:

```
commit_author:  expected "Second <second@example.com>"
                actual   "First <first@example.com>"
signing:        An error is expected but got nil.
```

`internal/client` and the eight pipes that use this client (brew, cask, scoop,
krew, nix, winget, aur, aursources) pass.

**Known, not addressed here**

The checkout directory is keyed on `repo.Name + "-" + branch` rather than on
the git URL, so two publishers with the same `repository.name` and different
`git.url` still share a checkout and the second pushes to the first one's
remote. That is a separate pre-existing defect and needs its own PR.

Supersedes #6821. Related to #6715.

AI assistance disclosure: GitHub Copilot CLI assisted with the implementation
and the tests. I reviewed and understand the resulting changes.


caarlos0:
gitleaks complaining about a fake key, ignoring

PR #6853, opened 2026-08-31
fix(chocolatey): a skipped chocolatey entry stops the ones after it

hktitof:
Chocolatey packages no longer stop midway when one chocolatey entry is skipped, every entry still runs, and the skips are reported together at the end

`chocolatey.Run` was returning on the first error, including a pipe skip, so a single chocolatey entry that skipped (its config resolves to multiple archives for the same platform) silently prevented every entry after it from generating a package, winget already treats this same condition as a per-entry skip and keeps going, and this brings chocolatey in line with that pattern

`errMultipleArchives` becomes a `pipe.Skip` (same message), and the `Run` loop now uses `pipe.SkipMemento` so a skipped entry does not stop the ones after it, `errNoWindowsArchive` deliberately stays a hard error, a chocolatey config with no windows archives at all is still a real mistake, same as winget's `errNoArchivesFound`

Before, with two entries where only the first has duplicate amd64 archives:
```
• chocolatey packages
  • pipe skipped or partially skipped   reason=found multiple archives for the same platform, please consider filtering by id
```
(no package produced for the valid second entry, release fails)

After:
```
• chocolatey packages
  • packing   nuspec=dist/valid.choco/valid.nuspec
  • pipe skipped or partially skipped   reason=found multiple archives for the same platform, please consider filtering by id
```
(valid entry still packages, release continues)

Tests: `TestRunSkippedChocolateyDoesNotStopOthers` proves the entry after the skipped one still produces its package, and it fails without this change

<details><summary>AI disclosure (required by CONTRIBUTING)</summary>
This change was written with AI assistance and reviewed, tested, and validated by me before sending
</details>

Pattern precedent: #6783, #6784, #6785, #6786, #6787, #6788, #6797 (the same SkipMemento continue change merged across the other package pipes), the guard being fixed comes from #6792

### DIFF
--- a/internal/pipe/chocolatey/chocolatey.go
+++ b/internal/pipe/chocolatey/chocolatey.go
@@ -12,6 +12,7 @@ import (
 	"github.com/caarlos0/log"
 	"github.com/goreleaser/goreleaser/v2/internal/artifact"
 	"github.com/goreleaser/goreleaser/v2/internal/client"
+	"github.com/goreleaser/goreleaser/v2/internal/pipe"
 	"github.com/goreleaser/goreleaser/v2/internal/skips"
 	"github.com/goreleaser/goreleaser/v2/internal/summary"
 	"github.com/goreleaser/goreleaser/v2/internal/tmpl"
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):

caarlos0:
thank you!

PR #6869, opened 2026-09-05
fix(git): report repository detection errors

### LINKED ISSUE #6847: Report VCS errors as git tells them
Hi there!

I recently faced an issue, that the git checkout was in a directory owned by user 1000 (it was setup by the GitHub actions runner), but the current user was `root` (since the action was run inside a container). This results in git treating the repo as an unsafe directory, which means despite it having cloned the repo to this location seconds earlier, now every call to git fails.

The issue is described at length in https://github.com/actions/checkout/issues/2031

However, during this I noticed that this unsafe directory was not reported nicely by goreleaser - it just listed an unknown previous version and set the current version to v0.0.0:

```
  • skipping validate...
  • cleaning distribution directory
  • loading environment variables
  • getting and validating git state
    • accepting to run without a git repository because this is a snapshot
    • using tags                                     previous=<unknown> current=v0.0.0
    • pipe skipped or partially skipped              reason=disabled during snapshot mode
  • parsing tag
  • setting defaults
  • snapshotting
    • building snapshot...                           version=0.0.0-SNAPSHOT-none
  • running before hooks
    • running                                        hook=go mod tidy
    • running                                        hook=go generate ./...
  • ensuring distribution directory
  • setting up metadata
  • writing release metadata
  • loading go mod information
  • build prerequisites
  • building binaries
    • building                                       paths=cmd/editorconfig-checker binaries=editorconfig-checker target=linux_arm64_v8.0
    • building                                       paths=cmd/editorconfig-checker binaries=editorconfig-checker target=linux_arm_6
    • building                                       paths=cmd/editorconfig-checker binaries=editorconfig-checker target=linux_386_sse2
    • building                                       paths=cmd/editorconfig-checker binaries=editorconfig-checker target=darwin_arm64_v8.0
    • building                                       paths=cmd/editorconfig-checker binaries=editorconfig-checker target=linux_amd64_v1
    • building                                       paths=cmd/editorconfig-checker binaries=editorconfig-checker target=windows_386_sse2
    • building                                       paths=cmd/editorconfig-checker binaries=editorconfig-checker target=darwin_amd64_v1
    • building                                       paths=cmd/editorconfig-checker binaries=editorconfig-checker target=windows_arm64_v8.0
    • building                                       paths=cmd/editorconfig-checker binaries=editorconfig-checker target=windows_amd64_v1
  ⨯ build failed after 0s                           
    error=
    │ build failed: exit status 1: error obtaining VCS status: exit status 128
    │ 	Use -buildvcs=false to disable VCS stamping.
    target=windows_386_sse2
```

This example was reproduced with `sudo chown root .git` - and while git is able to read the directory, it just is owned by the wrong UID.

In this case you can see that goreleaser doesn't really care that no info comes from git, but it also doesn't report the error from git at all, only the `go build` later fails, since it tries to embed the git commit id into the binary being built.

---

Now, what would I like to see goreleaser do?

I wish that goreleaser would see that a `.git` exists, but the call to git failed. And in this case it should at least transport the stderr from git to help the user understand what happened. Especially in a CI environment it's hard to debug this particular problem, when the checkout was freshly done without error, and other steps in the CI pipeline like a `go test ./...` worked nicely, and locally goreleaser runs fine without error.

I know this particular problem is hard to properly handle, after all what differentiates a broken repo from no repo at all, I just wish that users have an easier time debugging this problem.

Best regards,
Andre

caarlos0:
Report failed Git repository checks directly from `IsRepo`, including Git's
original diagnostic and recovery instructions. Preserve its boolean API and
the existing release and snapshot behavior.

`IsRepo` discarded Git's stderr. An unsafe checkout therefore looked like a
missing repository, and snapshot builds continued without showing the cause.
Log the error inside the helper so callers remain unchanged and the cause is
visible without `--verbose` or parsing Git's error text.

Each failed repository check emits the warning "git repository check failed",
with Git's original stderr, for both missing and unsafe repositories.

Add serial log-capture coverage for unsafe repositories in release and snapshot
modes, plus snapshots outside a repository. Isolate Git configuration, including
environment overrides, without changing filesystem ownership.

Fixes #6847.

### DIFF
--- a/internal/git/config.go
+++ b/internal/git/config.go
@@ -20,7 +20,7 @@ func ExtractRepoFromConfig(ctx context.Context) (result config.Repo, err error) 
 		// repository without a remote and a directory that is not a
 		// repository at all, so it takes a second command to tell the user
 		// which one it is. Only the failing path pays for it.
-		if isRepo, _ := IsRepo(ctx); !isRepo {
+		if !IsRepo(ctx) {
 			return result, errors.New("current folder is not a git repository")
 		}
 		return result, errors.New("no remote configured to list refs from")
--- a/internal/git/config.go
+++ b/internal/git/config.go
@@ -20,10 +20,7 @@ func ExtractRepoFromConfig(ctx context.Context) (result config.Repo, err error) 
 		// repository without a remote and a directory that is not a
 		// repository at all, so it takes a second command to tell the user
 		// which one it is. Only the failing path pays for it.
-		if isRepo, err := IsRepo(ctx); !isRepo {
-			if err != nil {
-				log.WithError(err).Warn("could not check if the current directory is a git repository")
-			}
+		if isRepo, _ := IsRepo(ctx); !isRepo {
 			return result, errors.New("current folder is not a git repository")
 		}
 		return result, errors.New("no remote configured to list refs from")
--- a/internal/git/config.go
+++ b/internal/git/config.go
@@ -20,7 +20,10 @@ func ExtractRepoFromConfig(ctx context.Context) (result config.Repo, err error) 
 		// repository without a remote and a directory that is not a
 		// repository at all, so it takes a second command to tell the user
 		// which one it is. Only the failing path pays for it.
-		if !IsRepo(ctx) {
+		if isRepo, err := IsRepo(ctx); !isRepo {
+			if err != nil {
+				log.WithError(err).Warn("could not check if the current directory is a git repository")
+			}
 			return result, errors.New("current folder is not a git repository")
 		}
 		return result, errors.New("no remote configured to list refs from")
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -11,13 +11,14 @@ import (
 	"github.com/caarlos0/log"
 )
 
-// IsRepo reports whether the current folder is inside a git work tree.
-func IsRepo(ctx context.Context) (bool, error) {
+// IsRepo returns true if current folder is a git repository.
+func IsRepo(ctx context.Context) bool {
 	out, err := Run(ctx, "rev-parse", "--is-inside-work-tree")
 	if err != nil {
-		return false, err
+		log.WithError(err).Warn("git repository check failed")
+		return false
 	}
-	return strings.TrimSpace(out) == "true", nil
+	return strings.TrimSpace(out) == "true"
 }
 
 func RunWithEnv(ctx context.Context, env []string, args ...string) (string, error) {
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -14,7 +14,11 @@ import (
 // IsRepo returns true if current folder is a git repository.
 func IsRepo(ctx context.Context) bool {
 	out, err := Run(ctx, "rev-parse", "--is-inside-work-tree")
-	return err == nil && strings.TrimSpace(out) == "true"
+	if err != nil {
+		log.WithError(err).Warn("could not check if the current directory is a git repository")
+		return false
+	}
+	return strings.TrimSpace(out) == "true"
 }
 
 func RunWithEnv(ctx context.Context, env []string, args ...string) (string, error) {
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -11,14 +11,13 @@ import (
 	"github.com/caarlos0/log"
 )
 
-// IsRepo returns true if current folder is a git repository.
-func IsRepo(ctx context.Context) bool {
+// IsRepo reports whether the current folder is inside a git work tree.
+func IsRepo(ctx context.Context) (bool, error) {
 	out, err := Run(ctx, "rev-parse", "--is-inside-work-tree")
 	if err != nil {
-		log.WithError(err).Warn("could not check if the current directory is a git repository")
-		return false
+		return false, err
 	}
-	return strings.TrimSpace(out) == "true"
+	return strings.TrimSpace(out) == "true", nil
 }
 
 func RunWithEnv(ctx context.Context, env []string, args ...string) (string, error) {
--- a/internal/git/git_test.go
+++ b/internal/git/git_test.go
@@ -1,6 +1,7 @@
 package git_test
 
 import (
+	"os"
 	"testing"
 
 	"github.com/goreleaser/goreleaser/v2/internal/git"
@@ -41,12 +42,43 @@ func TestGitWarning(t *testing.T) {
 }
 
 func TestRepo(t *testing.T) {
-	ctx := t.Context()
-	require.True(t, git.IsRepo(ctx), "goreleaser folder should be a git repo")
+	t.Setenv("GIT_CONFIG_GLOBAL", os.DevNull)
+	t.Setenv("GIT_CONFIG_NOSYSTEM", "1")
 
-	tmpDir := t.TempDir()
-	t.Chdir(tmpDir)
-	require.False(t, git.IsRepo(ctx), tmpDir+" folder should be a git repo")
+	t.Run("work tree", func(t *testing.T) {
+		testlib.Mktmp(t)
+		testlib.GitInit(t)
+		isRepo, err := git.IsRepo(t.Context())
+		require.NoError(t, err)
+		require.True(t, isRepo)
+	})
+
+	t.Run("not a repository", func(t *testing.T) {
+		testlib.Mktmp(t)
+		isRepo, err := git.IsRepo(t.Context())
+		require.ErrorContains(t, err, "not a git repository")
+		require.False(t, isRepo)
+	})
+
+	t.Run("bare repository", func(t *testing.T) {
+		dir := testlib.GitMakeBareRepository(t)
+		t.Chdir(dir)
+		t.Setenv("GIT_DIR", dir)
+		isRepo, err := git.IsRepo(t.Context())
+		require.NoError(t, err)
+		require.False(t, isRepo)
+	})
+
+	t.Run("unsafe repository", func(t *testing.T) {
+		testlib.Mktmp(t)
+		testlib.GitInit(t)
+		// Exercise Git's ownership check without changing filesystem ownership.
+		t.Setenv("GIT_TEST_ASSUME_DIFFERENT_OWNER", "1")
+		isRepo, err := git.IsRepo(t.Context())
+		require.ErrorContains(t, err, "fatal: detected dubious ownership")
+		require.ErrorContains(t, err, "git config --global --add safe.directory")
+		require.False(t, isRepo)
+	})
 }
 
 func TestClean(t *testing.T) {
--- a/internal/git/git_test.go
+++ b/internal/git/git_test.go
@@ -1,7 +1,6 @@
 package git_test
 
 import (
-	"os"
 	"testing"
 
 	"github.com/goreleaser/goreleaser/v2/internal/git"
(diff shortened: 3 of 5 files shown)


internal/pipe/git/git_test.go
```diff
@@ -25,6 +25,29 @@ func TestNotAGitFolder(t *testing.T) {
...
+
+			ctx := testctx.Wrap(t.Context())
+			ctx.Snapshot = mode == "snapshot"
+			err := Pipe{}.Run(ctx)
+			if ctx.Snapshot {
+				testlib.AssertSkipped(t, err)
+				require.Equal(t, fakeInfo, ctx.Git)
+			} else {
+				require.ErrorIs(t, err, ErrNotRepository)
+			}
```
caarlos0, line 55: **This test passes with the fix fully removed, so it is not a regression test for this PR.**

Both assertions describe control flow that already existed: old `IsRepo` returned `false` when `Run` failed, so `getInfo` took the exact same `!isRepo` branch. `TestNotAGitFolder:22` and `TestSnapshotWithoutRepo` already cover it.

Proven with `go test -overlay` (no files touched), three ways:

| source under test | this test |
|---|---|
| merge base `f5edd7395`, no fix at all | **PASS** |
| this branch with both `Warn` calls deleted | **PASS** |
| `internal/git` `TestRepo/unsafe_repository` against merge base | FAIL — real regression test |

So the user-visible behavior the PR exists for — git's message reaching the log — is asserted nowhere. Deleting lines 74-76 of `internal/pipe/git/git.go` and lines 24-26 of `internal/git/config.go` leaves the whole suite green and reopens #6847.

`ad866ac74` had the coverage and this commit dropped it. **Smallest fix:** restore the log buffer and the two `require.Contains(t, logs.String(), ...)` assertions from that commit. Note `internal/testlib/log.go` sets `log.Log = log.New(io.Discard)` in `init()`, so swapping `log.Log` is the only way to observe output here.

internal/pipe/git/git.go
```diff
@@ -70,7 +70,10 @@ var fakeInfo = context.GitInfo{
 }
 
 func getInfo(ctx *context.Context) (context.GitInfo, error) {
-	if !git.IsRepo(ctx) {
+	if isRepo, err := git.IsRepo(ctx); !isRepo {
+		if err != nil {
+			log.WithError(err).Warn("could not check if the current directory is a git repository")
```
caarlos0: **This warning also fires in the ordinary "not a git repository" case.**

`git rev-parse --is-inside-work-tree` exits **128** with `fatal: not a git repository (or any of the parent directories): .git` when the directory is not a repo, so `err != nil` is the *normal* not-a-repo signal, not the exceptional one. The only input reaching `(false, nil)` is a bare repo.

**Trigger:** `goreleaser release --snapshot` anywhere outside a git repo — a supported path with its own reassuring log line (`TestSnapshotWithoutRepo`). Captured output:

```
 • could not check if the current directory is a git repository
   error=
   │ fatal: not a git repository (or any of the parent directories): .git
 • accepting to run without a git repository because this is a snapshot
```

**Impact:** every snapshot build outside a repo now prints a `fatal:` line it did not print before, and the message is untrue — git *did* check, and answered no.

**Smallest fix:** skip the warning for the plain not-a-repo error, e.g. gate on `!strings.Contains(err.Error(), "not a git repository")`. #6847 already calls stderr matching messy, so if you would rather not match, at least reword the message to something true in both cases.

**Test:** assert log output for the plain not-a-repo case as well as the dubious-ownership case, so the noise level is pinned rather than incidental.

internal/git/config.go
```diff
@@ -20,7 +20,10 @@ func ExtractRepoFromConfig(ctx context.Context) (result config.Repo, err error)
 		// repository without a remote and a directory that is not a
 		// repository at all, so it takes a second command to tell the user
 		// which one it is. Only the failing path pays for it.
-		if !IsRepo(ctx) {
+		if isRepo, err := IsRepo(ctx); !isRepo {
+			if err != nil {
+				log.WithError(err).Warn("could not check if the current directory is a git repository")
```
caarlos0: **Second copy of the same message, and it makes a deliberately silent path shout.**

In the reported scenario the warning is emitted **twice**: `getInfo` logs it, then `internal/pipe/project.gitRemote` (project.go:78) calls `ExtractRepoFromConfig`, whose `ls-remote` fails with the same dubious-ownership text and which then calls `IsRepo` again. Triggered whenever `project_name` is unset and there is no `go.mod`/`Cargo.toml`.

`gitRemote` discards its error on purpose — it is a best-effort project-name guess — so a `Warn` reached from there is noise by construction.

The outer `err` from `ls-remote` on line 17 already carries the identical git stderr and is discarded, so this second `IsRepo` call adds nothing the caller did not already have.

**Smallest fix:** drop this warn and log only in `internal/pipe/git/git.go`, so the message has one owner and one occurrence.

**Test:** the log-capture test from the sibling comment, asserting the message appears exactly once.

(The `err` shadow of the named return is harmless — both branches return explicit values — but it only exists because of the signature change; see the comment on `internal/git/git.go`.)

internal/git/git.go
```diff
@@ -11,10 +11,13 @@ import (
 	"github.com/caarlos0/log"
 )
 
-// IsRepo returns true if current folder is a git repository.
-func IsRepo(ctx context.Context) bool {
+// IsRepo reports whether the current folder is inside a git work tree.
+func IsRepo(ctx context.Context) (bool, error) {
```
caarlos0: **The signature change buys nothing over logging inside `IsRepo`.**

`IsRepo` has two callers, both internal, and both do the *identical* thing with the error: write the same warning, then continue on `isRepo == false`. No caller uses it differently. Costs:

- the same message string duplicated in two files (`internal/git/config.go:25`, `internal/pipe/git/git.go:75`), free to drift;
- each call site grows from one line to five, and needs an `err` shadow of a named return in `config.go`;
- `(bool, error)` is a trap here, because the error must **not** abort the caller — `isRepo` controls the flow. The next reader writes `if err != nil { return err }` and silently changes snapshot behavior.

The usual "a helper must not log" objection does not apply: this package already logs, in `RunWithEnv` and `ExtractRepoFromURL`.

`ad866ac74` put the same behavior in one place, kept `IsRepo` a predicate, needed no call-site change, and carried a test that actually failed without it. **Smallest fix:** keep that version.

If you keep this one, the `bare repository` subtest in `internal/git/git_test.go` exists only to distinguish `(false, nil)` from `(false, err)` — it has no reason to exist under the predicate signature.

internal/git/git_test.go
```diff
@@ -41,12 +42,43 @@ func TestGitWarning(t *testing.T) {
 }
 
 func TestRepo(t *testing.T) {
-	ctx := t.Context()
-	require.True(t, git.IsRepo(ctx), "goreleaser folder should be a git repo")
+	t.Setenv("GIT_CONFIG_GLOBAL", os.DevNull)
+	t.Setenv("GIT_CONFIG_NOSYSTEM", "1")
```
caarlos0: **These two do not neutralize env-scoped config, so the ownership tests can be defeated.**

`ensure_valid_ownership` reads `git_protected_config()`, which includes the command-line/env scope. `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_NOSYSTEM` do not clear `GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_*`, and git's own `t0033-safe-directory.sh` uses exactly that channel to set `safe.directory`.

Measured with `GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=safe.directory GIT_CONFIG_VALUE_0='*'`: `TestRepo/unsafe_repository` fails, and `TestUnsafeRepository/snapshot` passes **vacuously**. Not hypothetical — a plain dev shell in this checkout already exports `GIT_CONFIG_COUNT=8`. That ambient `safe.bareRepository=explicit` is also what makes `GIT_DIR` load-bearing in the `bare repository` subtest below. CI is unaffected.

**Fix:** add `t.Setenv("GIT_CONFIG_COUNT", "0")` and `t.Setenv("GIT_CONFIG_PARAMETERS", "")` here and in `internal/pipe/git/git_test.go:34`.

caarlos0 (commented):
**Verdict: comment — five findings.** The behavior change does what #6847 asks: git's stderr now reaches the user without `--verbose`. Two problems are worth fixing before merge: the pipe-level test does not test this change, and the new warning fires on the ordinary not-a-repo path. Details are inline.

One note with no single site: **the description is now stale.** It says the fix works "without ... changes to the helper's API", but `26da2b9b2` changes `IsRepo` from `bool` to `(bool, error)`. It also says "Add regression coverage"; that is true of `internal/git/git_test.go`, which pins the new error return, but not of `internal/pipe/git/git_test.go` — see the inline comment there.

**Verification gaps.** Build, lint and gofmt were not run (assumed OK, per instruction). `goreleaser` itself was never executed end to end: the not-a-repo warning is proven from the code path plus raw `git` exit codes and captured log output from an overlay probe, not from the shipped binary. `GIT_TEST_ASSUME_DIFFERENT_OWNER` requires git >= 2.36, so contributors on older git (Debian bullseye ships 2.30.2) will see the new tests fail — the same dependency the earlier commit had, and Windows CI was green on it. The four new `internal/git/git_test.go` subtests have not yet run on Windows CI.

_Reviewed by a bot._

internal/pipe/git/git_test.go
caarlos0: Restored the output assertions in a6081ff15aa4a2b03a35dd5873b92ec1a1c712be. The test asserts the ownership diagnostic and safe.directory instructions immediately after the pipe runs, then asserts exactly one diagnostic after repository extraction. Changing the pipe warning to Debug makes TestUnsafeRepository/snapshot fail on the missing diagnostic. These tests run serially with t.Chdir/t.Setenv, and t.Cleanup restores the previous logger.

internal/git/git.go
caarlos0: Keeping (bool, error) intentionally. After a6081ff15aa4a2b03a35dd5873b92ec1a1c712be the callers use it differently: the Git-state pipe reports the diagnostic, while ExtractRepoFromConfig remains silent and returns its existing classification error. Moving warning output back into IsRepo would reintroduce the unwanted output from best-effort project-name discovery. The duplicate warning and named-return shadow are gone. Direct error coverage and the bare-repository distinction remain useful under this contract.

internal/git/git_test.go
caarlos0: Fixed in a6081ff15aa4a2b03a35dd5873b92ec1a1c712be by setting GIT_CONFIG_COUNT=0 and GIT_CONFIG_PARAMETERS to an empty value in both ownership-test setups. Reproduced the original failure with an inherited safe.directory=* entry. The Git helper, Git pipe, and project-default packages now pass with safe.directory=* supplied through both environment channels, with race detection and shuffle seed 6847.

internal/pipe/git/git.go
caarlos0: Used the wording alternative in a6081ff15aa4a2b03a35dd5873b92ec1a1c712be: the warning is now "git repository check failed", which describes the failed Git command in both cases. Git stderr remains visible for an ordinary missing repository; it is not filtered by locale-dependent text matching. TestSnapshotWithoutRepo now asserts one diagnostic and the existing snapshot-continuation message, while the unsafe-repository cases assert the ownership diagnostic.

internal/git/config.go
caarlos0: Removed this warning in a6081ff15aa4a2b03a35dd5873b92ec1a1c712be. Only the Git-state pipe reports the diagnostic; ExtractRepoFromConfig keeps its existing returned error without warning. The regression runs the pipe followed by repository extraction and checks that the ownership diagnostic appears exactly once. It failed with an actual count of two before this fix.

PR #6871, opened 2026-09-05
fix(cask): deprecate url.verified

caarlos0:
Deprecate `homebrew_casks.url.verified`, and stop writing the `verified:` parameter into generated Casks.

## Why

Homebrew deprecated the `verified:` parameter of the Cask `url` stanza, and now uses the default URL verification behaviour. Every generated Cask made `brew` print a warning to the end user, and told them to report it to the tap:

```
Warning: Calling the `verified` parameter in the `url` stanza is deprecated! Use the default URL verification behaviour instead.
Please report this issue to the caarlos0/homebrew-tap tap (not Homebrew/* repositories), or even better, submit a PR to fix it:
  /opt/homebrew/Library/Taps/caarlos0/homebrew-tap/Casks/pinata.rb:13
```

The parameter only affected `brew audit`, so removing it from the output does not change how a Cask installs. It will become an error in a future Homebrew release, so we remove it now.

## What

- `homebrew_casks.url.verified` now prints a deprecation notice, and `goreleaser check` reports it.
- The `verified:` parameter is no longer written into the Cask.
- Removed the option from the Homebrew Casks documentation and from the `brews` migration guide.
- Added the deprecation notice page entry.
- Removed `url.verified` from this project's own `.goreleaser.yaml`.

## Links

- https://github.com/Homebrew/brew/pull/23280

### DIFF
--- a/internal/pipe/cask/cask_test.go
+++ b/internal/pipe/cask/cask_test.go
@@ -482,6 +482,7 @@ func TestFullPipe(t *testing.T) {
 				ctx.Config.Casks[0].Repository.Name = "test"
 				ctx.Config.Casks[0].Homepage = "https://dummyhost-url-parameters.com/"
 				ctx.Config.Casks[0].URL.Using = ":post"
+				ctx.Config.Casks[0].URL.Verified = "https://dummyhost/download/"
 				ctx.Config.Casks[0].URL.Headers = []string{"Accept: application/octet-stream"}
 				ctx.Config.Casks[0].URL.Data = map[string]string{"payload": "hello_world"}
 			},
@@ -1233,6 +1234,7 @@ func TestDefaultDeprecated(t *testing.T) {
 	require.True(t, ctx.Deprecated)
 	require.Equal(t, []string{"bin"}, ctx.Config.Casks[0].Binaries)
 	require.Equal(t, []string{"man"}, ctx.Config.Casks[0].Manpages)
+	require.Contains(t, ctx.NotifiedDeprecations, "homebrew_casks.url.verified")
 }
 
 func TestGHFolder(t *testing.T) {
--- a/internal/pipe/cask/cask_test.go
+++ b/internal/pipe/cask/cask_test.go
@@ -482,7 +482,6 @@ func TestFullPipe(t *testing.T) {
 				ctx.Config.Casks[0].Repository.Name = "test"
 				ctx.Config.Casks[0].Homepage = "https://dummyhost-url-parameters.com/"
 				ctx.Config.Casks[0].URL.Using = ":post"
-				ctx.Config.Casks[0].URL.Verified = "https://dummyhost/download/"
 				ctx.Config.Casks[0].URL.Headers = []string{"Accept: application/octet-stream"}
 				ctx.Config.Casks[0].URL.Data = map[string]string{"payload": "hello_world"}
 			},
@@ -1223,6 +1222,9 @@ func TestDefaultDeprecated(t *testing.T) {
 			{
 				Binary:  "bin",
 				Manpage: "man",
+				URL: config.HomebrewCaskURL{
+					Verified: "github.com/foo/bar",
+				},
 			},
 		},
 	}, testctx.GitHubTokenType)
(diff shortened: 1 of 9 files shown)


internal/pipe/cask/cask_test.go
```diff
@@ -1223,6 +1222,9 @@ func TestDefaultDeprecated(t *testing.T) {
 			{
 				Binary:  "bin",
 				Manpage: "man",
+				URL: config.HomebrewCaskURL{
+					Verified: "github.com/foo/bar",
+				},
```
caarlos0, line 1228: These lines don't actually test the new code. `deprecate.Notice` for `binary` (`cask.go:73`) already sets `ctx.Deprecated = true`, and this test only asserts `require.True(t, ctx.Deprecated)` on line 1233. I deleted the new `if brew.URL.Verified != ""` block at `cask.go:86-88` in a scratch copy and `TestDefaultDeprecated` still passed.

So the new notice has no regression protection: a later refactor can drop it and users silently stop being warned.

Smallest fix — one assertion after line 1233:

```go
require.Contains(t, ctx.NotifiedDeprecations, "homebrew_casks.url.verified")
```

I checked it both ways: with the block removed it fails (`map[...binary, ...manpage] does not contain "homebrew_casks.url.verified"`), with the block present it passes.

internal/pipe/cask/cask_test.go
```diff
@@ -482,7 +482,6 @@ func TestFullPipe(t *testing.T) {
 				ctx.Config.Casks[0].Repository.Name = "test"
 				ctx.Config.Casks[0].Homepage = "https://dummyhost-url-parameters.com/"
 				ctx.Config.Casks[0].URL.Using = ":post"
-				ctx.Config.Casks[0].URL.Verified = "https://dummyhost/download/"
```
caarlos0, line 485: I'd keep this line. Without it, nothing proves that a value that *is* set gets dropped from the output — `TestFullPipe/url_parameters_post` now only shows that an empty field writes nothing. Putting `Verified` back into `dataFor` and the template would fail no test.

Since `dataFor` no longer reads the field, keeping the line changes no output: the new `testdata/TestFullPipe/url_parameters_post.rb.golden` stays exactly as committed. It costs nothing and turns that golden into a real regression test for the removal.

See also the comment at `cask_test.go:1225`.

caarlos0 (commented):
**Verdict: no functional defect. Two test gaps.**

The premise checks out: Homebrew/brew#23280 is merged, and `Library/Homebrew/cask/dsl.rb` on `main` now calls `odeprecated` for the `verified` parameter. `verified:` is still an accepted keyword there, so Casks already in taps keep working. Dropping it from the output is the right call.

What I verified:

- Nothing reads `HomebrewCaskURL.Verified` any more; only `pkg/config/config.go` declares it, so old configs still parse.
- Every block in `additional_url_params.rb` carries its own leading comma, so removing the first block cannot break the Ruby. Rendering 9 URL-parameter combinations (none, each alone, all) and running `ruby -c` on each passed.
- `goreleaser check` on a config with `url.verified` prints the notice and exits 2.
- The notice link resolves: `deprecate.Notice` builds `#homebrew_casksurlverified`, which matches the new heading.
- `since v2.18.1` is right — the last release is `v2.18.0` and all 37 commits since are `fix`/`chore`/`test`/`docs`/`perf`/`ci`. If a `feat` lands before the release, bump the entry to `v2.19.0`.
- `www/static/schema.json` is left alone, as it should be — the scheduled `generate` workflow owns it.

Verification gap, as requested: build, lint, and format were not run. Test runs were limited to `internal/pipe/cask`. `TestFullPipe/git_remote` fails locally, but it fails the same way on a clean `origin/main`, so it is environmental and unrelated.

Two inline comments below. Posted by a bot.

internal/pipe/cask/cask_test.go
caarlos0: Correct, thank you. Added `require.Contains(t, ctx.NotifiedDeprecations, "homebrew_casks.url.verified")` in 247adbc7c.

I verified it both ways: with the `if brew.URL.Verified != ""` block removed from `cask.go`, `TestDefaultDeprecated` now fails; with the block present it passes.

internal/pipe/cask/cask_test.go
caarlos0: Agreed, restored in 247adbc7c. The golden file is unchanged, as you expected.

I verified it: after putting `Verified` back into `dataFor`, `template.go` and the template, `TestFullPipe/url_parameters_post` fails with three `+ verified: "https://dummyhost/download/",` lines. Without the line it passed either way.

PR #6873, opened 2026-09-05
feat(cask): add hooks install_steps, deprecate raw Ruby hooks

### LINKED ISSUE #6870: homebrew_casks: hooks emit Homebrew's deprecated `postflight` stanza
### Summary

`homebrew_casks` with `hooks.post.install` generates a `postflight do ... end` block. Homebrew now deprecates that stanza, so every `brew` command touching the cask prints:

```
Warning: Calling `postflight` is deprecated! Use `postflight_steps` instead.
Please report this issue to the <owner>/homebrew-<tap> tap (not Homebrew/* repositories), or even better, submit a PR to fix it:
  /opt/homebrew/Library/Taps/<owner>/homebrew-<tap>/Casks/<cask>.rb:36
```

The cask is generated and carries `# This file was generated by GoReleaser. DO NOT EDIT.`, so a tap maintainer cannot fix it in the tap — the next release overwrites it — and there is no config option to change what is emitted.

### Config that produces it

```yaml
homebrew_casks:
  - name: ipscout
    repository:
      owner: jonhadfield
      name: homebrew-ipscout
    # the binaries are unsigned, so clear the quarantine attribute on
    # install to avoid Gatekeeper blocking them
    hooks:
      post:
        install: |
          if system_command("/usr/bin/xattr", args: ["-h"]).exit_status == 0
            system_command "/usr/bin/xattr", args: ["-dr", "com.apple.quarantine", "#{staged_path}/ipscout"]
          end
```

Generated cask (GoReleaser v2.18.0, Homebrew 6.0.21):

```ruby
  postflight do
    if system_command("/usr/bin/xattr", args: ["-h"]).exit_status == 0
      system_command "/usr/bin/xattr", args: ["-dr", "com.apple.quarantine", "#{staged_path}/ipscout"]
    end
  end
```

### Why it matters beyond the warning

Clearing `com.apple.quarantine` is the common use of this hook for projects shipping unsigned binaries, because Gatekeeper kills a quarantined ad-hoc-signed binary — it exits 137 and prints nothing. If `postflight` is removed rather than deprecated, that hook silently stops running: releases keep succeeding, the cask keeps installing, and users get a binary that dies with no output. The failure surfaces at install time on other people's machines, not at release time.

### The replacement is not a rename

`postflight_steps` takes a declarative, normalised list of typed steps and runs them in a sandbox (`Library/Homebrew/install_steps.rb`, `Library/Homebrew/cask/artifact/install_steps.rb`), rather than a block of arbitrary Ruby. So emitting the new stanza needs the hook body mapped onto the step model, or a new field that accepts steps directly — for example a `run` step invoking `/usr/bin/xattr`.

### Versions

- GoReleaser v2.18.0 (latest at time of writing)
- Homebrew 6.0.21
- macOS 26.5

caarlos0:
Add `homebrew_casks.hooks.{pre,post}.{install,uninstall}_steps`, and deprecate the raw-Ruby `install`/`uninstall` options.

## Why

Homebrew deprecated the `preflight`, `postflight`, `uninstall_preflight` and `uninstall_postflight` stanzas, so every generated Cask that uses a hook now prints a warning to the end user, and tells them to report it to the tap:

```
Warning: Calling `postflight` is deprecated! Use `postflight_steps` instead.
```

The replacement stanzas do **not** accept arbitrary Ruby. They use Homebrew's declarative install steps DSL, which allows 26 methods for Casks (`run`, `remove`, `move`, `symlink`, `set_permissions`, `on_macos`, …). GoReleaser passes the hook body through untouched, so it cannot translate it. Renaming the stanza alone would break every existing hook at install time, with no way for GoReleaser to detect it.

For that reason this adds new options rather than changing the existing ones.

## What

- New: `hooks.pre.install_steps`, `hooks.post.install_steps`, `hooks.pre.uninstall_steps`, `hooks.post.uninstall_steps`. They write the `*_steps` stanzas.
- Deprecated: `hooks.pre.install`, `hooks.post.install`, `hooks.pre.uninstall`, `hooks.post.uninstall`. They keep writing the old stanzas, and `goreleaser check` now reports them.
- When both are set for the same hook, the steps body wins.
- New Cask template fields `.StagedPath` and `.AppDir`. Homebrew's tokens use the same delimiters as GoReleaser templates, so without them users would have to write `{{ "{{staged_path}}" }}`.

```yaml
homebrew_casks:
  - name: foo
    hooks:
      post:
        install_steps: |
          on_macos do
            run "/usr/bin/xattr", args: ["-dr", "com.apple.quarantine", "{{ .StagedPath }}/foo"]
          end
```

## Notes

- Existing golden files are byte-identical, so the deprecated path produces exactly the same output as before.
- The generated Cask was fed to a real Homebrew tap: it parses, and prints no deprecation warning.
- `warn` is deliberately not used in the examples: it is not in Homebrew's `CASK_ALLOWED_STEP_METHODS`.

## Links

- https://github.com/Homebrew/brew/pull/23366
- closes #6870


internal/pipe/cask/template.go
```diff
@@ -99,6 +100,37 @@ func uninstallString(u config.HomebrewCaskUninstall) string {
...
+	"AppDir":     "{{appdir}}",
+}
+
+// flightString renders a Cask flight block. The `steps` body uses the
+// declarative install steps DSL and goes into the `_steps` stanza; the
+// deprecated `legacy` body is raw Ruby and keeps the old stanza. When both are
+// set, the steps body wins.
+func flightString(stanza, steps, legacy string) string {
+	body := legacy
+	if steps != "" {
```
caarlos0: **A whitespace-only `install_steps` silently drops the deprecated body.**

`steps != ""` accepts a body that is only whitespace, but `split` (line 52) trims before it splits. So `body` is replaced by the blank steps value, `lines` is empty, and the function returns `""`. No stanza is written at all, and the `install` body disappears with no error and no warning.

Trigger:
```yaml
hooks:
  pre:
    install: |
      system_command "/usr/bin/true"
    install_steps: |
      
```

Impact: a hook that the user configured is missing from the published Cask, and nothing reports it.

Smallest fix:
```go
if strings.TrimSpace(steps) != "" {
```

Test: a unit test that calls `flightString` with a whitespace-only `steps` and a real `legacy` body, and requires the legacy `preflight do ... end` block. There is no unit test for `flightString` at all today.

caarlos0 (commented):
**Changes requested** (posted as a comment: GitHub does not allow a change request on your own pull request).

The approach is right: Homebrew's `*_steps` stanzas take only the declarative DSL, so a rename would break every existing hook. New options plus a deprecation is the correct trade.

I checked the change against Homebrew's source, not only against the PR text:

- `preflight_steps`, `postflight_steps`, `uninstall_preflight_steps` and `uninstall_postflight_steps` are the correct stanza names, and they hold the same positions in `rubocops/cask/constants/stanza.rb`, so the template order stays valid.
- `mkdir_p`, `remove`, `run` and `on_macos` are all in `CASK_ALLOWED_STEP_METHODS`. `warn` is not, so leaving it out of the examples is correct.
- Literal `{{staged_path}}` and `{{appdir}}` in a step string do work: `default_base_for` treats a path that starts with a token as absolute, and `run` arguments go through `expand_template_tokens`.
- The deprecated path stays byte-identical. `TestFullCask` and `TestFullPipe/hooks_templated` still contain `preflight` and are unmodified.
- `goreleaser check` does report the four new notices: `cmd/check.go` runs `defaults.Pipe{}`, which runs `cask.Pipe{}.Default`.

Four findings are inline. One is a silent hook drop; three are documentation claims that do not hold.

Verification gaps:
- Build, lint and fmt were not run, at the request of the person who asked for this review.
- `TestFullPipe/git_remote` fails in my sandbox, and `internal/pipe/krew` `TestFullPipe/git_remote` fails the same way on code this PR does not touch. I treat it as environmental, not as a result of this change.

This review is from an automated reviewer (GitHub Copilot CLI).

www/content/customization/general/templates.md
```diff
@@ -147,6 +147,19 @@ In the nFPM name template field, you can use those extra fields:
...
+In the [Homebrew Cask](/customization/publish/homebrew_casks/) fields, you can
+use these extra fields {{< g_inline_version "v2.19-unreleased" >}}:
+
+| Key           | Description                                                        |
+| ------------- | ------------------------------------------------------------------ |
+| `.StagedPath` | Homebrew's `staged_path` token: where the cask contents are staged |
+| `.AppDir`     | Homebrew's `appdir` token: where `.app` bundles are installed      |
+
+Homebrew uses the same delimiters as GoReleaser, so these fields let you write
+its tokens without quoting them.
```
caarlos0: **Two claims in this section do not hold.**

**1. Only 2 of Homebrew's 13 tokens work.** The Cask Cookbook documents `{{HOMEBREW_BREW_FILE}}`, `{{HOMEBREW_CELLAR}}`, `{{HOMEBREW_PREFIX}}`, `{{appdir}}`, `{{caskroom_path}}`, `{{name}}`, `{{staged_path}}`, `{{temp}}`, `{{token}}`, `{{user}}`, `{{version}}`, `{{version.major}}` and `{{version.major_minor}}`. A user who copies any of the other eleven into `install_steps` fails at `internal/pipe/cask/cask.go:390`, because the whole rendered Cask goes through `tmpl.Apply`:

```
template: failed to apply "<the entire generated Cask file>": function "token" not defined
```

`internal/tmpl/errors.go` removes the line number and quotes the full file, so the message does not say which option is at fault. The cause is very hard to find.

**2. "In the Homebrew Cask fields" is too broad.** `cask.go:303` applies templates to `hooks.*.install` and `hooks.*.uninstall` *without* these extra fields, so `{{ .StagedPath }}` there fails with `map has no entry for key "StagedPath"`. That is a real migration trap, because `deprecations.md` tells the user to replace `#{staged_path}` with `{{ .StagedPath }}`. In `desc` or `caveats` the field is accepted and writes a literal `{{staged_path}}` that Homebrew never expands.

Smallest fix: limit the sentence to `hooks.*.{install,uninstall}_steps`, and add one line with the escape form for the other tokens: `{{ "{{version}}" }}` renders `{{version}}`.

Test: a `TestFullPipe` case whose `install_steps` uses the escape form, with a golden that shows the raw `{{version}}` token in the output.

www/content/customization/publish/homebrew_casks.md
```diff
@@ -200,18 +200,26 @@ homebrew_casks:
 
     # Hooks for the cask lifecycle.
     #
+    # The body of each option uses Homebrew's declarative install steps DSL.
+    # It is not arbitrary Ruby.
+    # See https://docs.brew.sh/Cask-Cookbook for the allowed steps.
+    #
+    # Use `{{ .StagedPath }}` and `{{ .AppDir }}` for Homebrew's own
+    # `staged_path` and `appdir` tokens.
```
caarlos0: **The precedence rule is not documented here.**

`flightString` drops the deprecated `install` body when `install_steps` is set (see `internal/pipe/cask/template.go:117`). The rule is reasonable, and the PR description states it, but the reference document does not. A user who sets both during a migration gets only a "deprecated option" notice, which does not say that the old body is ignored.

Every other Cask deprecation in this pipe merges instead of discarding: `binary` into `binaries` and `manpage` into `manpages`, at `cask.go:72` and `cask.go:82`.

Smallest fix: add one comment line to this block, for example: "If both `install` and `install_steps` are set, `install_steps` wins and `install` is ignored."

Test: none needed for the document. If you prefer a warning in the code instead, extend `TestDefaultDeprecated` with a case that sets both fields.

www/content/customization/publish/homebrew_casks.md
```diff
@@ -200,18 +200,26 @@ homebrew_casks:
...
     hooks:
       pre:
-        install: |
-          system_command "/usr/bin/defaults", args: ["write", "com.example.app", "key", "value"]
-        uninstall: |
-          system_command "/usr/bin/defaults", args: ["delete", "com.example.app"]
+        # {{< g_inline_version "v2.19-unreleased" >}}
+        install_steps: |
+          run "/usr/bin/defaults", args: ["write", "com.example.app", "key", "value"]
+        uninstall_steps: |
```
caarlos0, line 222: **Three of the four new options have no version marker.**

`{{< g_inline_version "v2.19-unreleased" >}}` is above `pre.install_steps` only. `pre.uninstall_steps`, `post.install_steps` and `post.uninstall_steps` are new in the same release and carry no marker, so a reader cannot tell when they became available. The `# Templates: allowed. {{< g_inline_version "v2.13" >}}` line above `hooks:` makes this more confusing, because it dates the block to v2.13.

Smallest fix: move the marker into the `hooks:` comment block, or repeat it above each of the four new options.

internal/pipe/cask/template.go
caarlos0: Confirmed and fixed in d1b06a1cb.

I reproduced it: `flightString("preflight", "  \n ", `system_command "/usr/bin/true"`)` returned `""`. Now uses `strings.TrimSpace(steps) != ""`.

Also added `TestFlightString`, which covers steps, legacy, precedence, the blank-steps case and empty. I checked the new case fails when I revert the guard to `steps != ""`.

www/content/customization/general/templates.md
caarlos0: Both claims confirmed, fixed in d1b06a1cb.

I reproduced the first one: `{{version}}` fails with `function "version" not defined`. The section now says only the two fields exist, names the failure message, and shows the escape form for the other tokens.

On the second: you are right that `hooks.*.{install,uninstall}` are templated at `cask.go:303` without these fields. The section is now scoped to `hooks.*.{install,uninstall}_steps`.

Added the golden coverage you asked for: the `hooks_steps` case now writes `write_file "{{staged_path}}/v", "{{version}}"` through the escape form, and the golden shows the raw token. I also fed that golden to a real tap: `brew info` reads it.

www/content/customization/publish/homebrew_casks.md
caarlos0: Agreed, documented in d1b06a1cb. The `hooks:` comment now states that `install_steps` wins and `install` is ignored, and the same for the uninstall pair.

I kept the discard behavior rather than merging: unlike `binary`/`binaries`, the two bodies are different languages, so concatenating raw Ruby into a `_steps` block would produce a Cask that Homebrew rejects.

www/content/customization/publish/homebrew_casks.md
caarlos0: Fixed in d1b06a1cb. I moved the marker into the `hooks:` comment block, as you suggested: "All four `_steps` options are {{< g_inline_version "v2.19-unreleased" >}}". The stray marker above `pre.install_steps` is gone, and the `v2.13` line now reads as being about templates only.

internal/pipe/cask/cask.go
```diff
@@ -86,6 +86,18 @@ func (Pipe) Default(ctx *context.Context) error {
...
+		}
+		if brew.Hooks.Pre.Uninstall != "" {
+			deprecate.Notice(ctx, "homebrew_casks.hooks.pre.uninstall")
+		}
+		if brew.Hooks.Post.Install != "" {
+			deprecate.Notice(ctx, "homebrew_casks.hooks.post.install")
+		}
+		if brew.Hooks.Post.Uninstall != "" {
+			deprecate.Notice(ctx, "homebrew_casks.hooks.post.uninstall")
+		}
```
caarlos0, line 100: **Low, non-blocking: the notice does not say that the deprecated body is ignored.**

When a user sets `install` and `install_steps` on the same hook, `flightString` (`template.go:117`) drops the `install` body, and the only signal is:

```
DEPRECATED: homebrew_casks.hooks.pre.install should not be used anymore, check ... for more info
```

"Should not be used anymore" and "is being ignored right now" are different statements. During a migration — which is the whole point of this change — both options coexist exactly when the user is least sure of the outcome. The reference page now records the rule, which is the main fix, but a user who does not read it loses hook behaviour with no warning.

Smallest fix, using the existing helper and the `nfpms.maintainer` precedent at `internal/pipe/nfpm/nfpm.go:77`:

```go
if brew.Hooks.Pre.Install != "" {
	if brew.Hooks.Pre.InstallSteps != "" {
		deprecate.NoticeCustom(ctx, "homebrew_casks.hooks.pre.install",
			"{{ .Property }} is ignored, because install_steps is set, check {{ .URL }} for more info")
	} else {
		deprecate.Notice(ctx, "homebrew_casks.hooks.pre.install")
	}
}
```

That is four times over, so a small helper is probably better than repeating it.

Test: a case beside `TestDefaultDeprecated` (`cask_test.go:1293`) that sets both fields and asserts the deprecation is still recorded. Take it or leave it — the documentation fix already covers the main gap.

caarlos0 (commented):
**Looks good.** This supersedes my earlier review, [#pullrequestreview-5122805435](https://github.com/goreleaser/goreleaser/pull/6873#pullrequestreview-5122805435). GitHub does not allow an approval or a change request on your own pull request, so this is a comment. I cannot dismiss the earlier review either, because it is also a comment.

`d1b06a1cb` fixes all four findings from that review. I verified each one independently, not from the commit message:

1. **Blank `install_steps` no longer drops the deprecated body.** `strings.TrimSpace(steps) != ""` now agrees with `split`. I reimplemented the pre-fix function and ran the new table against both: only the `blank steps keep legacy` subtest fails on the old code, which is what a regression test must do. End to end, `InstallSteps: "   \n  "` with a real `Install` body now emits `preflight do ... end`, and the output contains no `preflight_steps`.
2. **`templates.md` is now accurate.** The fields are scoped to the `_steps` options, and the escape form is documented. The quoted error is verbatim: a bare `{{version}}` produces `... function "version" not defined`.
3. **The precedence rule is documented** in the reference page.
4. **One version marker now covers all four options.**

Other checks, all clean:

- The new `write_file "{{ .StagedPath }}/v", "{{ "{{version}}" }}"` fixture really survives the pipeline; the golden is byte-generated, not hand-written. `write_file` is in `CASK_ALLOWED_STEP_METHODS`, and its content goes through `expand_template_tokens`, which resolves `{{version}}`. So the escape example teaches something that works in Homebrew.
- No golden of the deprecated path changed. `TestFullCask` still contains all four legacy stanzas.
- CRLF bodies and interior blank lines render correctly; `bufio.Scanner` strips the `\r` and the sanitizer trims the blank line.
- `schema.json` is genuinely regenerated: `go run . schema` produces an identical file.
- `caskTokens` is not aliased. `WithExtraFields` copies into a copied template.

One low-priority suggestion is inline. It does not block.

**Rejected suggestion:** a guard that scans a `_steps` body for `system_command`, `if OS.mac?` or `#{` and warns. GoReleaser has never validated hook bodies, a three-pattern blocklist cannot cover a 26-method DSL, and the failure it would catch is loud at cask load time. Not worth the code.

Verification gaps:
- Build, lint and fmt were not run, at the request of the person who asked for this review.
- `TestFullPipe/git_remote` fails in my sandbox. It fails identically in `internal/pipe/krew` on code this pull request does not touch, so I treat it as environmental. All other cask tests pass, and the new and changed tests pass with `-count=2`.

This review is from an automated reviewer (GitHub Copilot CLI).

www/content/customization/publish/homebrew_casks.md
```diff
@@ -200,18 +200,31 @@ homebrew_casks:
...
+    # The body of each `_steps` option uses Homebrew's declarative install
+    # steps DSL. It is not arbitrary Ruby.
+    # See https://docs.brew.sh/Cask-Cookbook for the allowed steps.
+    #
+    # Use `{{ .StagedPath }}` and `{{ .AppDir }}` for Homebrew's own
+    # `staged_path` and `appdir` tokens.
+    #
+    # The `install` and `uninstall` options are deprecated. If you set both
+    # `install` and `install_steps`, `install_steps` wins, and `install` is
+    # ignored. The same applies to `uninstall` and `uninstall_steps`.
```
caarlos0: **Low, non-blocking: "ignored" is not quite true — the deprecated body is still template-validated, and can fail the release.**

`doRun` renders the four deprecated bodies unconditionally at `internal/pipe/cask/cask.go:304-309`, before `flightString` decides that steps win. So an `install` body that this page calls "ignored" is still a build gate.

The migration guide invites exactly this: `deprecations.md` tells the user that `#{staged_path}` becomes `{{ .StagedPath }}`. Someone who edits the body in place and then adds `install_steps`, or who renames only some of the four options, gets a failed release from a body that is never written to the Cask.

Reproduced on the merged tree, with a valid `post.install_steps` and `post.install: system_command "{{ .StagedPath }}/x"`:

```
template: failed to apply "system_command \"{{ .StagedPath }}/x\"": map has no entry for key "StagedPath"
```

Smallest fix is the wording here, not the code — for example "`install_steps` wins, and `install` is not written to the Cask. It is still checked as a template." Skipping the render instead would copy the precedence rule into `doRun`, which is why I withdrew my earlier suggestion about the notice text; the same argument applies here.

Test: if you do change the code rather than the wording, a `TestFullPipe` case with a valid `install_steps` and `install: '{{ .Nope }}'` asserting no error. For a wording-only fix, no test.

caarlos0 (commented):
**Still good.** This supersedes [#pullrequestreview-5123093632](https://github.com/goreleaser/goreleaser/pull/6873#pullrequestreview-5123093632). GitHub does not allow an approval on your own pull request, so this is a comment.

The branch has not changed. What changed is the base: `main` is now 43 commits ahead of the merge base, and three of them touch the same files — `#7086` (`rubyString`, `HasMixedPackageTypes`, `caskNameFor` in the cask pipe), `#7066` (rewrote YAML examples in `deprecations.md`, moved rows in `templates.md`), and `#7090` (`config.go`, `schema.json`). **The green checks on this pull request predate all of them**, so I merged locally and tested the result instead of trusting them.

Merge verification (scratch worktree built from `git merge-tree origin/main HEAD`, since removed):

- No textual conflict. The merge builds.
- `go test ./internal/pipe/cask/...` passes, except the known environmental `TestFullPipe/git_remote`.
- `go run . schema` reproduces `www/static/schema.json` byte for byte on the merged tree, so the regenerated schema is not stale.
- Both new goldens are unchanged on the merged tree, with `-count=2`. `#7086`'s `HasMixedPackageTypes` gate leaves `binary "hooks_steps"` in place, because the fixture is archive-only.
- The four flight stanzas keep their canonical position. `#7086` only touched `desc` and the `binary` guard, both above them.
- All four deprecation notices are still reachable, and their anchors still match headings in the merged `deprecations.md`. `#7066`'s only explicit `{#id}` is `{#snapsbuilds}`, which is unrelated.
- The new `deprecations.md` and `templates.md` sections already follow the conventions `#7066` established: every YAML example uses the `- ` list form.

On the `#7086` interaction specifically: `rubyString` pulls `.Description` into the Go-template pass and escapes `\`, `"`, newlines and `#{`. That **reduces** the double-render surface rather than widening it. This pull request's only change to the whole-file pass is two extra field names, so it can turn a previous error into a substitution but cannot change an existing success. I checked that `\#{`, `\"` and `\\` survive the second pass untouched, because they contain no `{{`.

I also checked the Homebrew runtime, because a stanza rename is not proof of equal behaviour: `UninstallPreflightSteps#uninstall_phase` and `UninstallPostflightSteps#uninstall_phase` call `run_steps(command)` with the default `:install` phase, so their steps really execute. The migration guide is correct. `PreflightSteps`/`PostflightSteps` do gain an `uninstall_phase`, but `run_uninstall_step` is a no-op except for `symlink` steps with `uninstall: true`, which GoReleaser never writes by itself.

One new low finding is inline. It does not block.

**I withdraw my round-2 finding** (make the notice say `install` is ignored). The fix would put the precedence rule in a second place: `flightString` decides it with `strings.TrimSpace(steps) != ""`, and `Default()` would have to repeat that condition four times. Two copies drift. The reference page already records the rule. Not worth it.

**Rejected suggestion:** pre-render the `_steps` fields in `dataFor` so a bad template gives a short error instead of quoting the whole Cask. That breaks the feature — pre-rendering turns `{{ .StagedPath }}` into `{{staged_path}}`, and the whole-file pass then fails with `function "staged_path" not defined`. The error shape is also the same for every other field rendered only in that pass, such as `caveats` and `custom_block`.

Verification gaps:
- Build, lint and fmt were not run on the branch itself, at the request of the person who asked for this review. I did build the merge result.
- `TestFullPipe/git_remote` fails in my sandbox. It fails identically in `internal/pipe/krew` on code this pull request does not touch.

This review is from an automated reviewer (GitHub Copilot CLI).

www/content/customization/publish/homebrew_casks.md
caarlos0: Confirmed and fixed in 7e1edbfcf, wording only, as you recommended.

I reproduced it: `tmpl.ApplyAll` on a deprecated body containing `{{ .StagedPath }}` fails with `map has no entry for key "StagedPath"`, because `cask.go` renders the four deprecated fields without the cask extra fields.

The page now reads: `install_steps` wins, and `install` is not written to the Cask; it is still checked as a template, so it must remain valid.

internal/pipe/cask/cask.go
caarlos0: Leaving this one, per your own withdrawal in the `homebrew_casks.md` thread. The reference page records the rule, which you called the main fix.

One concrete reason beyond the duplication argument: the precedence rule is `strings.TrimSpace(steps) != ""`, not `steps != ""`. I measured the naive form you sketched against a blank `install_steps`: it reports `steps win = true`, while `flightString` correctly keeps the legacy body. A second copy of the rule in `Default` would therefore print `is ignored` for a hook that is in fact still used, which is worse than the current generic notice.

Keeping the rule in one place. Happy to revisit if you would rather have the precedence extracted into a shared helper that both call.

sc250024:
Thank you for doing this @caarlos0 !

Dedsecproject:
Confirming from the field: [localstack/lstk](https://github.com/localstack/lstk) uses exactly the `post`-\>`install` hook you deprecate, so every `brew` operation on `localstack/tap/lstk` prints the `postflight` deprecation warning and points users here.

I've already patched the generated cask directly in localstack's tap ([localstack/homebrew-tap#8](https://github.com/localstack/homebrew-tap/pull/8)) with a manual `postflight_steps` conversion — the same shape this PR's golden files produce (`run` step + `on_macos` guard + `{{staged_path}}` token). That cask: `brew style` clean, loads as `Cask::Artifact::PostflightSteps`, and `brew install` successfully strips the quarantine attribute on macOS arm64.

For anyone watching this space, the permanent fix for tools like lstk is switching their `.goreleaser.yaml` to `hooks.post.install_steps` once this PR ships. One UX note on the deprecation here: until this lands, the only fix available to tap users is hand-editing generated casks, so the sooner this releases, the fewer noisy warnings propagate through third-party taps.

yermulnik:
Hey @caarlos0,
Just a heads up: now, that Homebrew `7.0.0` is out, it spits out deprecation notice on every run — hopefully you can expedite `v2.19` release if you have capacity and time for that 🙏🏻 

Appreciate your time and effort working on this tool ❤️

Issue #7060, opened 2026-09-05
refactor(tests): use a fixed minimal project in the Zig build integration test


caarlos0:
### Focused enhancement

[zig.TestBuild](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/builders/zig/build_test.go#L101-L151) runs `zig init` to create its project, although its assertions concern GoReleaser's build execution and artifacts, not Zig's generated example project.

An independent control intercepted only initialization and supplied a fixed minimal `build.zig` and `main.zig`. It then ran the unchanged existing test with real Zig compilation. Every existing assertion passed: custom environment and flags, artifact fields, output existence, and modification timestamp. The control recorded one fixture setup and one real build. No BEFORE/AFTER speedup was measured.

### Narrow change

Store or write that minimal project as the fixture instead of invoking `zig init`. Keep the real compiler invocation and all current assertions.

Preserve the per-test local cache and shared global cache introduced by [#6829](https://github.com/goreleaser/goreleaser/pull/6829). This is not a request to revisit the historical cache-corruption issue or to replace the integration test with a fake compiler.

---
Audit priority: **P3**.

Audit source revision: [628c889d7](https://github.com/goreleaser/goreleaser/commit/628c889d7f2a34066649833be419fdb9abe01174). AI disclosure: this report comes from a Copilot-assisted audit, with discovery by `caarlos0` agents and independent verification by `anvil` agents. Reproduction methods and limits are stated above; no fix is included in this issue.

Issue #7057, opened 2026-09-05
refactor(golang): remove the unused buildOutput helper and its isolated test


caarlos0:
### Focused enhancement

The private [buildOutput helper](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/builders/golang/build.go#L473-L482) removes `go: downloading` lines from a byte slice. Repository search and language-server references found no production caller; its only calls are in [TestBuildOutput](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/builders/golang/build_test.go#L2122-L2139). Actual Go builds execute through `base.Exec` instead.

Remove the helper and the test that exercises only that dead implementation. Do not transfer the filtering into `base.Exec`: that would be a new output behavior, not dead-code removal.

### Preserve coverage

Keep the real builder success, failure, and command-construction tests, including `TestBuildFailed` and `TestBuildGoBuildLine`. Their production path is unaffected. No exported symbol, configuration field, diagnostic text, or compiler invocation needs to change. This enhancement removes code with no reachable product behavior; it does not claim an output bug or runtime performance gain.

---
Audit priority: **P3**.

Audit source revision: [628c889d7](https://github.com/goreleaser/goreleaser/commit/628c889d7f2a34066649833be419fdb9abe01174). AI disclosure: this report comes from a Copilot-assisted audit, with discovery by `caarlos0` agents and independent verification by `anvil` agents. Reproduction methods and limits are stated above; no fix is included in this issue.

Issue #7055, opened 2026-09-05
refactor(tests): use mktmp for invalid build and release configuration


caarlos0:
### Focused enhancement

[TestBuildInvalidConfig](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/cmd/build_test.go#L46-L52) and [TestReleaseInvalidConfig](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/cmd/release_test.go#L36-L42) call `setup`, then overwrite the configuration with invalid YAML. Loading fails before either pipeline needs the generated Go module or Git history.

The existing [mktmp helper](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/cmd/util_test.go#L20-L53) provides the required directory and environment setup without nine Git launches per case. Independent controls omitted the project setup and preserved both exact YAML errors, including their different line numbers. Elapsed savings were not measured.

### Narrow change

Replace `setup(t)` with `mktmp(t)` only in these two tests. Keep their invalid inputs, command flags, and exact assertions. Leave valid-pipeline and broken-project tests alone.

This follows [#6831](https://github.com/goreleaser/goreleaser/pull/6831), which added the helper and simplified config-discovery tests; these two remaining callers were not converted.

---
Audit priority: **P3**.

Audit source revision: [628c889d7](https://github.com/goreleaser/goreleaser/commit/628c889d7f2a34066649833be419fdb9abe01174). AI disclosure: this report comes from a Copilot-assisted audit, with discovery by `caarlos0` agents and independent verification by `anvil` agents. Reproduction methods and limits are stated above; no fix is included in this issue.

PR #7067, opened 2026-09-05
test: make tests deterministic and repair fuzz tasks

### LINKED ISSUE #6957: task fuzz: aggregate task exits successfully without running either fuzz target
### Reproduction

With Task 3.53.1, run:

```sh
task --dry fuzz
task --summary fuzz
task --summary cmds
```

### Observed and expected

`task --dry fuzz` produced no output and exited 0. Its summary showed a description but no commands. The separate `cmds` task summary contained the intended `fuzz:tmpl` and `fuzz:artifact` child tasks.

Executing the actual `task fuzz` command with a recording Go boundary also exited 0 without invoking Go. Controls for `task --dry fuzz:tmpl` and `task --dry fuzz:artifact` each printed their expected `scripts/fuzz.sh` invocation. No fuzz engine or Go build was needed to demonstrate the missing aggregate dispatch.

The advertised aggregate task should invoke both child tasks, rather than report success without doing the requested work.

### Cause and correction

In [Taskfile.yml](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/Taskfile.yml#L56-L60), `cmds:` is aligned with `fuzz:`. It is parsed as another task instead of the command list belonging to `fuzz`.

Indent `cmds:` and its entries beneath `fuzz`. Add a regression checking the aggregate's dry-run output and actual child dispatch with a recording Go executable. Keep the direct child-task controls.

---
Audit priority: **P2**.

### LINKED ISSUE #6958: fuzz wrapper: failing fuzz invocations are hidden by a successful final ordinary test
### Reproduction

Put a recording `go` stub first on PATH. Make it return 42 for calls containing `-fuzz=`, and 0 for the final ordinary test call. Run the unchanged wrapper:

```sh
bash scripts/fuzz.sh ./internal/tmpl 1ms
```

### Observed and expected

The wrapper made five fuzz invocations, each returning 42, then called `go test ./internal/tmpl/...`, which returned 0. The wrapper itself exited 0. Calling it through the actual `task fuzz:tmpl` task also returned 0.

Running the same unchanged script under `bash -eo pipefail` stopped at the first injected fuzz failure and returned 42. These were controlled process failures, not a claim that the real fuzz engine found a crashing input.

A fuzz-only failure must make the wrapper and task fail even when a later ordinary test succeeds. This does not imply that every persisted crashing seed is hidden: such a seed can also fail the final ordinary test.

### Cause and correction

[The wrapper](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/scripts/fuzz.sh#L5-L11) neither stops on failed loop commands nor preserves their status past the final command.

Enable fail-fast pipeline handling, such as `set -eo pipefail`. Add failing-fuzz and all-success boundary controls, and verify failure propagation through the child Task command.

---
Audit priority: **P2**.

### LINKED ISSUE #6962: tests: bound and stabilize FuzzChecksumLargeData inputs
### Problem

[FuzzChecksumLargeData](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/artifact/artifact_fuzz_test.go#L43-L65) allocates `make([]byte, size)` directly from a fuzzed integer. Adding only the seed `("sha256", -1)` to the existing target produced `panic: runtime error: makeslice: len out of range` at line 52, before `Artifact.Checksum` was called. This is a harness failure, not a checksum defect. Unbounded positive sizes can also make allocation size, rather than checksum behavior, dominate the target.

The bytes come from `crypto/rand`, so the recorded algorithm and size do not reproduce the actual file contents.

### Narrow change

Handle negative and oversized inputs before allocation, and generate deterministic contents from recorded fuzz input. Keep negative, zero, and bounded-size seeds, all supported algorithms, and the production checksum call. Preserve `FuzzChecksum`, which already covers caller-supplied byte contents. The negative seed should no longer terminate the fuzz worker in fixture setup.

---
Audit priority: **P2**.

### LINKED ISSUE #6963: tests(sign): isolate the Git configuration in TestBinarySignDefault
### Problem

[TestBinarySignDefault](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/sign/sign_binary_test.go#L20-L30) expects the command to be `gpg`, but production intentionally honors Git's configured signing program.

The unchanged test failed under an environment-only override: `GIT_CONFIG_COUNT=1`, `GIT_CONFIG_KEY_0=gpg.program`, and `GIT_CONFIG_VALUE_0=gpg2`. Its assertion reported expected `gpg`, actual `gpg2`. The empty-program control passed. No signing operation or external service is needed to trigger this difference.

### Narrow change

Give the test controlled Git configuration, including environment-provided configuration, rather than inheriting the contributor's global settings. Keep an explicit configured-program case as well as the default-program case. Do not change production to ignore a legitimate `gpg.program` value.

Preserve the existing signature-name, argument, and artifact-selection assertions. Running `go test ./internal/pipe/sign -run '^TestBinarySignDefault$' -count=1` under either host configuration should test the same declared defaults.

---
Audit priority: **P2**.

### LINKED ISSUE #6964: tests(sign): require cosign only for the real-cosign subtest
### Problem

[TestDockerSignArtifacts](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/sign/sign_docker_test.go#L79-L139) calls `testlib.CheckPath(t, "cosign")` before selecting any subtest. Most cases now use a shell-based fake signer or a certificate copy; only `no signature file` needs real cosign.

With CI mode unset, the same fake-only selector passed when cosign was available and skipped the parent test when it was absent. `TestDockerSignArtifacts/sign_all` is an existing example: it checks artifact filtering and generated signature files without executing cosign.

### Narrow change

Move the prerequisite into the real-cosign case. Keep the fake cases runnable without cosign, including their exact artifact names, nonempty file assertions, and certificate parsing.

This is a remaining coverage gap after [#6842](https://github.com/goreleaser/goreleaser/pull/6842) and [#6836](https://github.com/goreleaser/goreleaser/pull/6836), which introduced and consolidated the fake signer. It does not claim that the remaining real-GHCR case currently flakes.

---
Audit priority: **P2**.

### LINKED ISSUE #6965: tests(winget): make error cases assert the intended failure
### Problem

The existing `TestRunPipe/mixed-formats` passes for the wrong reason. Its [binary fixture](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/winget/winget_test.go#L978-L989) points to `doesnt-matter`, so generation fails while checksumming a missing file, before reaching `errMixedFormats`. Fixing only the path then exposes missing `artifact.ExtraBinary` metadata.

The [run-error assertion](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/winget/winget_test.go#L1008-L1012) also uses `expectPublishErrorIs`, not the run expectation. Taking a pointer to that `error` interface makes `ErrorAs` accept any nonnil error. The publish assertion has the same broad target.

### Narrow change

Create a real binary fixture with the metadata required to reach the intended branch. Assert sentinels with `ErrorIs`, and use concrete error types where type matching is intended. Use the matching run or publish expectation.

Retain the successful manifest golden checks. The mixed-format case must fail if format validation is removed, rather than remaining green because unrelated fixture setup failed.

---
Audit priority: **P2**.

### LINKED ISSUE #6966: tests(docker): remove public-service setup from the denied-push case
### Observed failure

[CI job 101364790990](https://github.com/goreleaser/goreleaser/actions/runs/33987896138/job/101364790990), running `task test TEST_OPTIONS=-count=1`, failed in `TestRunPipe/no_permissions_on_docker`. Docker Hub returned `502 Bad Gateway` to `HEAD /v2/library/alpine/manifests/latest`. The image build failed before the intended denied-push assertion ran.

The [case configuration](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/docker/docker_test.go#L892-L907) and [Dockerfile](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/docker/testdata/Dockerfile#L1-L2) match the audited revision. The fixture uses Alpine only to add the test binary. This was an observed external-service failure, not a Go race-detector or scheduling failure.

### Narrow change

Use a scratch-based fixture for the unrelated image-build step. To remove public-service dependence from the whole case, also use a controlled denied-write registry and test-owned setup resources; changing the base image alone is not sufficient.

Keep a real Docker publication attempt and assert that the controlled registry denied the push. Do not accept arbitrary build failures, add retries, or weaken the expected error.

---
Audit priority: **P2**.

### LINKED ISSUE #6967: tests(changelog): keep the GitLab selector case off the public network
### Problem

The GitLab case in [TestGetChangeloger](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/changelog/changelog_test.go#L764-L774) only needs to assert which changeloger implementation is selected. Constructing that implementation also initializes a GitLab client, using the public endpoint when no API URL is supplied.

A local recording control observed `GET /api/v4/version` when `CI_SERVER_VERSION` was empty. Supplying the fixed value `18.0.0` avoided HTTP entirely while selecting the same implementation. This establishes an unnecessary network dependency; it is not a measured slowdown or a report of random CI failure.

### Narrow change

Set and restore the version input for this selector test, or provide a controlled local version response. Keep the GitLab-with-previous-tag and no-previous-tag type assertions intact. Test version discovery separately at the client boundary, not as a prerequisite for selecting a changelog implementation. Keep this change separate from the misplaced header/footer assertions.

---
Audit priority: **P2**.

### LINKED ISSUE #6968: tests(gitlab): move header and footer error assertions to the rendering layer
### Problem

[TestGitLabCreateReleaseHeaderTemplateError and TestGitLabCreateReleaseFooterTemplateError](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/client/gitlab_test.go#L1521-L1536) call `client.CreateRelease(ctx, "body")`. That method receives an already-rendered body; it does not render the configured header or footer.

Redirecting the calls to controlled local responses demonstrated the gap. Valid and invalid templates both satisfied the existing `require.Error` with HTTP 401. Both succeeded with successful release responses. The original destination was public GitLab. A separate control through the real release-body rendering layer rejected the missing key and accepted the valid template.

### Narrow change

Place these template assertions in the release-body tests, with explicit template-error checks and valid-template controls. Keep client HTTP error coverage separate and local.

This remains after [#6848](https://github.com/goreleaser/goreleaser/pull/6848) and [#6854](https://github.com/goreleaser/goreleaser/pull/6854): those repaired the download-URL test, not header/footer rendering coverage. No production rendering change is requested.

---
Audit priority: **P2**.

### LINKED ISSUE #6969: tests(cmd): give TestSetupPipeline independent expected storage
### Problem

In [TestSetupPipeline](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/cmd/build_test.go#L93-L102), the expected slice is created by appending to `pipeline.BuildCmdPipeline`. [setupPipeline](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/cmd/build.go#L147-L152) appends to the same backing array when constructing the actual result.

The existing test passed with expected output `"foobar"` and input output `"."`. An independent control recorded spare capacity and the shared tail: evaluating the actual result changed the expected output from `"foobar"` to `"."`. Copying expected storage independently exposed the mismatch.

### Narrow change

Construct expected pipeline slices with independent storage before calling production code. Correct this case's expected output to match its input, and retain the separate explicit-output cases.

The demonstrated defect is the test assertion, not production handling of `"."`. Preserve the complete pipeline-order assertions and show that an intentionally wrong output value is detected.

---
Audit priority: **P2**.

### LINKED ISSUE #6970: tests(github): synchronize rate-limit cancellation after handler entry
### Problem

[TestGitHubCheckRateLimit](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/client/github_test.go#L972-L1002) starts a 250 ms context before client and request setup, then requires the HTTP handler to have run. If the deadline expires first, production correctly avoids making a canceled request, but the final assertion fails.

The unchanged test passed in the recorded control. An injected 251 ms virtual scheduling pause in a controlled reproduction produced `context deadline exceeded` and `called=false`, failing `should have checked rate limit`. This proves deadline sensitivity, not an observed CI flake or its frequency.

### Narrow change

Synchronize on handler entry before cancellation, or use a controlled clock and transport that establish the same order. Retain the request assertion and the cancellation behavior; do not merely increase the deadline or remove the assertion.

This is separate from [#6846](https://github.com/goreleaser/goreleaser/pull/6846), which fixed a GitLab retry-after expectation. The GitHub test still contains the setup-before-deadline assumption.

---
Audit priority: **P2**.

### LINKED ISSUE #6971: tests(github): isolate the bad target_commitish template test
### Problem

[TestGitHubCreateReleaseTargetCommitishBadTemplate](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/client/github_test.go#L2280-L2303) configures an invalid template but leaves the GitHub API endpoint at its public default. `CreateRelease` performs a rate-limit check before reaching the intended missing-key error.

A recording-only transport captured `GET https://api.github.com/rate_limit`, followed by the expected template error. No real GitHub request was sent during that verification. The result shows that an otherwise local template assertion depends on network behavior before it reaches the condition it is meant to test.

### Narrow change

Use the existing `githubTestServer` helper and provide only the local responses required before template evaluation. Retain `RequireTemplateError` and the invalid `target_commitish` input. Assert that no release mutation request is made after the template fails.

Do not alter production rate-limit ordering to make this unit test local. The defect is its uncontrolled test boundary.

---
Audit priority: **P2**.

### LINKED ISSUE #6972: tests(upload): inject connection refusal instead of assuming port 1234 is unused
### Problem

[TestRunPipe_ServerDown](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/upload/upload_test.go#L375-L406) and [TestRunPipe_ArtifactoryDown](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/artifactory/artifactory_test.go#L341-L372) assume that `localhost:1234` refuses connections. A contributor or CI service can legitimately own that port.

The original tests were not run for this verification. Instead, an isolated probe exclusively reserved both loopback port-1234 endpoints and pinned HTTP traffic to its owned listener. For both providers, the same production publication call received HTTP 201 and returned nil after one request. That contradicts the original `require.Error` and connection-refused assumptions.

### Narrow change

Inject the expected connection failure through a controlled HTTP transport or dialer, restoring any shared client state in cleanup. Do not select another supposedly unused fixed port or probe-and-close a port before using it.

Keep both providers' error propagation checks, including the intended `errors.Is` behavior, and retain their normal successful-upload coverage.

---
Audit priority: **P2**.

### LINKED ISSUE #6973: tests(announce): clear webhook variables in missing-environment cases
### Problem

The existing `TestAnnounceMissingEnv` cases for [Slack](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/slack/slack_test.go#L82-L90) and [Mattermost](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/mattermost/mattermost_test.go#L40-L48) assume the relevant webhook variable is absent, but do not clear it.

The unchanged tests passed with empty variables. With inherited `SLACK_WEBHOOK=://invalid` or `MATTERMOST_WEBHOOK=://invalid`, they failed before network access: expected the missing-variable error, actual `failed new request: parse "://invalid": missing protocol scheme`.

### Narrow change

Use `t.Setenv` to clear each provider's webhook before constructing its context. Preserve the exact missing-environment assertion and keep the configured-webhook cases separate.

A focused reproduction is `go test ./internal/pipe/slack ./internal/pipe/mattermost -run '^TestAnnounceMissingEnv$' -count=1` under those inherited values. Limit the issue to these two reproduced cases; analogous code in other providers is not evidence that those tests also fail.

---
Audit priority: **P2**.

### LINKED ISSUE #6974: tests(docker): restore fake imager registrations after retry tests
### Problem

[TestDockerBuildRetries](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/docker/docker_test.go#L1574-L1603) registers `test-build-retriable` and `test-build-fatal` globally without cleanup. [TestValidateImager](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/docker/docker_test.go#L1535-L1554) expects only the normal registered implementations.

Running the unchanged tests with `go test ./internal/pipe/docker -run '^(TestValidateImager|TestDockerBuildRetries)$' -count=2` passed the first iteration and failed the second. The expected options were `[buildx docker]`; the actual options also contained both fake names. This is persistent test state, not a Docker daemon or network failure.

### Narrow change

Restore the prior registry entries through `t.Cleanup`, deleting only test-owned entries when no prior value existed. Keep the real registry defaults and production registration behavior unchanged.

Retain the transient/fatal retry-count assertions and the repeated-run reproduction so a later test cannot silently depend on fake implementations left by an earlier one.

---
Audit priority: **P2**.

### LINKED ISSUE #6975: tests(docker): control provenance capability detection in TestBuildCommand
### Problem

[TestBuildCommand](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/docker/docker_test.go#L1146-L1189) expects provenance and SBOM flags on every non-Windows host. Production instead uses a [capability probe](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/docker/api_docker.go#L89-L96).

With Docker absent from PATH, the unchanged `TestBuildCommand` failed all four cases. Actual command lines correctly omitted `--provenance=false` and `--sbom=false`, while the expected slices included them. The test therefore depends on the installed Docker CLI and its cached probe result even though it is checking command construction.

### Narrow change

Control and restore the capability function in this test. Cover supported and unsupported results explicitly, preserving the existing image, flag, and buildx argument assertions.

[#6827](https://github.com/goreleaser/goreleaser/pull/6827) retained the production probe for valid reasons. This request does not remove that probe or infer a product failure when Docker is absent; it isolates the unit test's input.

---
Audit priority: **P2**.

### LINKED ISSUE #6976: tests(git): pin the committer timestamp in TestSingleCommit
### Problem

[TestSingleCommit](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/git/git_test.go#L63-L92) pins an old author date but leaves the committer date inherited, then requires the reported commit date to be within one minute of wall time.

The unchanged test failed with `GIT_COMMITTER_DATE=2001-01-01T00:00:00Z go test ./internal/pipe/git -run '^TestSingleCommit$' -count=1`. Production returned that exact committer timestamp; the wall-time comparison failed. A long scheduling delay can invalidate the same assertion, but no CI flake frequency is claimed.

### Narrow change

Pin distinct author and committer dates and assert the exact committer instant. Keep the full/short hash assertions and the distinction between `%ct` and `%at`.

This preserves the regression coverage added by [#6850](https://github.com/goreleaser/goreleaser/pull/6850), which consolidated Git queries. It removes ambient clock and environment inputs without changing production commit-date selection or weakening the assertion.

---
Audit priority: **P2**.

### LINKED ISSUE #6977: tests(env): isolate inherited token selection and token files
### Problem

The environment package's [TestMain](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/env/env_test.go#L16-L30) clears three token variables but leaves `GORELEASER_FORCE_TOKEN` and default token-file locations uncontrolled.

The unchanged [TestValidGithubEnv](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/env/env_test.go#L115-L121) failed under an inherited GitLab selector: it set its fake GitHub token, production selection discarded it, and the assertion received `missing GITHUB_TOKEN, GITLAB_TOKEN and GITEA_TOKEN`. The focused reproduction uses `GORELEASER_FORCE_TOKEN=gitlab go test ./internal/pipe/env -run '^TestValidGithubEnv$' -count=1` with isolated empty token-file locations.

### Narrow change

Clear provider-selection state before tests load it, and isolate token files from the contributor's home directory. Restore process-global state after the tests that change it. Keep explicit force-provider, multiple-token, file-token, and missing-token cases.

Do not change production token-selection semantics to accommodate this test assumption. The failure concerns inherited test inputs, separate from functional force-token error handling.

---
Audit priority: **P2**.

### LINKED ISSUE #6978: tests(golang): build once in TestBuildModTimestamp
### Redundant work

[TestBuildModTimestamp](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/builders/golang/build_test.go#L1381-L1434) loops over five platform labels, but every call passes `mustParse(t, runtimeTarget)` to `Default.Build`. Only output directories change; this is not five-target coverage.

A recording wrapper ran the unchanged test and forwarded every invocation to real Go. All five builds used the host target, `darwin/arm64`. Individual commands took 0.10-0.13 seconds, and the recorded test body took approximately 0.647 seconds, including wrapper overhead. There is no AFTER measurement or claimed speedup.

### Narrow change

Keep one host build with the same templated binary path, environment, compiler flags, and `ModTimestamp`, then assert the resulting binary's exact timestamp. Preserve the separate tests that actually exercise target parsing and cross-target build behavior.

This is distinct from [#6834](https://github.com/goreleaser/goreleaser/pull/6834), which removed duplicate Python build-mode tests. Do not remove the timestamp integration assertion or replace the real compiler with a fake.

---
Audit priority: **P2**.

caarlos0:
## Summary

Address 19 remaining test-suite and fuzz-task issues, plus focused CI and review fixes.

- Control inherited environment, Git configuration, timestamps, shared registries, and HTTP failures.
- Replace unintended public-service dependencies with local responses and a scratch-based denied-push registry.
- Match full container names during cleanup so legacy Docker setup cannot remove Docker v2's registry.
- Correct error assertions and fixtures so they fail for the intended reason.
- Remove repeated timestamp-test builds and give pipeline expectations independent storage.
- Pin CI to Zig 0.16.0 and copy only the minimal Zig project fixture.
- Bound deterministic fuzz inputs, select exact targets, preserve failures, handle packages without fuzz functions, and restore aggregate task dispatch.

## Why

The affected tests depended on host state, timing, or public services. Some assertions accepted unrelated failures, while the fuzz tasks could report success without doing the requested work.

CI also exposed cross-test container deletion: Docker's partial-name lookup let legacy registry setup remove `alt_registry-v2`. Cleanup now matches the full literal name.

These changes preserve release behavior and public APIs. The proposed `goreleaser test` command in #2355 is outside this PR. Restoring Go module-download output filtering is tracked separately in #7072 and is also outside this PR.

## Validation

Targeted Go tests passed, including race detection and repeated/shuffled isolation cases. Real Go and Zig builds, Docker/buildx denied pushes, and fuzz-wrapper execution passed. Lint reported no new issues in the changed packages.

The WinGet mixed-format test fails when its production guard is removed. The copyright URL case now fails when its invalid URL template is cleared. The invalid fuzz seed and incorrect pipeline expectation also reproduced their reported failures before correction.

A local Docker API regression reproduced wrong-container deletion and now covers neighboring names, literal dots, a missing target, and fatal-error reporting for failed deletions without a live Docker daemon.

The Go harness for shell scripts was removed during review. Manual controls confirm that packages without fuzz functions still run ordinary tests, failing fuzz targets retain exit code 42, and source-read errors stop before invoking Go.

Coverage requirements are unchanged.

## Related issues

- Environment, Git, and signing: Fixes #6963, fixes #6964, fixes #6973, fixes #6976, fixes #6977.
- Client and changelog boundaries: Fixes #6967, fixes #6968, fixes #6970, fixes #6971.
- Docker: Fixes #6966, fixes #6974, fixes #6975.
- Upload and WinGet assertions: Fixes #6965, fixes #6972.
- Command and builder tests: Fixes #6969, fixes #6978.
- Fuzz tasks and inputs: Fixes #6957, fixes #6958, fixes #6962.

#7055, #7057, and #7060 already landed in #7068. This PR preserves those changes and adds the reviewed Zig version pin and fixture-directory isolation.

AI assistance: implemented with GitHub Copilot.


internal/pipe/env/env_test.go
```diff
@@ -10,31 +10,28 @@ import (
...
-			_ = os.Unsetenv(key)
-			restores[key] = prevValue
-		}
-	}
-
-	m.Run()
-
-	for k, v := range restores {
-		_ = os.Setenv(k, v)
+func isolateEnv(t *testing.T) {
```
caarlos0:
```suggestion
func isolateEnv(tb *testing.TB) {
```

scripts/fuzz_test.go
```diff
@@ -0,0 +1,108 @@
+package scripts
```
caarlos0: we dont need to write tests for shell scripts... remove

internal/testlib/docker_test.go
```diff
@@ -0,0 +1,108 @@
+package testlib
```
caarlos0, line 1: isn't this already tested "by proxy" by its callers?

if not, can we remove it?

if not, we should run something cheaper than the registry for the purposes of this test... (e.g. `scratch` with a `sleep` or something similar)

internal/pipe/env/env_test.go
caarlos0: Changed in 50f3e2a2cdd1f51570fb6c55292889bbbf640a69: the helper now takes `tb testing.TB`, matching the other test helpers. I used the interface directly, rather than `*testing.TB`. The environment tests still pass with inherited token variables.

scripts/fuzz_test.go
caarlos0: Removed `scripts/fuzz_test.go` in bd3692b78387d6b7722aedcca18fd8efdc971f1f as requested. The fixes in `scripts/fuzz.sh` and `Taskfile.yml` remain unchanged.

internal/testlib/docker_test.go
caarlos0: The callers cover ordinary cleanup, but not the neighboring-name collision or deletion-error reporting. This test already uses only `httptest.NewServer` with fake Docker API responses: it starts no Docker process, registry, or container. All four cases passed with `PATH=/nonexistent`, and the test body took 0.01s. I kept the focused regressions from 88a90d4f7 and 690d257f8 for those uncovered cases. The separate denied-push integration fixture already uses `FROM scratch`.

internal/pipe/winget/winget_test.go
```diff
@@ -560,8 +573,8 @@ func TestRunPipe(t *testing.T) {
 			},
 		},
 		{
-			name:             "bad-copyright-url-tmpl",
-			expectRunErrorIs: &template.Error{},
+			name:           "bad-copyright-url-tmpl",
+			expectRunError: &tmpl.Error{},
```
caarlos0, line 577: This case still passes for the wrong reason, which is the thing this PR set out to remove.

`Name` on line 579 is also `{{ .Nope }}`. `doRun` calls `tp.ApplyAll(&winget.Publisher, &winget.Name, ..., &winget.CopyrightURL, ...)` and `ApplyAll` returns on the first failure (`internal/tmpl/tmpl.go:334-344`), so it fails on `Name`, the second entry, and never templates `CopyrightURL`, the seventeenth. `assertError`'s `*tmpl.Error` branch only checks `errors.As`, so the case is satisfied by the `Name` failure. Delete `CopyrightURL` templating from production and this test still passes; it is a duplicate of `bad-name-tmpl`.

Smallest fix: set `Name: "foo"` on line 579, as `bad-license-url-tmpl` and the other single-field cases already do.

Test: with `Name` valid, `CopyrightURL` becomes the only bad template, so the existing `&tmpl.Error{}` expectation becomes decisive. Confirm by clearing `CopyrightURL` as well and checking the case fails.

caarlos0 (commented):
Reviewed `bbe26fa5..bd3692b78`. The branch moved during the review; this covers `50f3e2a2c` and `bd3692b78` too.

The core of this change is correct, and two of the repairs are real. `require.ErrorAs(t, err, &tt.expectPublishErrorIs)` took a `*error` target, so every one of the ~30 WinGet run-error assertions matched any error. In `cmd/build_test.go`, `pipeline.BuildCmdPipeline` is itself built with `append`, so `cap > len`; the expected slice and `setupPipeline` shared one backing array and the expectation was overwritten before comparison. The `testlib` container filter is a genuine cross-package fix: `docker_test.go` starts `registry`/`alt_registry` and `v2/docker_integration_test.go` starts `registry-v2`/`alt_registry-v2`, and Docker's `name` filter is an unanchored regex. The two deleted GitLab tests were both vacuous and networked, because `gitlabClient.CreateRelease` never renders `Release.Header` or `Release.Footer`; `body_test.go` puts that coverage on `describeBody`, where the templates actually run.

Five notes are inline. Only the WinGet one is in scope for this PR.

**Out of scope, for a later PR.** Eight sibling announce packages still have the exact ambient-environment bug that `slack` and `mattermost` got fixed here: `discord`, `teams`, `twitter`, `mastodon`, `bluesky`, `reddit`, `linkedin`, `discourse`. Reproduced:

```
$ DISCORD_WEBHOOK_ID=x DISCORD_WEBHOOK_TOKEN=y go test ./internal/pipe/discord/ -run TestAnnounceMissingEnv
    expected: "env: environment variable \"DISCORD_WEBHOOK_ID\" should not be empty; ..."
    actual  : "400 Bad Request"
FAIL
```

The test reaches `discord.com`. One `t.Setenv("<VAR>", "")` each, as in `slack_test.go:83`.

**Missing coverage.** `scripts/fuzz.sh` and the `Taskfile.yml` `fuzz` indentation now have no automated test, because the harness was removed on request. That is your call and I am not reopening it; I verified both by hand instead. Unanchored `-fuzz=FuzzChecksum` makes `go test` exit 1 because it matches two targets, so `fuzz:artifact` never fuzzed; the anchored form runs both targets and a failing target now aborts the script with its own exit code.

**Verification gaps.** No Docker daemon on this host, so `TestPushDenied` and the Docker-gated `TestRunPipe` cases never ran. `internal/pipe/sign` `TestSignArtifacts` and `TestDockerSignArtifacts/no_signature_file` fail here for environment reasons unrelated to the diff: the installed cosign rejects `--upload=false` without `--bundle`, and this sandbox denies `~/.docker/config.json`. Everything else passed: `testlib`, `env`, `git`, `client`, `release`, `upload`, `artifactory`, `slack`, `mattermost`, `changelog`, `artifact`, `docker`, `cmd`, `builders/golang`, `builders/zig` (zig 0.16.0), and `TestBinarySignDefault`.

Disclosure: this review was written by an automated reviewer, not by a human.

internal/builders/zig/testdata/proj/build.zig
```diff
@@ -0,0 +1,13 @@
+const std = @import("std");
+
+pub fn build(b: *std.Build) void {
+    const exe = b.addExecutable(.{
+        .name = "proj",
+        .root_module = b.createModule(.{
```
caarlos0, line 6: Non-blocking, but this trades a flaky fixture for a fixture that expires.

`.root_module = b.createModule(...)` pins the `std.Build` API. `zig init` produced a `build.zig` that matched whichever compiler was installed, so this test used to survive Zig releases on its own. `.github/workflows/build.yml:162` uses `mlugg/setup-zig` with no `version:`; the action documents the empty default as "use `minimum_zig_version` from `build.zig.zon`, with a fallback to `latest`", and there is no `build.zig.zon` in this repository. CI therefore floats to the newest tagged Zig, and the next `std.Build` break lands as red CI on an unrelated PR.

Smallest fix: pin `version:` on the `setup-zig` step.

Test: none needed beyond the existing `TestBuild`, which passes on zig 0.16.0 here (4.9s).

Separately, `gio.Copy("testdata", ...)` in `build_test.go` copies the whole directory, so `main.c`, `targets.sh`, `version.txt`, and `.gitignore` also land in the generated project. Harmless, but a `testdata/proj/` subdirectory would keep the fixture minimal.

internal/builders/golang/build.go
```diff
@@ -470,17 +470,6 @@ func validateUniqueFlags(details config.BuildDetails) {
...
-	for line := range strings.SplitSeq(strings.TrimSpace(string(out)), "\n") {
-		if strings.HasPrefix(line, "go: downloading") {
-			continue
-		}
-		lines = append(lines, line)
-	}
-	return strings.Join(lines, "\n")
-}
-
 func artifactType(t Target, buildmode string) artifact.Type {
```
caarlos0, line 473: The removal above is correct: `git log -S buildOutput` shows the only caller went away in `516d9bccf` (bun support, #5403), which routed the Go builder through `base.Exec`, and the helper has been unreachable since. No behavior change here.

Worth recording before the evidence disappears: the fix from `9cf3bbbc5` (#4869, "do not print 'go: downloading' bits of go build output") went with it. `internal/builders/base/build.go:119` logs the raw `CombinedOutput`, and `grep -rn "go: downloading" --include=*.go` finds no remaining filter outside tests. `goreleaser build` against a cold module cache prints those lines again.

Smallest fix, in a separate PR: filter inside `base.Exec` so every builder benefits, and move the deleted `TestBuildOutput` cases to `internal/builders/base`. Please open an issue rather than growing this one.

scripts/fuzz.sh
```diff
@@ -1,11 +1,13 @@
 #!/bin/bash
+set -eo pipefail
```
caarlos0, line 2: Non-blocking. `set -eo pipefail` is the right call and I confirmed a failing target now aborts the loop and exits with the target's own code.

It also couples `grep` to script failure, though. `grep "func Fuzz" "$pkg"/*.go` exits 1 when a package has no fuzz function, and with `pipefail` plus `-e` that aborts before the trailing `go test "$pkg"/...`, so the package's unit tests are silently skipped and the task fails with no output. Both current packages have fuzz targets, so this is not reachable from `fuzz:tmpl` or `fuzz:artifact`; it will bite the first `fuzz:<pkg>` target added for a package without one.

Smallest fix: `grep "func Fuzz" "$pkg"/*.go || true |` — or collect the names into a variable first, which also drops the subshell.

internal/pipe/docker/denied_test.go
```diff
@@ -0,0 +1,114 @@
...
+	logctx, cancel := stdctx.WithCancel(t.Context())
+	t.Cleanup(cancel)
+	logs := exec.CommandContext(logctx, "docker", "logs", "--follow", containerID)
+	stdout, err := logs.StdoutPipe()
+	require.NoError(t, err)
+	logs.Stderr = logs.Stdout
+	require.NoError(t, logs.Start())
+	t.Cleanup(func() {
+		cancel()
+		require.Error(t, logs.Wait(), "the log follower should stop on cancellation")
```
caarlos0: Low severity, and I could not execute this file: no Docker daemon on this host.

The log follower is scaffolding, not the subject under test, but this line can fail the test on its own. If the registry container dies during the run, `docker logs --follow` exits 0 by itself and `logs.Wait()` returns nil, so the cleanup reports "the log follower should stop on cancellation" instead of whatever actually killed the registry. That is exactly the run where you need the real error.

Smallest fix: `_ = logs.Wait()`. The readiness assertion on line 70 already covers the case where the registry never starts.

Test: none; this only affects diagnosis.

internal/pipe/winget/winget_test.go
caarlos0: Fixed in c876a2f59709f71ed8f5396c0220353ce7a935de. Name is now valid, leaving CopyrightURL as the only invalid template. The focused case passes; clearing CopyrightURL in an overlay now makes its template-error assertion fail. Production templating is unchanged.

internal/builders/zig/testdata/proj/build.zig
caarlos0: Fixed in 1986e0603eee80613b130f51a33d050f37b86d77. CI now pins Zig 0.16.0, and the two project files live under testdata/proj so the unrelated fixtures are not copied. The existing TestBuild passes on Zig 0.16.0. Local and shared cache configuration is unchanged.

internal/builders/golang/build.go
caarlos0: Confirmed with a real cold-cache GoReleaser build: the download message appears, while the same build with a warm cache is quiet. Filed #7072 with a minimal project, the introducing history, and the separate follow-up scope. No output-filtering behavior was added to this PR.

scripts/fuzz.sh
caarlos0: Fixed in 2a3d328fdb9689390f2ea89ad964bb71b3b9e2dc. Discovery now uses sed, which succeeds with no matching functions but still fails on unreadable inputs. scripts/fuzz.sh ./internal/ids 1x now runs the ordinary tests. An injected failing fuzz target still exits 42 after one Go invocation, and the all-success path runs both fuzz targets and ordinary tests. No Go test harness was added for the script.

internal/pipe/docker/denied_test.go
caarlos0: Fixed in d1b995b72b42272ee986aa1637b08b4792d8d46b. Cleanup still cancels and joins the follower, but no longer requires an error from Wait. Readiness and denied-push assertions remain unchanged. Both real Docker and buildx subcases passed locally.

PR #5403, opened 2024-12-31
feat(build): add bun support

caarlos0:
very WIP: adding bun support

seemed easy enough

PR #7068, opened 2026-09-05
refactor: simplify internal code and test setup

### LINKED ISSUE #7051: refactor(archive): remove the unused EnhancedArchive file map
### Focused enhancement

[NewEnhancedArchive and EnhancedArchive](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/archive/archive.go#L364-L381) allocate and store a private `files` map that is never read or populated. The decorator's `Add` method only adjusts the destination, preserves file information, logs the operation, and delegates to the underlying archive writer.

Remove the unused field and its constructor allocation. This is a behavior-preserving cleanup, not a claim of a measured performance improvement or a missing duplicate-file check.

### Preserve coverage

Keep [TestDuplicateFilesInsideArchive](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/archive/archive_test.go#L1059-L1084): it wraps a real archive writer and expects `fs.ErrExist` on the second destination. The underlying writers still own duplicate detection. Preserve wrapping, Windows separator normalization, file metadata, delegation errors, and close behavior. Do not introduce a replacement map or change the public archive interface.

---
Audit priority: **P3**.

### LINKED ISSUE #7052: refactor(checksums): rely on Artifact.Checksum for checksum metadata
### Focused enhancement

The [checksums helper](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/checksums/checksums.go#L214-L226) calls `Artifact.Checksum` and then repeats that method's metadata initialization and checksum assignment. Its private [metadata-key constant](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/checksums/checksums.go#L24) also duplicates `artifact.ExtraChecksum`.

Independent controls exercised both nil and populated metadata. The artifact checksum call already stored `crc32:72d7748e`; calling the pipe helper afterwards produced `72d7748e  demo\n` without changing serialized artifact metadata. Existing IDs were preserved.

### Narrow change

Remove only the redundant map initialization and assignment, and use the shared constant in the remaining test. Retain checksum calculation, error propagation, logging, filename formatting, and the two-space checksum-file separator.

Keep coverage for nil metadata, existing metadata, and checksum failures. This removes duplicate ownership of the same metadata update; it does not change algorithms, refresh rules, or checksum caching behavior.

---
Audit priority: **P3**.

### LINKED ISSUE #7053: refactor(snapcraft): remove the duplicate filepath.Base call
### Focused enhancement

The default-app branch in [Snapcraft metadata generation](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/snapcraft/snapcraft.go#L327-L337) sets the command with `filepath.Base(filepath.Base(binaries[0].Name))`. The second call does not change the first call's result.

The independent control checked ordinary, empty, root, dot, and trailing-separator inputs. It also ran production metadata generation with a fake pack command, a binary named `subdir/tool`, and both default and explicit Snap names. Both generated the command `tool` under the expected app name.

### Narrow change

Use one `filepath.Base` call. Leave app-name selection, explicitly configured apps, binary copying, modes, metadata, and pack behavior unchanged.

Retain `TestRunPipeMetadataWithoutApps` and configured-name coverage. This is a readability enhancement with demonstrated equivalent metadata, not a performance claim or a request to remove real Snapcraft integration coverage.

---
Audit priority: **P3**.

### LINKED ISSUE #7054: refactor(changelog): inline the one-use prefixItem helper
### Focused enhancement

[prefixItem](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/changelog/changelog.go#L222-L224) contains only `return li + s`. Its sole caller is [formatEntry](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/changelog/changelog.go#L259-L270), which already has the rendered line and template error available.

Replace the call with `return li + line, err` and remove the private helper. Keep the existing list-prefix constant rather than duplicating its literal. This reduces one level of indirection without introducing a new abstraction or changing any public interface.

### Preserve behavior

Keep the current output on both successful and failed template application, including the prefix on partial output and the original error. Retain `TestFormatEntry`, `TestChangelogFormat`, and grouped changelog coverage. Do not fold in sorting, author cleanup, whitespace, or template-diagnostic changes. The request is an intact-output refactor, not a functional correction.

---
Audit priority: **P3**.

### LINKED ISSUE #7055: refactor(tests): use mktmp for invalid build and release configuration

### LINKED ISSUE #7056: refactor(healthcheck): use an ordinary map for invocation-local deduplication
### Focused enhancement

The healthcheck command creates a [local sync.Map](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/cmd/healthcheck.go#L58-L70) and passes it through sequential dependency checks. [checkPath](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/cmd/healthcheck.go#L107-L125) uses it only as a set of complete tool-command strings. There is no concurrent user of this invocation-owned state.

Use an ordinary map for the same set. Keep it local to each command invocation; [#6823](https://github.com/goreleaser/goreleaser/pull/6823) correctly fixed the previous process-wide lifetime, and that fix must remain intact.

### Preserve behavior

Record a tool before checking it, including failed checks, exactly as today. Retain `TestCheckPathChecksEachToolOnce`, which checks that a repeated failing tool is skipped in one invocation and checked again with fresh state. Preserve command arguments, logs, error results, and check order. This is a simpler representation of existing sequential state, not a measured speedup or a change to healthcheck policy.

---
Audit priority: **P3**.

### LINKED ISSUE #7057: refactor(golang): remove the unused buildOutput helper and its isolated test

### LINKED ISSUE #7058: refactor(commitauthor): copy the value struct directly before templating
### Focused enhancement

[commitauthor.Get](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/commitauthor/author.go#L16-L37) reconstructs every field of the supplied author, including its nested signing configuration, before applying templates. The current [configuration types](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/pkg/config/config.go#L475-L489) contain only strings, booleans, and the value-only signing struct.

Replace that field-by-field reconstruction with `author := og`. This makes the existing value-copy operation explicit and keeps the original argument separate from fields changed by templating.

### Preserve behavior

Retain all current template calls and their error return, including signing key, program, and format. Preserve `UseGitHubAppToken` and signing-enabled values. Keep `TestGet` and `TestDefault`, including invalid-template and signing cases. Do not change default names, email addresses, signing formats, token selection, or public signatures. This is a focused maintenance enhancement, not a fix for an observed publication failure.

---
Audit priority: **P3**.

### LINKED ISSUE #7059: refactor(tmpl): initialize Fields directly from its literal
### Focused enhancement

[tmpl.New](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/tmpl/tmpl.go#L103-L151) creates an empty map, constructs a second map literal containing all template fields, and copies that literal into the first map. The source contains no filtering or conversion between those two maps.

Initialize `Fields` directly with the same literal. Keep every key, computed value, timestamp format, and conditional Git-tree value identical. In particular, preserve the current environment-map aliasing; do not introduce an unrelated deep copy or change `SetEnv` behavior.

### Preserve coverage

Retain `TestWithArtifact`, `TestWithBuildOptions`, `TestReuseTpl`, and environment/template-field tests. This request is limited to field-map construction. It does not include changes to `Apply`, `ApplyAll`, `Slice`, error wrapping, or diagnostic output. No runtime speedup is asserted; the benefit is removing a redundant construction-and-copy step while keeping the same template API.

---
Audit priority: **P3**.

### LINKED ISSUE #7060: refactor(tests): use a fixed minimal project in the Zig build integration test

### LINKED ISSUE #7061: refactor(ko): remove the unused private buildOptions.main field
### Focused enhancement

The private [buildOptions.main field](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/ko/ko.go#L178-L208) is never populated. Its only production use forwards the empty string to `build.Config.Main` in `makeBuilder`. Repository references confirm that the useful input is the separate public `config.Ko.Main`, which participates in import-path selection.

Remove the private field and the explicit `Main: o.main` forwarding. The downstream configuration then retains the same zero value it receives today. Do not remove, rename, or change the public `ko.main` option.

### Preserve coverage

Keep the build-option tests for configured and empty main paths, including `TestBuildBuildOptionsEmptyMain`, and the publication coverage. Preserve import-path resolution, working directory, flags, environment, and errors. This is dead-state removal only; it does not change default-main behavior, image contents, or the ko builder API.

---
Audit priority: **P3**.

### LINKED ISSUE #7062: refactor(mcp): copy the legacy MCPDetails struct during migration
### Focused enhancement

The deprecated `mcp.github` migration in [Pipe.Default](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/mcp/mcp.go#L53-L70) copies nine fields individually between values of the same `config.MCPDetails` type.

Replace those assignments with `ctx.Config.MCP.MCPDetails = ctx.Config.MCP.GitHub`. Independent controls compared all migrated fields with that direct assignment, including packages, transports, repository, authentication, and overwrite of partially populated top-level details. A nonempty top-level name remained unchanged in the control.

### Preserve behavior

Keep the existing migration guard, deprecation notice, legacy value, and subsequent authentication default. Do not merge individual nonempty fields or change slice/reference aliasing: either would differ from the current assignment semantics.

Retain `TestDefault` with legacy-only, partial-top-level, explicit-top-level, and missing-auth controls. This is a focused maintenance enhancement. It does not alter the published MCP payload, supported transport schema, authentication lifecycle, or registry interaction.

---
Audit priority: **P3**.

caarlos0:
Resolve all 12 refactor issues in separate commits, plus the requested review simplification:

- Remove unused archive and ko state, the unused Go build-output helper, and redundant checksum, changelog, and Snapcraft operations.
- Template the commit-author value parameter directly, copy legacy MCP details directly, and construct template fields directly.
- Use an invocation-local map for healthcheck deduplication.
- Reduce invalid-config test setup and replace `zig init` with a fixed minimal project, while keeping real compilation and the existing cache setup.
- Expand MCP migration coverage for legacy and partial configuration, explicit top-level configuration, missing authentication, and slice aliasing.

Remove unnecessary code and test setup without changing public APIs, configuration, outputs, or defaults.

AI disclosure: implemented with GitHub Copilot.

Fixes #7051
Fixes #7052
Fixes #7053
Fixes #7054
Fixes #7055
Fixes #7056
Fixes #7057
Fixes #7058
Fixes #7059
Fixes #7060
Fixes #7061
Fixes #7062


internal/commitauthor/author.go
```diff
@@ -14,17 +14,7 @@ const (
...
-		Email:             og.Email,
-		UseGitHubAppToken: og.UseGitHubAppToken,
-		Signing: config.CommitSigning{
-			Enabled: og.Signing.Enabled,
-			Key:     og.Signing.Key,
-			Program: og.Signing.Program,
-			Format:  og.Signing.Format,
-		},
-	}
+	author := og
```
caarlos0: can probably remove teh `author := og` and just rename the function parameter to `author`

caarlos0: Addressed in afaac85d3356a219abd315c6a98ed73b934aa687. Renamed the value parameter to `author` and removed the redundant local copy. Templating now uses the parameter directly; the caller’s struct remains unchanged because Go passes it by value. The commit-author tests pass.

caarlos0 (commented):
## Review: no findings, approve

I reviewed the complete diff against the merge base `bbe26fa`. I found no correctness, compatibility, scope, or test problem. There is no inline comment because there is no finding to anchor.

This is a comment review, not a formal approval, because GitHub does not let the author approve their own pull request.

### What I verified independently

- **checksums**: `(*artifact.Artifact).Checksum` already writes `Extra[ExtraChecksum]` with the same `"%s:%s"` format on the same pointer (`internal/artifact/artifact.go:414-419`). The deleted write was a duplicate. The split-checksum path never used it. No template, SBOM, signature, or `artifacts.json` output changes.
- **mcp**: `config.MCPDetails` has exactly the nine fields that the old code copied one by one. The struct assignment is equivalent. `MCP.GitHub` stays independent, and `Auth.Type = cmp.Or(...)` still applies to the top level only. The slice aliasing is safe: `Publish` ranges over a copy and templates the loop copy, so no code mutates a shared backing array.
- **commitauthor**: `CommitAuthor` and `CommitSigning` contain only strings, bools, and a nested value struct. The by-value parameter is a full copy, so no caller can see the mutation. All eight callers pass a config field by value.
- **healthcheck**: the loop is sequential and `skip.Maybe` starts no goroutine. The pre-existing unsynchronized `errs = append(...)` in the same closure confirms this. `go test -race ./cmd/ -run 'TestCheckPath|TestHealthcheck'` passes.
- **ko / archive**: `buildOptions.main` was never assigned, so `Main: o.main` always passed the zero value. `EnhancedArchive.files` lost its last reader in #3904. Both removals are dead code.
- **snapcraft / changelog / tmpl**: `filepath.Base` is idempotent, `li + line` is identical, and `Fields` is `map[string]any`, so the literal equals the old empty map plus `maps.Copy`. `maps` is still used elsewhere in `tmpl.go`.
- **cmd tests**: both commands call `loadConfig` first and return on error, before any git or module work. Both tests assert the exact YAML error with `require.EqualError`, so a different cause fails loudly. `mktmp` still unsets the tokens and restores the working directory.
- **zig fixture**: I ran the exact fixture with the same invocation that the builder issues (`zig build -Dtarget=aarch64-macos -p zig-out/aarch64-macos -Doptimize=Debug`) on Zig 0.16.0. It produced `zig-out/aarch64-macos/bin/proj`, which is the path `Build()` copies from. No `build.zig.zon` is necessary, and the Zig builder never reads zon metadata.

Targeted tests pass with `-race` for `cmd`, `mcp`, `checksums`, `commitauthor`, `tmpl`, `changelog`, and `archive`.

### Note for a separate change, not a blocker

Removing `buildOutput` is correct, but the helper lost its only caller in #5403, not in this PR. `base.Exec` logs the command output verbatim, so `goreleaser build` has printed `go: downloading ...` lines again since that change, and #4869 is effectively reverted. `TestBuildOutput` stayed green only because it called the orphan directly, so deleting it removes false confidence. The impact is log noise. A follow-up issue could filter those lines in `base.Exec`.

One maintenance tradeoff worth naming: `zig init` followed the installed compiler, but the hand-written `build.zig` pins the `std.Build` API. CI uses `mlugg/setup-zig` with no `version` input and the repository has no root `build.zig.zon`, so CI tracks the latest stable Zig. The next breaking `std.Build` release will need a manual edit here. This is acceptable, but it is a real cost.

### Verification gaps

- The reviewer was asked not to build, lint, or format locally, so compilation, `go vet`, `gofmt`, and lint are assumed clean.
- I did not run the full test suite, and I did not run `TestBuild` in `internal/builders/zig` through Go. I reproduced the `zig build` invocation directly instead.
- The Zig version that CI resolves is inferred from the action default and the absence of a root `build.zig.zon`. I did not read a CI run log.

---

This review was produced by an AI agent (GitHub Copilot CLI), not by a human reviewer.

PR #7085, opened 2026-09-06
fix(flatpak,docker): quote install paths and respect project environment

### LINKED ISSUE #6924: docker and flatpak: inherited environment overrides project settings
### Problem

Export `DOCKER_CONFIG=ambient-config`, then configure top-level `env: [DOCKER_CONFIG=project-config]`. The project context contains the configured value, but Docker commands receive the ambient value. Flatpak commands have the same precedence error with conflicting environment keys.

The affected code appends the inherited environment after configured values in [Docker v1's two command helpers](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/docker/api.go#L47-L73), [Docker v2](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/docker/v2/docker.go#L281-L294), and [Flatpak](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/flatpak/flatpak.go#L271-L279). Go keeps the last value for duplicate environment keys.

### Evidence

Recording child executables invoked through each production helper received the ambient value despite the correct value in `ctx.Env`. Without the ambient key, the configured value reached the child. No Docker daemon, image push, or Flatpak installation was involved.

### Correction

Use one consistent precedence rule at all affected call sites: inherited values first, explicit project overrides last. Preserve unrelated inherited keys, command arguments, and output handling. Add process-boundary regressions for conflicting, configured-only, and inherited-only values.

---
Audit priority: **P2**.

### LINKED ISSUE #6931: Flatpak install commands do not quote binary paths
Build a binary named `my app` and select it for Flatpak packaging. The pipe copies that filename correctly and records it as a manifest source, but generates `install -Dm755 my app /app/bin/my app` for the install step.

Executing that exact generated command against a local argument-checking install recorder split the source and destination into separate words. The recorder received five arguments after the command instead of the mode, source, and destination it expected, and exited 64. A quoted control succeeded with the same existing source file. Verification covered manifest generation and shell interpretation only; it did not install a Flatpak or use a remote repository.

[The command construction](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/flatpak/flatpak.go#L183-L195) inserts the binary name without shell quoting. Supported filenames should remain single arguments when the generated build command executes.

Shell-quote both paths using an established helper. Add regression cases for spaces and quote characters, execute the generated command against an argument recorder, and retain the ordinary-name control.

---
Audit priority: **P2**.

caarlos0:
Fixes a group of related bugs in Flatpak pipe and inherited environment handling.

- Closes #6931: shell-quotes Flatpak install source and destination paths in generated build commands.
- Closes #6924: lets project environment values override inherited environment values for Docker and Flatpak subprocesses.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,6 @@ module github.com/goreleaser/goreleaser/v2
 go 1.27.1
 
 require (
-	al.essio.dev/pkg/shellescape v1.6.0
 	charm.land/lipgloss/v2 v2.0.6
 	code.gitea.io/sdk/gitea v0.25.1
 	dario.cat/mergo v1.0.2
@@ -64,6 +63,7 @@ require (
 )
 
 require (
+	al.essio.dev/pkg/shellescape v1.6.0 // indirect
 	cel.dev/expr v0.25.2 // indirect
 	cloud.google.com/go v0.123.0 // indirect
 	cloud.google.com/go/auth v0.20.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/goreleaser/goreleaser/v2
 go 1.27.1
 
 require (
+	al.essio.dev/pkg/shellescape v1.6.0
 	charm.land/lipgloss/v2 v2.0.6
 	code.gitea.io/sdk/gitea v0.25.1
 	dario.cat/mergo v1.0.2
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/goreleaser/goreleaser/v2
 go 1.27.1
 
 require (
+	al.essio.dev/pkg/shellescape v1.6.0
```
caarlos0: maybe use shellwords instead?

PR #7088, opened 2026-09-06
fix(builder/node,builder/bun): per-target SEA config, prebuild templates and target aliases

### LINKED ISSUE #6895: node: target-dependent main templates fail in prebuild
### Trigger

Configure `builder: node`, `targets: [linux-x64, linux-arm64]`, and `main: 'build/{{.Target}}.js'`. Provide the corresponding entrypoint files. The Node builder supports target-aware main rendering, but global prebuild runs first.

### Observed

Verification used strict YAML loading, real build defaults, target parsing, and the production prebuild pipe. Prebuild rejected the accepted configuration with `map has no entry for key "Target"`. Equivalent `.Os` and `.Arch` cases also failed. The same main rendered correctly as `build/linux-x64.js` with production target-aware template fields. Literal and global `{{.ProjectName}}` controls passed. Native SEA compilation was not involved because failure occurs before Node invocation.

[Prebuild renders all mains globally](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/prebuild/prebuild.go#L18-L27), before [the Node builder's own rendering](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/builders/node/sea.go#L16-L19).

### Smallest fix

Leave Node main rendering to the target-aware builder without changing other builders' prebuild behavior. Cover target, OS, and architecture fields alongside literal and global-template controls.

---
Audit priority: **P2**.

### LINKED ISSUE #6896: node: concurrent targets share the SEA configuration file
### Trigger

```yaml
builds:
  - builder: node
    targets: [linux-x64, linux-arm64]
    binary: 'app-{{.Target}}'
    main: index.js
    no_unique_dist_dir: true
```

The binary names are distinct, but [both jobs write `sea-config.json` beside their outputs](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/builders/node/build.go#L165).

### Observed

Strict configuration loading, defaults, prebuild, and the concurrent production build pipe accepted this configuration. Barrier-controlled recording child processes showed ARM64 overwriting x64's config before either child read it. Both children received ARM64's executable and output path. The pipe returned success and registered two artifacts, but the x64 output was absent. The same schedule with unique distribution directories preserved both outputs.

The production resolver, downloader, extractor, and config writer ran unchanged against fixture HTTP responses and checksums. Native `node --build-sea` compilation was replaced with a recording process; the evidence establishes the shared-file collision, not native binary execution.

### Smallest fix

Write the SEA config in the already-private target Node directory. Retain the barrier-controlled regression, assert distinct config/executable/output paths, and keep the unique-directory control.

---
Audit priority: **P2**.

### LINKED ISSUE #6901: partial: automatic target selection rejects valid Node and Bun aliases
### Trigger

Request automatic single-target selection for a Node build with `targets: [darwin-arm64]` on Darwin ARM64, or `targets: [win-x64]` on Windows AMD64. Bun targets such as `linux-x64-modern` have the same problem. Leave `TARGET` unset.

### Observed

Verification first passed each target through its actual Node or Bun defaults, then ran the production partial pipe with explicit `GGOOS`/`GGOARCH` inputs. All twelve accepted target cases failed with `could not setup the target filter, maybe set TARGET=[something]`. Supplying the exact `TARGET` passed in every control. No Node/Bun compilation or full CLI run was used; this isolates automatic selection.

[The alias tables and matching logic](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/partial/partial.go#L77-L99) omit supported `arm64`, `x64`, and Node's `win` spellings.

### Smallest fix

Add the missing aliases without changing target formats. Cover accepted Node/Bun targets and retain explicit-target controls. This is a narrow existing-behavior defect related to #5327, not a request to standardize separators.

---
Audit priority: **P2**.

caarlos0:
Fixes a group of related bugs in the Node and Bun builders and partial target aliases.

- Closes #6896: writes each Node SEA config in its per-target scratch directory so concurrent builds do not share it.
- Closes #6895: leaves Node main templates for the target-aware builder instead of globally rendering them in prebuild.
- Closes #6901: recognizes Node and Bun x64, arm64, and win target aliases during automatic partial target selection.


internal/builders/node/build_test.go
```diff
@@ -150,6 +155,112 @@ func TestBuild(t *testing.T) {
 	require.True(t, modTime.Equal(fi.ModTime()))
 }
 
+func TestBuildUsesPerTargetSEAConfig(t *testing.T) {
```
caarlos0: i think we can simply assert the internal file structure instead of all of this? 

we can do it on a pre-existing test, too, to avoid too manu builds

internal/builders/node/build_test.go
```diff
@@ -184,6 +295,80 @@ func createFakeNodeAlias(tb testing.TB, name string) {
 	)
 }
 
+func createFakeNodeBuildHelper(tb testing.TB) string {
```
caarlos0: seems too much for too little change

internal/builders/node/download.go
```diff
@@ -34,9 +34,12 @@ func ensureNode(ctx context.Context, dir, target, destDir string) (string, error
 		return "", fmt.Errorf("node: resolve node version: %w", err)
 	}
 
-	return downloadHostBinary(ctx, version, target, destDir)
+	return downloadNodeBinary(ctx, version, target, destDir)
 }
 
+//nolint:gochecknoglobals
+var downloadNodeBinary = downloadHostBinary
+
```
caarlos0: also not necessary per pervious commetns

internal/builders/node/build_test.go
caarlos0: Done in 07ad8f40247d3dbb226a340886226e1dc3cf214e. Dropped the extra build and now assert the internal file structure in the existing `TestBuild`: after the build, `dist/proj_darwin-arm64` must contain only `proj`. Before the fix, `sea-config.json` was written next to the binary, so this fails on the old behaviour without an extra build.

internal/builders/node/build_test.go
caarlos0: Agreed. Removed the fake node helper process, the HTTP barrier and the concurrent build in 07ad8f40247d3dbb226a340886226e1dc3cf214e.

internal/builders/node/download.go
caarlos0: Removed in 07ad8f40247d3dbb226a340886226e1dc3cf214e. The `downloadNodeBinary` variable only existed as a seam for the test that is now gone.

internal/pipe/prebuild/prebuild.go
```diff
@@ -17,6 +17,11 @@ func (Pipe) String() string {
 func (Pipe) Run(ctx *context.Context) error {
 	tpl := tmpl.New(ctx)
 	for i := range ctx.Config.Builds {
+		// the node builder renders Main per target, because it accepts
+		// target-aware fields such as .Target, .Os and .Arch.
+		if ctx.Config.Builds[i].Builder == "node" {
+			continue
```
caarlos0, line 23: Minor, non-blocking: nothing tests the behaviour that this `continue` now relies on.

The new prebuild subtest proves that `main` stays raw. No test proves that the node builder then renders it. The only caller of `tpl.Apply(build.Main)` is `createSEAConfig`, and the only test that reaches it is `TestBuild`, which uses the literal `index.js` and is skipped when `node` is not on `PATH`.

Trigger: any later change that drops `WithBuildOptions` from the template in `internal/builders/node/build.go:131-135`, or moves the `main` rendering. Impact: `main: 'build/{{ .Target }}.js'` (the config in #6895) fails at build time with `node: main "build/{{ .Target }}.js" not found`, and no test fails.

Smallest fix: a table test on `createSEAConfig` in `internal/builders/node`. It needs no `node` binary. Build the template exactly as `Build` does, pass a `config.Build` with a templated `main`, create the expected file, then assert the `main` field of the written JSON. Cover `.Target`, `.Os` and `.Arch`.

internal/builders/node/sea.go
```diff
@@ -96,8 +96,8 @@ func loadUserSEAConfig(buildDir string) (map[string]any, error) {
 // rewriteAssetPaths converts relative asset values in cfg["assets"]
 // into absolute paths anchored at buildDir. Node resolves `assets`
 // paths relative to the directory containing sea-config.json, but
-// goreleaser writes the merged config into the dist directory, so
-// relative user paths would otherwise break.
+// goreleaser writes the merged config into a per-target scratch
+// directory, so relative user paths would otherwise break.
```
caarlos0: Minor, non-blocking: the rewritten sentence keeps a wrong premise.

Node does not resolve `assets` against the directory of `sea-config.json`. `ParseSingleExecutableConfig` stores the raw strings and `BuildAssets` passes them to `ReadFileSync`, so relative asset paths resolve against the working directory of the `node` process ([node_sea.cc](https://github.com/nodejs/node/blob/main/src/node_sea.cc)). GoReleaser runs `node` with `base.Exec(ctx, command, env, "")`, so that directory is GoReleaser's own, not `build.Dir`.

The function stays correct, and the move to the scratch directory changes nothing here: the rewrite is necessary because the user writes the paths relative to their own `sea-config.json` in `dir`, while `node` runs elsewhere. Only the stated reason is wrong.

Suggested text: "Node resolves relative `assets` paths against the working directory of the node process, but the user writes them relative to their `sea-config.json` in the build dir, so anchor them at buildDir."

`www/content/customization/builds/builders/node.md` carries the same wrong reason ("keep resolving after GoReleaser moves the merged configuration into a scratch directory").

caarlos0 (commented):
## Verdict: approve (posted as a comment: GitHub does not allow approving your own pull request)

I traced the three fixes against the merge base `e0418e8d5` and found no correctness defect.

**SEA config relocation is safe.** Node does not rebase config paths on the config file. `ParseSingleExecutableConfig` keeps `main`, `output` and `executable` as raw strings, and `WriteSingleExecutableBlob` / `BuildAssets` pass them straight to `WriteFileSync` / `ReadFileSync`, so relative values resolve against the working directory of the `node` process ([node_sea.cc](https://github.com/nodejs/node/blob/main/src/node_sea.cc)). GoReleaser writes `main`, `executable` and the rewritten `assets` as absolute paths, and `base.Exec(..., "")` leaves the working directory unchanged, so the relative `output` still lands on `options.Path`.

**The prebuild skip loses no template field.** `tmpl.New(ctx).WithBuildOptions(options).WithArtifact(a)` in the node builder is a superset of the prebuild context. `defaults` runs before `prebuild` in the pipeline, so `node`'s `WithDefaults` has already replaced an empty `main` with `index.js`; the `m == ""` fallback is not needed for node.

**The alias tables cannot make a false match.** `findRuntime` splits the target on `-` and compares complete parts with `slices.Contains`, so `win` cannot match `darwin`. The `go` builder returns before `findRuntime`, so the `GOOS_GOARCH` path is unchanged. All 11 new partial subcases need at least one of the added aliases, so none of them is vacuous.

Two minor comments are inline. Neither blocks the merge.

**Verification gaps.** I could not execute `TestBuild`: the local `node` (v26.8.1, Homebrew) reports `Single executable application is disabled`, so `--build-sea` cannot run here. I did not run the Go test suite, as requested. The conclusions above come from reading the Node source and the GoReleaser call chain.

_Reviewed by GitHub Copilot CLI (bot)._

internal/pipe/prebuild/prebuild.go
caarlos0: Good catch. Added `TestCreateSEAConfigRendersMain` in 7fd34ec87b369f68c4bcda6cb021a525eb168f5e: a table test on `createSEAConfig` covering `.Target`, `.Os` and `.Arch`, plus a `win-x64` case that records that `.Os`/`.Arch` carry the Go names (`windows`/`amd64`), because `WithArtifact` runs after `WithBuildOptions` and overwrites them. It needs no `node` binary. I checked it is not vacuous: replacing `tpl.Apply(build.Main)` with the raw `build.Main` fails 4 of the 5 cases.

internal/builders/node/sea.go
caarlos0: You are right. I confirmed it in `node_sea.cc`: `BuildAssets` passes each raw path straight to `ReadFileSync`, and `ParseSingleExecutableConfig` stores the strings unchanged, so there is no anchoring to the config location. Corrected the comment and `www/content/customization/builds/builders/node.md` in 7fd34ec87b369f68c4bcda6cb021a525eb168f5e to give the working-directory reason instead.

Issue #6882, opened 2026-09-05
MCP authentication failure deletes pre-existing token files


caarlos0:
Place synthetic sentinel contents in `.mcpregistry_github_token` and `.mcpregistry_registry_token` in an isolated working directory. Configure an MCP name and `auth.type: github-oidc`, with the Actions OIDC environment variables absent.

The actual pipe's login step returns successfully because OIDC needs no interactive login. Token acquisition then fails on the missing environment variable before any HTTP request. Despite that failure, both pre-existing files are deleted by the deferred cleanup. A local reproduction observed the authentication error and both removals. Only synthetic files were used; no real token or external registry was involved.

[The deferred removals](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/mcp/mcp.go#L114-L121) do not establish ownership. The pinned authentication providers hold tokens in memory and did not create these legacy working-directory files during this operation.

Remove the obsolete cleanup and its file-path constants. Authentication should not remove existing files that it did not create. Add a regression that calls the real OIDC path with missing environment variables and requires unchanged sentinel contents after the error, without permitting any HTTP request.

---
Audit priority: **P1**.

Audit source revision: [628c889d7](https://github.com/goreleaser/goreleaser/commit/628c889d7f2a34066649833be419fdb9abe01174). AI disclosure: this report comes from a Copilot-assisted audit, with discovery by `caarlos0` agents and independent verification by `anvil` agents. Reproduction methods and limits are stated above; no fix is included in this issue.

caarlos0:
this is on purpose and will stay as is for now.

PR #7090, opened 2026-09-06
fix(mcp): support package transport URLs

### LINKED ISSUE #6940: MCP HTTP and SSE package transports cannot configure a URL
Configure an MCP package with `registry_type: npm`, `identifier: '@fixture/server'`, and `transport.type: streamable-http` or `sse`. Both transport types are accepted configuration choices.

Production request capture showed a transport containing only `type`. Running the captured payloads through the pinned registry v1.8.1 semantic validator rejected both because `packages[0].transport.url` is required. Adding a URL to the captured request made semantic validation pass. However, adding `transport.url` to GoReleaser configuration fails strict loading with `field url not found in type config.MCPTransport`. A stdio control passed without a URL. Requests were captured locally and validation was offline; no server was published to the live registry.

The missing field is in [MCPTransport](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/pkg/config/config.go#L1714-L1716), and [request construction](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/mcp/mcp.go#L152-L158) copies only the type.

Add a templated URL field and transfer it into the request model. Update the schema and documentation together. Regressions should cover strict configuration loading, template rendering, captured URLs for both HTTP transports, and unchanged stdio behavior.

---
Audit priority: **P2**.

### LINKED ISSUE #6941: MCPB packages cannot provide the required SHA-256 hash
Configure an MCP package with `registry_type: mcpb`, a release-asset URL such as `https://github.com/fixture/server/releases/download/v1.0.0/server.mcpb`, and `transport.type: stdio`.

The production request omits `fileSha256`. The pinned registry v1.8.1 publish validator, with its default ownership-validation policy enabled, rejected the captured request because the hash is mandatory. It made zero network calls before that rejection. Adding a synthetic hash to the captured payload advanced validation to URL accessibility, with real network access still blocked. Users cannot supply the missing value: strict loading rejects `file_sha256` in `config.MCPPackage`. This does not claim a successful live MCPB publication or a remote download result.

The gap connects [the package configuration](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/pkg/config/config.go#L1708-L1712) and [the serialized package](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/mcp/mcp.go#L152-L158).

Expose a templated checksum field and serialize it as `FileSHA256`. Update schema and docs. Add a configuration-to-payload regression requiring the exact hash, an absent-hash failure case, and controls showing that other registry types do not acquire an unwanted hash.

---
Audit priority: **P2**.

caarlos0:
Adds a templated `transport.url` field to MCP packages and sends it in the publish request, so `streamable-http` and `sse` servers can be published. Before this, the URL was dropped and those packages were unusable.

Also removes `registry_type: mcpb`.

### Net change

- `mcp.go`: template `transport.url`, and pass it to the registry.
- `config.go`: add `MCPTransport.URL`; drop `mcpb` from the `registry_type` enum.
- Docs, `schema.json`, `schema-pro.json`.
- Tests: `TestPublishPackageTransportURL`, `TestMCPTransportStrictConfig`.

### Why mcpb goes away

GoReleaser cannot build `.mcpb` files. The archive pipe accepts a fixed format list and names every file `<template>.<format>`, so a `.mcpb` extension is unreachable. The registry also makes `fileSha256` mandatory for mcpb, and that value can only come from a local file.

So `registry_type: mcpb` has been rejected by the registry since it shipped in v2.13.0. Removing it is not a breaking change, because nothing that worked stops working.

Closes #6940.
Closes #6941 — resolved by removing mcpb rather than adding the hash field.

### Note on the commit history

The branch history is longer than the result. It adds an mcpb SHA-256 implementation and a token-file change, then removes both. The net diff above is what actually lands. Squash on merge.

`#6882` was closed as not planned; that work was reverted.

### DIFF
--- a/internal/pipe/mcp/mcp.go
+++ b/internal/pipe/mcp/mcp.go
@@ -7,10 +7,12 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"path"
 	"strings"
 
 	"github.com/caarlos0/log"
+	"github.com/goreleaser/goreleaser/v2/internal/artifact"
 	"github.com/goreleaser/goreleaser/v2/internal/deprecate"
 	"github.com/goreleaser/goreleaser/v2/internal/logext"
 	"github.com/goreleaser/goreleaser/v2/internal/pipe"
@@ -18,6 +20,7 @@ import (
 	"github.com/goreleaser/goreleaser/v2/internal/skips"
 	"github.com/goreleaser/goreleaser/v2/internal/summary"
 	"github.com/goreleaser/goreleaser/v2/internal/tmpl"
+	"github.com/goreleaser/goreleaser/v2/pkg/config"
 	"github.com/goreleaser/goreleaser/v2/pkg/context"
 	"github.com/modelcontextprotocol/registry/cmd/publisher/auth"
 	proto "github.com/modelcontextprotocol/registry/cmd/publisher/commands"
@@ -123,13 +126,13 @@ func (p Pipe) Publish(ctx *context.Context) error {
 	for _, pkg := range mcp.Packages {
 		if err := tmpl.New(ctx).ApplyAll(
 			&pkg.Identifier,
-			&pkg.FileSHA256,
 			&pkg.Transport.URL,
 		); err != nil {
 			return fmt.Errorf("could not apply templates: %w", err)
 		}
-		if pkg.RegistryType == "mcpb" && pkg.FileSHA256 == "" {
-			return fmt.Errorf("mcpb package %q requires file_sha256", pkg.Identifier)
+		fileSHA256, err := mcpbFileSHA256(ctx, pkg)
+		if err != nil {
+			return err
 		}
 		version := ctx.Version
 		if pkg.RegistryType == "oci" {
@@ -139,7 +142,7 @@ func (p Pipe) Publish(ctx *context.Context) error {
 			RegistryType: pkg.RegistryType,
 			Identifier:   pkg.Identifier,
 			Version:      version,
-			FileSHA256:   pkg.FileSHA256,
+			FileSHA256:   fileSHA256,
 			Transport: model.Transport{
 				Type: pkg.Transport.Type,
 				URL:  pkg.Transport.URL,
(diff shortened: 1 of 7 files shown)


www/content/customization/publish/mcp.md
```diff
@@ -113,6 +113,12 @@ mcp:
         # Required.
         type: stdio
 
+        # URL for streamable-http or sse transports.
+        #
+        # Required for streamable-http and sse transports.
+        # Templates: allowed.
+        # url: "https://example.com/myserver/mcp"
```
caarlos0:
```suggestion
        # Templates: allowed.
        url: "https://example.com/myserver/mcp"
```

also missing the `since` version shortcode

internal/pipe/mcp/mcp.go
```diff
@@ -196,6 +195,58 @@ func (p Pipe) Publish(ctx *context.Context) error {
...
+	name, err = url.PathUnescape(name)
+	if err != nil {
+		return "", fmt.Errorf("parse mcpb package identifier: %w", err)
+	}
+	return name, nil
+}
+
+func findArtifact(ctx *context.Context, name string) (*artifact.Artifact, error) {
+	var matches []*artifact.Artifact
+	for _, art := range ctx.Artifacts.Filter(artifact.ByTypes(artifact.ReleaseUploadableTypes()...)).List() {
```
caarlos0: should filter only for the actual types that can be used? or improve the filter with byext, or pass the function art.Name == name as a filter as well

internal/pipe/mcp/mcp.go
```diff
@@ -103,11 +100,6 @@ func (p Pipe) Publish(ctx *context.Context) error {
 	if err := provider.Login(ctx); err != nil {
 		return fmt.Errorf("could not login: %w", err)
 	}
-	defer func() {
-		// logout...
-		_ = os.Remove(gitHubTokenFilePath)
-		_ = os.Remove(registryTokenFilePath)
-	}()
```
caarlos0, line 110: we should probably keep this - otherwise some CI might keep it logged in.

this was on purpose.

www/content/customization/publish/mcp.md
caarlos0: Fixed in a5230e579: moved the URL docs onto the concrete HTTP transport example and added `{{< g_inline_version "v2.19-unreleased" >}}`. I also synced `schema-pro.json` so the generated schemas both accept `transport.url`.

internal/pipe/mcp/mcp.go
caarlos0: Fixed in 89dd0d091: MCPB checksum lookup now requires the configured `.mcpb` artifact name and limits matches to uploadable package artifacts, so same-name checksum/signature artifacts are ignored. The regression adds a same-name checksum artifact and still hashes the MCPB file.

internal/pipe/mcp/mcp.go
caarlos0: Fixed in fadb9a5b9/4aba069e4: cleanup is kept for legacy token files created during publish, while files that existed before the run are restored. The regressions cover both preserving pre-existing files and removing files created by login.

PR #7091, opened 2026-09-06
fix(sign,sbom,universalbinary): universal binary signing, per-variant names and duplicate slices

### LINKED ISSUE #6887: sbom: default binary document names overwrite CPU variants
### Reproduction

Build `demo` for Linux AMD64 with `goamd64: [v1, v2]`, then configure `sboms: [{artifacts: binary}]` without overriding `documents`.

### Actual and expected results

Both inputs receive `demo_1.2.3_linux_amd64.sbom.json`. The generator runs for both binaries, but the second document overwrites the first. Two SBOM artifact records remain, both pointing to the last input's document. Each target should retain a distinct SBOM associated with its own binary.

Verification built actual v1/v2 binaries and checked their embedded GOAMD64 settings. A recording generator then exercised production cataloging and wrote its input into each document. A variant-aware document-name control retained two paths and two correct input associations. This verifies output handling, not Syft's catalog contents or remote upload behavior.

The [default template omits the variant](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/sbom/sbom.go#L70-L71), while [cataloging registers each result](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/sbom/sbom.go#L150-L158).

### Narrow correction

Make default binary document names variant-aware without changing ordinary default-variant names. Test distinct output paths and surviving input associations, including a working custom-template control.

---
Audit priority: **P2**.

### LINKED ISSUE #6888: sign: binary_signs skips replacement macOS universal binaries
### Reproduction

Build a Darwin executable for AMD64 and ARM64, configure `universal_binaries: [{replace: true}]`, and enable `binary_signs: [{}]`. Keep IDs at their matching defaults.

### Actual and expected results

Universal-binary creation removes the two thin artifacts and registers one `UniversalBinary`. Binary signing then warns that no artifacts match, returns success, calls no signer, and creates no signature. The replacement executable should be included in binary signing, subject to the configured ID filter.

Verification used actual Go builds and production universal-binary generation. Mach-O inspection confirmed a valid fat binary containing both CPU architectures. A recording signer established that no signing command ran. This is an artifact-selection defect, not a cryptographic verification failure.

The [binary-signing filter](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/sign/sign_binary.go#L82) selects only `artifact.Binary`, although universal-binary processing has already run.

### Narrow correction

Include `artifact.UniversalBinary` alongside `artifact.Binary`, preserving ID filtering. Add replacement-universal, ordinary-binary, and excluded-ID regressions that assert signer calls and resulting signature artifacts.

---
Audit priority: **P2**.

### LINKED ISSUE #6909: universalbinary: duplicate CPU slices produce an invalid fat binary
### Trigger

Configure one Go build for `darwin_amd64_v1`, `darwin_amd64_v2`, and `darwin_arm64`, then enable a universal binary selecting that build ID. Both AMD64 variants have the same Mach-O CPU and subtype, despite their distinct Go target strings.

### Observed

Real Go compilation produced both AMD64 inputs with CPU `CpuAmd64`, subtype `3`, and an ARM64 input. The production universal-binary pipe accepted and combined them. `lipo -info` rejected the output with `duplicate x86_64 slices`; Go's Mach-O parser reported a duplicate architecture. Keeping only AMD64 v1 plus ARM64 produced a valid two-architecture control. This verifies file-format rejection, not a claim about executing the malformed binary on every macOS version.

[Collection and assembly](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/universalbinary/universalbinary.go#L157-L179) use [a filter without CPU/subtype uniqueness](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/universalbinary/universalbinary.go#L253-L259).

### Smallest fix

Reject duplicate CPU/subtype pairs before writing or registering a universal output, rather than silently choosing a variant. Add the three-input failure and valid two-input control, asserting an actionable error and no registered invalid artifact.

---
Audit priority: **P2**.

caarlos0:
Fixes a group of related bugs in sign, sbom and universal binary pipes.

- Closes #6888: includes replacement universal binaries in `binary_signs` artifact selection while preserving ID filtering. The filter now matches `artifact.Binary` and `artifact.UniversalBinary`, the same shape `archives` already uses, so a universal binary is selected by its own `universal_binaries.id`.
- Closes #6887: makes default binary SBOM document names disambiguate build variants, so they no longer collide and now track the binary names.
- Closes #6909: rejects duplicate Mach-O CPU/subtype slices before writing or registering universal binaries, instead of emitting an invalid fat binary.

### Note on SBOM document names

The default `documents` for `sboms.artifacts: binary` now uses the same suffix as the `archives`, `nfpms`, `flatpaks`, `makeselfs`, `snapcrafts` and `binary_signs` defaults:

```
{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ with .Arm }}v{{ . }}{{ end }}{{ with .Mips }}_{{ . }}{{ end }}{{ if not (eq .Amd64 "v1") }}{{ .Amd64 }}{{ end }}.sbom.json
```

Previously every `goamd64`, `goarm` and `gomips` variant cataloged to a single document that later runs overwrote, leaving several SBOM artifacts pointing at one file.

**This renames a release asset** for users who set a non-default variant, do not override `documents`, and use `artifacts: binary`, for example `goamd64: [v3]` now produces `app_1.0.0_linux_amd64v3.sbom.json` instead of `app_1.0.0_linux_amd64.sbom.json`. Users on the defaults (`goamd64: [v1]`, no `goarm`/`gomips` list) see identical names.

### Docs

- `sbom.md` documents the new default.
- `binary_sign.md` records that `artifacts: binary` covers universal binaries, and that `ids` takes the `universal_binaries.id` rather than the source build IDs.


internal/pipe/universalbinary/universalbinary_test.go
```diff
@@ -497,5 +497,8 @@ func checkUniversalBinary(tb testing.TB, unibin *artifact.Artifact) {
 	require.True(tb, strings.HasSuffix(unibin.Path, unibin.ID()+"_darwin_all/foo"))
 	f, err := macho.OpenFat(unibin.Path)
 	require.NoError(tb, err)
+	defer func() {
```
caarlos0: should be t.Cleanup

caarlos0: Done in 92e91a7 — switched to `tb.Cleanup`.

internal/pipe/sbom/sbom.go
```diff
@@ -68,7 +68,7 @@ func setConfigDefaults(cfg *config.SBOM) error {
 	if len(cfg.Documents) == 0 {
 		switch cfg.Artifacts {
 		case "binary":
-			cfg.Documents = []string{"{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}.sbom.json"}
+			cfg.Documents = []string{`{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ if not (eq .Amd64 "v1") }}{{ .Amd64 }}{{ end }}.sbom.json`}
```
caarlos0: **Incomplete fix — `goarm` and `gomips` variants still collide on one filename.**

`goamd64` is one of several list-valued matrix dimensions. `goarm` and `gomips` also keep `.Arch` constant, so they hit the exact bug this line is fixing.

**Trigger:** `builds: [{goos: [linux], goarch: [arm], goarm: ["6", "7"]}]` with `sboms: [{artifacts: binary}]` and no `documents` override.

**Impact:** both binaries catalog to `demo_1.2.3_linux_arm.sbom.json`. The second run overwrites the first, and two SBOM artifacts are registered pointing at one file whose contents describe only the `v7` binary. Same for `gomips: [softfloat, hardfloat]`. Rendering this template directly confirms it:

```
arm  goarm 6        -> demo_1.2.3_linux_arm.sbom.json
arm  goarm 7        -> demo_1.2.3_linux_arm.sbom.json
mips softfloat      -> demo_1.2.3_linux_mips.sbom.json
mips hardfloat      -> demo_1.2.3_linux_mips.sbom.json
```

Secondary effect: the SBOM filename no longer matches the binary it describes, since the binary is named `demo_1.2.3_linux_armv6`.

**Smallest fix** — this repo already has one canonical suffix for exactly this, used in `archive.go:30`, `nfpm.go:38`, `flatpak.go:40`, `makeself.go:31`, `snapcraft.go:106`, and in `sign/sign_binary.go:16` right next door. Use it verbatim so SBOM names track binary names:

```go
cfg.Documents = []string{`{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ with .Arm }}v{{ . }}{{ end }}{{ with .Mips }}_{{ . }}{{ end }}{{ if not (eq .Amd64 "v1") }}{{ .Amd64 }}{{ end }}.sbom.json`}
```

(`goarm64`/`go386`/`goppc64`/`goriscv64` are uncovered by that suffix repo-wide — pre-existing, leave them out of this PR.)

**Needed test:** add a `goarm: 6/7` case to `TestSBOMCatalogBinaryGoamd64Variants` in `sbom_test.go:556`. It fails today against this line.

internal/pipe/sign/sign_binary.go
```diff
@@ -79,7 +79,10 @@ func (BinaryPipe) Run(ctx *context.Context) error {
 			default:
 				return fmt.Errorf("invalid list of artifacts to sign: %s", cfg.Artifacts)
 			}
-			filters := []artifact.Filter{artifact.ByType(artifact.Binary)}
+			filters := []artifact.Filter{artifact.ByTypes(
+				artifact.Binary,
+				artifact.UniversalBinary,
+			)}
```
caarlos0, line 85: **The fix does not apply when `binary_signs.ids` is set — the original bug survives for those users.**

Widening the type filter is right, but line 87 then applies `artifact.ByIDs(cfg.IDs...)`, and the universal binary's ID is not a build ID. `universalbinary.go:252` sets `extra[artifact.ExtraID] = unibin.ID`, and `universalbinary.go:39-43` defaults `unibin.ID` to the **project name** while `unibin.IDs` holds the build IDs.

**Trigger** — the CGO layout that `www/content/customization/builds/universalbinaries.md:19-27` recommends:

```yaml
builds:
  - id: darwin-amd64
  - id: darwin-arm64
universal_binaries:
  - ids: [darwin-amd64, darwin-arm64]   # id defaults to the project name
    replace: true
binary_signs:
  - ids: [darwin-amd64, darwin-arm64]
```

**Impact:** `replace: true` removes both `Binary` artifacts, and the surviving `UniversalBinary` has ID `<project name>`, so `ByIDs` drops it. Nothing is signed. `sign.go:146` only emits `log.Warn("no artifacts matching the given filters found")`, so the release still succeeds with no signatures — precisely the silent failure this commit set out to fix.

This is masked by the new test: `sign_binary_test.go:225` filters on `IDs: ["foo"]` and `:255` hand-sets `ExtraID: "foo"` on the `UniversalBinary`, which production code never does unless `universal_binaries.id` happens to equal the build ID.

**Smallest fix:** decide and document the contract. Either match a universal binary when `cfg.IDs` contains any of its source build IDs (its `unibin.IDs`), or keep today's semantics and state in `www/content/customization/sign/binary_sign.md:63` — which currently says "Build IDs of the binaries to sign" — that `ids` must also list the `universal_binaries.id`.

**Needed test:** a case where the `UniversalBinary` carries `ExtraID: "<project name>"` while `binary_signs.ids` lists only the build IDs, asserting whichever behaviour you choose.

internal/pipe/sbom/sbom_test.go
```diff
@@ -115,7 +115,7 @@ func TestSBOMCatalogDefault(t *testing.T) {
 			},
 			artifact: "binary",
 			cmd:      defaultCmd,
-			sboms:    []string{"{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}.sbom.json"},
+			sboms:    []string{`{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ if not (eq .Amd64 "v1") }}{{ .Amd64 }}{{ end }}.sbom.json`},
```
caarlos0: This assertion pins the new default, but the published documentation for it was not updated, so it is now false.

`www/content/customization/sbom.md:43` still states:

```
#   When "binary":   ["{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}.sbom.json"]
```

Users read that line to predict their release asset names, and for non-`v1` `goamd64` builds the name is now different. Compare `www/content/customization/sign/binary_sign.md:36`, which does spell out the full suffix for the signature default.

**Smallest fix:** update `sbom.md:43` to the final template — including the `Arm`/`Mips` parts if you take the change suggested on `sbom.go:71`.

caarlos0 (commented):
Three findings, all inline. Two are incomplete fixes of the exact bugs this PR targets; one is stale docs.

Verified independently: pipeline order is correct (`universalbinary.Pipe` at `pipeline.go:102` runs before `sign.BinaryPipe` at `:106`); `.Amd64` is always a string via `tmpl.go:202`, so `eq .Amd64 "v1"` cannot error, and arm64/386/universal SBOM names are byte-identical to before; signature names do not collide when `replace: false` (`_darwin_amd64` / `_darwin_arm64` / `_darwin_all`); `data[4:8]` cannot panic on a short file on this branch or on main (`os.ReadFile` allocates `cap >= 512`), unchanged by this PR. All three regression tests genuinely fail on `origin/main`, so the coverage is real.

Cross-cutting notes, no single site:

- **Scope.** Three unrelated fixes with different risk profiles: a new hard error (universalbinary), a release-asset rename (sbom), and a filter widening (sign). These want separate PRs and separate release notes.
- **Compatibility.** A user on `goamd64: [v3]` with default `documents` gets `app_1.0.0_linux_amd64.sbom.json` renamed to `app_1.0.0_linux_amd64v3.sbom.json`. That is a renamed release asset for a previously-working setup and needs a changelog note. Users on the default `v1` see no change.
- **`replace: false` behaviour change.** `binary_signs` now also produces a signature for each universal binary, so those users get one extra release asset. Intended, but undocumented — `www/content/customization/sign/binary_sign.md:56-58` still says `binary: the binaries`.
- **Branch is 27 commits behind `origin/main`**, which reworked `catalogArtifact`'s stdout/stderr `Close()` handling in `sbom.go`. Textually non-conflicting, but merge `main` in before merging.

Verification gap: `TestSignArtifacts/*` and `TestBinarySign/{default,filter}` fail locally with `exit status 2` from gpg. I reproduced them identically on unmodified `origin/main` in a scratch worktree, so they are environmental and unrelated to this change. They were not re-verified in CI as part of this review.

(Reviewer is a bot.)

internal/pipe/sbom/sbom.go
caarlos0: Fixed in faa5e9e. Adopted the canonical suffix verbatim, so the SBOM name now tracks the binary name:

```go
{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ with .Arm }}v{{ . }}{{ end }}{{ with .Mips }}_{{ . }}{{ end }}{{ if not (eq .Amd64 "v1") }}{{ .Amd64 }}{{ end }}.sbom.json
```

Generalised `TestSBOMCatalogBinaryGoamd64Variants` into `TestSBOMCatalogBinaryVariants` with `goarm: 6/7` and `gomips: softfloat/hardfloat` cases, and added an `ElementsMatch` on the resulting document names so a collision cannot pass silently. Confirmed both new cases fail against the previous template and pass against the new one:

```
--- FAIL: TestSBOMCatalogBinaryVariants/default_documents_include_goarm
--- FAIL: TestSBOMCatalogBinaryVariants/default_documents_include_gomips
```

Left `goarm64`/`go386`/`goppc64`/`goriscv64` alone: that gap is repo-wide and out of scope here.

internal/pipe/sign/sign_binary.go
caarlos0: Kept the current semantics and fixed the docs instead, in faa5e9e.

`archive.go:118-126` already uses the identical shape — `ByTypes(Binary, UniversalBinary, ...)` plus `ByIDs(archive.IDs...)` — so a universal binary is matched by its own `universal_binaries.id` everywhere in the codebase. Making `binary_signs` alone also match the source build IDs would make it inconsistent with `archives.ids` and would need the artifact to carry its source IDs, which it deliberately does not (`universalbinary.go:250-252` overwrites `ExtraID`). Not worth that surface.

So `www/content/customization/sign/binary_sign.md` now says under `ids`:

> To sign a macOS universal binary, use the `universal_binaries.id` here, not the IDs of the builds it was made from.

and `artifacts: binary` now reads `the binaries, including universal binaries`.

Added `TestBinarySignUniversalBinaryReplaced` with a realistic fixture — `replace: true`, no `Binary` artifacts left, universal binary carrying `ExtraID: "proj"` — covering all three cases: no `ids` signs it, the unibin ID signs it, and build IDs alone do not. That pins the contract instead of relying on the ID coincidence in the older test.

internal/pipe/sbom/sbom_test.go
caarlos0: Fixed in faa5e9e. `www/content/customization/sbom.md:43` now shows the full suffix, matching the final template from the `sbom.go:71` thread.

PR #7093, opened 2026-09-06
fix(exec,publish): cancellation, empty commands, SRPM artifacts and upload URLs

### LINKED ISSUE #6934: HTTP upload URL construction corrupts queries and filenames
Configure an HTTP or Artifactory target ending in `/files?token=abc`, leave `custom_artifact_name` false, and upload `notes.txt`. The token here is a synthetic example.

Recording local HTTP endpoints received `/files?token=abc/notes.txt`, not `/files/notes.txt?token=abc`. Filename cases also showed `notes#draft.txt` truncated to `/files/notes`, `notes?draft.txt` interpreted as a query, and `notes%.txt` rejected before any request. A parsed-and-escaped full-target control preserved each filename, the query, and an existing `%2F` path component. Both real publishing pipes exercised the shared helper; no external upload or authentication was performed.

[The shared construction](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/http/http.go#L304-L308) appends the filename to the raw URL string. A filename should be encoded as path data rather than change the URL's query or fragment syntax.

Append an escaped filename through the parsed URL path. Add request-target regressions for queries, literal reserved characters, and already escaped project paths. Preserve the explicitly supplied full target when `custom_artifact_name` is enabled.

---
Audit priority: **P2**.

### LINKED ISSUE #6935: Source RPMs are omitted from HTTP and custom publishing
Generate a source RPM before publication, then configure an HTTP or Artifactory upload in archive mode with `exts: [src.rpm]`. A custom publisher operating on the same artifact collection also misses the source RPM.

Local reproductions registered producer-shaped `artifact.SourceRPM` entries with format `src.rpm` and extension `.src.rpm`. Both HTTP-based pipes returned success with zero requests, with or without the extension filter. The custom publisher returned success with zero command executions. Changing only the HTTP artifact type to `LinuxPackage`, or adding a binary-RPM custom-publisher control, caused the expected operation. Network requests stayed on loopback and custom commands used a recorder.

The shared contract is missing from the [HTTP type list](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/http/http.go#L208-L222) and [custom-publisher type list](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/exec/exec.go#L135-L149), despite source RPM generation occurring before publication.

Include `artifact.SourceRPM` in both filters. Add focused regressions requiring one upload or command invocation for the source RPM, plus existing type and extension controls. This issue does not request a general expansion of every publisher's artifact types.

---
Audit priority: **P2**.

### LINKED ISSUE #6936: Empty custom publisher commands crash the release
Configure a custom publisher that matches at least one uploadable artifact, but omit `cmd`, set it to whitespace, or use `cmd: '{{ printf "" }}'`.

The command parser returns an empty argument slice without an error. The worker then constructs the process from `c.Args[0]` and `c.Args[1:]`. Isolated subprocess reproductions exercised the actual publisher and exited 2 with a slice-bounds panic for all three inputs. A control with the same empty command but no matching artifacts returned nil because no worker ran. No publishing command or network request was executed.

The unsafe assumption connects [argument parsing](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/exec/exec.go#L177-L212) to [process construction](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/exec/exec.go#L94-L100). Invalid user configuration should produce a normal error identifying the publisher, not terminate the release process with a worker panic.

Reject an empty parsed argument list before starting the command. Add omitted, whitespace, and template-to-empty cases with a matching artifact, and retain a valid-command control. Keep the no-matching-artifact behavior unchanged.

---
Audit priority: **P2**.

### LINKED ISSUE #6939: exec: cancellation waits for descendant-held output pipes
### Problem

A hook or custom publisher can start a descendant that inherits stdout/stderr. Canceling the release kills the direct command, but its runner can remain blocked while the descendant keeps the output descriptors open.

Both [the shared shell runner](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/shell/shell.go#L27-L47) and [the custom-publisher runner](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/exec/exec.go#L100-L122) use `CommandContext` without a bound on post-exit pipe draining.

### Evidence

On macOS/arm64, barrier-controlled subprocess tests observed the direct shell's exit event, confirmed that the descendant remained alive, and found both production runners waiting in `Cmd.Wait` for output-copy goroutines. They remained blocked during a 250 ms observation window, then returned immediately after the descendant was released. The observation window is not an asserted product SLA. Linux and Windows behavior was not exercised.

### Correction

Bound pipe draining with an appropriate `Cmd.WaitDelay`, preserving captured output and cancellation errors in both runners. Add regressions synchronized on direct-process exit and descendant release. This does not itself terminate descendants; broader process-tree termination needs a separate policy.

---
Audit priority: **P2**.

caarlos0:
Fixes a group of related bugs in command execution and publishing.

- Closes #6939: Bound command pipe draining after cancellation for shell hooks and custom publishers.
- Closes #6936: Reject empty custom publisher commands with a normal error before process construction.
- Closes #6935: Include Source RPM artifacts in HTTP uploads and custom publishing filters.
- Closes #6934: Append artifact names as escaped URL path data without corrupting queries or filenames.

### Behaviour notes

- `cmd.WaitDelay` is 1s. It also bounds the normal post-exit pipe drain, so a shorter value breaks
  successful commands: measured with a slow reader, `1ms` and `10ms` returned `ErrWaitDelay` on 10/10
  successful runs with truncated output, while `1s` gave 0/10 and full output.
- A command that exits 0 but leaves a descendant holding stdout and stderr now **succeeds** with a
  warning, instead of failing with `ErrWaitDelay`. A publisher ending in `something &` did its work,
  so a red release would be wrong. Cancellation still fails, because a killed process is not successful.
- `artifact.ByID` now exempts `SourceRPM`, next to `UploadableSourceArchive`. Source RPMs carry no ID,
  so an `ids:` filter previously dropped them and left #6935 only half closed.
- Artifact names are escaped with `(&url.URL{Path: name}).EscapedPath()`. That escapes each path
  segment, so a `name_template` containing a directory keeps its layout on the server
  (`/files/sub/dir/notes.txt`, not `/files/sub%2Fdir%2Fnotes.txt`), while leaving `;` and `,`
  unescaped as before, which matters for Artifactory matrix parameters.


internal/exec/exec_test.go
```diff
@@ -312,6 +346,93 @@ func TestExecute(t *testing.T) {
...
+func TestExecuteCommandCancellationWithDescendantHeldOutputPipe(t *testing.T) {
+	testlib.SkipIfWindows(t, "uses unix shell and fifo")
+
+	dir := t.TempDir()
+	ready := filepath.Join(dir, "ready")
+	release := filepath.Join(dir, "release")
+	require.NoError(t, mkfifo(ready, 0o600))
+	t.Cleanup(func() {
+		require.NoError(t, os.WriteFile(release, nil, 0o600))
+	})
```
caarlos0: **The test leaks a shell process that runs for ever.**

`t.TempDir()` registers its own cleanup before this `t.Cleanup`. Go calls cleanups in reverse order. This function therefore writes `release`, and `RemoveAll` deletes `release` some microseconds later. The descendant tests for the file one time each second, so it almost never finds it.

I measured this with a standalone test that copies the code shape of this one:

```
go test returned at:  1788670668
last heartbeat:       1788670674   now: 1788670674
```

The descendant was still alive 6 seconds after `go test` returned.

Impact: each run of `go test ./internal/exec/... ./internal/shell/...` leaves two orphan `sh` + `sleep` pairs on the developer machine. `-count=N` multiplies them. CI is safe, because the container stops.

Smallest fix: let the descendant stop by itself, then delete the marker file and this cleanup:

```go
Args: []string{"sh", "-c", `sh -c ': > "$READY"; sleep 30' & while :; do sleep 1; done`},
```

The descendant holds the pipes for much more than `WaitDelay`, thus the test continues to fail if you remove the `WaitDelay` line.

No new test is necessary. The change removes the hazard.

The same problem is at `internal/shell/shell_test.go:73`.

internal/shell/shell_test.go
```diff
@@ -60,6 +63,40 @@ func TestRunCommand(t *testing.T) {
...
+	t.Run("cancellation with descendant-held output pipe", func(t *testing.T) {
+		testlib.SkipIfWindows(t, "uses unix shell and fifo")
+
+		dir := t.TempDir()
+		ready := filepath.Join(dir, "ready")
+		release := filepath.Join(dir, "release")
+		require.NoError(t, mkfifo(ready, 0o600))
+		t.Cleanup(func() {
+			require.NoError(t, os.WriteFile(release, nil, 0o600))
+		})
```
caarlos0: **Same process leak as `internal/exec/exec_test.go:387`.**

`t.TempDir()` registers its cleanup first, so `RemoveAll` deletes `release` microseconds after this function writes it. The descendant polls one time each second and does not find the file. It then loops for ever after `go test` returns.

Smallest fix: use a self-terminating descendant and delete the marker file and this cleanup:

```go
[]string{"sh", "-c", `sh -c ': > "$READY"; sleep 30' & while :; do sleep 1; done`},
```

See `internal/exec/exec_test.go:387` for the measurement.

internal/http/http.go
```diff
@@ -341,6 +343,29 @@ func uploadAsset(ctx *context.Context, upload *config.Upload, artifact *artifact
...
+			target += "/"
+		}
+		return target + url.PathEscape(name), nil
+	}
+
+	path := u.EscapedPath()
+	if !strings.HasSuffix(path, "/") {
+		path += "/"
+	}
+	path += url.PathEscape(name)
```
caarlos0: **`url.PathEscape` escapes `/`. This changes the destination of extra files that have a directory in their name.**

Trigger:

```yaml
uploads:
  - name: x
    target: https://h/files/
    extra_files:
      - glob: ./notes.json
        name_template: "metadata/{{ .Version }}.json"
```

`extrafiles.Find` (`internal/extrafiles/extra_files.go:47`) copies `name_template` into `artifact.Name` without a check, and `uploadWithFilter` adds it as an artifact.

Measured, old against new:

```
old = "http://h/files/sub/dir/notes.txt"
new = "http://h/files/sub%2Fdir%2Fnotes.txt"
```

The file goes to a different location on the server.

Smallest fix: escape each segment.

```go
parts := strings.Split(name, "/")
for i, p := range parts {
	parts[i] = url.PathEscape(p)
}
path += strings.Join(parts, "/")
```

This keeps the fix for `#`, `?`, `%`, and space.

Test needed: add a case to `TestUploadArtifactNameTargetURL` with the name `sub/dir/notes.txt`, the target `/files/`, and the expected result `/files/sub/dir/notes.txt`.

internal/http/http.go
```diff
@@ -216,6 +217,7 @@ func uploadOne(ctx *context.Context, upload config.Upload, kind string, check Re
 			artifact.Flatpak,
 			artifact.PySdist,
 			artifact.PyWheel,
+			artifact.SourceRPM,
```
caarlos0, line 220: **`ids:` continues to remove source RPMs, thus #6935 is only half closed.**

`internal/pipe/srpm/srpm.go:239-246` does not set `artifact.ExtraID`, and the `SRPM` config has no `id` field. `artifact.ByID` (`internal/artifact/artifact.go:714-723`) accepts an artifact that has no matching ID only if its type is `Checksum`, `UploadableSourceArchive`, `UploadableFile`, or `Metadata`. `SourceRPM` is not in that list, and `a.ID()` is always empty.

Trigger:

```yaml
uploads:
  - name: x
    target: https://h/files/
    ids: [mypkg]
```

The filter at line 228 then removes the source RPM, and the user sees the same failure as in #6935. `publishers[].ids` at `internal/exec/exec.go:187` has the same problem.

Smallest fix: add `SourceRPM` to the exempt list in `ByID`, adjacent to `UploadableSourceArchive`. A source RPM is a source-level artifact, like the source archive.

Test needed: a copy of `TestUploadSourceRPM` with `IDs: []string{"other"}` that requires one request.

internal/exec/exec.go
```diff
@@ -94,6 +96,7 @@ func executePublisher(ctx *context.Context, publisher config.Publisher) error {
 func executeCommand(c *command, artifact *artifact.Artifact) error {
 	//nolint:gosec
 	cmd := exec.CommandContext(c.Ctx, c.Args[0], c.Args[1:]...)
+	cmd.WaitDelay = time.Second
```
caarlos0, line 100: **Decide what a successful command with a held pipe must do, then pin the decision with a test.**

I measured the two states with `cmd.Stdout` set to an `io.Writer`, as it is here:

```
process exits 0, descendant holds the pipes
  WaitDelay=0  (old)  Wait() blocked for ever
  WaitDelay=1s (new)  err = "exec: WaitDelay expired before I/O complete"  (errors.Is ErrWaitDelay == true)
```

This is not a regression. A hang becomes an error, and an error is better. But the PR makes this choice without a test and without a message.

Trigger: a publisher command or a hook that exits 0 and leaves a descendant with stdout and stderr open, for example `something &` with no `wait`. The user sees `publishing failed`, `cmd=<name>`, and output that stops at the 1-second mark. Nothing tells the user that the command was successful.

Smallest fix, if you want to keep the success result:

```go
if errors.Is(runErr, exec.ErrWaitDelay) && cmd.ProcessState.Success() {
	log.Warn("command left its output open; the output can be incomplete")
	runErr = nil
}
```

A test is necessary for either decision: a command that exits 0 and leaves a descendant with the pipes open, with an assertion on the result that you select. No test covers this path today, thus a later change can reverse the behavior without a failure.

`internal/shell/shell.go:28` has the same open question.

internal/http/http_test.go
```diff
@@ -750,3 +750,116 @@ func TestManyUploads(t *testing.T) {
...
+				Type: artifact.UploadableArchive,
+			})
+
+			require.NoError(t, Upload(ctx, []config.Upload{{
+				Name:               "target-url",
+				Mode:               ModeArchive,
+				Target:             srv.URL + tt.target,
+				CustomArtifactName: tt.customArtifactName,
+			}}, "test", func(*http.Response) error { return nil }))
+			require.Equal(t, tt.want, <-requestURIs)
```
caarlos0, line 922: **This test blocks for 10 minutes. It does not fail.**

If the artifact stops matching the filter, `uploadWithFilter` logs `no artifacts found` and returns `nil` (`internal/http/http.go:277-279`). `require.NoError` is then satisfied, and this receive blocks until the `go test` panic. The panic message does not identify the cause.

Smallest fix: count the requests, as `TestUploadSourceRPM` does:

```go
require.Equal(t, int64(1), requests.Load())
```

Alternatively, replace the receive with a `select` that has a `time.After` case and a `t.Fatal`.

internal/http/http.go
```diff
@@ -341,6 +343,29 @@ func uploadAsset(ctx *context.Context, upload *config.Upload, artifact *artifact
...
+	path := u.EscapedPath()
+	if !strings.HasSuffix(path, "/") {
+		path += "/"
+	}
+	path += url.PathEscape(name)
+
+	u.Path, err = url.PathUnescape(path)
+	if err != nil {
+		return "", err
+	}
```
caarlos0: **This error cannot occur. Remove it.**

`path` is `u.EscapedPath()`, plus `"/"`, plus `url.PathEscape(name)`. Each part is always valid escaping, thus `url.PathUnescape` always succeeds. I tested 26 inputs, which include `://bad`, `https://h/100%discount/`, `https://h:notaport/x/`, and `""`. The error was nil each time.

The `(string, error)` signature and the wrap at lines 307-309 are therefore dead code. Your local commit `ab6db71` already removes them. Include it in this PR.

Related: the `url.Parse` error branch at lines 348-353 is reachable (`https://h/100%discount/` gives a parse error), but it has no use. The string that it builds fails in `http.NewRequest` with the same parse error. Return the error, or delete the branch.

caarlos0 (commented):
Review by an AI agent (GitHub Copilot CLI), started by @caarlos0.

**Verdict: changes requested.** The four bugs are real. I confirmed each premise with a reproduction. The direction is correct. Seven items need work. Two of them are important.

Each finding is on the line that it concerns.

### Cross-cutting items

**1. The merge commit `1c47409` reverted `0c437ec` "chore: remove fifo".**
`git diff --stat 0c437ec 1c47409` shows the four `fifo_*_test.go` files as *added*, although the first parent deletes them. `main` never had these files. Redo the merge, or delete the files again.

**2. The documentation is now incorrect.**
`www/content/customization/publish/upload.md:36` and the `custom_artifact_name` comment at lines 204-207 tell the user that GoReleaser adds the artifact name "to the end of the URL". GoReleaser now adds the name to the URL path, before the query. Add one sentence to `upload.md` and to `artifactory.md`.

**3. `appendArtifactNameToTargetURL` has no direct unit test.**
Each test goes through `httptest` with a valid `srv.URL`. These branches are never used: the `url.Parse` error path, a target with an empty path, and names that contain a space, a `/`, or `..`. A table test of approximately 20 lines covers all of them.

### Verification gaps

- I did not run `go build`, `go vet`, or the repository test suite. The system load average was above 12. I got my evidence from small standalone Go programs that copy the code shape of the changed functions.
- All probes ran on macOS/arm64. I did not test Linux or Windows behavior of `WaitDelay`.

### Findings that I rejected

- *"`WaitDelay` breaks a pipeline that operates correctly today."* It does not. I measured the old behavior: a command that exits 0 and leaves a descendant with the pipes open blocks in `Wait` for ever. There is no pipeline that operates correctly today.
- *"One second can truncate the output of a normal command."* It cannot. The timer starts only when the process exits or when the context ends. A 5 MB output drained in 50 ms, with the full output and a nil error.
- *"An opaque target URL loses the artifact name."* This is true, but it has no effect. `http.NewRequest` gives an empty `Host`, so the transport rejects the request with the old code and with the new code.

internal/exec/exec_test.go
caarlos0: Confirmed and fixed in 001b0f5f7.

I reproduced the leak before changing anything. After one `go test -count=1 -run TestExecuteCommandCancellation... ./internal/exec/`, the machine held **43** orphan `sh` processes. `ps -eo pid,etime,args` showed two of them at `03:01:29` — they were still polling three hours after the run that started them.

Your ordering analysis is exactly right: `t.TempDir()` registers its cleanup first, cleanups run LIFO, so the marker write happens and `RemoveAll` deletes it microseconds later, well inside the 1-second poll interval.

I took your fix:

```go
Args: []string{"sh", "-c", `sh -c ': > "$READY"; sleep 30' & while :; do sleep 1; done`},
```

The `release` marker and the `t.Cleanup` are gone. I verified both properties afterwards:

- **No leak:** ran both packages, waited 35s, then counted. `0` descendants left.
- **Test keeps its teeth:** with `cmd.WaitDelay` removed, the test still fails with `command did not return while descendant held stdout and stderr open`.

I also killed the 43 orphans that earlier runs had left on this machine.

internal/shell/shell_test.go
caarlos0: Fixed in 001b0f5f7, same change as in `internal/exec/exec_test.go`.

The descendant is now self-terminating, and the `release` marker plus the `t.Cleanup` are removed:

```go
[]string{"sh", "-c", `sh -c ': > "$READY"; sleep 30' & while :; do sleep 1; done`},
```

Verified after the change: `0` leaked descendants once both packages finish, and the subtest still fails with `command did not return while descendant held stdout and stderr open` when `cmd.WaitDelay` is removed.

internal/http/http.go
caarlos0: Confirmed and fixed in 0df4dd7ae.

I checked the premise first: `extrafiles.Find` sets `name = n` straight from `name_template` with no check for `/` (`internal/extrafiles/extra_files.go:47`), so an artifact name really can carry a directory.

Applied your fix as a small helper:

```go
func escapePath(name string) string {
	parts := strings.Split(name, "/")
	for i, part := range parts {
		parts[i] = url.PathEscape(part)
	}
	return strings.Join(parts, "/")
}
```

I added the `sub/dir/notes.txt` case you asked for, plus one that mixes both concerns (`sub dir/notes#draft.txt` → `/files/sub%20dir/notes%23draft.txt`) so the segment escaping cannot silently regress.

The new case has teeth. With `escapePath` reverted to plain `url.PathEscape`:

```
expected: "/files/sub/dir/notes.txt"
actual  : "/files/sub%2Fdir%2Fnotes.txt"
```

internal/http/http.go
caarlos0: Confirmed and fixed in 0df4dd7ae.

I verified both halves: `internal/pipe/srpm/srpm.go:239` adds the artifact with only `ExtraFormat` and `ExtraExt`, and `grep ExtraID internal/pipe/srpm/srpm.go` returns nothing, so `a.ID()` is always empty.

Applied your fix in `ByID`, next to `UploadableSourceArchive`:

```go
return a.Type == Checksum ||
	a.Type == UploadableSourceArchive ||
	a.Type == SourceRPM ||
	...
```

Added `TestUploadSourceRPMWithUnrelatedIDs` with `IDs: []string{"other"}`, asserting one request. Without the `ByID` change it fails with `expected: 1, actual: 0`.

This also covers `publishers[].ids` in `internal/exec/exec.go:187`, since that path goes through the same filter.

One note: `ByID` is shared, so I ran the whole suite. The only failures are pre-existing and environmental on this machine (docker socket missing, node SEA disabled, git exit 128). I confirmed they fail identically with my changes stashed.

internal/exec/exec.go
caarlos0: Agreed, and decided in 001b0f5f7: **a command that exits 0 keeps its success**, with a warning.

I took your suggested shape in both packages:

```go
runErr := cmd.Run()
if errors.Is(runErr, exec.ErrWaitDelay) && cmd.ProcessState.Success() {
	log.Warn("command exited successfully but left its output open: output may be incomplete")
	runErr = nil
}
```

Reasoning: the command did the work. Reporting `publishing failed` for a publisher that ends with `something &` would turn a working release into a red one. Cancellation is unaffected, because a killed process gives `ProcessState.Success() == false`.

Pinned with a test in each package (`TestExecuteCommandSucceedsWithDescendantHeldOutputPipe`, and the `success with descendant-held output pipe` subtest). Both use a command that exits 0 while a descendant holds the pipes. With the new branch removed they fail:

```
Received unexpected error: exec: WaitDelay expired before I/O complete
Messages: a successful command must not fail because a descendant held its pipes
```

On the 1-second value itself: I kept it deliberately. The delay also bounds the normal post-exit pipe drain, and shortening it breaks successful commands. Measured with a slow reader — at `WaitDelay=1ms` and `10ms`, 10/10 successful runs returned `ErrWaitDelay` with truncated output; at `1s`, 0/10 and full output.

internal/http/http_test.go
caarlos0: Fixed in 0df4dd7ae, with the counter approach you suggested.

```go
require.Equal(t, int64(1), requests.Load(), "the artifact must have been uploaded")
require.Equal(t, tt.want, <-requestURIs)
```

The counted assertion runs first, so a filtered-out artifact fails immediately instead of blocking on the receive.

This proved itself straight away. When I ran the new `TestUploadSourceRPMWithUnrelatedIDs` against the unfixed `ByID`, it failed in 0.00s with `expected: 1, actual: 0` rather than hanging until the panic.

internal/http/http.go
caarlos0: First half done: `ab6db71` is now in this PR (pushed with 0df4dd7ae). The `(string, error)` signature and the wrap at 307-309 are gone.

Independent of your 26 inputs, I fuzzed the function for 45s — **5.6M executions, zero errors** — which matches your result that `url.PathUnescape` cannot fail there.

On the `url.Parse` branch, I did not take either option, so I am leaving this thread open for your call.

I checked what actually reaches the user with `https://h/100%discount/`:

```
url.Parse error : parse "https://h/100%discount/": invalid URL escape "%di"
NewRequest error: parse "https://h/100%discount/notes.txt": invalid URL escape "%di"
```

`http.NewRequest` rejects it with the same diagnosis, wrapped as an upload error. So:

- **Deleting the branch** is not available as written — `u` is nil there, so we would panic.
- **Returning the error** restores the `(string, error)` signature and the call-site wrap that we just removed, to report a failure three lines earlier that the user already gets clearly.

I kept the branch and documented why:

```go
// target is unparseable, so http.NewRequest will reject it later with
// the same error. Keep the old naive behaviour rather than guess.
```

That also preserves the pre-PR behaviour for unparseable targets. If you would rather fail fast with a dedicated message, say so and I will add the error return — it is a live path, so unlike the one we removed it would not be dead code.

internal/http/http.go
caarlos0: Small correction to my previous reply: `semgrep` blocked `strings.Join(parts, "/")` with `rules.use-strings-join-path`, so the helper now builds the path with a `strings.Builder` (36303a0be).

`path.Join` was not an option, since it cleans the result and would alter already-escaped segments. Behaviour and tests are unchanged; the full check suite is now green.

internal/http/http.go
```diff
@@ -341,6 +340,40 @@ func uploadAsset(ctx *context.Context, upload *config.Upload, artifact *artifact
...
+		sb.WriteByte('/')
+	}
+	for i, part := range strings.Split(name, "/") {
+		if i > 0 {
+			sb.WriteByte('/')
+		}
+		sb.WriteString(url.PathEscape(part))
+	}
+	return sb.String()
+}
```
caarlos0, line 366: **The standard library does this in one expression.**

`(&url.URL{Path: name}).EscapedPath()` escapes each path segment and keeps `/` as a separator, which is exactly the contract of this helper:

```go
func appendEscapedName(base, name string) string {
	if !strings.HasSuffix(base, "/") {
		base += "/"
	}
	return base + (&url.URL{Path: name}).EscapedPath()
}
```

I compared the two forms over 144 base/name pairs. They agree on every case in `TestUploadArtifactNameTargetURL` and on `#`, `?`, `%`, space, non-ASCII, `+`, `[`, `]`, `|`, `..`, `./`, a leading `/`, a trailing `/`, `//`, an empty name, and `%2F`.

They differ on three inputs only, and there the standard library keeps the behavior that GoReleaser had before this PR:

```
name="a;b.txt"    current="a%3Bb.txt"      proposed="a;b.txt"
name="a,b.txt"    current="a%2Cb.txt"      proposed="a,b.txt"
name="a;b,c.txt"  current="a%3Bb%2Cc.txt"  proposed="a;b,c.txt"
```

`url.PathEscape` uses `encodePathSegment`, which escapes `;` and `,`. `EscapedPath` uses `encodePath`, which does not. `;` has a meaning for Artifactory matrix parameters, and this repository shows one at `www/content/customization/publish/upload.md:206`. Escaping it was not an aim of #6934.

This also removes the reason for commit `36303a0be`: there is no `strings.Join` and no loop, thus the `use-strings-join-path` semgrep rule does not apply, and the comment about `path.Join` becomes unnecessary.

Test needed: one case in `TestUploadArtifactNameTargetURL`, target `/files/`, artifact `a;b,c.txt`, want `/files/a;b,c.txt`. This pins the difference.

internal/artifact/artifact.go
```diff
@@ -713,9 +713,10 @@ func Not(filter Filter) Filter {
 // ByID filter artifacts by an `ID` extra field.
 func ByID(id string) Filter {
 	return func(a *Artifact) bool {
-		// checksum and source archive are always for all artifacts, so return always true.
+		// checksum and source artifacts are always for all artifacts, so return always true.
 		return a.Type == Checksum ||
 			a.Type == UploadableSourceArchive ||
+			a.Type == SourceRPM ||
```
caarlos0, line 719: **Add this exemption to `TestByIDs`.**

This line changes a shared filter. A source RPM now passes every `ids:` filter in the repository. Through `ReleaseUploadableTypes()` that reaches four pipes: `release/release.go:184`, `checksums/checksums.go:187`, `sign/sign.go:129` with `artifacts: all`, and `blob/upload.go:196`. A user who sets `checksum.ids` gets a different `checksums.txt` after this change.

`TestByIDs` (`internal/artifact/artifact_test.go:666-706`) covers only `Checksum` of the five exempt types. `TestReleaseUploadableTypes` pins the type list, not the exemptions. So the only cover for this line is `TestUploadSourceRPMWithUnrelatedIDs`, which is in one pipe and reaches the filter indirectly. A person who removes this line to tidy the condition gets a green `internal/artifact` package.

Smallest fix: add an artifact of type `SourceRPM` with no `ExtraID` to the `TestByIDs` fixture, and correct the expected counts.

The change itself is correct. `internal/pipe/srpm/srpm.go:238-247` never sets `ExtraID`, and `config.SRPM` (`pkg/config/config.go:985`) has no `id` field and is one object, so `a.ID()` is always empty and no user can name it. I checked the other 26 `ByIDs` call sites: each one ANDs a type filter that has no `SourceRPM`, thus nothing else changes.

caarlos0 (commented):
Review by an AI agent (GitHub Copilot CLI), started by @caarlos0. This review supersedes https://github.com/goreleaser/goreleaser/pull/7093#pullrequestreview-5124219403. That review is a comment review, so the API cannot dismiss it; ignore it and use this one.

**Verdict: no blocking defect. This is good to merge after the documentation change below.** Each finding from the previous review is fixed, and I confirmed each fix independently. Two small items and one documentation item remain. None of them changes behavior.

### Previous findings, confirmed fixed

| Finding | Fixed by | How I confirmed it |
|---|---|---|
| Tests leaked `sh` + `sleep` processes for ever | `001b0f5f7` | The release-marker race is gone. The descendant is now `sleep 30` and stops by itself. |
| A successful command failed with `ErrWaitDelay` | `001b0f5f7` | Measured: exit 0 with a held pipe gives `ErrWaitDelay`, and the new branch turns it into `nil`. A non-zero exit gives `exit status 1` and a cancellation gives `signal: killed`, so neither is hidden. |
| `url.PathEscape` destroyed directories in artifact names | `0df4dd7ae` | `sub/dir/notes.txt` is now byte-identical to the old output. |
| `ids:` dropped source RPMs | `0df4dd7ae` | `srpm.go` sets no `ExtraID` and `config.SRPM` has no `id` field, thus the `ByID` exemption is the correct fix. |
| Unreachable error in the URL builder | `ab6db71e2` | Removed. |
| The merge reverted `0c437ec` | `10cefd3bf` | No fifo code remains. |
| `TestUploadArtifactNameTargetURL` blocked instead of failing | `0df4dd7ae` | The request counter is now asserted first. |

### Verification that I did

- **CI is green on `e554d959`**: `test (ubuntu-latest)`, `test (windows-latest)`, `lint`, `semgrep / scan`, `check`, and `license-check` all pass. This replaces the local build that I did not run.
- **`appendArtifactNameToTargetURL`**: I copied the current function into a standalone program and compared it with the old raw concatenation. The `Path`/`RawPath` invariant holds in each case, thus `URL.String()` always uses `RawPath` and never re-escapes `Path`. An empty name, `/lead.txt`, `trail/`, `a//b.txt`, `../evil.txt`, `./x.txt`, and `sub/dir/notes.txt` each give the same URL as before the change. Only `#`, `?`, `%`, space, `;`, and `,` change, which is the intention of #6934.
- **`ByID` blast radius**: I read each of the 30 `artifact.ByIDs` call sites. Only four have a type filter that contains `SourceRPM`, all through `ReleaseUploadableTypes()`: `release/release.go:184`, `checksums/checksums.go:187`, `sign/sign.go:129` with `artifacts: all`, and `blob/upload.go:196`. For a user who sets `ids:`, a source RPM now goes into the release, the checksums file, the signatures, and the blob bucket. This is correct and consistent with the `UploadableSourceArchive` precedent, and the PR description records it. `checksums.txt` content changes for such a user, so put one line in the release notes.

### Documentation, still missing

`www/content/customization/publish/upload.md:203-207` continues to tell the user that GoReleaser adds the artifact name "to the end of the URL". After this change the name goes into the URL path, before the query. Line 36 stays correct, and `artifactory.md` stays correct. Replace the comment with:

```yaml
    # Custom artifact name.
    # By default, GoReleaser adds the artifact name to the target URL path,
    # before the query string, and escapes it as URL path data. A `/` in the
    # artifact name continues to be a path separator.
    # If you enable this, you must supply the artifact name as part of the
    # target URL, because GoReleaser will not add it. The pre-computed name is
    # available as _ArtifactName_, for example:
    # target: https://some.server/some/path/example-repo-local/{{ .ArtifactName }};deb.distribution=xenial
    custom_artifact_name: true
```

### Smaller items

- `appendArtifactNameToTargetURL` has no direct unit test. Each case goes through `httptest` with a valid `srv.URL`, thus the `url.Parse` error path at `http.go:345-349` is never used. A table test on the function is a few lines and is much faster than the HTTP round trip.
- Commit `0df4dd7ae` contains three concerns: the `internal/artifact` exemption, the per-segment escaping in `internal/http`, and a test change. The repository convention is one logical change for each commit. The scope `upload` is also incorrect for the `internal/artifact/artifact.go` hunk; use `fix(artifact)`.

### Findings that I rejected

- *"Remove `&& cmd.ProcessState.Success()`, because `ErrWaitDelay` always implies success."* The condition is indeed always true: `Cmd.Wait` assigns `ErrWaitDelay` only when `err == nil`, which needs `state.Success()`. But the guard is one cheap term, it makes the intention local, and it does not depend on a reader knowing that rule. Removing it is a refactor with no user effect.
- *"An opaque target URL loses the artifact name."* True, but it has no effect. I measured that the old URL and the new URL both fail in the transport with `http: no Host in request URL`.
- *"Increase the 3-second and 5-second test budgets."* The measured times are 1.00 s and 1.01 s against a fixed 1-second `WaitDelay`. The margin is sufficient.

### Verification gaps

- I did not run the repository test suite or a local build. I used the CI result on `e554d959` and standalone Go programs that copy the exact shape of the changed functions.
- All my probes ran on macOS/arm64. The Windows production path for `WaitDelay` has no test, because the new tests call `testlib.SkipIfWindows`. The Windows CI job is green, but it does not exercise this path.

internal/http/http.go
caarlos0: Adopted in 7bb2a2099. You are right, and the `;`/`,` difference is the deciding point.

I reproduced your comparison independently over 26 names, including the whole existing case list plus `+`, `[`, `]`, `|`, non-ASCII, `..`, `./`, leading `/`, trailing `/`, `//`, empty, `%2F`, `:`, `@`, `=`, `&`, `$`. Exactly three disagree, and each is one you named:

```
name="a;b.txt"    current="a%3Bb.txt"      proposed="a;b.txt"
name="a,b.txt"    current="a%2Cb.txt"      proposed="a,b.txt"
name="a;b,c.txt"  current="a%3Bb%2Cc.txt"  proposed="a;b,c.txt"
```

Everything else is byte-identical, `/` included. So the standard library form keeps the pre-PR behaviour for `;` and `,` while still fixing `#`, `?`, `%` and space.

The helper is now:

```go
func appendEscapedName(base, name string) string {
	if !strings.HasSuffix(base, "/") {
		base += "/"
	}
	return base + (&url.URL{Path: name}).EscapedPath()
}
```

Added the case you asked for — target `/files/`, artifact `a;b,c.txt`, want `/files/a;b,c.txt`. It fails against the old escaping with `actual: "/files/a%3Bb%2Cc.txt"`.

As you noted, this also retires the reason for 36303a0be: no loop, no `strings.Join`, so the `use-strings-join-path` rule no longer applies and the `path.Join` comment is gone. `semgrep` stays green.

internal/artifact/artifact.go
caarlos0: Fixed in 7bb2a2099. Your point about the missing cover is the important one: I had changed a shared filter whose only guard was one pipe test.

`TestByIDs` now carries a source RPM with no `ExtraID`:

```go
{
	// source rpms carry no ID, so they must pass every ids filter.
	Name: "pkg.src.rpm",
	Type: SourceRPM,
},
```

Counts corrected to 3 / 4 / 4 / 5 / 6. I verified it now guards the line — deleting `a.Type == SourceRPM ||` fails `internal/artifact` directly:

```
--- FAIL: TestByIDs
    Error: "[check checksum]" should have 3 item(s), but has 2
```

So the tidy-up you described would no longer pass silently.

Thanks also for auditing the 26 other `ByIDs` call sites; that matches what I found for the upload and publish paths.

internal/http/http.go
caarlos0: Status update, since the helper changed in 7bb2a2099. This thread still needs your call, so I am leaving it unresolved.

The dead `(string, error)` signature and the wrap are gone, as agreed. `appendEscapedName` is now `(&url.URL{Path: name}).EscapedPath()`, per your other comment.

The `url.Parse` fallback is unchanged and still reads:

```go
u, err := url.Parse(target)
if err != nil {
	// target is unparseable, so http.NewRequest will reject it later with
	// the same error. Keep the old naive behaviour rather than guess.
	return appendEscapedName(target, name)
}
```

My position is unchanged too: deleting the branch is not possible as written, because `u` is nil there, and returning the error would restore the signature we just removed in order to report a failure the user already receives verbatim from `http.NewRequest`.

Say the word if you want the error return and I will add it in one commit — it is a live path, so it would not be dead code.

internal/http/http_test.go
```diff
@@ -750,3 +750,170 @@ func TestManyUploads(t *testing.T) {
...
+		"custom artifact name": {
+			target:             "/files/notes.txt?token=abc",
+			artifact:           "ignored#draft.txt",
+			customArtifactName: true,
+			want:               "/files/notes.txt?token=abc",
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			var requests atomic.Int64
+			requestURIs := make(chan string, 1)
```
caarlos0: **A duplicate-upload regression makes this test block for 10 minutes instead of failing.**

My earlier review asked for a request counter **or** a receive with a timeout. The counter at line 915 is good, but it cannot run in the failure case that matters: the handler sends on this channel while it holds the request. With a capacity of 1, a second request blocks the handler, the client waits for a response that never arrives, `Upload` never returns, and `require.NoError` at line 909 never completes. The test then reaches the package timeout, and the panic output does not identify the cause.

Trigger: any future change that uploads an artifact twice, for example a retry that becomes unconditional in `retryx.Do` (`http.go:373`), or a filter that stops removing duplicates.

Smallest fix: give the channel more room than the test can ever need, so the handler never blocks and the counter assertion is the thing that fails:

```go
requestURIs := make(chan string, 10)
```

`TestUploadSourceRPMWithUnrelatedIDs` at line 801 has the same capacity of 1 and the same exposure. `TestUploadSourceRPM` at line 756 uses 2, which is enough for one unexpected extra request but not for two.

No new test is needed. This only changes a hang into a readable failure.

caarlos0 (commented):
Review by an AI agent (GitHub Copilot CLI), started by @caarlos0, at `7bb2a20`. This supersedes review [5124415894](https://github.com/goreleaser/goreleaser/pull/7093#pullrequestreview-5124415894) and review [5124219403](https://github.com/goreleaser/goreleaser/pull/7093#pullrequestreview-5124219403). Both are comment reviews, thus the API cannot dismiss them; use this one.

**Verdict: the code is correct and complete. One documentation change remains before merge.** I would approve, but GitHub does not permit an approval on your own pull request.

### The two findings from the last review are fixed, and I checked each fix

**`appendEscapedName` (`http.go:361-366`)** now uses `(&url.URL{Path: name}).EscapedPath()`. I compared the new form with the previous per-segment loop over 144 base/name pairs. They agree everywhere except `;` and `,`, which now stay raw, as before this PR. The new case `"artifact name with matrix parameters"` pins that. The `semgrep / scan` job is green, which confirms the `use-strings-join-path` workaround from `36303a0b` is no longer necessary.

**`TestByIDs` (`artifact_test.go:696-714`)** now covers the exemption. I checked each count by hand against the fixture:

| Filter | Artifacts that pass | Expected |
|---|---|---|
| `ByID("check")` | check, checksum, pkg.src.rpm | 3 |
| `ByID("foo")` | foo, foobar, checksum, pkg.src.rpm | 4 |
| `ByIDs("foo")` | same as above | 4 |
| `ByIDs("foo", "bar")` | foo, foobar, bar, checksum, pkg.src.rpm | 5 |
| `ByIDs()` | all, because `autoOr` gives a nil filter | 6 |

Each count is correct, and the test is load-bearing: remove `a.Type == SourceRPM ||` and the first assertion gives 2 instead of 3.

### The one item that remains

`www/content/customization/publish/upload.md:203-207` still tells the user that GoReleaser adds the artifact name "to the end of the URL". The name now goes into the URL path, before the query. The pull request contains no change under `www/`. The replacement text is in my previous review; nothing about it has changed.

While you are in that file, `upload.md:64-65` says archive mode "will also include the `LinuxPackage` type which is generated by `nfpm` and the like". That sentence was already incomplete, and this change adds one more type to it. Archive mode now uploads `UploadableArchive`, `UploadableSourceArchive`, `Makeself`, `LinuxPackage`, `Flatpak`, `PySdist`, `PyWheel`, and `SourceRPM`. Give the list, or say "archive mode uploads all release artifacts except raw binaries".

Also, the last bullet of the pull request description no longer agrees with the code: it says "Artifact names are escaped per path segment", which was the rule before `7bb2a20`. The description becomes the squash-merge message, so please change it to something like: "Artifact names are escaped as URL path data. A `/` keeps the directory structure, and `;` and `,` stay raw for matrix-parameter targets."

### Verification that I did

- CI is green on `7bb2a20`: `test (ubuntu-latest)`, `test (windows-latest)`, `lint`, `semgrep / scan`, `check`, `license-check`. Codecov reports that every modified line is covered.
- `exts` filtering: I traced `ByExt` and `ByFormat`. For a source RPM, `Ext()` gives `.src.rpm` and `Format()` gives `src.rpm`, thus `exts: [rpm]` does **not** collect source RPMs, and `exts: [src.rpm]` and `exts: [.src.rpm]` both do. An empty `exts:` collects them, because `autoOr` gives a nil filter and `Or` treats nil as a match. There is no accidental inclusion for a user who already writes `exts: [rpm]`.
- Templates: a source RPM has empty `Goos` and `Goarch`, thus a target that uses `{{ .Os }}` gives empty path segments. This is not new. `UploadableSourceArchive` has the same empty metadata and has been in archive mode for a long time, and `upload.md:59-62` documents the rule.
- Concurrency: `Cmd.Wait` waits for every output-copy goroutine before it returns, on the `ErrWaitDelay` path also. Therefore `redact.Writer`, which is not safe for concurrent use, is not newly exposed by `WaitDelay`.

### Findings that I rejected

- *"Add `Makeself`, `Flatpak`, `MSIX`, or `SBOM` to the two type lists."* Those gaps are real, but they are older than this change, and #6935 says it does not request a general expansion. Open a separate issue. When you do, do not unify the lists on `ReleaseUploadableTypes()`: `exec` needs the three Docker types, which that list does not hold, and `http` archive mode would get `UploadableBinary`, which removes the archive/binary separation.
- *"Add a test with an empty `exts:`."* `TestUploadSourceRPM` already fails if `artifact.SourceRPM` leaves the archive type list. An empty-`exts` test protects no line that this change touches.
- *"Give `ExtraFormat: \"rpm\"` to the control artifact in `TestUploadSourceRPM`."* The control is excluded because its extension is `rpm`, not `src.rpm`. It stays excluded when `ExtraFormat` is present, so the test proves what it claims.

### Verification gaps

I did not run a local build or the test suite; I used the CI result on `7bb2a20` and standalone Go programs that copy the changed functions. All my probes ran on macOS/arm64. The Windows `WaitDelay` path has no test, because the new tests call `testlib.SkipIfWindows`.

internal/http/http_test.go
caarlos0: Fixed in f0cb1a480. Agreed on the mechanism: the handler sends while it still holds the request, so a full channel stops it from ever replying, and the counter I added cannot run because `Upload` has not returned.

All three now use a capacity the tests can never fill:

```go
// roomy on purpose: an unexpected extra upload must fail the request
// count, not block the handler and hang the test.
requestURIs := make(chan string, 10)
```

That covers `TestUploadSourceRPM` (was 2), `TestUploadSourceRPMWithUnrelatedIDs` (was 1) and `TestUploadArtifactNameTargetURL` (was 1).

Verified by adding a second artifact so the upload really does happen twice. The test now fails immediately and names the cause:

```
--- FAIL: TestUploadSourceRPMWithUnrelatedIDs (0.00s)
    http_test.go:836: Not equal:
        expected: 1
        actual  : 2
```

One honest caveat: in that particular duplicate scenario the old capacity of 1 also failed fast rather than hanging, so I did not reproduce an actual ten-minute block. The change still removes the whole class of risk at no cost, which is why I took it.

PR #7098, opened 2026-09-06
fix(builder/go): module download output, env ordering, ellipsis and c-archive artifacts

### LINKED ISSUE #6897: go: clean C-library builds omit generated header artifacts
### Trigger

Build a small Go main package exporting a C function with `buildmode: c-shared` or `buildmode: c-archive`, using a fresh output directory. Generated headers should be registered alongside their libraries so downstream release packaging can include them.

### Observed

The production Go builder ran real local compiler commands for both modes. Compilation produced a library and a nonempty `cexport.h`, but the artifact list contained only the library. Calling the existing header-discovery helper after compilation found that same header. This confirms artifact omission; a complete release upload was not part of verification.

[Header discovery runs before compilation](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/builders/golang/build.go#L282-L293). The [compiler runs later](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/builders/golang/build.go#L337), while [discovery returns nil for an absent file](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/builders/golang/build.go#L496-L503).

### Smallest fix

Discover generated headers after successful compilation. Add clean-directory regressions for both supported library modes, asserting that each existing header is registered with the correct target and build ID. Also assert that a failed compiler invocation does not register stale header artifacts.

---
Audit priority: **P2**.

### LINKED ISSUE #6898: go: ellipsis WASM builds register filenames that do not exist
### Trigger

Use `main: ./...` and `targets: [js_wasm]` in a module containing `cmd/foo/main.go`. Run with a fresh output directory, first without `mod_timestamp`, then with `mod_timestamp: '1700000000'`.

### Observed

The production Go builder ran real WASM compilation. Without timestamping, it successfully created a nonempty `foo` executable but registered the nonexistent path `foo.wasm`. With timestamping, compilation still produced `foo`, then the builder failed at `chtimes` because `foo.wasm` was absent. This was a real builder reproduction, not a full CLI release or an execution of the WASM program.

[Ellipsis compilation passes an output directory to Go](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/builders/golang/build.go#L453-L454), while [artifact construction adds the WASM suffix](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/builders/golang/build.go#L623-L624).

### Smallest fix

Align the actual output names with registered paths before timestamping, retaining the documented `.wasm` extension. Extend ellipsis build coverage to assert file existence for every artifact, with and without timestamping, rather than checking artifact metadata alone.

---
Audit priority: **P2**.

### LINKED ISSUE #6899: go: ellipsis discovery ignores the effective build context
### Trigger

Use `main: ./...` for a package whose only main file has `//go:build feature`, and enable `feature` through build tags, flags, build-level `GOFLAGS`, or a target override. Another trigger is cross-compilation when separate main packages contain host-only and target-only files.

### Observed

The production Go builder rejected all four configured-tag cases with `directory does not contain any main function`. The same package compiled successfully with explicit `main: .`, or when the tag was supplied through the ambient environment. On Darwin ARM64, discovery selected the Darwin-only main for a Linux ARM64 build; the real compiler then rejected it with `build constraints exclude all Go files`. An ambient Linux-target control selected and compiled the correct main.

[Package loading receives neither the effective environment nor selection flags](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/builders/golang/gomain/gomain.go#L87-L91).

### Smallest fix

Resolve and pass target environment and build-selection flags, including overrides, into main discovery. Preserve explicit-main behavior. Add tagged-only and cross-platform fixtures asserting both the selected packages and the real artifact paths.

---
Audit priority: **P2**.

### LINKED ISSUE #6900: go: build overrides randomly reorder dependent environment entries
### Trigger

Configure ordered environment entries `A=value` and `B={{.Env.A}}`. Add a matching target override that changes only `ldflags`, leaving environment entries unchanged. A matching override should not alter whether this dependency chain can be evaluated.

### Observed

A real production-builder control without overrides produced a binary printing `value`. With the override, verification captured the merged order `B={{.Env.A}}, A=value` and the error `map has no entry for key "A"`. The whole production builder also failed after a successful real build. Replaying the captured reversed order reproduced the templating failure deterministically; Go map iteration itself has no controllable seed.

[Override merging converts the ordered entries through a map](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/builders/golang/build.go#L375), and [conversion back iterates that map](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/pkg/context/context.go#L51-L57).

### Smallest fix

Preserve stable entry order and last-value override precedence. Add a merged-order regression plus a dependent-template build control. Do not use retries to hide the random failure or require a particular random iteration seed.

---
Audit priority: **P2**.

### LINKED ISSUE #7072: Go builds print module-download output again with a cold cache
## Problem

A successful `goreleaser build` prints `go: downloading ...` messages again when the Go module cache is cold. These messages were filtered by #4869.

This is a pre-existing regression, not a behavior change from removing the unused `buildOutput` helper in #7067. The follow-up was requested during that PR's review.

## Reproduction

Create a project with these files.

`go.mod`:

```go
module example.com/download-output

go 1.27.1

require github.com/caarlos0/env/v11 v11.4.1
```

`main.go`:

```go
package main

import _ "github.com/caarlos0/env/v11"

func main() {}
```

`.goreleaser.yaml`:

```yaml
version: 2
project_name: download-output
```

Run:

```sh
go mod tidy
cache="$(mktemp -d)"
GOMODCACHE="$cache" goreleaser build --snapshot --single-target --clean
GOMODCACHE="$cache" goreleaser build --snapshot --single-target --clean
```

The first build succeeds but includes this in its logged build output:

```text
go: downloading github.com/caarlos0/env/v11 v11.4.1
```

The second build, using the same cache, succeeds without the download message.

## Expected behavior

Restore the successful-build output filtering introduced by #4869. Keep compiler diagnostics and build failures visible.

## Cause and follow-up scope

- Commit `9cf3bbbc5` (#4869) introduced `buildOutput` and called it from the Go builder's execution function.
- Commit `516d9bccf` (#5403) removed that caller when execution moved to the shared builder layer.
- The current Go builder calls `base.Exec`, which logs raw `CombinedOutput`. The old filter has had no production caller since that move.

Consider restoring the filtering at the shared execution boundary in `internal/builders/base/build.go`, with focused coverage for empty, download-only, and mixed output. Keep this separate from #7067.

## Environment

Reproduced locally from commit `bd3692b78387d6b7722aedcca18fd8efdc971f1f`, on macOS/arm64 with Go 1.27.1. Both the cold-cache and warm-cache builds used the real Go compiler and GoReleaser CLI.

AI assistance: investigated with GitHub Copilot.

caarlos0:
Fixes a group of related bugs in the Go builder.

- Closes #7072: filters successful `go: downloading` lines from Go build output while preserving failure diagnostics.
- Closes #6900: merges target override `env` deterministically, keeping each variable at the position of its first definition so that base entries referencing it still template.
- Closes #6899: uses the effective target environment and build-selection flags for ellipsis main discovery. Only flags that can change package selection reach `go list`, which rejects the `go test` flags `command: test` makes reachable.
- Closes #6898: renames ellipsis outputs to the registered artifact paths before timestamping. This covers `c-shared` and `c-archive` as well as WASM, and finds the file `go build` actually wrote instead of assuming its name.
- Closes #6897: registers generated C header artifacts after successful c-shared and c-archive builds, with the build mod timestamp, so archives that contain them stay reproducible.

### Behaviour change

With an ellipsis `main` and no explicit `id`, every binary now gets its own artifact ID (its binary name), including when the ellipsis resolves to a single main package. Before, the ID was assigned only when more than one main package was found — but build constraints make that count target dependent, so a package excluded on one target silently changed the IDs of its siblings there.

This is a silent break for a config that references such a build in an `ids` field of `archives`, `nfpms`, `brews`, or `dockers`, so GoReleaser now warns once when it builds an affected configuration, and points at setting `id` explicitly. Documented in the Go builder page.

### Known limitation

`env` entries are templated in order, so an override cannot introduce a new variable that a *redefined* base variable references. Supporting both directions needs ordering by reference, which is out of scope here. Documented in the Go builder page.

### Internal

`execGo` is replaced by `base.Exec` with a `WithLogFilter` option, so the Go builder keeps sharing the helper with the other builders.


internal/builders/golang/build.go
```diff
@@ -581,14 +664,21 @@ func checkBuild(build config.Build, options api.Options) (map[string]string, []*
...
 ) (map[string]string, []*artifact.Artifact, error) {
 	logFindingMains(build, main)
 
 	var binaries []*artifact.Artifact
-	mains, err := gomain.All(dir, main)
+	buildFlags, err := buildSelectionFlags(ctx, details, options, getBinaryArtifact(options.Target.(Target), build, options.Name, options.Path, options.Ext), env)
+	if err != nil {
+		return nil, nil, err
+	}
+	mains, err := gomain.All(dir, env, buildFlags, main)
```
caarlos0, line 716: **Ellipsis discovery is now per-target, so auto-assigned artifact IDs differ between targets.**

This line is the fix for #6899 and it is correct, but it makes `len(mains)` a per-target value. Two unchanged decisions below depend on that count and are now target-dependent:

* line 710 — `if len(mains) > 1 && build.InternalDefaults.ID { a.Extra[ExtraID] = bin }`
* lines 686-692 — the two `'binary' is also set` / `'id' is set` validation errors

Trigger: `main: ./...`, no explicit `id` or `binary`, `cmd/foo` (portable) plus `cmd/bar` guarded by `//go:build !windows`, targets `linux_amd64` and `windows_amd64`.

Verified with `go list -e -f '{{.ImportPath}} name={{.Name}}' ./cmd/...`:

```
GOOS=linux   -> idt/cmd/bar name=main ; idt/cmd/foo name=main
GOOS=windows -> idt/cmd/foo name=main
```

`gomain.All` drops packages whose `pkg.Name != "main"` (gomain.go:100), so `len(mains)` is 2 on linux and 1 on windows. Result:

| target | ExtraID of `foo` |
|---|---|
| linux_amd64 | `foo` |
| windows_amd64 | `<project name>` (`build.ID`, set at internal/pipe/build/build.go:123-125) |

Impact: `archives.ids`, `nfpms.ids`, `brews.ids` and `dockers.ids` that name `foo` silently exclude the Windows build. Before this PR discovery ignored `GOOS`, so the count was identical for every target and the IDs were consistent; the Windows build simply failed loudly at `go build`.

The validation errors have the same problem: the identical config errors for a linux target and passes for a windows-only one.

Smallest fix: make the ID scheme independent of the per-target count, e.g. key it off "`main` is an ellipsis and `InternalDefaults.ID` is set" rather than `len(mains) > 1`. That does change the single-main ellipsis case, so it needs a deliberate call.

Test needed: two `cmd/` packages, one guarded by `//go:build !windows`, build `linux_amd64` and `windows_amd64`, assert `Extra[artifact.ExtraID]` for `foo` is the same on both.

caarlos0 (commented):
Reviewed the full diff against `origin/main` (merge base `9c27633`).

All five issue premises check out, and the two central regression tests genuinely pin their production changes: `TestBuildCgoLibraryFailureDoesNotRegisterStaleHeader` fails on the old code because the header was registered before the compiler ran, and `TestBuildVariadicWasmArtifactsExistWithModTimestamp` fails because `gio.Chtimes` cannot stat the missing `.wasm`.

Six findings inline. The first one is the one I would fix before merging: the per-target discovery introduced by #6899 makes auto-assigned artifact IDs differ between targets, which silently drops binaries from `ids`-filtered archives and packages.

**Cross-cutting notes**

* The two `TestOverrides` env-order subtests are statistical rather than deterministic regression tests — on the old code they fail roughly 96% and 50% of the time respectively, since they are asserting against Go map iteration order. That is inherent to proving an ordering fix and I would not change it, but it is worth knowing before someone concludes that reverting `mergeEnv` keeps the suite green.
* `ensureWASMEllipsisOutputs` is only covered with a single main package. The actual #6898 shape is multi-main, and neither the early-return branch (`mains != nil && ext != ".wasm"`) nor the non-ellipsis wasm path is covered.
* Headers now flow into archives, nfpm packages, makeself and reportsizes. That is the point of #6897, but it adds a file to existing users' archives, so it is worth a release note.
* Env templating for ellipsis builds now resolves `{{ .Binary }}`, `{{ .ArtifactName }}`, `{{ .ArtifactPath }}` and `{{ .ArtifactExt }}` from `options` rather than from the first discovered main. `{{ .Name }}`, `{{ .Path }}` and `{{ .Ext }}` are unchanged, and non-ellipsis builds are byte-identical. This is an improvement — the old value was arbitrary and env must exist before discovery — but it is an undocumented behaviour change with no test.

**Verification gaps**

* I did not build, lint, format or run the test suite; the requester asked me to assume those are fine. The checks below are code traces plus targeted `go build` / `go list` probes with go1.27.1 on darwin/arm64.
* I could not confirm that `-buildmode=c-archive` and `-buildmode=c-shared` with cgo work on the `windows-latest` job in the CI matrix. `TestBuildCgoLibraryRegistersGeneratedHeader` sets `CGO_ENABLED=1` for both modes and has no skip guard, so it fails hard if that runner has no C toolchain.
* `TestBuildVariadicUsesTargetEnvForDiscovery` triggers a cold `plan9_amd64` cross-compile of the standard library on first run. No toolchain download is needed, but I did not measure the wall-clock cost on CI.

_Review by an automated agent (GitHub Copilot CLI); please treat the findings as leads and verify before acting._

internal/builders/golang/build.go
```diff
@@ -274,80 +275,135 @@ var go118FirstClassTargets = []string{
...
+	}
 	for _, a := range allbinaries {
 		if a.Type == artifact.CShared || a.Type == artifact.CArchive {
 			fullPathWithoutExt := strings.TrimSuffix(a.Path, options.Ext)
-			if ha := getHeaderArtifactForLibrary(
-				build,
-				options.Target.(Target),
-				fullPathWithoutExt,
-			); ha != nil {
+			if ha := getHeaderArtifactForLibrary(build, t, fullPathWithoutExt); ha != nil {
```
caarlos0, line 322: **Header artifacts are registered without `mod_timestamp`, so archives that contain them stop being reproducible.**

Moving this loop after `execGo` is the right fix for #6897. But headers are added here at line 320 while `base.ChTimes` is only applied to binaries in the next loop (line 326), so the header keeps its wall-clock mtime.

Trigger: `buildmode: c-archive` (or `c-shared`) on a package with an `//export`ed symbol, `mod_timestamp` set, and an `archives` block.

Chain, all verified by reading:
* internal/pipe/archive/archive.go:122 puts `artifact.Header` into the archive alongside the library.
* pkg/archive/tar/tar.go:73-80 and pkg/archive/zip/zip.go:92-100 take the on-disk mtime; `ParsedMTime` only overrides it when `archives.builds_info.mtime` is explicitly configured, which is empty by default.

So the library entry is deterministic and the `.h` entry is not. This path is new — before this PR no header ever reached the artifact list, so the gap was unreachable.

Smallest fix: fold the header registration into the loop at line 325, or call `base.ChTimes(build, tpl.WithArtifact(ha), ha)` before `ctx.Artifacts.Add(ha)`.

Test needed: add `ModTimestamp` to `TestBuildCgoLibraryRegistersGeneratedHeader` and assert `os.Stat(headers[0].Path).ModTime()`.

internal/builders/golang/build.go
```diff
@@ -274,80 +275,135 @@ var go118FirstClassTargets = []string{
...
 	}
+	return nil
+}
 
-	env := []string{}
-	// used for unit testing only
-	testEnvs := []string{}
-	env = append(env, ctx.Env.Strings()...)
+func ensureWASMEllipsisOutputs(mains map[string]string, binaries []*artifact.Artifact, ext string) error {
+	if mains == nil || ext != ".wasm" {
```
caarlos0: **This covers `.wasm` only; `buildmode: c-shared` has exactly the same bug.**

Verified with go1.27.1 on darwin/arm64, single main package exporting a C symbol:

```
$ go build -buildmode=c-shared -o out/ ./a && ls out/
a    a.h          # no .dylib

$ go build -buildmode=c-archive -o out2/ ./a && ls out2/
a.a  a.h          # .a is added
```

Trigger: `main: ./...` resolving to one main package plus `buildmode: c-shared`. `extFor` (internal/pipe/build/build.go:259-266) returns `.dylib`/`.so`/`.dll`, `checkBuildElipsis` registers `dist/<id>_<target>/a.dylib`, `buildGoBuildLine` passes `-o <dir>`, and Go writes `a`. `ext != ".wasm"` so this returns early.

Impact is the same as #6898: with `mod_timestamp` the build fails at `base.ChTimes` (line 326); without it the artifact is registered pointing at a file that does not exist and a later pipe fails.

This is pre-existing rather than introduced, but it is the same bug class this function exists to fix, and the neighbouring hunk makes `c-shared`/`c-archive` a subject of this PR — so it seems worth closing here rather than leaving half of it open.

Note the `strings.TrimSuffix(a.Path, ext)` strategy does not generalise blindly: on Windows `extFor` returns `.lib` for `c-archive` while Go writes `.a`, so the rename source is not `TrimSuffix(path, ext)` there. Comparing against what Go actually wrote would be more robust than a per-extension list.

Test needed: a subtest shaped like `TestBuildVariadicWasmArtifactsExistWithModTimestamp` with `Buildmode: "c-shared"`, `Main: "./cmd/..."` and `runtimeTarget`, asserting the registered library path exists.

internal/builders/golang/build.go
```diff
@@ -382,6 +438,27 @@ func withOverrides(ctx *context.Context, build config.Build, target Target) (con
 	return build.BuildDetails, nil
 }
 
+func mergeEnv(base, overrides []string) []string {
```
caarlos0: **A key redefined by an override keeps the base's position, which breaks ordering for variables the override introduces.**

Preserving insertion order does fix #6900 for the reported case. But because an existing key keeps its original index and only its value is replaced, an override that adds a new variable and redefines an existing one that depends on it comes out in the wrong order:

```yaml
env: ["CC=gcc"]
build_details_overrides:
  - goos: darwin
    env: ["SDK=/opt/sdk", "CC={{.Env.SDK}}/bin/cc"]
```

`mergeEnv` yields `["CC={{.Env.SDK}}/bin/cc", "SDK=/opt/sdk"]`. `base.TemplateEnv` applies entries sequentially and chains with `tpl.SetEnv` (internal/builders/base/build.go:92-106), so `CC` is templated before `SDK` exists and the build fails with `map has no entry for key "SDK"`.

Before this PR the order was random, so this config failed intermittently; now it fails every time. Determinism is the goal of #6900, so this looks like the wrong half of the ordering to lock in.

Smallest fix: when a key reappears in `overrides`, move it to the override's position instead of updating it in place (drop it from `keys` and re-append). I checked this against the new `TestOverrides/"env order is stable"` expectation `["A=value","B={{.Env.A}}","C=override","D={{.Env.B}}"]` — `C` is already last among the base keys, so that assertion still holds.

Test needed: the `CC`/`SDK` case above, asserting `dets.Env == ["SDK=/opt/sdk", "CC={{.Env.SDK}}/bin/cc"]`.

Minor, same line: the parameter name `base` shadows the imported `internal/builders/base` package (line 20). It compiles because this function does not use the package, but it is a trap for the next edit.

internal/builders/golang/build.go
```diff
@@ -627,6 +717,29 @@ func checkBuildElipsis(
...
 
+func buildSelectionFlags(
+	ctx *context.Context,
+	details config.BuildDetails,
+	options api.Options,
+	a *artifact.Artifact,
+	env []string,
+) ([]string, error) {
+	tpl := tmpl.New(ctx).WithEnvS(env).WithArtifact(a)
+	flags, err := tpl.Slice(details.Flags, tmpl.NonEmpty())
```
caarlos0, line 768: **Build-only flags are now forwarded to `go list`, which rejects them.**

`details.Flags` goes into `packages.Config.BuildFlags` (gomain.go:91), and `go/packages` shells out to `go list`. Most build flags are shared, but not all.

Verified with go1.27.1:

```
$ go list -c ./a
flag provided but not defined: -c
usage: go list [-f format] [-json] [-m] [list flags] [build flags] [packages]
```

(`-cover`, `-trimpath`, `-race`, `-mod`, `-ldflags`, `-gcflags` are all accepted, so this is a narrow set — mainly `-c` and `-o`.)

Trigger: `command: test` with `main: ./...` and `flags: ["-c"]`. That is an explicitly supported configuration — line 486 special-cases a user-supplied `-c` to avoid adding it twice.

The failure is loud but the message is misleading: golang.org/x/tools@v0.49.0/go/packages/golist.go:926-929 maps any `flag provided but not defined` stderr to `goTooOldError`, so the user sees `could not load packages: unsupported version of go: exit status 1: flag provided but not defined: -c`.

Smallest fix: drop flags that cannot affect package selection (`-c`, `-o`) before passing them to `packages.Load`.

Test needed: a `checkBuild` test with `Command: "test"`, `Main: "./cmd/..."`, `Flags: []string{"-c"}`.

Unrelated nit at line 734: tags are templated with `.WithBuildOptions(options)` here, but `buildGoBuildLine` templates them without it (line 505). The values happen to match today because `a` is built from `options.Name`/`Path`/`Ext` and `Target.String()` returns `t.Target`, so this is only a latent divergence — but dropping `.WithBuildOptions(options)` here would keep the two paths identical by construction.

internal/builders/golang/build.go
```diff
@@ -274,80 +275,135 @@ var go118FirstClassTargets = []string{
...
+}
 
-	if len(testEnvs) > 0 {
-		for i := range allbinaries {
-			allbinaries[i].Extra["testEnvs"] = testEnvs
-		}
-	}
-
-	cmd, err := buildGoBuildLine(ctx, build, details, options, allbinaries[0], mains, env)
+func execGo(ctx *context.Context, command, env []string, dir string) error {
```
caarlos0: `execGo` is a line-for-line copy of `base.Exec` (internal/builders/base/build.go:104-118) with one changed line — `buildOutput(out)` instead of `string(out)`.

Seven other builders (zig, rust, bun, deno, node, uv, poetry) call `base.Exec`. Forking it here means the Go builder silently stops inheriting any future change to it.

Suggest an option on the shared helper instead, e.g. `base.Exec(ctx, cmd, env, build.Dir, base.WithOutputFilter(buildOutput))`, which keeps the #7072 fix and drops the copy.

Not a correctness problem — the failure path is preserved correctly (the raw unfiltered output is returned in the error at line 391), which is the part that mattered for #7072.

internal/builders/golang/build.go
caarlos0: Confirmed, and fixed in 91720d327.

Took the deliberate call you asked for: `ExtraID` is now set whenever the main is an ellipsis and the `id` is a default, dropping the `len(mains) > 1` condition. The count is target dependent and cannot be used for anything that must be stable across targets.

This does change the single-main ellipsis case — the ID becomes the binary name instead of the project name. That is the cost of a rule that holds for every target. The loud checks keep the count, as an explicit `id` with several mains is still a config error, and it fails the whole build on the first target that sees it.

Documented in `www/content/customization/builds/builders/go.md`.

Regression test: `TestCheckBuildElipsisIDsAreTargetIndependent` — `cmd/foo` plus `cmd/bar` guarded by `//go:build !windows`, `linux_amd64` and `windows_amd64`, asserts the ID of `foo` is equal on both. Verified it fails on the old condition (`windows_amd64` gave `proj`).

internal/builders/golang/build.go
caarlos0: Correct, fixed in 91720d327.

Folded the header registration into the `ChTimes` loop, and the header now gets `base.ChTimes` before it is added, so it has the same mod timestamp as the library.

Regression test: added `ModTimestamp` to `TestBuildCgoLibraryRegistersGeneratedHeader` and it asserts the header mtime for both `c-archive` and `c-shared`. Verified it fails without the `ChTimes` call.

internal/builders/golang/build.go
caarlos0: Reproduced on go1.27.1 darwin/arm64 and fixed in 91720d327.

```
$ go build -buildmode=c-shared -o out/ ./a && ls out/
a    a.h
```

`ensureWASMEllipsisOutputs` is now `ensureEllipsisOutputs` and applies to every non-empty extension. Took your second point about not generalising `TrimSuffix`: `findBuildOutput` looks in the output directory for the file Go actually wrote — the bare prefix, or `prefix.*` with `.h` excluded — instead of assuming the source name. That covers `foo` for `foo.wasm` and `foo.dylib`, and `foo.a` for `foo.lib` on Windows. It fails with a clear message if the candidate is not unique.

Regression test: `TestBuildEllipsisLibraryArtifactsExist`, subtests for `c-archive` and `c-shared`, with `Main: "./cmd/..."`, `runtimeTarget` and `ModTimestamp`. Verified the `c-shared` subtest fails on the old wasm-only condition.

internal/builders/golang/build.go
caarlos0: Agreed, fixed in 91720d327.

`mergeEnv` now removes a key from the order list before it appends it again, so the last definition sets both the value and the position. Applied it uniformly instead of only to the override side: the rule is one sentence, and a variable redefined in terms of a later one must move in both cases.

`TestOverrides/"env order is stable"` still passes unchanged, as you predicted.

The parameter is now `defaults`, so it no longer shadows the `base` package.

Regression test: `TestOverrides/"redefined env moves to the override position"` with your `CC`/`SDK` case, asserting `["SDK=/opt/sdk", "CC={{.Env.SDK}}/bin/cc"]`.

internal/builders/golang/build.go
caarlos0: Reproduced both flags on go1.27.1 and fixed in 91720d327.

```
$ go list -c ./a
flag provided but not defined: -c
$ go list -o /tmp/xx ./a
flag provided but not defined: -o
```

`dropListIncompatibleFlags` removes `-c` and `-o` before the flags go to `packages.Load`. It also drops the value of `-o` when it is a separate argument. Neither flag can change package selection.

Took the nit too: the tags here are now templated without `.WithBuildOptions(options)`, so the two paths are identical by construction. The `options` parameter is gone, as nothing else used it.

Regression tests: `TestCheckBuildElipsisDropsListIncompatibleFlags` with `Command: "test"`, `Main: "./cmd/..."` and `Flags: ["-c", "-o", "somewhere", "-trimpath"]`, plus a unit test for the filter. Verified the first one fails without the filter.

internal/builders/golang/build.go
caarlos0: Agreed, fixed in 91720d327.

`execGo` is deleted. `base.Exec` takes variadic options and the go builder calls `base.Exec(ctx, cmd, env, build.Dir, base.WithLogFilter(buildOutput))`. Named it `WithLogFilter` rather than `WithOutputFilter`, as it only changes what is logged; the failure path still reports the raw output, which is the behaviour the other seven builders keep by default.

`TestExecGoKeepsFailureOutput` now exercises the shared helper with the filter.

caarlos0 (commented):
Re-reviewed at 91720d3. Supersedes my earlier review (#pullrequestreview-5124211911) — all six findings there are addressed, and the two structural ones came out well: `base.Exec` growing `WithLogFilter` is the right shape and keeps the seven other builders sharing the helper, and folding the header registration into the binaries loop fixes the `mod_timestamp` gap in the smallest possible way.

Four new findings inline, all introduced by this commit or newly decided by it.

The first one is a regression I caused: my suggestion on `mergeEnv` was wrong. "Last definition sets position" fixes the narrow case I raised and breaks the canonical cross-compile config, which is far more common. I would restore the ordering from c48de35e and drop the subtest that pins the new rule.

**Verified this round**

* `mergeEnv` — ran the new implementation against four configurations; results in the inline comment.
* `findBuildOutput` — reproduced the ambiguous glob with a real `GOOS=wasip1` build of `cmd/app` and `cmd/app.v2`.
* `dropListIncompatibleFlags` — probed `go list` with six `go test` flags.
* The ID change — confirmed `main: ./...` shipped in v2.15.0, and traced that an `ids:` filter matching nothing yields zero archives with no error.
* `ensureEllipsisOutputs` — walked the `ext == ""` early return (ordinary unix binaries, unchanged), `.exe`, `.wasm`, `.so`/`.dylib`, `.a` and `.lib`. Apart from the glob ambiguity, the extension handling is right, and skipping the rename when the file is already in place is correct now that `dist` is guaranteed empty.
* `base.Exec` — the default `logFilter` is `string(out)`, so the other builders are byte-identical, and the failure path still returns the raw unfiltered output.
* `tmpl.WithExtraFields` copies rather than mutates, so the added `tpl.WithArtifact(ha)` call cannot leak header fields into the next iteration.

**Verification gaps**

* Still no build, lint, format or test-suite run, as requested. Everything above is code tracing plus targeted `go build` / `go list` probes with go1.27.1 on darwin/arm64.
* `TestBuildEllipsisLibraryArtifactsExist` adds a second cgo test to the `windows-latest` job. I still could not confirm a C toolchain is available there, and neither cgo test has a skip guard.
* `TestBuildEllipsisLibraryArtifactsExist` asserts the library but not the header, so the ellipsis path has no coverage for header registration.

_Review by an automated agent (GitHub Copilot CLI); please treat the findings as leads and verify before acting._

internal/builders/golang/build.go
```diff
@@ -382,6 +455,34 @@ func withOverrides(ctx *context.Context, build config.Build, target Target) (con
...
+	all := append(slices.Clone(defaults), overrides...)
+	values := make(map[string]string, len(all))
+	keys := make([]string, 0, len(all))
+	for _, env := range all {
+		key, value, ok := strings.Cut(env, "=")
+		if !ok || key == "" {
+			continue
+		}
+		// a redefined key drops its old position.
+		keys = slices.DeleteFunc(keys, func(k string) bool { return k == key })
```
caarlos0: **This re-breaks the common override case, and it is more common than the one I asked you to fix. My earlier suggestion was wrong — sorry.**

"Last definition sets both value and position" fixes the case I raised, but it breaks the opposite and much more frequent one: an override that redefines a base variable which *other base variables* depend on.

```yaml
env:
  - SYSROOT=/usr
  - CGO_CFLAGS=-I{{ .Env.SYSROOT }}/include
  - CGO_LDFLAGS=-L{{ .Env.SYSROOT }}/lib
build_details_overrides:
  - goos: darwin
    env:
      - SYSROOT=/opt/osxcross/SDK/MacOSX.sdk/usr
```

`SYSROOT` is redefined, so it is deleted from `keys` and re-appended at the end:

```
[CGO_CFLAGS=-I{{ .Env.SYSROOT }}/include  CGO_LDFLAGS=-L{{ .Env.SYSROOT }}/lib  SYSROOT=/opt/osxcross/...]
```

`base.TemplateEnv` applies entries in order and chains with `tpl.SetEnv` (internal/builders/base/build.go:92-106), and templates use `missingkey=error` (internal/tmpl/tmpl.go:274), so `CGO_CFLAGS` is evaluated before `SYSROOT` exists and the build fails with `map has no entry for key "SYSROOT"` — the exact failure #6900 reports.

I ran the new `mergeEnv` against four configs:

| config | result | ok? |
|---|---|---|
| base `[PREFIX, CC={{.Env.PREFIX}}]`, override `[PREFIX]` | `[CC={{.Env.PREFIX}}, PREFIX]` | ❌ fails |
| base `[CC=gcc]`, override `[SDK, CC={{.Env.SDK}}]` | `[SDK, CC={{.Env.SDK}}]` | ✅ (the case I raised) |
| base `[A=value]`, override `[B={{.Env.A}}]` | `[A, B]` | ✅ |
| the new `env order is stable` test | `[A, B, C=override, D]` | ✅ |

Note the first row **worked** in c48de35e (first-occurrence order) and fails now, so this is a regression inside the PR, not a pre-existing one.

No purely positional rule satisfies both directions:

* first-occurrence (c48de35e) breaks only "override introduces a var that a redefined base var references"
* last-definition (this commit) breaks "override redefines a base var that other base vars reference"

The second is the canonical cross-compile config, so it is the one that must keep working.

Recommendation: restore the first-occurrence ordering from c48de35e, drop the new `redefined env moves to the override position` subtest, and note the limitation in go.md — an override cannot introduce a variable that a *redefined* base variable references. If both must work, it needs dependency-aware ordering (a topological sort over `{{ .Env.X }}` references), which is more machinery than this PR should carry.

internal/builders/golang/build.go
```diff
@@ -350,6 +334,95 @@ func (*Builder) Build(ctx *context.Context, build config.Build, options api.Opti
...
+	}
+	return nil
+}
+
+// findBuildOutput looks for the file `go build` wrote in place of the expected
+// path, e.g. `foo` for `foo.wasm`, or `foo.a` for `foo.lib`.
+func findBuildOutput(expected, ext string) (string, error) {
+	name := filepath.Base(expected)
+	prefix := strings.TrimSuffix(expected, ext)
+	matches, err := filepath.Glob(prefix + ".*")
```
caarlos0: **The glob makes this ambiguous when one binary name is a dotted prefix of a sibling, and it fails a case that worked in the previous commit.**

Generalising past `.wasm` is right, and the Windows `c-archive` case (Go writes `.a`, GoReleaser expects `.lib`) is exactly why the plain `TrimSuffix` was not enough. But `filepath.Glob(prefix + ".*")` also matches *other* binaries of the same ellipsis build.

Reproduced with go1.27.1, packages `cmd/app` and `cmd/app.v2`:

```
$ GOOS=wasip1 GOARCH=wasm go build -o dist/ ./cmd/...
$ ls dist/
app  app.v2
$ ls dist/app.*
dist/app.v2
```

For the `app.wasm` artifact: `prefix` is `dist/app`, the glob returns `dist/app.v2`, `os.Stat(prefix)` also succeeds, so `candidates` has two entries and this returns `could not find the build output for app.wasm`. The build fails. 1847e6aa7's `TrimSuffix` handled this.

Smallest fix: prefer the extensionless path and only fall back to the glob.

```go
prefix := strings.TrimSuffix(expected, ext)
if _, err := os.Stat(prefix); err == nil {
    return prefix, nil
}
// only c-archive on Windows gets here: go writes .a, we want .lib
```

I checked that against every shape this function has to handle:

| case | `prefix` exists? | result |
|---|---|---|
| wasm ellipsis | yes | `prefix` ✓ |
| `c-shared` unix (`.so`/`.dylib`) | yes | `prefix` ✓ |
| `c-archive` Windows (`.lib`) | no | glob → `foo.a`, `.h` filtered ✓ |
| `app` + `app.v2` wasm | yes | `prefix` ✓ (fixed) |

Test needed: an ellipsis wasm build with `cmd/app` and `cmd/app.v2`, asserting both `app.wasm` and `app.v2.wasm` are registered and exist.

internal/builders/golang/build.go
```diff
@@ -617,7 +731,10 @@ func checkBuildElipsis(
 		bins = append(bins, name)
 		pkgs = append(pkgs, mains[bin])
 		a := getBinaryArtifact(t, build, name, path, options.Ext)
-		if len(mains) > 1 && build.InternalDefaults.ID {
+		// the set of main packages is target dependent, as build constraints
+		// may exclude some of them, so the ID cannot depend on how many were
+		// found for this target.
+		if build.InternalDefaults.ID {
```
caarlos0, line 748: Dropping the `len(mains) > 1` condition does make the ID target-independent, which was my point, and the go.md note is good. One compatibility consequence worth deciding on explicitly before merge.

This also changes the **single-main** ellipsis case, which is released behaviour — `main: ./...` landed in v2.15.0 and the current release is v2.18.1. `TestCheckBuildElipsisSingleMain` records the change: the ID goes from `foo` (the build ID / project name) to `a` (the package name).

So this config, working today, breaks on upgrade:

```yaml
project_name: myapp
builds:
  - main: ./cmd/...        # single main in cmd/cli
archives:
  - ids: [myapp]           # was the auto-assigned ID; becomes "cli"
```

It breaks silently rather than loudly: `ids.Validate()` (internal/pipe/archive/archive.go:102) only rejects duplicate archive IDs, it does not check that `ids:` reference existing builds, and `checkArtifacts` returns `nil` for an empty group map (archive.go:157-166). The `ids:` filter matches nothing, `GroupByPlatform` is empty, the loop body never runs, and the release simply ships **no archives**. The same applies to `nfpms.ids`, `brews.ids` and `dockers.ids`.

I think the new behaviour is the right one — consistency across targets and between the 1-main and n-main cases is worth more than the old ID. But it is a silent break of released behaviour, so it deserves more than a docs line: a release-note callout, or a `deprecate` warning when `main` is an ellipsis, resolves to one package, and the auto-assigned ID would change.

internal/builders/golang/build.go
```diff
@@ -627,6 +744,49 @@ func checkBuildElipsis(
...
+}
+
+// dropListIncompatibleFlags removes the flags `go list` rejects, as they only
+// make sense when actually building. They cannot change package selection.
+func dropListIncompatibleFlags(flags []string) []string {
+	result := make([]string, 0, len(flags))
+	for i := 0; i < len(flags); i++ {
+		name, _, hasValue := strings.Cut(flags[i], "=")
+		switch name {
+		case "-c":
```
caarlos0: The deny-list is incomplete for the one command it was written for. `-c` is a `go test` flag, so `command: test` is the case that reaches here — but `go list` rejects several other `go test` flags too.

Verified with go1.27.1:

```
-count=1        flag provided but not defined: -count
-run=X          flag provided but not defined: -run
-timeout=1s     flag provided but not defined: -timeout
-exec=echo      flag provided but not defined: -exec
-bench=.        flag provided but not defined: -bench
-covermode=set  ok
```

So `command: test` + `main: ./...` + `flags: ["-count=1"]` still produces the misleading `could not load packages: unsupported version of go: ... flag provided but not defined: -count`.

An allow-list would be both smaller and closed, since only a few flags can change which packages `go list` returns: `-tags`, `-race`, `-msan`, `-asan` (they set the matching build tags), and `-mod`, `-modfile`, `-overlay` (module resolution). Everything else is code generation and cannot affect selection. That also keeps #6899's `flags: ["-tags=feature"]` case working.

If you prefer to keep the deny-list, adding `-run`, `-count`, `-timeout`, `-bench` and `-exec` covers the realistic `command: test` set — but it will need revisiting whenever `go test` grows a flag.

Test needed: extend `TestDropListIncompatibleFlags` with the flags above.

internal/builders/golang/build.go
caarlos0: The simplification is equivalent — `keys` and `values` are always written together, so the unconditional `DeleteFunc` is a no-op when the key is new. No behaviour change here.

**The ordering problem from my previous review is still open**, though, and this line is where it lives. Re-ran the current implementation at 3ba574f:

```
base [SYSROOT, CGO_CFLAGS={{.Env.SYSROOT}}, CGO_LDFLAGS={{.Env.SYSROOT}}], override [SYSROOT]
  -> [CGO_CFLAGS=-I{{.Env.SYSROOT}}/include  CGO_LDFLAGS=-L{{.Env.SYSROOT}}/lib  SYSROOT=/opt/osxcross]
```

`base.TemplateEnv` evaluates in order with `missingkey=error`, so this fails with `map has no entry for key "SYSROOT"` — an override redefining a base variable that other base variables reference is the canonical cross-compile config, and it worked in c48de35e.

Full reasoning and the four-case comparison table are in https://github.com/goreleaser/goreleaser/pull/7098#discussion_r5124276858 — my recommendation is still to restore the first-occurrence ordering and drop the `redefined env moves to the override position` subtest, since no purely positional rule satisfies both directions and this is the more common one.

internal/builders/golang/build.go
```diff
@@ -350,6 +334,95 @@ func (*Builder) Build(ctx *context.Context, build config.Build, options api.Opti
...
+// path, e.g. `foo` for `foo.wasm`, or `foo.a` for `foo.lib`.
+func findBuildOutput(expected, ext string) (string, error) {
+	name := filepath.Base(expected)
+	prefix := strings.TrimSuffix(expected, ext)
+	matches, err := filepath.Glob(prefix + ".*")
+	if err != nil {
+		return "", fmt.Errorf("find build output for %s: %w", name, err)
+	}
+	// the header is generated alongside c-archive/c-shared libraries.
+	candidates := slices.DeleteFunc(matches, func(m string) bool {
```
caarlos0: The rewrite is equivalent: `filepath.Glob` returns a freshly allocated slice, so mutating it in place is safe, and candidate order only changes the result when `len(candidates) != 1`, which is the error case either way.

**The glob ambiguity from my previous review is still open.** Re-confirmed against this code with go1.27.1 — packages `cmd/app` and `cmd/app.v2`, `GOOS=wasip1 go build -o dist/ ./cmd/...` writes `dist/app` and `dist/app.v2`, and `dist/app.*` matches `dist/app.v2`. For the `app.wasm` artifact both `dist/app.v2` (glob, extension `.v2`, not filtered) and `dist/app` (the `os.Stat(prefix)` branch below) end up in `candidates`, so this returns `could not find the build output for app.wasm` and the build fails. 1847e6aa7 handled it.

The fix from the previous round still applies and gets smaller against this shape — return early when the extensionless path exists, before globbing:

```go
prefix := strings.TrimSuffix(expected, ext)
if _, err := os.Stat(prefix); err == nil {
	return prefix, nil
}
// only c-archive on Windows gets here: go writes .a, we want .lib
```

Details and the case table are in https://github.com/goreleaser/goreleaser/pull/7098#discussion_r5124276859

caarlos0 (commented):
Re-reviewed at 3ba574f. That commit is a pure cleanup and I confirmed it is behaviour-preserving:

* `mergeEnv` — dropping the `if _, exists` guard around `slices.DeleteFunc` is equivalent, since `keys` and `values` are always written together and deleting an absent key is a no-op.
* `findBuildOutput` — filtering `matches` in place is safe (`filepath.Glob` returns a fresh slice) and the reordered candidates only matter when the count is not 1, which errors either way.
* `ensureEllipsisOutputs` — the `if`/`else if` split and the moved `InternalDefaults.ID` comment are cosmetic.

**It does not address any of the four findings from my previous review, and all four still reproduce at this commit.** Two of them are at lines this commit touched, so I re-anchored them there; the other two are unchanged:

* `internal/builders/golang/build.go:737` — the single-main ellipsis artifact ID still changes from the build ID to the package name. `main: ./...` shipped in v2.15.0, and an `ids:` filter that matches nothing produces zero archives with no error, so this breaks silently on upgrade. Still worth a release note or a `deprecate` warning rather than only the go.md line.
* `internal/builders/golang/build.go:777` — the deny-list still misses the `go test` flags `go list` also rejects: `-run`, `-count`, `-timeout`, `-bench`, `-exec`. An allow-list of the flags that can actually change package selection (`-tags`, `-race`, `-msan`, `-asan`, `-mod`, `-modfile`, `-overlay`) would be smaller and closed.

**One verification gap from the previous rounds is now closed.** I was unable to confirm that the two cgo tests would work on the `windows-latest` job. The GitHub Actions Windows runner images ship MinGW-w64 with `gcc` on `PATH`, and Go supports `c-archive` and `c-shared` on `windows/amd64`, so `TestBuildCgoLibraryRegistersGeneratedHeader` and `TestBuildEllipsisLibraryArtifactsExist` should pass there. Worth knowing that these are the first tests in the suite that invoke a C toolchain, and that `testlib.CheckPath(t, "gcc")` is the existing idiom if you want a guard — but I am not reporting it as a finding.

**Remaining gaps**: still no build, lint, format or test-suite run, as requested; everything above is code tracing plus targeted probes with go1.27.1 on darwin/arm64. `TestBuildEllipsisLibraryArtifactsExist` still asserts the library but not the header, so header registration has no coverage on the ellipsis path.

_Review by an automated agent (GitHub Copilot CLI); please treat the findings as leads and verify before acting._

internal/builders/golang/build.go
caarlos0: You are right, and thanks for the correction. Reverted in b4a730e6f.

I reproduced your case against 3ba574f before changing anything:

```
merged order: [CGO_CFLAGS=-I{{ .Env.SYSROOT }}/include CGO_LDFLAGS=-L{{ .Env.SYSROOT }}/lib SYSROOT=/opt/osxcross]
templated: [] err: template: failed to apply "CGO_CFLAGS=-I{{ .Env.SYSROOT }}/include": map has no entry for key "SYSROOT"
```

So `mergeEnv` is back to first-occurrence ordering, and the `redefined env moves to the override position` subtest is gone. I agree the two directions cannot both work without ordering by reference, and that a topological sort over `{{ .Env.X }}` is more machinery than this PR should carry.

Documented the limitation in `go.md`: an override cannot introduce a variable that a redefined base variable references; define it in the base `env` and let the override replace its value.

Replacement test: `TestOverrides/"redefined env keeps the base position"` uses your `SYSROOT`/`CGO_CFLAGS` config, asserts the order, and then runs `base.TemplateEnv` over the result to prove it templates. Verified it fails on the last-definition rule.

caarlos0: Same thread as https://github.com/goreleaser/goreleaser/pull/7098#discussion_r5124276858 — fixed in b4a730e6f by restoring the first-occurrence ordering. Reproduced your `SYSROOT` failure first, dropped the `redefined env moves to the override position` subtest, and documented the remaining limitation in `go.md`. Details in the other thread.

internal/builders/golang/build.go
caarlos0: Confirmed and fixed in b4a730e6f, with your suggested shape.

`findBuildOutput` now returns the extensionless path when it exists, and only globs after that:

```go
prefix := strings.TrimSuffix(expected, ext)
if _, err := os.Stat(prefix); err == nil {
	return prefix, nil
}
// only c-archive on Windows gets here: go writes `.a`, we want `.lib`.
```

Regression test: `TestBuildVariadicWasmDottedSiblingNames` builds `cmd/app` and `cmd/app.v2` for `js_wasm` and asserts both `app.wasm` and `app.v2.wasm` are registered and exist on disk. Verified it fails on the previous shape with `could not find the build output for app.wasm`.

One residual case I decided not to chase: Windows `c-archive` with the same dotted sibling names still globs, so `app.lib` sees both `app.a` and `app.v2.a`. It fails loudly with the same message rather than producing a wrong artifact, and narrowing it further needs the sibling list passed down. Happy to open a follow-up issue if you want it closed.

internal/builders/golang/build.go
caarlos0: Same thread as https://github.com/goreleaser/goreleaser/pull/7098#discussion_r5124276859 — fixed in b4a730e6f with the early return you suggested, plus `TestBuildVariadicWasmDottedSiblingNames` covering `cmd/app` and `cmd/app.v2`. Details in the other thread.

internal/builders/golang/build.go
caarlos0: Agreed, took the allow-list. Fixed in b4a730e6f.

Confirmed your findings on go1.27.1 and checked the rest of the set:

```
REJECT: -c -count=1 -run=X -timeout=1s -bench=. -exec=echo -o=/tmp/zz
accept: -tags=x -race -msan -asan -mod=mod -modfile=go.mod -overlay=...
accept: -trimpath -ldflags=-s -gcflags=-N -cover -covermode=set -buildvcs=false -pgo=off
```

`keepListFlags` now keeps only `-tags`, `-race`, `-msan`, `-asan`, `-mod`, `-modfile` and `-overlay` — the first four set a build tag, the rest change module resolution. The allow-list is closed, so it does not need revisiting when `go test` grows a flag.

One thing the deny-list got for free and the allow-list has to handle: a value passed as a separate argument. `flags: ["-o", "somewhere"]` must drop `somewhere` too, otherwise `go list` reads it as a package pattern. A bare token now shares the fate of the flag before it, which also keeps `["-tags", "foo"]` intact.

Tests: `TestKeepListFlags` covers the kept set, the rejected set including all the flags you listed, and the separate-argument case. `TestCheckBuildElipsisDropsListIncompatibleFlags` now passes `["-c", "-count=1", "-run=X", "-timeout=1s", "-bench=.", "-exec", "echo", "-o", "somewhere", "-trimpath"]` through the real `checkBuild`. Verified it fails without the filter.

internal/builders/golang/build.go
caarlos0: Agreed on both points, and added the warning in b4a730e6f.

There is no release-notes file in the repo to add a callout to — release notes come from the commit log — so a docs line alone would not reach anyone upgrading. GoReleaser now warns once per build when the ellipsis resolves to a single main package and the auto-assigned ID actually changes:

> the artifact ID of this build is now `cli` instead of `myapp`, because `main` is an ellipsis path and `id` is not set: set `id` if you reference it in an `ids` field

The condition is `InternalDefaults.ID && len(mains) == 1 && bin != build.ID`, so it only fires for the affected set. Multi-main builds, where the ID was already the binary name, stay quiet. Setting `id` explicitly is the migration path and silences it, which is why I used a plain `log.Warn` rather than `deprecate.NoticeCustom` — nothing here is deprecated, and `deprecate` also sets `ctx.Deprecated`, which would be wrong.

Also added a `Changed in v2.19` callout to the ellipsis section of `go.md`, naming the four `ids` fields.

I did not touch the silent-empty-`ids` behaviour in `archive.go`. You are right that it is the reason this breaks quietly, but it spans archives, nfpms, brews and dockers, and is not caused by this PR — that reads like a separate issue. Say the word and I will open one.

internal/builders/golang/build.go
```diff
@@ -274,74 +274,58 @@ var go118FirstClassTargets = []string{
...
 		return err
 	}
 
 	for _, a := range allbinaries {
 		if err := base.ChTimes(build, tpl.WithArtifact(a), a); err != nil {
 			return err
 		}
+		if a.Type == artifact.CShared || a.Type == artifact.CArchive {
+			fullPathWithoutExt := strings.TrimSuffix(a.Path, options.Ext)
+			if ha := getHeaderArtifactForLibrary(build, t, fullPathWithoutExt); ha != nil {
```
caarlos0, line 322: **The generated header keeps the build ID while its library takes the binary name, so `ids:` filters drop the header.**

`getHeaderArtifactForLibrary` sets `artifact.ExtraID: build.ID` (build.go:596), but the library artifact now gets `a.Extra[ExtraID] = bin` unconditionally in `checkBuildElipsis`. Before this pull request a single-main ellipsis kept `build.ID` for the library too, so the two always agreed. They no longer do.

Trigger — the configuration the new docs callout tells affected users to write:

```yaml
project_name: myproject
builds:
  - main: ./cmd/...      # one main package, cmd/mylib
    buildmode: c-archive
    env: [CGO_ENABLED=1]
archives:
  - ids: [mylib]
```

Confirmed against the real `Build()`:

```
artifact type=C Header          name=mylib.h id=myproject
artifact type=C Archive Library name=mylib.a id=mylib
```

Impact: `artifact.ByIDs` is applied to `artifact.Header` in internal/pipe/archive/archive.go:122, so the archive ships `mylib.a` without `mylib.h`, which no C consumer can use. `nfpms` filters the same way, and the mismatched group sizes can also trip `ErrArchiveDifferentBinaryCount`.

Smallest fix: give the header the ID of the library it belongs to, here at the registration site:

```go
if ha := getHeaderArtifactForLibrary(build, t, fullPathWithoutExt); ha != nil {
    ha.Extra[artifact.ExtraID] = a.Extra[artifact.ExtraID]
```

Test: `TestBuildEllipsisLibraryArtifactsExist` masks this, because it uses `ID: "cexport"` with package `cmd/cexport`, so the two IDs coincide. Give the build an ID that differs from the package name and assert the header ID equals the library ID.

internal/builders/golang/build.go
```diff
@@ -350,6 +334,69 @@ func (*Builder) Build(ctx *context.Context, build config.Build, options api.Opti
...
+// With more than one main package, `-o` has to be a directory, and go names
+// each output after its package. That matches the name GoReleaser registered
+// on every target except wasm, where go writes no extension and GoReleaser
+// uses `.wasm`.
+//
+// Builds that resolve to a single main package get an exact `-o` and never
+// need this, which is why the buildmodes that change the extension are not
+// handled here: go requires exactly one main package for `c-archive` and
+// `c-shared`.
+func ensureEllipsisOutputs(binaries []*artifact.Artifact, ext string) error {
```
caarlos0: **Deciding with `os.Stat` lets one binary's output satisfy another's check, which silently orphans a binary.**

The loop skips an artifact when its registered path already exists, but with several main packages that file may be a *different* package's output.

Trigger: `main: ./cmd/...` on `js_wasm` with sibling packages `cmd/app` and `cmd/app.wasm`. Registered names are `app.wasm` and `app.wasm.wasm`; go writes `app` and `app.wasm`.

- `app` → `os.Stat("dist/js_wasm/app.wasm")` succeeds, because that is `cmd/app.wasm`'s output, so it is skipped.
- `app.wasm` → renames `app.wasm` to `app.wasm.wasm`.

Reproduced with the real `Build()`, which returns no error:

```
ON DISK: app
ON DISK: app.wasm.wasm
ARTIFACT dist/js_wasm/app.wasm  exists=false
ARTIFACT dist/js_wasm/app.wasm.wasm exists=true
```

That is #6898 again in a new shape, and it is silent unless `mod_timestamp` is set, which is the only thing that would make `base.ChTimes` fail.

Note the sources and destinations can chain (`app` → `app.wasm` → `app.wasm.wasm`), so renaming in place in order also clobbers. The smallest correct fix is to stop inferring from `os.Stat` — go's output name is known — and to move every output aside before moving any into place, so no rename consumes another artifact's source.

Test: `TestBuildVariadicWasmDottedSiblingNames` uses `cmd/app.v2`, which cannot collide; only a suffix equal to `ext` does. Change it to `cmd/app.wasm`, or add that case.

internal/builders/golang/build.go
```diff
@@ -350,6 +334,69 @@ func (*Builder) Build(ctx *context.Context, build config.Build, options api.Opti
...
+	if ext != ".wasm" {
```
caarlos0: **The `.wasm`-only guard, and the comment above it, are wrong for `command: test`.**

`go test -c` does not name outputs the way `go build` does — it appends `.test` (`testBinaryName` in `cmd/go/internal/test/test.go`), and `-c` is always added for `command: test` (build.go:486). Verified:

```
$ go test -c -o out/ ./cmd/a ./cmd/b
$ ls out
a.test  b.test
$ GOOS=windows go test -c -o w/ ./cmd/...
a.test.exe  b.test.exe
```

Trigger:

```yaml
builds:
  - command: test
    main: ./...     # two or more main packages
```

GoReleaser registers `dist/<id>_<target>/a`, go writes `a.test`, and this returns early because `ext` is not `.wasm`. The artifact points at a missing file and a later stage fails with `no such file or directory`.

The bug predates this pull request — the several-mains branch is unchanged — but the doc comment now asserts it cannot happen ("matches the name GoReleaser registered on every target except wasm"), so it must not stay as written. Either handle the suffix or reject the configuration.

Separately, this function is called for every build (build.go:312), including `mains == nil` and `no_main_check`. Since only several-mains builds can reach go's own naming, it should be guarded on that instead of inferring from the extension; today a `command: test` wasm build can reach the rename and report `rename ...: no such file or directory` rather than the real problem.

Test: an end-to-end build with two main packages and `command: test`, asserting each registered path exists.

internal/builders/golang/build.go
```diff
@@ -627,6 +731,64 @@ func checkBuildElipsis(
...
+
+// keepListFlags filters build flags down to the ones `go list` understands and
+// that can change which packages it returns.
+func keepListFlags(flags []string) []string {
+	var result []string
+	keep := false
+	for _, flag := range flags {
+		if strings.HasPrefix(flag, "-") {
+			name, _, _ := strings.Cut(flag, "=")
+			keep = slices.Contains(listFlags, name)
```
caarlos0: **The allow-list misses the double-dash spelling, so tag-gated main packages are silently skipped.**

Go accepts both spellings, for `go build` and for `go list`:

```
$ go list --tags=feature ./cmd/...   # honoured
$ go build --tags=feature            # honoured
```

`keepListFlags` compares the exact token, so `--tags` is not in `listFlags` and is dropped. Confirmed through the real `checkBuild` with `flags: ["--tags=feature"]` and a main package behind `//go:build feature`:

```
--tags  -> map[always:./cmd/always]
-tags   -> map[always:./cmd/always gated:./cmd/gated]
```

Impact: the tag reaches `go build` but not discovery, so the gated binary is never built and never appears in the release. No error. This is the exact build-context divergence #6899 set out to remove.

Smallest fix: normalise the leading dashes before the lookup, e.g. compare `"-" + strings.TrimLeft(name, "-")`.

Test: add `--tags=foo` and `--mod=vendor` to `TestKeepListFlags/keeps selection flags`.

caarlos0 (commented):
**Verdict: changes needed.** Three defects, one of which loses an artifact silently and is introduced by this pull request. Reviewed `origin/main...f73b72e5` (merge base `34619ef1`). Findings are inline.

Cross-cutting notes with no single site:

**The ID warning reports a change that did not happen.** `logIDChange` fires when `len(mains) == 1`, but that count is target dependent — which is the very reason the ID no longer uses it. With `cmd/foo` portable and `cmd/bar` behind `//go:build !windows`, the `windows_amd64` target sees one main and warns that the ID "changed from `proj` to `foo`", while the ID of `foo` was already `foo` before this pull request. Wording that states the current ID instead of claiming a change is true in every case it fires.

**A regression test stopped guarding its bug.** `TestBuildVariadicWasmArtifactsExistWithModTimestamp` uses a single main, so after `a8a3483c` it takes the exact `-o` path and no longer covers #6898. Verified by mutation: disabling `ensureEllipsisOutputs` leaves it green. `TestBuildVariadicWasmDottedSiblingNames` is now the only guard.

**Verification gaps, disclosed:**
- The `c-archive` half of `TestBuildEllipsisLibraryArtifactsExist` cannot exercise the extension mismatch on macOS or Linux, because go and `extFor` agree on `.a` there. Only Windows (`.lib` vs `.a`) reaches it.
- CI last ran green on `d1c92e87`, two commits before `HEAD`. `a8a3483c` changed the `-o` selection and has not completed a full run.
- I did not measure the cost of `gomain.All` now running `packages.Load` per target with a distinct `GOOS`. `mode` in `gomain.go` requests `NeedTypes|NeedTypesInfo|NeedDeps` while `All` reads only `Name`, `Syntax`, `Dir` and `Module`, so a large target matrix may pay for type checking it does not use.

Reviewed by an automated agent (GitHub Copilot CLI); findings were reproduced against the code before reporting.

caarlos0:
Closing in favour of four focused pull requests — this had grown to cover five issues in one diff, which made it hard to review.

Splitting it also meant re-testing each premise on its own, and two of the five did not survive:

| Issue | Outcome |
|---|---|
| #6900 | goreleaser/goreleaser#7112 — env order under overrides |
| #6897 | goreleaser/goreleaser#7113 — C header registered after compiling |
| #6898 | goreleaser/goreleaser#7114 — ellipsis builds register paths that exist |
| #6899 | goreleaser/goreleaser#7115 — documented instead of fixed |
| #7072 | dropped, see below |

Two conclusions from this branch were wrong and are not carried over:

- **The artifact ID change is gone.** I had concluded that fixing ellipsis discovery forced auto-assigned artifact IDs to change, which broke released behaviour for `main: ./...` resolving to a single main package. That conflated the tags half of #6899 with the GOOS half; only the second creates target dependence, and #6899 is not being fixed in code at all now.
- **#6897 was understated.** It is not limited to clean builds. GoReleaser refuses a non-empty dist, so the header could never exist when discovery ran — every `c-archive` and `c-shared` build shipped a library without its header.

#7072 is dropped as a deliberate call: the `go: downloading` lines are the only explanation a user gets for a build that stalls on a slow link, so suppressing them turns an explained wait into an unexplained one.

PR #7103, opened 2026-09-06
fix(cmd): auto-snapshot ordering, schema error reporting and --output selection

### LINKED ISSUE #6949: build: --output accepts multiple selected builds and copies only the first binary
### Reproduction

Create a Go module with two main packages, `./a` and `./b`, which print `A` and `B`, and use:

```yaml
version: 2
project_name: audit
builds:
  - id: a
    main: ./a
    binary: a
  - id: b
    main: ./b
    binary: b
```

Run:

```sh
goreleaser build --snapshot --single-target --id a --id b -p2 -o picked
./picked
```

### Observed and expected

The real CLI exited 0, built both programs, logged `copying binary to "picked"`, and the copied program printed `A`. A separate call through the production output pipe, with the same two real binaries inserted in reverse order, copied `B` instead. Alternation between identical CLI runs was not established.

The [flag help](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/cmd/build.go#L99) limits output copying to a single selected build. Ambiguous selection should not silently choose one program.

### Cause and correction

The [pipeline gate](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/cmd/build.go#L148-L150) accepts any nonempty ID list, and the [copy pipe](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/cmd/build.go#L231-L241) selects `bins[0]`. Reject ambiguous multiple-build output. Add a regression with two distinguishable binaries and both insertion orders, retaining a successful single-build control.

---
Audit priority: **P2**.

### LINKED ISSUE #6952: jsonschema: command errors are silent while the schema alias reports them
### Reproduction

Run each command separately, capturing stdout, stderr, and exit status:

```sh
goreleaser jsonschema --not-a-flag
goreleaser schema --not-a-flag
```

### Observed and expected

The canonical `jsonschema` command exited 1 with both stdout and stderr empty. The `schema` alias also exited 1, but stderr correctly contained `unknown flag: --not-a-flag`.

Successful controls for both names exited 0 and produced valid JSON on stdout. The canonical command had empty stderr; the alias printed the usual thank-you message. Thus this is an error-reporting difference, not a failure to generate the schema.

Both names should preserve actionable error messages while keeping generated JSON separate from diagnostics. A caller should not need to know the alternate command name to discover why an invocation failed.

### Cause and correction

[Command setup](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/cmd/root.go#L39-L40) sets the logger to `FatalLevel` for names selected by [shouldDisableLogs](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/cmd/root.go#L151-L158). The [error handler](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/cmd/root.go#L79-L82) then logs at `ErrorLevel`, which is suppressed.

Suppress informational messages without disabling error output. Add subprocess regressions for both command names, covering invalid flags and successful JSON generation.

---
Audit priority: **P2**.

### LINKED ISSUE #6953: build and release: --auto-snapshot rejects Pro configuration before detecting a dirty tree
### Reproduction

In a dirty Git repository containing a buildable Go program, use this configuration on macOS/arm64:

```yaml
version: 2
pro: true
includes: []
project_name: audit
builds:
  - id: audit
    binary: audit
    goos: [darwin]
    goarch: [arm64]
archives:
  - formats: [binary]
changelog:
  disable: true
```

Compare:

```sh
goreleaser build --snapshot -f pro.yaml --clean -p1 --single-target
goreleaser build --auto-snapshot -f pro.yaml --clean -p1 --single-target
```

### Observed and expected

Explicit `--snapshot` exited 0 after warning that Pro-only fields would be ignored. `--auto-snapshot` exited 1 with `field includes not found in type config.Project` and the Pro/OSS configuration error, without reaching `implying --snapshot`.

The same comparison was confirmed for `release` with publishing and announcements skipped. Controls showed that dirty OSS configuration works with `--auto-snapshot`, while clean Pro configuration remains rejected. That clean-run strictness must remain intact.

### Cause and correction

Both [build](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/cmd/build.go#L119-L130) and [release](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/cmd/release.go#L103-L115) enforce strict loading before their later dirty-tree check sets snapshot mode.

Determine the effective snapshot state before deciding whether to reject the Pro configuration. Add regressions for explicit snapshot, dirty automatic snapshot, clean automatic snapshot, and ordinary OSS configuration in both commands.

---
Audit priority: **P2**.

caarlos0:
Fixes a group of related bugs in the build, release and schema commands.

- Closes #6953: load dirty --auto-snapshot build and release runs as snapshots before strict Pro config validation.
- Closes #6952: keep jsonschema command errors visible while still keeping generated JSON on stdout.
- Closes #6949: reject ambiguous build --output selections instead of copying the first binary.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -37,7 +37,7 @@ func (cmd *rootCmd) Execute(args []string) {
 	}
 
 	if shouldDisableLogs(args) {
-		log.SetLevel(log.FatalLevel)
+		log.SetLevel(log.ErrorLevel)
 	}
 
 	if err := fang.Execute(
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -37,7 +37,7 @@ func (cmd *rootCmd) Execute(args []string) {
 	}
 
 	if shouldDisableLogs(args) {
-		log.SetLevel(log.ErrorLevel)
+		log.SetLevel(log.FatalLevel)
 	}
 
 	if err := fang.Execute(
@@ -57,6 +57,10 @@ func (cmd *rootCmd) Execute(args []string) {
 }
 
 func errorHandler(_ io.Writer, _ fang.Styles, err error) {
+	// commands that disable logs still need to report their errors.
+	if logger, ok := log.Log.(*log.Logger); ok && logger.Level > log.ErrorLevel {
+		logger.Level = log.ErrorLevel
+	}
 	log := log.WithError(err)
 	var message string
 	if de, ok := errors.AsType[gerrors.ErrDetailed](err); ok {
--- a/cmd/schema_test.go
+++ b/cmd/schema_test.go
@@ -1,9 +1,14 @@
 package cmd
 
 import (
+	"bytes"
 	"encoding/json"
+	"errors"
 	"os"
+	"os/exec"
 	"path"
+	"path/filepath"
+	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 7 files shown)


cmd/schema_test.go
```diff
@@ -26,3 +31,59 @@ func TestGenerateSchema(t *testing.T) {
...
+			require.NoError(t, err)
+
+			schema := map[string]any{}
+			require.NoError(t, json.Unmarshal([]byte(stdout), &schema))
+			require.Equal(t, "https://json-schema.org/draft/2020-12/schema", schema["$schema"].(string))
+		})
+	}
+}
+
+func buildGoReleaserBinary(tb testing.TB) string {
```
caarlos0: this is dumb

cmd/root.go
```diff
@@ -37,7 +37,7 @@ func (cmd *rootCmd) Execute(args []string) {
 	}
 
 	if shouldDisableLogs(args) {
-		log.SetLevel(log.FatalLevel)
+		log.SetLevel(log.ErrorLevel)
```
caarlos0: this affects all commands, we should override only when its actually needed.

cmd/schema_test.go
caarlos0: Agreed. Removed in 5b8f2a425: the test no longer builds and executes a binary, it runs the command in process with `newRootCmd` and captures stdout/stderr.

cmd/root.go
caarlos0: Fixed in 83a7f851b. `shouldDisableLogs` sets `FatalLevel` again, so those commands stay as quiet as before. The level is lowered to `ErrorLevel` only inside `errorHandler`, i.e. only when there is an error to report, and only if the level is currently above it (so `--verbose` is not affected). `TestSchemaCommandErrors` fails without that override.

PR #7086, opened 2026-09-06
fix(brew,cask): formula escaping, token matching and binary stanzas

caarlos0:
Fixes a group of related bugs in Homebrew and Cask pipes.

- Closes #6988: escapes formula and cask descriptions after template rendering so quoted text remains valid Ruby.
- Closes #6987: writes generated cask files using the normalized token that appears in the cask header.
- Closes #6986: scopes cask binary stanzas to archive or raw-binary package selections in mixed releases.
- Closes #6985: relocates binaries from wrapped universal macOS archives before declaring them.
- Closes #6984: relocates wrapped manpage and shell completion sources along with binaries.

PR #7107, opened 2026-09-06
fix: address regressions since v2.18.1

caarlos0:
## Summary

Fix seven regressions found while reviewing the changes after `v2.18.1`:

- Preserve target and artifact template fields during Rust preparation, sharing build-option resolution with the build pipeline.
- Restore Docker v2 build-argument template context and empty-image skipping while retaining effective `FROM` overrides.
- Render Homebrew descriptions only once, preserving literal template delimiters.
- Install universal macOS binaries when a cask mixes archives and raw binaries.
- Accept literal executable paths containing apostrophes or spaces in healthcheck.
- Preserve directory-symlink traversal when a copy source ends with `/`.
- Restore normal-output redaction throughput without losing cross-write secret handling.

Also simplify Docker flag construction and close handling, replace blob regression polling with deterministic synchronization, and include `go mod tidy`.

The Windows `rustup` fixture now records environment values literally, so pipe characters in target-context fixtures are not interpreted as batch commands.

## Motivation

Recent fixes introduced compatibility regressions at template, path, and packaging boundaries. The streaming redactor also started checking every secret against every ordinary output byte.

Each fix is isolated in its own commit, with focused regression coverage. For normal-output redaction, benchstat reported these medians for 33.6 KB chunks with 64 secrets (`Go 1.27.1`, `GOMAXPROCS=2`, 10 samples):

| `v2.18.1` | Reviewed HEAD (`6ceb354cd`) | This branch |
|---:|---:|---:|
| 47.4 us | 5,331 us | 49.9 us |

Focused local regression and race runs and targeted golangci-lint runs passed. Blob unit coverage was exercised separately from its MinIO bootstrap locally. The full Ubuntu and Windows CI test jobs passed at `7ebed533b` in [build run 34032079529](https://github.com/goreleaser/goreleaser/actions/runs/34032079529); the other workflows also passed.

## References

- [Reviewed commit range](https://github.com/goreleaser/goreleaser/compare/v2.18.1...6ceb354cd7e212d7ca65592003564abe7ab65bad)
- Follow-ups to #7076, #7080, #7086, #7089, #7095, #7101, and #7104.

AI-assisted review and implementation with GitHub Copilot.

### DIFF
--- a/cmd/healthcheck_test.go
+++ b/cmd/healthcheck_test.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -68,7 +69,7 @@ func TestCheckPathChecksEachToolOnce(t *testing.T) {
 
 func TestCheckPathLiteralExecutable(t *testing.T) {
 	for _, name := range []string{"signer's-tool", "signer's tool"} {
-		if os.PathSeparator == '\\' {
+		if runtime.GOOS == "windows" {
 			name += ".exe"
 		}
 		path := filepath.Join(t.TempDir(), name)
--- a/cmd/healthcheck_test.go
+++ b/cmd/healthcheck_test.go
@@ -1,8 +1,10 @@
 package cmd
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -62,4 +64,19 @@ func TestCheckPathChecksEachToolOnce(t *testing.T) {
 	require.NoError(t, checkPath(t.Context(), checked, "some invalid command"))
 	// a cache of its own sees the failure again.
 	require.Error(t, checkPath(t.Context(), map[string]bool{}, "some invalid command"))
+}
+
+func TestCheckPathLiteralExecutable(t *testing.T) {
+	for _, name := range []string{"signer's-tool", "signer's tool"} {
+		if os.PathSeparator == '\\' {
+			name += ".exe"
+		}
+		path := filepath.Join(t.TempDir(), name)
+		require.NoError(t, os.WriteFile(path, nil, 0o755))
+		for _, tool := range []string{path, fmt.Sprintf("%q", path)} {
+			t.Run(tool, func(t *testing.T) {
+				require.NoError(t, checkPath(t.Context(), map[string]bool{}, tool))
+			})
+		}
+	}
 }
--- a/cmd/healthcheck.go
+++ b/cmd/healthcheck.go
@@ -121,14 +121,17 @@ func checkPath(ctx stdctx.Context, checked map[string]bool, tool string) error {
 		return nil
 	}
 	checked[tool] = true
-	// Parse gives no arguments back when it fails, and also for a line that is
-	// only shell syntax, such as `|`.
-	args, err := shellwords.Parse(tool)
-	if len(args) == 0 {
-		return warn(tool, "is not a valid command", cmp.Or(err, exec.ErrNotFound))
-	}
-	if _, err := exec.LookPath(args[0]); err != nil {
-		return warn(tool, "not present in path", err)
+	args := []string{tool}
+	if _, err := exec.LookPath(tool); err != nil {
+		// Dependencies can be either literal executable paths or commands.
+		// Parse also returns no arguments for bare shell syntax such as `|`.
+		args, err = shellwords.Parse(tool)
+		if len(args) == 0 {
+			return warn(tool, "is not a valid command", cmp.Or(err, exec.ErrNotFound))
+		}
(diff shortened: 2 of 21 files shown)


cmd/healthcheck_test.go
```diff
@@ -63,3 +65,18 @@ func TestCheckPathChecksEachToolOnce(t *testing.T) {
 	// a cache of its own sees the failure again.
 	require.Error(t, checkPath(t.Context(), map[string]bool{}, "some invalid command"))
 }
+
+func TestCheckPathLiteralExecutable(t *testing.T) {
+	for _, name := range []string{"signer's-tool", "signer's tool"} {
+		if os.PathSeparator == '\\' {
```
caarlos0: runtime.Goos == "windows"

caarlos0: Updated the fixture to use `runtime.GOOS == "windows"` in 8419da4728131004b0c48dc779c6bce20df1c41e. The focused literal-executable path regression passes with the race detector.

caarlos0 (commented):
## Review: no findings — approve

I reviewed `34619ef16..7ebed533b` (21 files) — the seven regression fixes, the Docker flag simplification, the blob `synctest` conversion, and `go mod tidy`. I found no correctness, scope, compatibility, or determinism problem that a user can trigger. No inline comments, because there is nothing to anchor.

This is posted as a comment, not a formal approval, only because the reviewing token is the pull request author and GitHub refuses self-approval. The verdict is approve.

### What I verified independently

- **`redact`**: the `hasIncompleteSuffix` fast path is output-identical, not merely faster. The old loop can only bail where `len(s)-i < maxLen`, so the scan window `[max(0, len(s)-maxLen+1), len(s))` is exact, and it is a superset of the positions the loop visits, so it can only over-trigger the slow path. The empty suffix is never tested, which keeps `maxLen == 1` correct. On the fast path `strings.Replacer` matches `match()` because `redact()` sorts secrets longest-first and trie priority follows pair order.
- **`brew`/`cask` `desc`**: I ran the `rubyString` → `printf "{{ %q }}"` → second-pass pipeline against hostile descriptions (`{{example}}`, `}} weird {{`, backslashes, embedded quotes, `%`, `#{}`, emoji, and a zero-width space). All twelve round-trip byte-exactly. `lexQuote` consumes the whole Go string constant, so embedded delimiters never re-lex. Zero golden-file churn confirms existing output is unchanged.
- **`cask` mixed packages**: `CaskBins` is set only on archive packages and `Binary` only on raw binaries, so the two template branches are mutually exclusive and cannot emit a duplicate stanza. The new block also makes `macos_packages.rb` symmetric with `linux_packages.rb`, which already had it.
- **`gio`**: `filepath.Clean(src)` came in with #7089, so this is a true regression revert, not a behavior change. `filepath.Rel` cleans both operands, so no target path moves. `Copy("", dst)` now errors instead of walking the working directory, which is strictly better.
- **`docker/v2`**: narrowing the FROM overrides to ARGs actually referenced by a `FROM` is equivalent, because `parseBaseImage` only consults args through `substituteArgs`. Dropping empty values matches the old `tplMapEntries`. `errors.AsType[tmpl.Error]` works because `tmpl.Error` is a value type returned unwrapped, and the deferred check sits before the tags check, so nothing is swallowed for a config that has images.
- **`builders/rust`**: `WithDefaults` already rejects targets outside `all_targets.txt`, so `Parse` cannot newly fail in `Prepare`; `t.clean()` is byte-identical to the removed inline `stripGlibcVersion`; and `Prepare`'s template chain now matches `Build`'s exactly. `base.OptionsForTarget` is a faithful extraction, so it can only fail where the build pipe would fail moments later.
- **`healthcheck`**: `exec.LookPath` searches only `PATH` for a string with no separator, so multi-word dependencies such as `go mod tidy` and `sh -c ...` still fall through to `shellwords.Parse` unchanged. A literal path is no longer split and, with `len(args) == 1`, is no longer executed.
- **`go mod tidy`**: correct — `internal/pipe/gomod/gomod_proxy.go` imports `golang.org/x/mod/modfile` directly.
- **Tests**: each of the seven fixes has a regression test that fails without its fix. The new tests use no sleeps, and never combine `t.Setenv` with `t.Parallel`.

### Verification gaps

- As instructed for this pass, I did not run the build, linters, formatters, or the test suite locally, and I took the redaction benchstat medians as given.
- I could not execute the Windows paths. Two test fixtures are Windows-only: the `%q`-quoted path in `TestCheckPathLiteralExecutable` reaches `go-shellwords`, whose `isEscapeRune` returns `false` on Windows, so the parsed path keeps doubled backslashes and depends on Win32 collapsing repeated separators; and the new `setlocal EnableDelayedExpansion` batch fixture in `createFakeRustup`. Both look correct by inspection, but Windows CI is the decisive check.
- The full integration suite was not run, consistent with the description.

Two earlier review passes on this PR (5125212386, 5125239252) are superseded by this one. Their only request, `runtime.GOOS`, is already addressed in 8419da4.

_Reviewed by GitHub Copilot CLI — this review is generated by a bot, so please weigh it accordingly._

PR #7109, opened 2026-09-06
test: reduce pipeline test overhead

caarlos0:
## Goal and current result

Reduce Linux and Windows CI test overhead while preserving correctness and readable fixtures. The original [source-backed audit](https://github.com/goreleaser/goreleaser/pull/7109#issuecomment-5562578387) covered 255 platform/test entries above one second, spanning 96 root tests in 34 packages.

Current validated revision: `20c59707c`. [CI run 34071824184](https://github.com/goreleaser/goreleaser/actions/runs/34071824184) passed. The branch includes current main through `be63647e5`, has no merge conflicts, and all three review threads are resolved.

| Measurement | Diagnostic baseline | Earlier optimized checkpoint | Current merged revision |
| --- | ---: | ---: | ---: |
| Linux test step | 196s | 139s | 121s |
| Windows test step | 139s | 124s | 107s |
| Linux test job | 386s | 369s | 380s |
| Windows test job | 322s | 295s | 280s |

Baseline: [34031392110](https://github.com/goreleaser/goreleaser/actions/runs/34031392110). Earlier checkpoint: [b9f377ce4, warm-cache attempt](https://github.com/goreleaser/goreleaser/actions/runs/34056030635/attempts/2).

The earlier checkpoint built only a Windows Node SEA. Review requested real Darwin, Linux, and Windows builds; all three now run on both CI platforms. Current timings include that broader coverage and upstream fixes merged from main. These are complete CI samples, not isolated measurements of the lint fixes. Cache/setup time varies, and package/subtest durations overlap.

## Changes

- Build real Node SEAs for darwin-arm64, linux-x64, and win-x64 in isolated parallel cases. Check Mach-O, ELF, and PE architecture metadata, retain timestamp and scratch-config assertions, and execute outputs matching the host OS/architecture. Tool templates use absolute paths without shared PATH or CWD mutation.
- Use standard-library httptest servers and recorders for HTTP fixtures. Preserve Node request/checksum/extraction assertions and GitLab probe counts and retry-budget bounds. Keep virtual time for the GitHub rate-limit tests without the httpmock dependency.
- Keep separate real Mach-O signing coverage, including the empty-signature placeholder.
- Replace unnecessary project generators, Git initialization, histories, verification clones, and fake input files with small explicit fixtures. Keep real compiler, package, and publication assertions.
- Run isolated Docker, ko, package, and process-lifecycle cases concurrently. Keep shared-state cases sequential and retain the real one-second descendant-pipe contract. Disable race exit sleeps only in synchronous helper processes, not race detection itself.
- Keep real cryptographic verification. Password-input cases use two artifacts so repeated input delivery remains covered; lintian-only cases no longer perform unrelated signing. Cosign signs immutable digests without unnecessary external signing services.
- Create MinIO buckets through the existing S3 SDK instead of another container, isolate the conditional-upload bucket, and use test-owned failure endpoints.

## Compiler cache

Keep Zig and Rust compiler data outside the workspace and save it only after successful tests. Failed or cancelled tests cannot publish a partial cache. Runner/platform/workflow keys separate incompatible caches. The validated restored archives were about 18 MiB and took 1–3 seconds to restore.

## Integration and correctness

Preserved the latest main changes, including lazy project-name evaluation, the nfpm temporary-file close, Go-builder documentation, and stable override-environment ordering. Fixed Gitea fixture helper attribution and handler-safe error assertions, and removed the JSON helper parameter that was always HTTP 200; no assertions or checks were disabled.

Existing AWS modules were reclassified as direct test dependencies, and the httpmock dependency was removed. Verbose output from #7108 remains enabled. Validation after the request to stop local builds has run entirely on GitHub-hosted runners.


caarlos0:
## CI measurements

Compared [baseline 34031392110](https://github.com/goreleaser/goreleaser/actions/runs/34031392110) with [8992066b4, run 34034301766, attempt 2](https://github.com/goreleaser/goreleaser/actions/runs/34034301766/attempts/2).

| Measurement | Before | After | Difference |
| --- | ---: | ---: | ---: |
| Ubuntu test step | 196s | 186s | -10s (-5.1%) |
| Windows test step | 139s | 133s | -6s (-4.3%) |
| Ubuntu test job | 386s | 381s | -5s |
| Windows test job | 322s | 309s | -13s |

The previous performance run also completed both test jobs: Ubuntu was 185s and Windows was 140s. Ubuntu shows a consistent modest gain; Windows remains close to normal run-to-run variation. Setup time varies and is not a code speedup.

Largest direct Ubuntu package reductions:

| Package | Before | After |
| --- | ---: | ---: |
| cmd | 30.370s | 4.355s |
| internal/pipe/docker | 49.076s | 25.587s |
| internal/pipe/flatpak | 31.052s | 6.772s |

These are package durations, not sums of subtest times. Packages overlap, and Go excludes parallel-child waiting time from the parent test duration. Docker v2 was roughly unchanged on Ubuntu (15.012s to 15.523s); its Windows package went from 11.452s to 7.978s. Windows cmd increased from 25.513s to 30.304s, so this is not a claim that every package improved.

Remaining slowest packages, in descending order:

| Rank | Ubuntu | Windows |
| --- | --- | --- |
| 1 | node: 59.728s | rust: 37.866s |
| 2 | rust: 31.894s | zig: 31.564s |
| 3 | ko: 26.955s | cmd: 30.304s |
| 4 | docker: 25.587s | git: 25.542s |
| 5 | snapcraft: 23.259s | client: 17.558s |
| 6 | zig: 21.232s | changelog: 15.463s |

All required checks now pass. The first attempt at this revision hit an nfpm TestSkipSign temporary-file lock on Windows and cancelled Ubuntu; rerunning the unchanged revision passed. No change was made to nfpm.

The Node buffering experiment was discarded because it did not give a useful gain. No production code or dependency changes are included. Further validation was done only on GitHub-hosted runners after the request to stop local builds.

caarlos0:
## Complete over-one-second audit

Accounted for **255 distinct platform/test entries**, covering **96 root tests in 34 packages**. Nested cases are included under their owning root test.

The final revision is `b9f377ce4`. Both cold and warm runs passed: [CI run 34056030635](https://github.com/goreleaser/goreleaser/actions/runs/34056030635). No test execution was done locally after the request to use CI.

“Implemented” means that the cost was addressed, not that the test was removed or necessarily fell below one second. Retained costs preserve real integrations, required lifecycle timing, or simpler and more reliable fixtures. The table summarizes the source analysis; the complete per-platform observations are recorded in the session audit.

<details><summary>All root-test decisions and source evidence</summary>

| Package / root test | Decision | Cost and action | Source |
| --- | --- | --- | --- |
| `cmd/TestBuildAutoSnapshot` | implemented | Repeated Git fixture histories and real command pipelines. Release cases also performed gzip work that was unrelated to snapshot behavior. Use uncompressed tar archives and two commits with two tags instead of four commits. Keep real Git state, compilation, archive/checksum creation, failure paths, and clean/dirty/Pro assertions. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/cmd/util_test.go) |
| `cmd/TestBuildAutoSnapshotWithProConfig` | implemented | Repeated Git fixture histories and real command pipelines. Release cases also performed gzip work that was unrelated to snapshot behavior. Use uncompressed tar archives and two commits with two tags instead of four commits. Keep real Git state, compilation, archive/checksum creation, failure paths, and clean/dirty/Pro assertions. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/cmd/util_test.go) |
| `cmd/TestBuildBrokenProject` | implemented | Repeated Git fixture histories and real command pipelines. Release cases also performed gzip work that was unrelated to snapshot behavior. Use uncompressed tar archives and two commits with two tags instead of four commits. Keep real Git state, compilation, archive/checksum creation, failure paths, and clean/dirty/Pro assertions. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/cmd/util_test.go) |
| `cmd/TestBuildSingleTarget` | implemented | Repeated Git fixture histories and real command pipelines. Release cases also performed gzip work that was unrelated to snapshot behavior. Use uncompressed tar archives and two commits with two tags instead of four commits. Keep real Git state, compilation, archive/checksum creation, failure paths, and clean/dirty/Pro assertions. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/cmd/util_test.go) |
| `cmd/TestInitExampleConfigsAreNotDeprecated` | retained | Loads all five language examples and applies defaults to ensure none are deprecated. No external project generation or Git setup occurs here. Retain coverage of all shipped examples; the measured duration is close to the threshold. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/cmd/init_test.go) |
| `cmd/TestReleaseAutoSnapshot` | implemented | Repeated Git fixture histories and real command pipelines. Release cases also performed gzip work that was unrelated to snapshot behavior. Use uncompressed tar archives and two commits with two tags instead of four commits. Keep real Git state, compilation, archive/checksum creation, failure paths, and clean/dirty/Pro assertions. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/cmd/util_test.go) |
| `cmd/TestReleaseAutoSnapshotWithProConfig` | implemented | Repeated Git fixture histories and real command pipelines. Release cases also performed gzip work that was unrelated to snapshot behavior. Use uncompressed tar archives and two commits with two tags instead of four commits. Keep real Git state, compilation, archive/checksum creation, failure paths, and clean/dirty/Pro assertions. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/cmd/util_test.go) |
| `cmd/TestReleaseBrokenProject` | implemented | Repeated Git fixture histories and real command pipelines. Release cases also performed gzip work that was unrelated to snapshot behavior. Use uncompressed tar archives and two commits with two tags instead of four commits. Keep real Git state, compilation, archive/checksum creation, failure paths, and clean/dirty/Pro assertions. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/cmd/util_test.go) |
| `cmd/TestRootReleaseVerbose` | implemented | Constructed a full buildable Git repository just to assert root-command release dispatch and a nonzero exit status. Use an empty temporary working directory and valid configuration; missing credentials still exercise the release error and exit-code path before Git/build work. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/cmd/root_test.go) |
| `internal/builders/bun/TestBuild` | implemented | bun init creates project scaffolding and can install development dependencies before the actual cross-compile. Write a dependency-free entry point and package.json module field directly. Preserve default main discovery, the real Darwin ARM64 build, and artifact assertions. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/bun/build_test.go) |
| `internal/builders/deno/TestBuild` | implemented | Runs deno init before compiling a trivial application. Write main.ts directly and retain the real cross-compile, target metadata, and timestamp assertions. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/deno/build_test.go) |
| `internal/builders/golang/TestBuild` | retained | The tests exercise distinct target, extension, environment, override, and multiple-main combinations using tiny programs and the real Go builder. Keep the matrices. Different targets and main-package expansion are the behavior asserted; CWD mutation prevents simply making these cases parallel. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/golang/build_test.go) |
| `internal/builders/golang/TestBuildCodeInSubdir` | retained | Each constructs a tiny program and performs one real build to check a distinct working-directory edge case: code in a configured subdirectory or an unrelated .go directory. Keep the isolated real build and CWD behavior. There is no redundant repository history, generator, network request, or explicit delay to remove. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/golang/build_test.go) |
| `internal/builders/golang/TestBuildVariadic` | retained | The tests exercise distinct target, extension, environment, override, and multiple-main combinations using tiny programs and the real Go builder. Keep the matrices. Different targets and main-package expansion are the behavior asserted; CWD mutation prevents simply making these cases parallel. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/golang/build_test.go) |
| `internal/builders/golang/TestBuildWithDotGoDir` | retained | Each constructs a tiny program and performs one real build to check a distinct working-directory edge case: code in a configured subdirectory or an unrelated .go directory. Keep the isolated real build and CWD behavior. There is no redundant repository history, generator, network request, or explicit delay to remove. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/golang/build_test.go) |
| `internal/builders/golang/TestCheckBuildElipsisSingleMain` | retained | Resolve actual Go packages from small per-test modules to distinguish proxied paths, explicit names, and single versus multiple mains. These cases are already parallel and have minimal file fixtures. Mocking package discovery would remove the regression guard. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/golang/build_test.go) |
| `internal/builders/golang/TestCheckBuildElipsisSingleMainWithExplicitBinary` | retained | Resolve actual Go packages from small per-test modules to distinguish proxied paths, explicit names, and single versus multiple mains. These cases are already parallel and have minimal file fixtures. Mocking package discovery would remove the regression guard. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/golang/build_test.go) |
| `internal/builders/golang/TestCheckBuildElipsisWithExplicitBinaryError` | retained | Resolve actual Go packages from small per-test modules to distinguish proxied paths, explicit names, and single versus multiple mains. These cases are already parallel and have minimal file fixtures. Mocking package discovery would remove the regression guard. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/golang/build_test.go) |
| `internal/builders/golang/TestCheckBuildElipsisWithProxiedSubpathMain` | retained | Resolve actual Go packages from small per-test modules to distinguish proxied paths, explicit names, and single versus multiple mains. These cases are already parallel and have minimal file fixtures. Mocking package discovery would remove the regression guard. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/golang/build_test.go) |
| `internal/builders/golang/TestRunPipeWithMainFuncNotInMainGoFile` | retained | Three real builds cover an omitted main, an explicit non-main.go filename, and a directory main. Keep the three distinct input forms and their real builds. They share a CWD and mutable build context, so parallel subtests would be unsafe. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/golang/build_test.go) |
| `internal/builders/golang/TestRunPipeWithProxiedRepo` | retained | Runs go mod tidy and a real build against the pinned goreleaser/test-mod module to exercise the proxy path. An unnecessary clone was already replaced with local files. Retain the actual module resolution and proxied build. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/golang/build_test.go) |
| `internal/builders/node/TestBuild` | implemented | The old integration test downloads and inflates a large Darwin Node distribution, then signs the full runtime under the race detector. Build and inspect a real Windows SEA, execute it on Windows, test tar downloads with small checksum-verified HTTP fixtures, and test real Mach-O signing and its empty-signature placeholder with a small Go binary. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/node/build_test.go) |
| `internal/builders/node/TestSignMachO` | implemented | The old integration test downloads and inflates a large Darwin Node distribution, then signs the full runtime under the race detector. Build and inspect a real Windows SEA, execute it on Windows, test tar downloads with small checksum-verified HTTP fixtures, and test real Mach-O signing and its empty-signature placeholder with a small Go binary. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/node/build_test.go) |
| `internal/builders/poetry/TestBuild` | implemented | Starts Poetry to generate a project before invoking the real wheel and sdist builds. Copy a minimal fixed project instead of running poetry new. Keep both package formats and all artifact and timestamp assertions. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/poetry/build_test.go) |
| `internal/builders/rust/TestBuild` | implemented | Cargo project generation, target preparation, and real cross-builds for distinct ARM64 GNU and ARMv7 hard-float GNU targets. The tests assert the ABI, libc version, artifact metadata, and timestamp. Replace cargo init and generated Git repositories with a minimal checked-in project. Keep both real target builds, optimization level zero, and local output isolation. Reuse the separately saved complete Zig compiler cache on CI. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/rust/build_test.go) |
| `internal/builders/rust/TestBuildArm` | implemented | Cargo project generation, target preparation, and real cross-builds for distinct ARM64 GNU and ARMv7 hard-float GNU targets. The tests assert the ABI, libc version, artifact metadata, and timestamp. Replace cargo init and generated Git repositories with a minimal checked-in project. Keep both real target builds, optimization level zero, and local output isolation. Reuse the separately saved complete Zig compiler cache on CI. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/rust/build_test.go) |
| `internal/builders/uv/TestBuild` | implemented | Generates a small project and performs both real wheel and sdist builds. UV also initializes a Git repository by default, which these packaging assertions do not use. Pass --vcs none to avoid unused Git setup while preserving UV's project format and both real package builds. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/uv/build_test.go) |
| `internal/builders/zig/TestBuild` | implemented | One real cross-compile of an empty main to aarch64-macos. The fixture already uses Debug optimization and a shared content-addressed global cache with isolated local outputs. Keep the supported compiler backend, empty program, and real target. Preserve the complete shared compiler cache across CI runs, outside the workspace, saving only after successful tests. Cold and warm CI runs passed and the cache transfer remained small. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/builders/zig/build_test.go) |
| `internal/client/TestGitClient` | retained | Real clone, add, commit, push, and history inspection. CreateFiles holds the global gil mutex, so measured durations can include waiting behind other Git tests. Do not add ineffective parallelism or remove production synchronization for test speed. Reused-checkout sequences and signing behavior need real ordered Git operations. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/client/git_test.go) |
| `internal/client/TestGitClientReconfiguresReusedCheckout` | retained | Real clone, add, commit, push, and history inspection. CreateFiles holds the global gil mutex, so measured durations can include waiting behind other Git tests. Do not add ineffective parallelism or remove production synchronization for test speed. Reused-checkout sequences and signing behavior need real ordered Git operations. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/client/git_test.go) |
| `internal/client/TestGitClientRemovesInlinePrivateKey` | retained | Real Git success/failure paths and verification that generated keys are removed but caller-owned keys remain. Cases share TMPDIR/TMP/TEMP and inspect that directory for leaked keys. Parallel cases would invalidate the cleanup assertion. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/client/git_test.go) |
| `internal/client/TestGitClientUsesRepositorySpecificCheckout` | retained | Real clone, add, commit, push, and history inspection. CreateFiles holds the global gil mutex, so measured durations can include waiting behind other Git tests. Do not add ineffective parallelism or remove production synchronization for test speed. Reused-checkout sequences and signing behavior need real ordered Git operations. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/client/git_test.go) |
| `internal/client/TestGitClientWithSigning` | retained | Real clone, add, commit, push, and history inspection. CreateFiles holds the global gil mutex, so measured durations can include waiting behind other Git tests. Do not add ineffective parallelism or remove production synchronization for test speed. Reused-checkout sequences and signing behavior need real ordered Git operations. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/client/git_test.go) |
| `internal/client/TestGitHubChangelogRetriesOnSecondaryRateLimit` | implemented | Wall-clock waits for the SDK secondary rate-limit window and the bounded GitLab version-probe retry schedule. Use in-memory HTTP transports and testing/synctest. Preserve response types, Retry-After, retry budgets, exact request counts, and elapsed-time assertions on the virtual clock. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/client/github_test.go) |
| `internal/client/TestGitLabVersionProbeIsBounded` | implemented | Wall-clock waits for the SDK secondary rate-limit window and the bounded GitLab version-probe retry schedule. Use in-memory HTTP transports and testing/synctest. Preserve response types, Retry-After, retry budgets, exact request counts, and elapsed-time assertions on the virtual clock. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/client/github_test.go) |
| `internal/client/TestKeyPath` | retained | Real OpenSSH key serialization, password-protected key handling, parsing, and temporary-file cleanup. Subtests already run in parallel. Keep real encrypted-key coverage and do not weaken key-generation settings for speed. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/client/git_test.go) |
| `internal/elf/TestIsDynamicallyLinked` | retained | Two tiny real Go programs produce authentic static and dynamic ELF inputs. Both fixture builds are already parallel and use the shared Go build cache. Keep the readable source fixtures and real ELF parser. Checking in opaque executable files or constructing a synthetic binary format would reduce clarity and authenticity for a marginal near-threshold saving. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/elf/elf_test.go) |
| `internal/exec/TestExecuteCommandCancellationWithDescendantHeldOutputPipe` | implemented | These tests were added on main after the feature branch diverged. Each exercises the actual one-second Cmd.WaitDelay while a descendant keeps output open longer than the failure deadline. Keep the real one-second bound, long-lived descendant, readiness handshake, and result assertions. Run the independent top-level cases in parallel and ensure cancellation also happens on test failure. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/exec/exec_test.go) |
| `internal/exec/TestExecuteCommandSucceedsWithDescendantHeldOutputPipe` | implemented | These tests were added on main after the feature branch diverged. Each exercises the actual one-second Cmd.WaitDelay while a descendant keeps output open longer than the failure deadline. Keep the real one-second bound, long-lived descendant, readiness handshake, and result assertions. Run the independent top-level cases in parallel and ensure cancellation also happens on test failure. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/exec/exec_test.go) |
| `internal/git/TestGitWarning` | retained | Creates two commits, same-looking branch/tag refs and a nightly tag, then executes describe on tags/1.2.3^ and a sorted points-at tag query. Real ref ambiguity is important because git.Run separates stderr... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/git/git_test.go) |
| `internal/git/TestRelativeRemote` | retained | Constructs a real commit, named upstream remote and local remote-tracking ref, configures main's upstream, then creates a relative tracking branch. After asserting branch.relative_branch.remote=.,... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/git/config_test.go) |
| `internal/http/TestUpload` | implemented | The root executes a broad sequential HTTP/TLS matrix with a new httptest server per transport case. It exercises actual requests, request bodies, authentication, headers, filtering, extra files, trusted... In the failed-request setup, close the supplied test server and retain its original ephemeral-port URL as the upload target. This preserves a real connection-failure path without contacting an unrelated service on loopback port 80 or 443. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/http/http_test.go) |
| `internal/pipe/aur/TestFullPipe` | implemented | All seventeen serial table entries create a local bare repository, generate an Ed25519 key and materialize an empty archive. Seven successful entries generate two package files, use the real Git upload... At internal/pipe/aur/aur_test.go:275-277, initialize url/key only when neither expectedRunError nor expectedRunErrorCheck is configured. The two Run-template-error entries can use empty unused publish fields; retain tt.prepare and the original runAll error... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/aur/aur_test.go) |
| `internal/pipe/aur/TestRunPipeMultipleConfigurations` | implemented | One disabled configuration and two enabled configurations share a single local bare repository/key. Both enabled configurations publish distinct directories through the real client. The client reuses its... Replace only the final verification clone and four FileExists assertions at internal/pipe/aur/aur_test.go:590-597 with one git.Run using -C url ls-tree -r --name-only master and assertions that its output contains foo/.SRCINFO, foo/PKGBUILD, bar/.SRCINFO and... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/aur/aur_test.go) |
| `internal/pipe/aursources/TestFullPipe` | implemented | Seventeen serial table entries unconditionally create bare repositories, Ed25519 keys and source archive placeholders. Seven successful entries generate PKGBUILD/.SRCINFO, run real local Git publishing and... At internal/pipe/aursources/aursources_test.go:259-261, skip GitMakeBareRepository and MakeNewSSHKey only when expectedRunError or expectedRunErrorCheck is set. Leave their otherwise-unused publish fields empty in those two entries, preserve tt.prepare and... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/aursources/aursources_test.go) |
| `internal/pipe/aursources/TestRunPipe` | retained | Creates one local bare repository and key, generates source-package files from one empty archive placeholder, publishes via real Git, then clones and compares PKGBUILD, .SRCINFO and the copied foo.install... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/aursources/aursources_test.go) |
| `internal/pipe/aursources/TestRunPipeMultipleConfigurations` | implemented | A templated disabled entry and two enabled source-package entries share one local bare repository/key. Both enabled entries publish directories through the real client, after which a further verification clone... Replace only internal/pipe/aursources/aursources_test.go:514-521 with one explicit -C url git ls-tree -r --name-only master query. Require the four pushed paths foo/.SRCINFO, foo/PKGBUILD, bar/.SRCINFO and bar/PKGBUILD in its output. Keep the real publishes... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/aursources/aursources_test.go) |
| `internal/pipe/blob/TestMinioUpload` | implemented | The test prepares seven tiny artifact files, uploads those plus one extra file to real MinIO, and lists the bucket to assert eight object paths (blob_minio_test.go:79-185). Before uploading, setupBucket... Replace setupBucket's minio/mc container launch and disappearance polling with a synchronous AWS SDK v2 S3 CreateBucket call against the existing local MinIO endpoint, using explicit test credentials and path-style addressing. The SDK config, credentials, and... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/blob/blob_minio_test.go) |
| `internal/pipe/brew/TestFullPipe` | retained | Runs the formula/template/publication matrix using tiny empty artifact files and a mock API. Only the Git-remote case constructs a real repository and key. Keep all template, error, API, Git-remote, and golden-output checks. Expensive publication fixtures are already limited to the case that exercises them. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/brew/brew_test.go) |
| `internal/pipe/changelog/TestAbbrev` | implemented | The root constructs eleven commits and two tags, then each of six subcases runs an additional rev-list through withFirstCommit, a real git log and a CHANGELOG.md write. All subcases inspect hash abbreviation... Reduce this root's fixture to two real commits: retain first/v0.0.1, one subsequent message and v0.0.2. Multiple log entries remain, but message categories unrelated to abbreviation no longer require nine extra commits. Remove withFirstCommit(t) from the six... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/changelog/changelog_test.go) |
| `internal/pipe/changelog/TestChangelog` | retained | Builds a twelve-commit, two-tag fixture and invokes the real Git changeloger once. Separate messages exercise tag-range exclusion, regex exclusions, case-insensitive matching, anchored merge filtering, quoted... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/changelog/changelog_test.go) |
| `internal/pipe/changelog/TestChangelogForGitlab` | retained | Creates nine commits and two tags, then generates a local Git changelog and writes it to disk. GitLab TokenType changes newline rendering; because Changelog.Use is empty this is not a GitLab API/network test.... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/changelog/changelog_test.go) |
| `internal/pipe/changelog/TestChangelogOnBranchWithSameNameAsTag` | implemented | Creates four commits, a tag and an identically named checked-out branch, then computes FirstCommit with rev-list and executes git log. Every message is asserted in the first-release output. The current git... Remove withFirstCommit(t) from the context options at internal/pipe/changelog/changelog_test.go:536. Retain all four commits, the real v0.0.1 tag and branch, and every changelog assertion. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/changelog/changelog_test.go) |
| `internal/pipe/changelog/TestGroup` | implemented | Creates eleven commits and two tags, computes the first commit with a separate rev-list, then logs and groups the real history. Unlike TestAbbrev, the expected multiline regexp enumerates the messages and... Remove only withFirstCommit(t) from the context options at internal/pipe/changelog/changelog_test.go:990. Keep all messages, tags, the full pipe run and the multiline output assertion. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/changelog/changelog_test.go) |
| `internal/pipe/docker/TestPushDenied` | retained | Builds a tiny real rejecting registry, runs it in a container reachable by the Docker daemon, and checks rejection with both imagers. The fixture already uses a scratch image, native builder, allocated port, and readiness output instead of sleeps. A host-only mock would not exercise daemon-visible registry errors. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/docker/denied_test.go) |
| `internal/pipe/docker/TestRunPipe` | implemented | Many real Docker/buildx build-and-push cases, manifest operations, label inspection, and image cleanup. Give successful cases separate image names and run cases in parallel while keeping imagers within a case sequential. Preserve all cases and reject unexpected detailed errors instead of silently accepting them. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/docker/docker_test.go) |
| `internal/pipe/docker/v2/TestGetBaseImage` | implemented | The error-retains-base-name case runs docker buildx against a nonexistent public image. Use an already-cancelled context for the resolution failure. Keep parsing and the base-name/empty-digest/error assertions without public-registry access. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/docker/v2/baseimage_test.go) |
| `internal/pipe/docker/v2/TestIsDockerDaemonAvailableNoDaemon` | retained | The real Docker CLI's unavailable-daemon probe remains slow on Windows even against a closed loopback port. The missing Unix socket was already replaced with loopback. Keep the real unavailable-daemon path rather than change production timeouts or make the test only cover a missing executable. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/docker/v2/docker_test.go) |
| `internal/pipe/docker/v2/TestPublish` | retained | Real multi-platform images, Python-wheel selection, OCI metadata, SBOMs, registry pushes, and manifest inspection. Keep the distinct integration contracts and cleanup. Replacing them with fake commands would lose image/index behavior. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/docker/v2/docker_integration_test.go) |
| `internal/pipe/docker/v2/TestRun` | retained | Real multi-platform images, Python-wheel selection, OCI metadata, SBOMs, registry pushes, and manifest inspection. Keep the distinct integration contracts and cleanup. Replacing them with fake commands would lose image/index behavior. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/docker/v2/docker_integration_test.go) |
| `internal/pipe/flatpak/TestRunPipe` | retained | Real Flatpak construction and validation that a disabled entry does not suppress a later enabled entry. Keep the real tool integration and enabled-after-disabled regression guard. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/flatpak/flatpak_test.go) |
| `internal/pipe/flatpak/TestRunPipeQuotesInstallPaths` | retained | Runs the generated install commands through a shell and synchronous re-executed helpers for ordinary, space-containing, quoted, and hostile names. The helper race-exit sleeps were already removed. Keep real shell parsing and exact install-argument checks; remaining startup cost is small. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/flatpak/flatpak_test.go) |
| `internal/pipe/flatpak/TestRunPipeSomeDisabled` | retained | Real Flatpak construction and validation that a disabled entry does not suppress a later enabled entry. Keep the real tool integration and enabled-after-disabled regression guard. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/flatpak/flatpak_test.go) |
| `internal/pipe/git/TestBranch` | implemented | Both branches independently recreate the same repository, remote, commit and tag before checking out their branch and running the full Git pipe. The template render itself is in-process. The apostrophe subtest... Move Mktmp, GitInit, GitRemoteAdd, GitCommit and GitTag from the subtest body to the root before the branch loop. Keep GitCheckoutBranch and every pipe/template assertion inside each serial t.Run. Both branches then start at the same unchanged commit,... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/git/git_test.go) |
| `internal/pipe/git/TestDirty` | retained | A real tracked file is committed, tagged and modified once. Three contexts then run the full pipe against that already-shared dirty repository. Normal validation calls CheckDirty after metadata collection has... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/git/git_test.go) |
| `internal/pipe/git/TestFilterTags` | implemented | The root creates four commits and four tags, then runs the Git pipe three times. Each successful run performs real Git metadata, tag-discovery and validation commands; even the invalid-template case reaches... Remove only testlib.GitCommit(t, "commit2") at internal/pipe/git/git_test.go:497. Keep commit1/v0.0.1, middle commit/nightly, and commit3 with both HEAD tags. This removes one fixture process without changing the tested tag graph relationships. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/git/git_test.go) |
| `internal/pipe/git/TestNoPreviousTag` | retained | One real repository, one commit, one tag, and the Git pipe verify the absence of a previous tag while preserving first-commit metadata. The repository graph is already minimal. Retain the full metadata integration; the Windows sample only slightly exceeds one second. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/git/git_test.go) |
| `internal/pipe/git/TestPreviousTagFromCI` | retained | One already-shared repository contains two separately tagged commits. Two complete pipe invocations exercise previous-tag ancestry discovery and the environment override. The default path performs describe,... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/git/git_test.go) |
| `internal/pipe/git/TestShallowClone` | retained | Creates a two-commit tagged upstream, performs a real depth-one file:// clone, asserts .git/shallow exists, then invokes the complete Git pipe three times. The local clone and repeated metadata subprocesses... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/git/git_test.go) |
| `internal/pipe/git/TestSnapshotNoTags` | implemented | The clean and dirty entries each recreate the same tagless repository, remote and committed file before invoking the snapshot pipe. The only fixture difference is whether that tracked file's contents change.... Hoist the temp repository, remote, tracked foo file, GitAdd and GitCommit into the root before the case loop. In each serial subtest explicitly write either the original "initial" contents or "dirty" contents before constructing a fresh context and calling... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/git/git_test.go) |
| `internal/pipe/git/TestTagFromCI` | retained | The repository, one commit and two same-commit tags are already shared. The loop runs Pipe.Run twice: normal tag discovery and GORELEASER_CURRENT_TAG selection. Environment selection can avoid tag discovery,... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/git/git_test.go) |
| `internal/pipe/git/TestTagIsNotLastCommit` | retained | The minimal history is one tagged commit followed by exactly one untagged commit. The full pipe must fall back from tags pointing at HEAD to git describe, then reject the non-exact tag while exposing a... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/git/git_test.go) |
| `internal/pipe/git/TestTagSortOrder` | implemented | Creates three empty commits and two tags on the last one, in an order different from the expected descending version order, then runs the complete pipe. Only the final tagged commit and an untagged parent are... Remove GitCommit("commit1") at internal/pipe/git/git_test.go:265. Keep commit2, commit3, both same-HEAD tags in their existing order, and the full current-tag assertion. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/git/git_test.go) |
| `internal/pipe/git/TestTagSortOrderPrerelease` | implemented | The fixture creates three empty commits although the release and prerelease tags both point to the last commit. The tested choice is Git's version sorting with versionsort.suffix=-, followed by a complete pipe... Remove GitCommit("commit1") at internal/pipe/git/git_test.go:284, retaining commit2 as an untagged parent and commit3 as the target of v0.0.1-rc.2 and v0.0.1. Keeping two commits also preserves a resolvable current-tag parent for ancillary previous-tag... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/git/git_test.go) |
| `internal/pipe/git/TestValidState` | retained | Two commits and three tags exercise a valid current tag, selection among two tags on the previous commit, remote extraction, first-commit discovery and cleanliness through the full pipe. The prior commit and... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/git/git_test.go) |
| `internal/pipe/gomod/TestGoModProxy` | retained | Four cases execute real go get against pinned test modules to verify generated proxy modules and main-path rewriting. The mixed-builder and environment-precedence cases already use small fake Go executables. Keep actual module resolution for the distinct proxy integration cases. Replacing it with a new local proxy or no-op command would add complexity or remove the behavior asserted; no measured cache or network problem justifies that change. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/gomod/gomod_proxy_test.go) |
| `internal/pipe/ko/TestDisable` | retained | Real local Docker base-image handling, snapshot loading without publishing, and enabled work following a disabled entry. Keep these distinct integration paths. Snapshot summary capture and shared local image names make simple test parallelism unsafe. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/ko/ko_test.go) |
| `internal/pipe/ko/TestPublishLocalBaseImage` | retained | Real local Docker base-image handling, snapshot loading without publishing, and enabled work following a disabled entry. Keep these distinct integration paths. Snapshot summary capture and shared local image names make simple test parallelism unsafe. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/ko/ko_test.go) |
| `internal/pipe/ko/TestPublishPipeError` | implemented | Distinct parsing/build failures plus a publication failure that depended on external DNS for a .invalid hostname. Keep all negative cases and the real build. Replace the DNS-dependent publication failure with a local HTTP server that rejects publication. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/ko/ko_test.go) |
| `internal/pipe/ko/TestPublishPipeSuccess` | implemented | Real image builds and registry pushes for SBOM, base-image, platform, labels, annotations, user, tag, and timestamp variants. Shared digest-addressed SBOM tags imposed ordering. Give cases separate repositories and run the cases in parallel, retaining both registries and all image/manifest/SBOM assertions. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/ko/ko_test.go) |
| `internal/pipe/ko/TestSnapshot` | retained | Real local Docker base-image handling, snapshot loading without publishing, and enabled work following a disabled entry. Keep these distinct integration paths. Snapshot summary capture and shared local image names make simple test parallelism unsafe. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/ko/ko_test.go) |
| `internal/pipe/makeself/TestRun` | retained | Uses a tiny shell-script payload, generates real self-extracting archives for platform/filter cases, inspects their contents and LSM metadata, and runs the installer. Keep the real packer, archive inspection, and executable-installer assertions. The fixture already avoids compiled binaries, and no unnecessary external service or wait is present. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/makeself/makeself_test.go) |
| `internal/pipe/nfpm/TestDebSpecificConfig` | implemented | The common setup installs an encrypted Debian signing key for every subcase, not only the passphrase cases (nfpm_test.go:1125-1176). The lintian case inherits that signing configuration, enables five formats,... Clear Deb.Signature in the lintian, lintian override only, and lintian override replaces top level cases, and remove their now-unnecessary passphrase environment maps. Alternatively, make the shared setup unsigned and add a small signed-context wrapper used... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/nfpm/nfpm_test.go) |
| `internal/pipe/nfpm/TestRPMSpecificConfig` | retained | The two successful cases each package and sign Linux amd64 and 386 artifacts using the encrypted RPM signing-key fixture. Darwin artifacts are excluded by the production artifact filter... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/nfpm/nfpm_test.go) |
| `internal/pipe/nfpm/TestRunPipe` | retained | The fixture uses empty files rather than compiling binaries, then exercises six formats across platform and architecture variants. It asserts 57 generated packages (nfpm_test.go:101-115,135,216-416). The... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/nfpm/nfpm_test.go) |
| `internal/pipe/nfpm/TestRunPipeConventionalNameTemplate` | implemented | Regular and snapshot children run sequentially. Each constructs its own temporary directory and context, writes a small fake executable, and generates 52 real packages (nfpm_test.go:569-609,725-727). The... Add t.Parallel at the start of the regular and snapshot child callbacks, leaving the root sequential and keeping the entire package matrix and assertions. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/nfpm/nfpm_test.go) |
| `internal/pipe/nix/TestRunPipe` | implemented | Every table entry recreates a broad artifact metadata matrix and repeatedly creates empty archive files, including cases that fail before package generation. Build and publish use alwaysZeroHasher and... Delete only the input-file materialization at internal/pipe/nix/nix_test.go:494-497 from createFakeArtifact: MkdirAll, os.Create, its error assertion and Close. Keep artifact Name/Path/Extra records and every matrix entry. Both injected hashers consume only... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/nix/nix_test.go) |
| `internal/pipe/sbom/TestSBOMCatalogRealSyft` | retained | The test already uses one four-byte fake binary in a private temporary directory and invokes syft once, asserting the resulting SBOM artifact name and count (sbom_test.go:520-556). Its explicit arguments omit... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/sbom/sbom_test.go) |
| `internal/pipe/sign/TestDockerSignArtifacts` | implemented | The no-signature case invokes real cosign with an encrypted-key fixture, stdin password, --upload=false, and three image/manifest references. sign processes selected artifacts sequentially, producing three... Copy signer fixtures only for cases that consume them; echo-only cases need neither the key nor certificate fixture. Consider changing the real cosign argument from ${artifact} to ${artifact}@${digest}, consistent with the default command and existing digest... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/sign/sign_docker_test.go) |
| `internal/pipe/sign/TestSignArtifacts` | implemented | Despite their 'sign single' names, all three password cases select all nine artifacts and expect nine signatures (sign_test.go:432-515). Each selected artifact launches a signer, and the helper subsequently... For the three password-input cases, select Artifacts: binary and expect artifact3.sig plus linux_amd64/artifact4.sig, with their existing corresponding artifact names. Preserve real password-protected GPG signing and verification. Keep two artifacts rather... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/sign/sign_test.go) |
| `internal/pipe/snapcraft/TestPublish` | implemented | Started the real Snapcraft CLI only to assert wrapping of a failed command. Clear PATH for a deterministic command failure. Keep the same package-path error assertion without CLI startup or store access. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/snapcraft/snapcraft_test.go) |
| `internal/pipe/snapcraft/TestRunPipe` | retained | Builds nine real snaps for multiple binary selections and architectures, plus a disabled configuration. Platform work already uses a bounded group. Retain the real packer validation across these combinations; fake metadata tests already cover cheaper paths separately. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/snapcraft/snapcraft_test.go) |
| `internal/pipe/sourcearchive/TestArchive` | implemented | For each of tar.gz, tar, zip and tgz the test recreates the same Unicode-containing tracked tree, attributes, ignore file, commit and tag, then adds the same untracked/ignored files. Each format runs both... Move the repository/tree setup currently at internal/pipe/sourcearchive/source_test.go:19-48 to the root before the format loop. Keep a single root-owned temp CWD, initial dist directory, commit, tag and ignored/untracked files. Keep all eight archive runs... | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/sourcearchive/source_test.go) |
| `internal/pipe/universalbinary/TestRun` | implemented | Real Mach-O fixtures for distinct CPU slices and duplicate AMD64 variants. Disabling optimization for all packages created a separate standard-library cache. Remove the unnecessary all-packages gcflags override. Keep every real CPU/variant build, duplicate rejection, merge, replacement, and hook assertion. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/universalbinary/universalbinary_test.go) |
| `internal/pipe/universalbinary/TestRunDuplicateCPUSlices` | implemented | Real Mach-O fixtures for distinct CPU slices and duplicate AMD64 variants. Disabling optimization for all packages created a separate standard-library cache. Remove the unnecessary all-packages gcflags override. Keep every real CPU/variant build, duplicate rejection, merge, replacement, and hook assertion. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/pipe/universalbinary/universalbinary_test.go) |
| `internal/shell/TestRunCommand` | implemented | The two descendant-held-output cases each wait for the production one-second WaitDelay. Main now contains the lifecycle cases that were absent from the earlier local checkout. Overlap the two independent lifecycle cases instead of shortening their deadlines. Keep their real processes, readiness signal, cancellation, and success/error semantics. The separate global-logger test remains serial. | [source](https://github.com/goreleaser/goreleaser/blob/b9f377ce4681018d81ae9aae1b7e84c7cf64fcb0/internal/shell/shell_test.go) |

</details>

internal/builders/node/build_test.go
```diff
@@ -102,7 +103,9 @@ func TestResolveVersionStringRejectsUnsupportedSEARelease(t *testing.T) {
 func TestBuild(t *testing.T) {
 	testlib.CheckPath(t, "node")
 
-	target := "darwin-arm64"
+	// Windows binaries are available without a large compressed archive.
+	// Download extraction and Mach-O signing have separate tests.
+	target := "win-x64"
```
caarlos0: we should probably test both darwin, linux, and windows, actually.

internal/builders/node/download_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"compress/gzip"
+	"crypto/sha256"
+	"fmt"
+	"net/http"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/goreleaser/goreleaser/v2/internal/nodedist"
+	"github.com/jarcoal/httpmock"
```
caarlos0: use httptest instead of this lib

internal/client/gitlab_test.go
```diff
@@ -13,13 +13,15 @@ import (
...
 	"testing"
+	"testing/synctest"
 	"text/template"
 	"time"
 
 	"github.com/goreleaser/goreleaser/v2/internal/artifact"
 	"github.com/goreleaser/goreleaser/v2/internal/retryx"
 	"github.com/goreleaser/goreleaser/v2/internal/testctx"
 	"github.com/goreleaser/goreleaser/v2/pkg/config"
+	"github.com/jarcoal/httpmock"
```
caarlos0: use stdlib http test pkg

internal/builders/node/build_test.go
caarlos0: Addressed in ccdc691e1. TestBuild now builds darwin-arm64, linux-x64, and win-x64 in isolated parallel cases without changing shared PATH or CWD. It checks each executable format and architecture, retains timestamp/config isolation assertions, and executes the output when it matches the runner. All three targets passed on both Linux and Windows CI at cb43948cb.

internal/builders/node/download_test.go
caarlos0: Addressed in 15590b6bc, with the distribution-path correction in cb43948cb. The test now uses httptest.NewServer and a small request redirector rather than httpmock. It retains the real HTTP request, host/path/method checks, checksum verification, extraction assertions, and exact request count. Linux and Windows CI passed.

internal/client/gitlab_test.go
caarlos0: Addressed in c08a301c0. The GitLab version-probe test now uses httptest.NewServer and the real HTTP client path. The mock-library and synctest imports were removed from this file; exact probe counts and the bounded retry-budget assertion remain. Linux and Windows CI passed at cb43948cb.

PR #7110, opened 2026-09-06
fix(github): honor short Retry-After delays

### LINKED ISSUE #6918: github: short Retry-After delays become one minute
### Problem

When GitHub returns a secondary-rate-limit response with `Retry-After: 1`, a release using the normal five-minute maximum retry delay waits 60 seconds rather than the supplied one second. Changelog generation and other calls through `githubDo` take this path.

[`githubError` applies `max(serverDelay, time.Minute)`](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/client/github.go#L72-L73), and [the retry layer uses the resulting duration](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/retryx/retryx.go#L104-L110). The fallback therefore also replaces valid shorter delays.

### Evidence

An independent reproduction used the real SDK's secondary-rate-limit parsing and production changelog/retry path with a recording transport under `testing/synctest`. With `max_delay: 5m`, requests occurred at virtual times `[0s, 60s]` for `Retry-After: 1`. Missing-header and 90-second controls yielded 60 and 90 seconds respectively. No live provider request or minute-long real sleep was used. This demonstrates excess latency, not a protocol violation from waiting longer.

### Correction

Honor a supplied usable delay and reserve the one-minute fallback for its absence. Keep the retry budget and configured maximum behavior. Add virtual-clock regressions for short, absent, and longer header values.

---
Audit priority: **P2**.

0jaspahwa:
Closes #6918.

`githubError` used `max(*must(arle.RetryAfter), time.Minute)`, where the minute was doing two jobs — fallback for a missing header, and floor for a present one. `must` returns a pointer to the zero value when the header is absent, so both cases collapsed into the same expression and a `Retry-After: 1` became 60s.

Now it branches on whether the value is usable, keeping the minute for absence only.

Added `TestGitHubSecondaryRateLimitHonorsRetryAfter` with the three virtual-clock cases from the issue — short (1s), absent (1m), longer (1m30s). It fails on current code with 1m0s for the short case.

Left `TestGitHubChangelogRetriesOnSecondaryRateLimit` alone; its `MaxDelay: 2s` comment was describing this bug, happy to update it if you'd like.

### DIFF
--- a/internal/client/github.go
+++ b/internal/client/github.go
@@ -70,7 +70,11 @@ func githubError(err error, resp *github.Response) error {
 	if rle, ok := errors.AsType[*github.RateLimitError](err); ok {
 		he.RetryAfter = max(time.Until(rle.Rate.Reset.Time), time.Second)
 	} else if arle, ok := errors.AsType[*github.AbuseRateLimitError](err); ok {
-		he.RetryAfter = max(*must(arle.RetryAfter), time.Minute)
+		if d := *must(arle.RetryAfter); d > 0 {
+			he.RetryAfter = d
+		} else {
+			he.RetryAfter = time.Minute
+		}
 	}
 	return he
 }
--- a/internal/client/github_test.go
+++ b/internal/client/github_test.go
@@ -167,6 +167,70 @@ func TestGitHubUploadReleaseIDNotInt(t *testing.T) {
 	)
 }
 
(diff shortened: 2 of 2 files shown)


caarlos0 (approved):

caarlos0:
thanks!

PR #7111, opened 2026-09-06
fix(project): evaluate name candidates lazily

### LINKED ISSUE #6921: project: name selection runs unnecessary fallback commands
### Trigger

Leave `project_name` unset, but provide a Cargo package name or a configured GitHub, GitLab, or Gitea release repository name. Those earlier candidates already determine the project name, so module and remote lookups should not run.

### Observed

The production project-default selector ran with local recording `go` and `git` executables. For each Cargo/provider winner it still invoked both `go list -m` and `git -c log.showSignature=false ls-remote --get-url`. A module-name winner also invoked the unnecessary Git fallback. An explicit `project_name` control invoked neither command. Verification measured command counts only; no wall-time speedup, network delay, or benchmark improvement is claimed.

[The candidate slice evaluates every fallback before selection starts](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/project/project.go#L27-L33).

### Smallest fix

Evaluate candidates lazily in the existing priority order and stop after the first nonempty result. Keep current error handling and name precedence. Add command-recorder regressions for explicit, Cargo, all three provider, module, and Git fallback cases, asserting both selected names and exact command counts.

---
Audit priority: **P2**.

0jaspahwa:
Closes #6921.

The candidate slice is a composite literal, so Go evaluates every element before the loop body runs. That means `moduleName` and `gitRemote` shell out to `go list -m` and `git ls-remote` even when an earlier candidate already won.

Changed it to `[]func() string` so each candidate only runs when it's reached. Same order, same precedence, same error handling.

The ten existing tests in this package cover every selection path and all still pass.

I haven't added the command count regressions. There's no command recorder in `testlib` right now, and faking `go`/`git` on `PATH` needs different shims on Windows and Linux. Happy to add them if you want, just wanted to check the approach first.

### DIFF
--- a/internal/pipe/project/project.go
+++ b/internal/pipe/project/project.go
@@ -24,18 +24,19 @@ func (Pipe) Default(ctx *context.Context) error {
 		return nil
 	}
 
-	for _, candidate := range []string{
-		cargoName(),
-		ctx.Config.Release.GitHub.Name,
-		ctx.Config.Release.GitLab.Name,
-		ctx.Config.Release.Gitea.Name,
-		moduleName(ctx),
-		gitRemote(ctx),
+	for _, candidate := range []func() string{
+		cargoName,
+		func() string { return ctx.Config.Release.GitHub.Name },
+		func() string { return ctx.Config.Release.GitLab.Name },
+		func() string { return ctx.Config.Release.Gitea.Name },
+		func() string { return moduleName(ctx) },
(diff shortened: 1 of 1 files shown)


caarlos0 (approved):

caarlos0:
thanks!

PR #7114, opened 2026-09-06
fix(builder/go): give a single ellipsis main an exact output path

### LINKED ISSUE #6898: go: ellipsis WASM builds register filenames that do not exist

caarlos0:
Closes #6898

## Problem

An ellipsis `main` always passed an output **directory**:

```go
cmd = append(cmd, "-o", filepath.Dir(options.Path))
```

go then names each output after its package and drops the extension that
`extFor` registered for the target. `GOEXE` is empty for `js/wasm` and
`wasip1/wasm`, so `go build -o <dir>` writes `foo` while GoReleaser registered
`foo.wasm`. Only wasm goes out of step: `main: ./cmd/foo` is fine, because it
uses `-o <path>`.

Three outcomes with `main: ./...` and `targets: [js_wasm]`:

| config | result |
|---|---|
| bare | **"build succeeded"**, registering a nonexistent `dist/x_js_wasm/foo.wasm` |
| `mod_timestamp` | `chtimes ...: no such file or directory` |
| `archives` | `failed to add ...: lstat ...: no such file or directory` |

Reporting success while writing a path that does not exist into
`artifacts.json` is the part that matters.

## Fix

When the ellipsis resolves to exactly one main package, give go the exact `-o`
path, the same as a non-ellipsis build. go honours the name even when it is not
the one it would choose (`-o out/custom.lib` writes `custom.lib`), so nothing
needs a rename afterwards.

```go
output := filepath.Dir(options.Path)
if len(mains) == 1 {
	output = artifact.Path
}
cmd = append(cmd, "-o", output)
```

Several main packages still need a directory. That case keeps the old
behaviour, and it stays wrong for wasm. It is left alone on purpose: no user
has reported it, and the alternative is to copy `cmd/go` output-naming rules
into GoReleaser, which is not a stable interface to copy.

## Tests

Both fail without the fix:

- `TestBuildVariadicWasmArtifactsExistWithModTimestamp` — the issue, end to end,
  with a real wasm build and `mod_timestamp`.
- `TestBuildGoBuildLineEllipsisOutput` — the `-o` shape for one main against
  several.

### DIFF
--- a/internal/builders/golang/build.go
+++ b/internal/builders/golang/build.go
@@ -452,7 +452,14 @@ func buildGoBuildLine(
 		// NOTE: build.Main will never be empty here
 		cmd = append(cmd, "-o", options.Path, build.Main)
 	} else {
-		cmd = append(cmd, "-o", filepath.Dir(options.Path))
+		output := filepath.Dir(options.Path)
+		if len(mains) == 1 {
+			// go names the output itself when -o is a directory, dropping the
+			// extension registered for the target, e.g. `.wasm`. A single main
+			// can get an exact path instead.
+			output = artifact.Path
+		}
+		cmd = append(cmd, "-o", output)
 		cmd = append(cmd, slices.Sorted(maps.Values(mains))...)
 	}
 	return cmd, nil
--- a/internal/builders/golang/build_test.go
+++ b/internal/builders/golang/build_test.go
@@ -2387,3 +2387,73 @@ func cSharedExt() string {
 		return ".so"
 	}
 }
+
+func TestBuildVariadicWasmArtifactsExistWithModTimestamp(t *testing.T) {
+	modTime := time.Date(2023, time.November, 14, 22, 13, 20, 0, time.UTC)
+
+	folder := testlib.Mktmp(t)
+	writeGoMod(t, folder, "github.com/foo/bar")
+	writeGoodMain(t, filepath.Join(folder, "cmd", "foo"))
+
+	target := mustParse(t, "js_wasm")
+	build := config.Build{
+		ID:           "foo",
+		Main:         "./cmd/...",
+		Tool:         "go",
+		Command:      "build",
+		ModTimestamp: fmt.Sprintf("%d", modTime.Unix()),
+	}
+	options := api.Options{
+		Target: target,
+		Name:   "foo.wasm",
+		Path:   filepath.Join(folder, "dist", target.Target, "foo.wasm"),
+		Ext:    ".wasm",
+	}
+	require.NoError(t, os.MkdirAll(filepath.Dir(options.Path), 0o755))
+
+	ctx := testctx.WrapWithCfg(t.Context(), config.Project{Builds: []config.Build{build}})
+	require.NoError(t, Default.Build(ctx, build, options))
+
+	bins := ctx.Artifacts.Filter(artifact.ByType(artifact.Binary)).List()
+	require.Len(t, bins, 1)
+	require.Equal(t, filepath.ToSlash(filepath.Join("dist", target.Target, "foo.wasm")), bins[0].Path)
+	info, err := os.Stat(bins[0].Path)
+	require.NoError(t, err)
+	require.NotZero(t, info.Size())
+	require.Equal(t, modTime, info.ModTime().UTC())
+}
+
+func TestBuildGoBuildLineEllipsisOutput(t *testing.T) {
+	t.Parallel()
+
+	dir := filepath.Join("dist", "foo_js_wasm")
+	path := filepath.Join(dir, "app.wasm")
+
+	// go names the output itself when -o is a directory, which drops the
+	// registered extension. A single main gets the exact path instead.
+	for name, tt := range map[string]struct {
+		mains map[string]string
+		want  []string
+	}{
+		"single main gets an exact output path": {
+			mains: map[string]string{"app": "./cmd/app"},
+			want:  []string{"go", "build", "-o", path, "./cmd/app"},
+		},
+		"several mains get an output directory": {
+			mains: map[string]string{"app": "./cmd/app", "other": "./cmd/other"},
+			want:  []string{"go", "build", "-o", dir, "./cmd/app", "./cmd/other"},
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			build := config.Build{Main: "./cmd/...", Tool: "go", Command: "build"}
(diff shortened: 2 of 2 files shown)


internal/builders/golang/build.go
```diff
@@ -350,6 +353,47 @@ func (*Builder) Build(ctx *context.Context, build config.Build, options api.Opti
...
+		}
+		renames = append(renames, rename{src: src, tmp: src + ".goreleaser-tmp", dst: a.Path})
+	}
+
+	// move every output aside before moving any into place: one output may be
+	// named like another output's destination, e.g. `app` and `app.wasm` when
+	// building `./cmd/app` and `./cmd/app.wasm` for wasm.
+	for _, r := range renames {
+		if err := os.Rename(r.src, r.tmp); err != nil {
+			return fmt.Errorf("rename build output: %w", err)
```
caarlos0: **`command: test` with `main: ./...`: the build now fails hard, and it orphans a binary, when a main package has no tests.**

`go test -c -o <dir> ./...` writes nothing for a main package that has no `_test.go` file, and it exits 0:

```
$ go test -c -o outt ./cmd/...
?   	example.com/foo/cmd/other	[no test files]
$ ls outt
app.test
```

`gomain.All` still returns both mains, so this loop renames `app.test` to `app.test.goreleaser-tmp` and then fails on the missing `other.test`. I reproduced it at `cedf10fb` with a copy of `TestBuildVariadicTestCommandOutputs` in which only `cmd/app` has a test file:

```
Build() err = rename build output: rename .../dist/darwin_arm64_v8.0/other.test
  .../dist/darwin_arm64_v8.0/other.test.goreleaser-tmp: no such file or directory
dist entry: "app.test.goreleaser-tmp"
```

Two user-visible problems:

1. **It is a regression.** The same configuration exits 0 on `main`. I ran the same repro against the base `build.go` through `go test -overlay`: `Build() err = <nil>`, `dist entry: "app.test"`. A `goreleaser build` that completed before now stops with an error. `command: test` is documented (`www/content/customization/builds/builders/go.md:252`), and a main package without tests is the normal case.
2. **`dist` is left worse than before the fix.** The binary that *was* built is orphaned under the internal `.goreleaser-tmp` name, and the message names that internal path instead of saying that `./cmd/other` produced no binary.

**Smallest fix:** `os.Rename` should not be the thing that discovers that go wrote no output. Stat `r.src` first. For `command: test`, drop the artifact for a package that produced no binary — that is the correct outcome, because `go test -c` legitimately writes nothing there, and registering a path that does not exist is the bug this PR fixes. If you prefer to stop the build instead, the error must name the package and the reason, and the temporaries already moved must go back.

**Test:** the variant above — two mains under `./cmd/...`, `command: test`, and only one of them with a `_test.go`. `TestBuildVariadicTestCommandOutputs` gives *both* packages a test file, so it cannot catch this.

**Same root cause, different trigger — please keep it out of this PR.** `gomain.BinaryNameFor` returns `filepath.Base(pkg.Dir)`, but go strips a `vN` element from *any* import path, not only from the module root. With `cmd/v2/main.go`, `GOOS=js GOARCH=wasm go build -o out ./cmd/...` writes `cmd`, not `v2`, so a wasm build hits the same `rename ... no such file or directory`. That fault already exists on `main`, where it registers a missing artifact instead, and the fix belongs in `gomain`.

caarlos0 (commented):
> [!NOTE]
> **Obsolete.** This review covered `cedf10fb`. The branch was reduced to the single-main `-o` fix at `3ff4ac49`, and `ensureEllipsisOutputs` — the subject of the inline comment below — no longer exists. Both the finding and the scope recommendation are now resolved. Kept for the record.

Reviewed `cedf10fb` against `main` (merge base `0087e682c`).

### Verified at the time

- go1.27.1 really does ignore the registered extension. `GOEXE` is empty for `js/wasm` and `wasip1/wasm`, and `go build -o <dir> ./cmd/...` writes `app`/`other` (and `app.exe`/`other.exe` for windows). `go test -c -o <dir>` writes `<bin>.test`, and `<bin>.test.exe` for windows.
- All four tests of that revision failed at the merge base and passed at `cedf10fb`.
- The two-phase rename was necessary for what it did: a single `Rename(src, dst)` loop made `TestBuildVariadicWasmCollidingSiblingNames` fail.
- `-buildmode=c-archive`, `c-shared` and `plugin` all refuse two or more main packages.

### Outcome

The inline comment reported that `command: test` with `main: ./...` stopped the build, and orphaned a binary under a `.goreleaser-tmp` name, when a main package had no tests. It also recommended cutting `ensureEllipsisOutputs`, because its two triggers — several mains with wasm, and several mains with `command: test` — were never reported, and because the rules it copied from `cmd/go` already disagreed with go for a `cmd/v2` directory.

`3ff4ac49` takes that route: it keeps only the `len(mains) == 1` exact `-o`, which is what #6898 needs.

_This review was produced by an automated reviewer._

internal/builders/golang/build.go
caarlos0: Fixed in 795ad3acfa21749f116c7fc93c513a03961c510d. The builder now checks all source paths before any rename and omits artifacts when `go test -c` produces no binary. Other filesystem errors still fail the build. The two-step rename remains in place to prevent name collisions.

The regression cases cover mixed packages, no test binaries, and a single main with or without tests. They also check the registered paths and the output directory, so missing artifacts and leftover temporary files are detected. The existing wasm and command-line cases still pass. The `vN` naming issue is unchanged and remains out of scope.

PR #7115, opened 2026-09-06
docs(builder/go): explain that ellipsis discovery uses the host context

### LINKED ISSUE #6899: go: ellipsis discovery ignores the effective build context

caarlos0:
Closes #6899

## What the issue reports

Both halves reproduce on `main`. `gomain.All` builds a `packages.Config` with only `Mode` and `Dir` — no `Env`, no `BuildFlags`:

- `main: ./...` with `tags: [feature]` and a main behind `//go:build feature` fails with `directory does not contain any main function`. Ambient `GOFLAGS=-tags=feature` succeeds.
- Releasing from macOS with `cmd/macos` and `cmd/linuxer`, discovery finds only `./cmd/macos` and builds it for both targets; the linux build then fails with `build constraints exclude all Go files`.

## Why this is documentation and not a code change

I started on the obvious fix — pass the target environment and selection flags into `packages.Load` — and backed it out:

- **Cost.** `gomain.All` runs `packages.Load` with `NeedTypes|NeedSyntax|NeedTypesInfo|NeedDeps`, once per target. Today every call shares one environment, so loads 2..N are nearly free. A per-target environment forces a distinct full type check per `GOOS`/`GOARCH` across the whole matrix.
- **It makes validation target dependent.** `checkBuildElipsis` rejects a config when an ellipsis resolves to more than one main and `binary` or `id` is set. If discovery is target aware, the same config errors on one target and passes on another.
- **It shifts artifact IDs.** Auto-assigned IDs key off the number of main packages found, which would become target dependent.
- **The payoff is small.** Both failures are loud, never silent, and the workaround is one line: name the package. Ellipsis is opt-in and documented as sugar for `go build -o dir . ./cmd/foo ./cmd/bar` — if the mains are conditionally compiled, list them.

So the limit is now stated where users meet it rather than worked around.

## Changes

- A host build context subsection in the ellipsis section of the Go builder page, with the failing config and the fix.
- A short section in the `no-main` error page that links to the Go builder subsection for details and examples.
- `ErrNoMains` now says to set `main` to the package path when the mains are behind build tags or OS constraints, so the error points at the cause.

No behaviour change.

### DIFF
--- a/www/content/customization/builds/builders/go.md
+++ b/www/content/customization/builds/builders/go.md
@@ -567,33 +567,32 @@ Binary names will be inferred just like `go build` does.
 If you have multiple build configurations that change only the `main` and
 `binary` portions, this might be your friend.
 
-> [!WARNING]
-> **Discovery uses the host build context**
->
-> GoReleaser looks for `main` packages using the environment it runs in.
-> It does not apply the `tags` and `flags` of the build, nor the `GOOS` and
-> `GOARCH` of each target.
->
-> Because of that, an ellipsis path does not see a `main` package that is
-> behind a build constraint:
->
-> ```yaml {filename=".goreleaser.yaml"}
-> builds:
->   - main: ./... # cmd/feature/main.go has `//go:build feature`
->     tags: [feature]
-> ```
->
-> This fails with `directory does not contain any main function`, and a
-> `main` package with `//go:build linux` is skipped when you release from
-> macOS.
->
-> Set `main` to the package path in that case, and use one build per package:
->
-> ```yaml {filename=".goreleaser.yaml"}
-> builds:
->   - id: feature
->     main: ./cmd/feature
->     tags: [feature]
-> ```
+### Discovery uses the host build context
+
+GoReleaser looks for `main` packages using the environment it runs in.
+It does not apply the `tags` and `flags` of the build, nor the `GOOS` and
+`GOARCH` of each target.
+
+Because of that, an ellipsis path does not see a `main` package whose build
+constraints are not satisfied in that environment:
+
+```yaml {filename=".goreleaser.yaml"}
+builds:
+  - main: ./... # cmd/feature/main.go has `//go:build feature`
+    tags: [feature]
+```
+
+This fails with `directory does not contain any main function`, and a
+`main` package with `//go:build linux` is skipped when you release from
+macOS.
+
+Set `main` to the package path in that case, and use one build per package:
+
+```yaml {filename=".goreleaser.yaml"}
+builds:
+  - id: feature
+    main: ./cmd/feature
+    tags: [feature]
+```
 
 {{< g_templates >}}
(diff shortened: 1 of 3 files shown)


www/content/customization/builds/builders/go.md
```diff
@@ -567,4 +567,33 @@ Binary names will be inferred just like `go build` does.
...
+>
+> Set `main` to the package path in that case, and use one build per package:
+>
+> ```yaml {filename=".goreleaser.yaml"}
+> builds:
+>   - id: feature
+>     main: ./cmd/feature
+>     tags: [feature]
+> ```
+
```
caarlos0: we should make this a sub-title instead of a warning like that

www/content/resources/errors/no-main.md
```diff
@@ -39,6 +39,30 @@ For more information, check the [builds documentation](/customization/builds/bui
...
+```yaml {filename=".goreleaser.yaml"}
+builds:
+  - id: feature
+    main: ./cmd/feature
+    tags: [feature]
+```
+
+The same applies to a `main` package that only builds on one operating system,
+for example one with `//go:build linux` when you release from macOS.
+
```
caarlos0, line 50: we should link to the sub title of go.md - and here we can keep it a lot shorter

www/content/customization/builds/builders/go.md
caarlos0: Changed the warning to a level-three subsection, "Discovery uses the host build context", under "Building ellipsis paths". The examples remain in that subsection. Fixed in e334e45a8397b1975f789f4ffaefbe8f059928f6.

www/content/resources/errors/no-main.md
caarlos0: Shortened the no-main section to a summary and the package-path workaround. It now links directly to the new Go builder subsection for details and examples, without repeating them. Fixed in e334e45a8397b1975f789f4ffaefbe8f059928f6.

PR #7121, opened 2026-09-08
fix(http): reuse clients for custom TLS uploads

### LINKED ISSUE #6919: http: custom TLS uploads retain one transport per attempt
### Problem

HTTP and Artifactory uploads configured with `trusted_certificates` or client-certificate settings construct a fresh transport for every artifact attempt. [Request execution calls the client factory each time](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/http/http.go#L387-L429), including [retries](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/http/http.go#L345-L364). Those custom transports have no idle timeout or owner cleanup.

### Evidence

Independent local TLS tests exercised both real publishing pipes with three sequential artifacts. Production created three TCP connections and TLS handshakes; one retry per artifact created six. Equivalent shared-client controls used one connection in either case, with matching payloads, authentication, headers, and TLS verification. Custom idle connections remained after publication, context cancellation, and GC, until explicit server cleanup. Empty responses excluded unread response bodies as the cause. Client-certificate and trusted-certificate cases were covered. These are connection/resource observations, not a wall-time benchmark or public-service failure.

### Correction

Share one client per upload configuration, then close only its owned custom transport after all workers finish, including error paths. Do not close the borrowed default client or share credentials across configurations. Add local regressions for connection reuse, retries, retained TLS/authentication behavior, and owned-resource cleanup.

---
Audit priority: **P2**.

vzer200:
Custom TLS uploads currently create a client and transport for every artifact and retry attempt. Keep one client per nonempty upload configuration and close its owned idle connections after all upload workers finish, including error and cancellation paths.

The client is passed through the private request/retry call chain. Each attempt still opens a fresh artifact stream, and each configuration retains its own TLS identity. Cleanup is optional so the borrowed default client stays available to other callers. Client setup errors retain the existing upload-name and kind context and support `errors.Is`.

Closes #6919.

Validation on Windows amd64 with Go 1.27.1:

- Local TLS regressions: three artifacts use one connection instead of three; retries use one instead of six. The tests verify request bodies, method, auth and artifact headers, cleanup after parallel errors/cancellation, default-client ownership, and isolation between two client certificates.
- A separate review caught a setup-error context regression before submission; its regression test passes with the final implementation.
- `go test -race ./internal/http ./internal/pipe/upload ./internal/pipe/artifactory ./internal/retryx -count=1` passes.
- Repository-configured golangci-lint reports zero issues; gofumpt and `git diff --check` pass. The complete application builds.

The connection-reuse tests use empty response bodies. The full repository test suite was not run; validation covers the changed HTTP package, its direct pipeline consumers and retry helper. All upload tests use local `httptest` servers.

AI assistance was used for implementation and a separate automated code review, and the listed validation was run locally. This work is submitted on behalf of the account owner; it does not claim human code review.

### DIFF
--- a/internal/http/http.go
+++ b/internal/http/http.go
@@ -260,25 +260,32 @@ func uploadWithFilter(ctx *context.Context, upload *config.Upload, filter artifa
 
 	if len(artifacts) == 0 {
 		log.Info("no artifacts found")
+		return nil
 	}
 	log.Debugf("will upload %d artifacts", len(artifacts))
+	client, cleanup, err := getHTTPClient(upload)
+	if err != nil {
+		return fmt.Errorf("%s: %s: upload failed: %w", upload.Name, kind, err)
+	}
+	if cleanup != nil {
+		defer cleanup()
+	}
+
 	g := semerrgroup.New(ctx.Parallelism)
 	for _, artifact := range artifacts {
 		g.Go(func() error {
-			return uploadAsset(ctx, upload, artifact, kind, check)
+			return uploadAsset(ctx, client, upload, artifact, kind, check)
 		})
 	}
 	if err := g.Wait(); err != nil {
 		return err
 	}
-	if len(artifacts) > 0 {
-		summary.Appendf("Uploaded %d files to `%s`", len(artifacts), upload.Name)
-	}
+	summary.Appendf("Uploaded %d files to `%s`", len(artifacts), upload.Name)
 	return nil
 }
 
 // uploadAsset uploads file to target and logs all actions.
-func uploadAsset(ctx *context.Context, upload *config.Upload, artifact *artifact.Artifact, kind string, check ResponseChecker) error {
+func uploadAsset(ctx *context.Context, client *h.Client, upload *config.Upload, artifact *artifact.Artifact, kind string, check ResponseChecker) error {
 	// username and secret are optional since the server may not support/need
 	// basic authentication always
 	username, err := getUsername(ctx, upload, kind)
@@ -329,7 +336,7 @@ func uploadAsset(ctx *context.Context, upload *config.Upload, artifact *artifact
 		WithField("file", artifact.Name).
 		Info("uploading")
 
-	res, err := uploadAssetToServer(ctx, upload, targetURL, username, secret, headers, kind, artifact, check)
+	res, err := uploadAssetToServer(ctx, client, upload, targetURL, username, secret, headers, kind, artifact, check)
 	if err != nil {
 		return fmt.Errorf("%s: %s: upload failed: %w", upload.Name, kind, err)
 	}
@@ -366,7 +373,7 @@ func appendEscapedName(base, name string) string {
 }
 
 // uploadAssetToServer uploads the asset file to target.
-func uploadAssetToServer(ctx *context.Context, upload *config.Upload, target, username, secret string, headers map[string]string, kind string, artifact *artifact.Artifact, check ResponseChecker) (*h.Response, error) {
+func uploadAssetToServer(ctx *context.Context, client *h.Client, upload *config.Upload, target, username, secret string, headers map[string]string, kind string, artifact *artifact.Artifact, check ResponseChecker) (*h.Response, error) {
 	var resp *h.Response
 	err := retryx.Do(ctx, ctx.Config.Retry, func() error {
 		a, err := assetOpen(kind, artifact)
@@ -380,7 +387,7 @@ func uploadAssetToServer(ctx *context.Context, upload *config.Upload, target, us
 			return retryx.Unrecoverable(err)
 		}
 
-		resp, err = executeHTTPRequest(ctx, upload, req, check) //nolint:bodyclose // closed by caller (uploadAsset)
+		resp, err = executeHTTPRequest(ctx, client, req, check) //nolint:bodyclose // closed by caller (uploadAsset)
 		if err != nil {
 			return retryx.HTTP(err, resp)
 		}
@@ -408,9 +415,10 @@ func newUploadRequest(ctx *context.Context, method, target, username, secret str
 	return req, err
 }
 
-func getHTTPClient(upload *config.Upload) (*h.Client, error) {
+func getHTTPClient(upload *config.Upload) (*h.Client, func(), error) {
 	if upload.TrustedCerts == "" && upload.ClientX509Cert == "" && upload.ClientX509Key == "" {
-		return h.DefaultClient, nil
+		// The default client is shared with other callers and must not be closed.
+		return h.DefaultClient, nil, nil
 	}
 	transport := &h.Transport{
 		Proxy:           h.ProxyFromEnvironment,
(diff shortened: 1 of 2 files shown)


internal/http/http.go
```diff
@@ -408,9 +415,10 @@ func newUploadRequest(ctx *context.Context, method, target, username, secret str
 	return req, err
 }
 
-func getHTTPClient(upload *config.Upload) (*h.Client, error) {
+func getHTTPClient(upload *config.Upload) (*h.Client, func(), error) {
```
caarlos0, line 418: Non-blocking simplification. The `func()` return value carries only one bit of information: the caller owns the client, or borrows it. `http.Client` already has `CloseIdleConnections()`.

Smallest change: keep the old two-value signature and let the caller decide:

```go
client, err := getHTTPClient(upload)
if err != nil {
	return fmt.Errorf("%s: %s: upload failed: %w", upload.Name, kind, err)
}
if client != h.DefaultClient {
	// only close the connections we own.
	defer client.CloseIdleConnections()
}
```

This removes one return value, the two `return nil, nil, err` lines, and the `if cleanup != nil` guard at line 270. No new test is necessary; the existing connection-count tests cover it. If you make this change, the assertion in the comment on `http_client_test.go:191` becomes `require.Same(t, h.DefaultClient, client)` only.

internal/http/http.go
```diff
@@ -408,9 +415,10 @@ func newUploadRequest(ctx *context.Context, method, target, username, secret str
...
 	if upload.TrustedCerts == "" && upload.ClientX509Cert == "" && upload.ClientX509Key == "" {
-		return h.DefaultClient, nil
+		// The default client is shared with other callers and must not be closed.
+		return h.DefaultClient, nil, nil
 	}
 	transport := &h.Transport{
```
caarlos0, line 423: Question, not a defect, and not a merge condition. The transport does not set `MaxIdleConnsPerHost`, so it uses `http.DefaultMaxIdleConnsPerHost`, which is 2 (`net/http/transport.go:62`). GoReleaser sets `ctx.Parallelism` to `GOMAXPROCS`, which is usually 8 or more. When more than two workers release a connection at the same moment, the extra connections close instead of going back into the pool, and the next artifact makes a new TLS handshake.

I could not measure the size of this effect, and my two independent checks disagreed about it. `tryPutIdleConn` gives a released connection directly to a worker that is already waiting, before it applies the cap, so in a steady pipeline most of the reuse can still occur. The unconditional wins of this PR (one `x509.SystemCertPool()` clone, one `LoadX509KeyPair`, and an owner for the idle connections) do not depend on the cap.

If you want the reuse that the description promises at the default parallelism, add `MaxIdleConnsPerHost: <parallelism>` here. Do not clone `http.DefaultTransport` instead: it sets `ForceAttemptHTTP2`, which would multiplex the requests and make every connection count in the new tests meaningless. Test: run `TestUploadReusesTLSClient` with `Parallelism = 4` and 8 artifacts, and assert `created <= 4`.

internal/http/http_client_test.go
```diff
@@ -0,0 +1,348 @@
...
+		w.WriteHeader(http.StatusCreated)
+	}))
+	srv.Config.ConnState = connections.record
+	srv.StartTLS()
+	t.Cleanup(srv.Close)
+	borrowed := srv.Client()
+	transport := &uploadCloseTracker{RoundTripper: borrowed.Transport}
+	borrowed.Transport = transport
+	previous := http.DefaultClient
+	http.DefaultClient = borrowed
```
caarlos0, line 191: This test writes to the process-global `http.DefaultClient`. It is safe today only because no test in this package calls `t.Parallel`. When a person adds `t.Parallel` later, this becomes a real data race against the read at `http.go:421`, and the failure will be far from this file.

The test plus the `uploadCloseTracker` type at the end of the file is about 40 lines, and it guards one `nil` return. You can assert the same contract directly, with no global state and no server:

```go
func TestBorrowedDefaultClient(t *testing.T) {
	client, cleanup, err := getHTTPClient(&config.Upload{})
	require.NoError(t, err)
	require.Same(t, h.DefaultClient, client)
	require.Nil(t, cleanup)
}
```

That removes the global write, the tracker type, and the TLS server. See also the comment on `http.go:418`: if you take that simplification, keep only the `require.Same` line.

internal/http/http_client_test.go
```diff
@@ -0,0 +1,348 @@
...
+			requestMu.Unlock()
+			require.Equal(t, want, got)
+			created, _, _ := connections.counts()
+			require.Equal(t, 1, created, "all artifacts and attempts should share a connection")
+			connections.requireClosed(t)
+		})
+	}
+}
+
+func TestUploadTLSClientCleanupWaitsForWorkers(t *testing.T) {
```
caarlos0, line 99: This test is about 70 lines, with a three-worker barrier, a blocking handler, a cancellation branch, and mid-flight polling of the connection state. What it proves is that a `defer` written after `g.Wait()` runs after `g.Wait()`. Go guarantees this. The only regression it can find is a person who moves `cleanup()` above `g.Wait()`, and `connections.requireClosed` in `TestUploadReusesTLSClient` already fails in that case, because a connection would still be open at the end.

There is also a hidden coupling at line 148: the exact assertion `idle == 2` is the same value as `http.DefaultMaxIdleConnsPerHost`. If a person later raises the artifact count in `uploadClientTestContext` or the parallelism, the third released connection closes instead of becoming idle, `open == 3` never becomes true, and the test fails after a 5-second `Eventually` with a message that does not name the cause.

Suggestion: delete this test. If you want to keep the protection against an early cleanup, keep only the `cancel=false` branch, or assert `idle >= 1 && open == created` instead of the exact counts.

caarlos0 (approved):
Approving. I found no correctness defect. The change is correct: `errgroup.Wait` joins all workers, so the deferred cleanup cannot run while a worker still holds the client, on the error path and on the cancellation path. The no-artifacts early return is equivalent to the old code, because the old `if len(artifacts) > 0` guard already suppressed the summary. The client-setup error keeps the same message shape as `uploadAsset`, so the user sees no new text. Sharing one `*http.Client` between the workers is safe.

The strongest justification is the resource leak: the old per-request transports had no `IdleConnTimeout` and no owner, so each pooled connection stayed open until the process stopped. This change gives them an owner.

The comments below are all non-blocking. They are about the size of the new test file, which is 348 lines for a 22-line production change, and one signature simplification. Please do not treat them as merge conditions.

Verification gaps I must disclose:
- I did not build and did not run the tests. I was told to assume the build is green. All statements above come from reading the diff, the repository, and the Go standard-library sources.
- I confirmed that the connection-count assertions are deterministic. Responses with `Content-Length: 0` put the connection back in the idle pool before the response reaches the caller (`net/http/transport.go`, the `hasBody` branch). HTTP/2 cannot occur, because `TLSClientConfig` is set and `httptest` does not enable it. So I make no flakiness claim about the counts.
- I could not measure the real reduction of TLS handshakes. See the comment about `MaxIdleConnsPerHost`.

Disclosure: this review was written by an automated agent (GitHub Copilot CLI), not by a human reviewer.

PR #7124, opened 2026-09-08
fix(git): tag not in the local repository is not an error

### LINKED ISSUE #7122: couldn't get tag contents: unexpected git tag output for "<my version>": ""
Please see the bug report here: https://github.com/osac-project/osac/pull/786
Changes introduced in #6850 

I'm facing the same after upgrading to goreleaser v2.18.1

My version number looks like this: v1.3.140-pullrequest2225080.6

I'm currently working around by pinning goreleaser v2.18.0, but I'd prefer to use latest

caarlos0:
A tag that is not in the local repository no longer fails the run.

`getTagContents` runs `git tag -l <tag>`, which prints nothing and exits 0 when the tag is not in the local repository. The output shape check added in #6850 turned that empty output into an error:

```
⨯ release failed after 0s   error=couldn't get tag contents: unexpected git tag output for "v0.17.0": ""
```

Users hit this with `GORELEASER_CURRENT_TAG` set to a tag that the checkout never fetched. Before #6850, the three separate `git tag -l` calls each returned an empty string in that case, so the subject, contents and body were empty and the release continued.

This restores that behavior: empty output gives empty strings. Output that is not empty and does not have the three fields is still an error.

`TestTagFromCINotInRepository` sets `GORELEASER_CURRENT_TAG` to a tag that is not in the repository. With the fix removed, it fails with the exact error from the issue.

Closes #7122

### DIFF
--- a/internal/pipe/git/git.go
+++ b/internal/pipe/git/git.go
@@ -259,6 +259,11 @@ func getTagContents(ctx *context.Context, tag string) (subject, contents, body s
 	if err != nil {
 		return "", "", "", err
 	}
+	if out == "" {
+		// the tag is not in the local repository, e.g. when
+		// GORELEASER_CURRENT_TAG names a tag that was never fetched.
+		return "", "", "", nil
+	}
 	parts := strings.Split(out, "\x00")
 	if len(parts) != 3 {
 		return "", "", "", fmt.Errorf("unexpected git tag output for %q: %q", tag, out)
--- a/internal/pipe/git/git_test.go
+++ b/internal/pipe/git/git_test.go
@@ -458,6 +458,22 @@ func TestTagFromCINotInRepository(t *testing.T) {
 	require.Empty(t, ctx.Git.TagBody)
 }
 
+func TestTagFromCINotInRepositoryValidates(t *testing.T) {
+	testlib.Mktmp(t)
+	testlib.GitInit(t)
+	testlib.GitRemoteAdd(t, "git@github.com:foo/bar.git")
+	testlib.GitCommit(t, "commit1")
+	testlib.GitTag(t, "v0.0.1")
+	t.Setenv("GORELEASER_CURRENT_TAG", "v0.0.2")
+
+	ctx := testctx.Wrap(t.Context())
+	err := Pipe{}.Run(ctx)
+	var wrongRef ErrWrongRef
+	require.ErrorAs(t, err, &wrongRef)
+	require.Equal(t, "v0.0.2", wrongRef.tag)
+	require.Equal(t, ctx.Git.Commit, wrongRef.commit)
+}
+
 func TestNoPreviousTag(t *testing.T) {
 	testlib.Mktmp(t)
 	testlib.GitInit(t)
--- a/internal/pipe/git/git_test.go
+++ b/internal/pipe/git/git_test.go
@@ -442,6 +442,22 @@ func TestTagFromCI(t *testing.T) {
 	}
 }
 
+func TestTagFromCINotInRepository(t *testing.T) {
+	testlib.Mktmp(t)
+	testlib.GitInit(t)
+	testlib.GitRemoteAdd(t, "git@github.com:foo/bar.git")
+	testlib.GitCommit(t, "commit1")
+	testlib.GitTag(t, "v0.0.1")
+	t.Setenv("GORELEASER_CURRENT_TAG", "v0.0.2")
+
+	ctx := testctx.Wrap(t.Context(), testctx.Skip(skips.Validate))
+	testlib.AssertSkipped(t, Pipe{}.Run(ctx))
+	require.Equal(t, "v0.0.2", ctx.Git.CurrentTag)
+	require.Empty(t, ctx.Git.TagSubject)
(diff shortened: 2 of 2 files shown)


caarlos0 (commented):
Reviewed by a bot (GitHub Copilot CLI), at the request of the repository owner.

**No blocking findings.** The fix is correct and minimal, and it restores exactly the behaviour that existed before #6850.

I verified the premise independently rather than accepting the description:

- `git tag -l --format=... <missing-tag>` writes **zero bytes** and exits 0, so `internal/git.Run` gives back exactly `""` and the guard is precise.
- The guard cannot misfire on a tag that exists. A lightweight tag on a commit with an empty message gives `\0\0\n`, thus three parts, and an annotated tag with an empty body gives three parts too.
- Real tag names can never contain `*`, `?` or `[`, because `git check-ref-format` refuses them. The `len(parts) != 3` branch is therefore only reachable through a hand-written glob in `GORELEASER_CURRENT_TAG`, where the run fails loudly with the pattern named in the message. Keep the branch.
- `TagSubject`, `TagContents` and `TagBody` are consumed only as template fields (`internal/tmpl/tmpl.go:143-145`). No pipe reads them for the changelog or the release notes, so an empty value cannot make a silently wrong release.
- Without `--skip=validate`, a tag that is not in the repository still stops the run with `ErrWrongRef`. The loud path stays for everyone who did not opt out.

Two comments below: one on test coverage, and one on a separate defect of the same class that this PR correctly does not try to fix.

internal/pipe/git/git.go
```diff
@@ -259,6 +259,11 @@ func getTagContents(ctx *context.Context, tag string) (subject, contents, body s
 	if err != nil {
 		return "", "", "", err
 	}
+	if out == "" {
```
caarlos0, line 262: **Follow-up, not for this PR: `column.ui` breaks the same merged format.**

I anchor this here because it is the same output-shape assumption that this line repairs, and it comes from the same commit (#6850). It is a different trigger, so it needs its own change.

`git tag -l` obeys `column.ui`. With columns on, git removes the `%00` separators and pads with spaces, so `out` is not empty, `len(parts)` is 1, and the `len(parts) != 3` branch below fires for a completely normal annotated tag:

```
⨯ error=couldn't get tag contents: unexpected git tag output for "v1": "subject here     body here      \n"
```

Trigger: `column.ui = always` in the user's `~/.gitconfig`, in the repository `.git/config`, or in the system gitconfig. The global file counts: `RunWithEnv` does `cmd.Env = append(cmd.Env, env...)` with an empty `env`, which leaves `cmd.Env` as `nil`, and `os/exec` then gives the child the full environment, `HOME` included. I confirmed this with a small Go program and with `HOME=<tmp> git tag -l --format=...`.

Impact: the release fails. A binary built from `4181c4f2b^` succeeds on the same repository, so #6850 changed this from working to a hard failure.

Smallest fix, one line in `internal/git/git.go:25-27`, beside the flag that is already forced there:

```go
extraArgs := []string{
	"-c", "log.showSignature=false",
	"-c", "column.ui=never",
}
```

The same configuration also joins the lines of `gitTagsPointingAt` (`internal/pipe/git/git.go:311`), so `CurrentTag` becomes `"v1.0.0-rc1  v1.0.0"` when two tags point at HEAD. That one is older than #6850.

Test needed: a case in `internal/pipe/git` that runs `git config column.ui always` in the fixture repository before `Pipe{}.Run`, and asserts `TagSubject`, `TagBody` and `CurrentTag`.

internal/pipe/git/git_test.go
```diff
@@ -442,6 +442,22 @@ func TestTagFromCI(t *testing.T) {
...
 
+func TestTagFromCINotInRepository(t *testing.T) {
+	testlib.Mktmp(t)
+	testlib.GitInit(t)
+	testlib.GitRemoteAdd(t, "git@github.com:foo/bar.git")
+	testlib.GitCommit(t, "commit1")
+	testlib.GitTag(t, "v0.0.1")
+	t.Setenv("GORELEASER_CURRENT_TAG", "v0.0.2")
+
+	ctx := testctx.Wrap(t.Context(), testctx.Skip(skips.Validate))
```
caarlos0, line 453: **Missing coverage, non-blocking: nothing tests `ErrWrongRef`.**

`skips.Validate` here is load-bearing and correct, because it mirrors the invocation in the issue. But it also means this test covers only the quiet half of the new behaviour.

After this change, a tag that is not in the repository no longer stops `getTagContents`. The only thing that still catches a misspelled `GORELEASER_CURRENT_TAG` is `validate` at `internal/pipe/git/git.go:186-193`, and `grep -rn ErrWrongRef --include=*.go .` finds no test anywhere in the repository. So the guard that this PR now depends on can be deleted or broken later and every test still passes.

I confirmed the guard works today: in a repository that holds `v0.0.1` only, `git describe --exact-match --tags --match v0.0.2` exits 128, so `validate` returns `ErrWrongRef`.

Test needed: the same fixture as this test, without `testctx.Skip(skips.Validate)`, asserting that `Pipe{}.Run` returns `ErrWrongRef`. Four lines, and it pins that an empty tag message does not make an absent tag acceptable.

PR #7127, opened 2026-09-09
fix(git): force column.ui=never so tag output is not columnized

### LINKED ISSUE #7125: git: column.ui breaks git tag output parsing
# Summary

`git tag` puts its output in columns when the user has `column.ui` or `column.tag` set to `always`. GoReleaser parses that output, so a release fails or gets a wrong tag. GoReleaser does not disable the setting.

This is separate from #7122, but it is the same class of problem: the output of `git tag` is not what the parser expects.

# The two calls that break

Both are in `internal/pipe/git/git.go`. They are the only calls in the code that `column` changes. I checked `rev-parse`, `show`, `rev-list`, `describe`, `status --porcelain` and `ls-remote --get-url`: `git` does not put those in columns.

## 1. `getTagContents` (git.go:258) — the release stops

The format holds three fields separated by NUL. Columns remove the NUL bytes and add spaces, thus the split gives 1 part and not 3:

```console
$ git config column.ui always
$ git tag -l --format='%(contents:subject)%00%(contents)%00%(contents:body)' v1 | xxd | head -2
00000000: 7375 626a 6563 7420 6865 7265 2020 2020  subject here
00000010: 2020 2020 2020 2020 2020 2020 626f 6479              body

$ git -c column.ui=never tag -l --format='%(contents:subject)%00%(contents)%00%(contents:body)' v1 | xxd | head -2
00000000: 7375 626a 6563 7420 6865 7265 0073 7562  subject here.sub
00000010: 6a65 6374 2068 6572 650a 0a62 6f64 7920  ject here..body
```

The user sees:

```
⨯ release failed after 0s   error=couldn't get tag contents: unexpected git tag output for "v1": "subject here     body here      \n"
```

The tag is normal and is in the repository. A binary from `4181c4f2b^` (before #6850) does the release. So #6850 made this a hard failure.

## 2. `gitTagsPointingAt` (git.go:311) — the tag is wrong

With two tags on `HEAD`, the columns put them on one line, and `CleanAllLines` gives one item:

```console
$ git config column.ui always
$ git tag --points-at HEAD --sort -version:refname
v1.0.0-rc1  v1.0.0

$ git -c column.ui=never tag --points-at HEAD --sort -version:refname
v1.0.0-rc1
v1.0.0
```

`CurrentTag` becomes `"v1.0.0-rc1  v1.0.0"`. `validate` stops this with `ErrWrongRef`, but not with `--skip=validate`. This problem is older than #6850.

# Why the user configuration reaches GoReleaser

`RunWithEnv` (`internal/git/git.go:24-37`) does:

```go
cmd.Env = append(cmd.Env, env...)
```

`env` is empty for every `Run`, so `cmd.Env` stays `nil`, and `os/exec` gives the child the full environment. `HOME` is thus set and `git` reads `~/.gitconfig`. The repository `.git/config` and the system gitconfig also do it.

`column.ui = auto` is safe, because the output goes to a pipe.

# Suggested fix

One line, beside the flag that is already forced there. `internal/git/git.go:25-27`:

```go
extraArgs := []string{
	"-c", "log.showSignature=false",
	"-c", "column.ui=never",
}
```

I checked that this repairs both calls, and that it also beats an explicit `column.tag = always` in the repository configuration.

`git for-each-ref` is immune to `column` and is the correct plumbing command for both jobs, but that is a bigger change. `-c column.ui=never` protects every `git` call that GoReleaser makes now or later, thus I prefer it.

# Tests

- `internal/pipe/git`: run `git config column.ui always` in the fixture repository, then assert `TagSubject` and `TagBody` for an annotated tag, and `CurrentTag` for two tags on `HEAD`.

# Environment

- `git` 2.55.0, macOS. I did not check Windows or `git` before 2.30. `column` is old, so I do not expect a difference.

KR-Ravindra:
## Problem

When the user has `column.ui = always` (or `column.tag = always`) in their git configuration, `git tag` prints its output in columns even when stdout is a pipe. GoReleaser parses that output, so:

- `getTagContents` fails hard with `couldn't get tag contents: unexpected git tag output for "v1": ...` because the NUL separators in `--format=%(contents:subject)%00%(contents)%00%(contents:body)` are replaced by column padding and the split yields 1 part instead of 3 (a hard failure since #6850).
- `gitTagsPointingAt` returns all tags pointing at `HEAD` on a single line, so with two tags on `HEAD` `CurrentTag` becomes e.g. `"v1.0.0-rc1  v1.0.0"`.

## Root cause

`internal/git/git.go:24-27` (`RunWithEnv`) forces `-c log.showSignature=false` but nothing for `column.*`. Since `cmd.Env` stays `nil` when no extra env is passed, the child `git` inherits the full environment and reads the user's `~/.gitconfig` (and the repository and system configs), so a `column.ui = always` there reaches every `git tag` invocation.

Affected callers: `internal/pipe/git/git.go` `getTagContents` (`git tag -l --format=...`) and `gitTagsPointingAt` (`git tag --points-at ...`).

## Fix

Add `-c column.ui=never` next to the existing `-c log.showSignature=false` in `RunWithEnv`, as suggested in the issue. `-c` is applied after all config files, so it also overrides a more specific `column.tag = always` set in the repository config (verified locally). This protects every `git` call GoReleaser makes now or later, not just the two known ones.

## How tested

Added `TestGitColumnUI` in `internal/git` and `TestAnnotatedTagsWithColumnUI` / `TestTagSortOrderWithColumnUI` in `internal/pipe/git`. Each sets `column.ui always` in the fixture repository's config (not the global one).

Before the fix (git 2.43.0, Linux):

```
--- FAIL: TestGitColumnUI (0.02s)
        	Error:      	elements differ
        	            	extra elements in list B:
        	            	 (string) (len=14) "nightly  1.2.3"
--- FAIL: TestAnnotatedTagsWithColumnUI (0.05s)
        	Error:      	Received unexpected error:
        	            	couldn't get tag contents: unexpected git tag output for "v0.0.1": "first version  lalalla        lah            lalal          \n               lalal                         lah\n"
--- FAIL: TestTagSortOrderWithColumnUI (0.07s)
        	Error:      	Received unexpected error:
        	            	git tag v0.0.2  v0.0.1 was not made against commit f454a831d7daa63ca258d6f6df329727c9022a26
```

After the fix:

```
--- PASS: TestGitColumnUI (0.02s)
--- PASS: TestAnnotatedTagsWithColumnUI (0.04s)
--- PASS: TestTagSortOrderWithColumnUI (0.04s)
ok  	github.com/goreleaser/goreleaser/v2/internal/git	1.179s
ok  	github.com/goreleaser/goreleaser/v2/internal/pipe/git	2.134s
ok  	github.com/goreleaser/goreleaser/v2/internal/testlib	1.066s
```

`gofumpt -l`, `go vet` and `golangci-lint run --tests` on the changed packages report no issues; `go build ./...` passes.

## Links

- Closes https://github.com/goreleaser/goreleaser/issues/7125
- Related: #6850 (made the tag-contents parse a hard failure), #7122 / #7124 (same class of problem, different cause)

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -24,6 +24,7 @@ func IsRepo(ctx context.Context) bool {
 func RunWithEnv(ctx context.Context, env []string, args ...string) (string, error) {
 	extraArgs := []string{
 		"-c", "log.showSignature=false",
+		"-c", "column.ui=never",
 	}
 	args = append(extraArgs, args...)
 	/* #nosec */
--- a/internal/git/git_test.go
+++ b/internal/git/git_test.go
@@ -70,3 +70,22 @@ func TestClean(t *testing.T) {
 		require.Empty(t, out)
 	})
 }
+
+func TestGitColumnUI(t *testing.T) {
+	ctx := t.Context()
+	testlib.Mktmp(t)
+	testlib.GitInit(t)
+	testlib.GitCommit(t, "foo")
+	testlib.GitTag(t, "1.2.3")
+	testlib.GitTag(t, "nightly")
+	_, err := git.Run(ctx, "config", "column.ui", "always")
+	require.NoError(t, err)
+
+	tags, err := git.CleanAllLines(git.Run(ctx, "tag", "--points-at", "HEAD", "--sort", "-version:refname"))
+	require.NoError(t, err)
+	require.ElementsMatch(t, []string{"1.2.3", "nightly"}, tags)
+
+	out, err := git.Run(ctx, "tag", "-l", "--format=%(refname:short)%00%(objecttype)", "1.2.3")
+	require.NoError(t, err)
+	require.Equal(t, "1.2.3\x00commit\n", out)
+}
--- a/internal/pipe/git/git_test.go
+++ b/internal/pipe/git/git_test.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/caarlos0/log"
+	"github.com/goreleaser/goreleaser/v2/internal/git"
 	"github.com/goreleaser/goreleaser/v2/internal/skips"
 	"github.com/goreleaser/goreleaser/v2/internal/testctx"
 	"github.com/goreleaser/goreleaser/v2/internal/testlib"
@@ -119,6 +120,40 @@ func TestAnnotatedTagsWithApostrophes(t *testing.T) {
 	require.Equal(t, "don't break it", ctx.Git.TagSubject)
 	require.Equal(t, "don't break it\n\nit's the user's message", ctx.Git.TagContents)
 	require.Equal(t, "it's the user's message", ctx.Git.TagBody)
+}
+
+func TestAnnotatedTagsWithColumnUI(t *testing.T) {
+	testlib.Mktmp(t)
+	testlib.GitInit(t)
+	testlib.GitRemoteAdd(t, "git@github.com:foo/bar.git")
+	testlib.GitCommit(t, "commit1")
+	testlib.GitAnnotatedTag(t, "v0.0.1", "first version\n\nlalalla\nlalal\nlah")
+	_, err := git.Run(t.Context(), "config", "column.ui", "always")
+	require.NoError(t, err)
+	ctx := testctx.Wrap(t.Context())
+	require.NoError(t, Pipe{}.Run(ctx))
+	require.Equal(t, "v0.0.1", ctx.Git.CurrentTag)
+	require.Equal(t, "first version", ctx.Git.TagSubject)
+	require.Equal(t, "first version\n\nlalalla\nlalal\nlah", ctx.Git.TagContents)
+	require.Equal(t, "lalalla\nlalal\nlah", ctx.Git.TagBody)
+}
+
+func TestTagSortOrderWithColumnUI(t *testing.T) {
+	testlib.Mktmp(t)
+	testlib.GitInit(t)
+	testlib.GitRemoteAdd(t, "git@github.com:foo/bar.git")
+	testlib.GitCommit(t, "commit1")
+	testlib.GitTag(t, "v0.0.2")
+	testlib.GitTag(t, "v0.0.1")
+	_, err := git.Run(t.Context(), "config", "column.ui", "always")
+	require.NoError(t, err)
+	ctx := testctx.WrapWithCfg(t.Context(), config.Project{
+		Git: config.Git{
+			TagSort: "-version:refname",
+		},
+	})
+	require.NoError(t, Pipe{}.Run(ctx))
+	require.Equal(t, "v0.0.2", ctx.Git.CurrentTag)
 }
 
 func TestBranch(t *testing.T) {


KR-Ravindra:
Round 1 self-review.

1. `-c` values are applied after every config file, so `-c column.ui=never` also overrides a repository-level `column.tag=always`; confirmed on git 2.50: with `column.tag=always` set, `git -c column.ui=never tag --points-at HEAD` prints one tag per line.
2. `go test ./internal/git/ ./internal/pipe/git/` passes at 4fe4e2f; the three new tests fail without the change with the outputs quoted in the description.
3. No other PR references #7125. Marking ready.

caarlos0 (approved):
**Approve.** I found no defect. This review is from an automated reviewer (a bot).

### The production change is correct

`-c column.ui=never` also overrides a more specific `column.tag = always`. This is not obvious, so I give the evidence here for the record. Git writes `column.ui` and `column.tag` into the same `colopts` variable, and git reads the command-line config layer last. Thus the last value wins, and the value from `-c` is always the last one. I made sure of this with git 2.55: with `column.tag = always` in the repository config, `git tag -l --format='%(refname:short)%00%(objecttype)' 1.2.3` gives `1.2.3` and drops the NUL byte and all data after it. With `-c column.ui=never`, the same command gives `1.2.3\0commit`. An additional `-c column.tag=never` is thus not necessary.

The flag is inert for users who have no column config, because `never` is the git default on a pipe. Of the commands that GoReleaser parses, only `git tag` can make columns. `status --porcelain`, `rev-parse`, `show`, `rev-list`, `describe` and `ls-remote` are not sensitive to `column.*`. `internal/git/git.go` holds the only `exec` of git in non-test code, thus there is no path that avoids the fix.

The position of the flag is also correct: `extraArgs` goes before `args`, and `args[0]` is always the subcommand. The `-c versionsort.suffix=` that `gitTagsPointingAt` adds stays between the extras and the subcommand, thus it stays valid.

### The tests are true regression tests

All three tests fail without the one-line change, at all terminal widths. This is important, because column layout is sensitive to the width: with a narrow terminal, `git tag --points-at` gives one tag per line even with `column.ui = always`. Each test has an assertion that does not depend on the width, because column rendering cuts the output at the first NUL byte at every width:

- `TestGitColumnUI` fails on the `%00` assertion.
- Both pipe tests fail in `Pipe{}.Run`, because `getTagContents` loses its NUL bytes. This includes `TestTagSortOrderWithColumnUI`, thus that test does not depend on the two tags being on one line.

The tests write `column.ui` into the repository config in a per-test temporary directory, thus there is no global state that can leak. The absence of `t.Parallel()` is correct, because `testlib.Mktmp` uses `tb.Chdir`.

### Verification gaps

- I did not run a build or the full test suite; I was told to assume that the build is green.
- The git behavior above comes from experiments with git 2.55 on macOS. The three new tests were executed with the one-line change temporarily removed and then restored, at `COLUMNS` unset, 10 and 200. I did not test older git versions or Windows. The risk is low: `column.ui` exists since git 1.7.11, and git ignores an unknown `-c` key.

PR #7134, opened 2026-09-13
fix(blob): stream unencrypted uploads instead of buffering files

### LINKED ISSUE #6920: Blob uploads retain full artifact buffers for each destination
Publish four 2 MiB files without KMS encryption, using explicit pipe parallelism 4. Compare one blob destination with four destinations, each receiving those same four files.

A controlled local blob driver held uploads after their file data was read. One destination had four active uploads retaining 8 MiB of file buffers; four destinations had sixteen active uploads retaining 32 MiB. Measured heap increases after GC were 8,401,936 and 33,697,712 bytes respectively. The run used Go 1.27.1 with GOMAXPROCS 1 and explicitly selected pipe parallelism 4. No cloud upload was made. These are retained-buffer and heap observations, not an OOM or throughput claim.

[Configuration concurrency](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/blob/blob.go#L42-L55) combines with per-destination artifact workers, and [getData](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/blob/upload.go#L231-L234) reads each complete file again for every destination.

Stream unencrypted files into the blob writer rather than retaining whole-file byte slices. Preserve content, writer-close error handling, and the existing KMS format. Add a bounded multi-destination allocation regression and byte-for-byte output controls before making any performance claim beyond this measured workload.

---
Audit priority: **P2**.

raviayadav:
Closes #6920.

Every blob upload called `getData`, which did `os.ReadFile` on the whole file, and then passed the `[]byte` to the writer. `Publish` runs destinations concurrently, and each destination runs its own upload workers, so memory held at once grew with file size × parallelism × destinations.

Unencrypted files are now opened and copied into the gocloud `blob.Writer` with `io.Copy`. `uploader.Upload` takes an `io.Reader` instead of `[]byte`. KMS-encrypted files still go through `getData`, because `keeper.Encrypt` needs the whole plaintext in one call and cannot be streamed without changing the encryption format — and the providers cap it anyway (4,096 bytes for `awskms`), so that path only ever carries tiny files. Its size check is unchanged; `getData` only lost the non-KMS branch that `openData` made unreachable.

Content type detection still works: `blob.Writer` sniffs the first 512 bytes whether they arrive in one write or several. The opened file is closed on every path.

Because the file is now read inside the writer's lifetime, a local read error mid-upload became a new way to leave a truncated object in the bucket: `blob.Writer.Close` commits and has no abort path, so the previous `defer w.Close()` would finalize a partial upload. `Upload` now cancels the writer context before closing on the error path, which aborts the write and also removes the double `Close()` on the success path.

### What the win actually is

Allocation becomes O(1) in artifact size rather than O(size). It is not free: gocloud's `readFromWriteTo` uses a flat 1 MiB buffer per upload, so the gain scales with how large the artifacts are. Measured per upload through `productionUploader` against `fileblob`:

| file size | before | after |
|---|---|---|
| 2 MiB | 2.036 MiB | 1.027 MiB |
| 64 MiB | 64.044 MiB | 1.027 MiB |

Peak concurrency is `parallelism × parallelism` (destinations × artifacts), so worst-case live buffering goes from `parallelism² × filesize` to `parallelism² × 1 MiB`.

Tests:

- `TestPublishAllocationDoesNotScaleWithFileSize` runs `Publish` twice, with four destinations and four files, at 1 MiB and at 4 MiB, and asserts the total does not grow with the file size. It goes through the `file://` provider, so `newProductionUploader`, `urlFor`, `Open`, the real `blob.Writer` and `Close` all run. Streaming: 17,845,656 -> 17,708,200 bytes (flat). Buffering: 17,845,656 -> 68,149,912 bytes (scales), which is how it fails without this change.
- `TestProductionUploaderUploadsFileContentsAndDetectsContentType` uploads empty, small (under the 512-byte sniff length), and multi-MiB files into a `memblob` bucket and compares them byte for byte, along with the detected content type.
- `TestProductionUploaderDoesNotCommitPartialUploads` fails a read part way through, at 100 and 100,000 bytes so both sides of the sniff threshold are covered, and asserts the error surfaces and no object is left behind.
- `TestUploadDataAWSKMSEncryptsBeforeUploading` runs the KMS path through `uploadData` into a `memblob` bucket and asserts the ciphertext — never the plaintext — is what lands there, and that an oversize file uploads nothing.
- `TestUploadDataMissingFile` keeps the `failed to open file` error.

Package coverage 76.1% -> 86.9%. MinIO tests pass locally.

AI disclosure: I used an AI assistant (Claude) to help write and test this change. I reviewed all of it and understand it.


caarlos0 (approved):

caarlos0:
thank you!

raviayadav:
Hi @caarlos0, thank you for approving the changes. The ubuntu failure in checks was the minio/minio image pull, fixed in #7136. Could you please update the branch so CI picks it up? Or let me know if you want me to rebase and force-push myself.

internal/pipe/blob/upload.go
```diff
@@ -324,7 +343,7 @@ func (u *productionUploader) Upload(ctx *context.Context, filepath string, data
 		return err
 	}
 	defer func() { _ = w.Close() }()
-	if _, err = w.Write(data); err != nil {
+	if _, err = io.Copy(w, data); err != nil {
```
caarlos0: **A read error mid-copy now commits a truncated object to the bucket.**

Trigger: `io.Copy` fails partway through reading the local artifact (genuine disk I/O error, or an NFS/FUSE/overlayfs hiccup on `dist/`). The deferred `defer func() { _ = w.Close() }()` two lines above still runs, and `blob.Writer.Close` has no abort path — it calls `w.w.Close()`, which *commits*, and only then runs `defer w.cancel()` (`gocloud.dev@v0.46.0/blob/blob.go:436-469`). Reproduced against `memblob` with a reader that fails after N bytes: `size=100 -> PARTIAL OBJECT COMMITTED: 100 bytes`, `size=100000 -> PARTIAL OBJECT COMMITTED: 100000 bytes`.

What is new here: the `defer Close()` pattern and bucket-side write failures leaving partials are both pre-existing. This PR newly makes a *local file read error* a trigger. Previously `os.ReadFile` failed inside `getData`, `up.Upload` was never called, `NewWriter` never ran, and no object could exist.

Impact: the release fails with `failed to write to bucket: ...` while a corrupt, truncated artifact sits in the bucket under the final key. Silent at the object level — it looks valid to anything polling the bucket, and it is unrecoverable if the bucket has object-lock/immutability.

Reachability is low: it needs a real I/O error. Note the tempting variant is *not* reachable — EISDIR cannot happen, because `internal/extrafiles/extra_files.go:36-40` skips directories and artifact paths are always files. So I am not treating this as a blocker, and it is defensible to leave it for a separate change.

Smallest fix (verified: `cancel()` before `Close()` aborts the write on both `memblob` and `fileblob`, `NO object committed` in every case), which also removes the current double-`Close()` on the success path:

```go
wctx, cancel := stdcontext.WithCancel(ctx)
defer cancel()
w, err := u.bucket.NewWriter(wctx, filepath, opts)
if err != nil {
	return err
}
if _, err := io.Copy(w, data); err != nil {
	cancel()
	_ = w.Close()
	return err
}
return w.Close()
```

Test: call `Upload` against `memblob` with a reader that yields N bytes then a non-EOF error, and assert both the returned error and `bucket.Exists(ctx, key) == false`.

internal/pipe/blob/upload.go
```diff
@@ -198,17 +199,35 @@ func artifactList(ctx *context.Context, conf config.Blob) []*artifact.Artifact {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to open file %s: %w", path, err)
+		}
+		return f, nil
+	}
+	data, err := getData(ctx, conf, path)
+	if err != nil {
+		return nil, err
+	}
+	return io.NopCloser(bytes.NewReader(data)), nil
```
caarlos0, line 229: **This branch — the encrypted half of the refactor — has zero test coverage.**

`go test ./internal/pipe/blob/ -coverprofile` reports 0 for every statement here:

```
upload.go:224.2,225.16 2 0
upload.go:226.3,227.1 1 0
upload.go:228.2,228.49 1 0
```

The reason is that the existing KMS test (`blob_test.go:337`) calls `getData` **directly**, never through `openData`/`uploadData`/`doUpload`. `grep -rn KMSKey internal/pipe/blob/` returns exactly that one test; `blob_minio_test.go` has none; and both new tests (`TestProductionUploaderStreamsFileContents`, `TestUploadDataMissingFile`) pass `config.Blob{}`, i.e. an empty `KMSKey`, so they only take the plaintext branch above.

Impact: if the `conf.KMSKey == ""` guard at line 217 ever regresses, GoReleaser streams **plaintext** to a bucket the user explicitly configured for KMS encryption, and the entire suite still goes green. That is the single most valuable assertion missing from this PR, and it is missing precisely on the branch the PR introduced. I hand-verified the branch does work today (fake awskms server + `memblob`: 0 B, 10 B and 4096 B files store the exact ciphertext; 4097 B fails with the expected message and writes nothing), so this is a coverage gap rather than a bug.

Smallest fix: repoint the existing KMS table test at `uploadData` with a `memblob`-backed `productionUploader` instead of calling `getData` directly, and assert the stored object equals the ciphertext — not the plaintext. That keeps the current size-limit cases and covers these three lines at the same time.

internal/pipe/blob/upload.go
```diff
@@ -198,17 +199,35 @@ func artifactList(ctx *context.Context, conf config.Blob) []*artifact.Artifact {
...
+// Unencrypted files are streamed, so they are never fully loaded into memory.
+func openData(ctx *context.Context, conf config.Blob, path string) (io.ReadCloser, error) {
+	if conf.KMSKey == "" {
+		f, err := os.Open(path)
+		if err != nil {
+			return nil, fmt.Errorf("failed to open file %s: %w", path, err)
+		}
+		return f, nil
+	}
+	data, err := getData(ctx, conf, path)
```
caarlos0, line 225: **This call makes `getData`'s own `KMSKey == ""` branch dead code.**

Anchored here because this is now the only production caller of `getData`, and it is reached exclusively when `conf.KMSKey != ""`. That makes the early return at `upload.go:267-269`:

```go
if conf.KMSKey == "" {
	return data, nil
}
```

unreachable. Nothing exercises it — not even the test suite, since `blob_test.go:337` passes a non-empty `awskms://` key. Coverage confirms it: `upload.go:268.3,269.1 1 0`.

Impact is maintenance, not behaviour: `getData` now reads as if it still serves both paths, which hides the contract this PR established (`getData` is the KMS-only path). The next person to touch it has to re-derive that from the call graph.

Smallest fix: delete the branch, so `getData` is unambiguously KMS-only. No new test needed — `TestGetDataAWSKMSPlaintextLimit` already covers the remaining path.

internal/pipe/blob/blob_test.go
```diff
@@ -392,6 +395,98 @@ func TestDoUploadUploadsArtifactsAndExtraFiles(t *testing.T) {
 	}, uploader.uploaded())
 }
 
+func TestProductionUploaderStreamsFileContents(t *testing.T) {
```
caarlos0: **This test's name overstates what it verifies.**

It passes unchanged against the old buffering implementation (`os.ReadFile` + `w.Write`), because asserting round-tripped bytes and `http.DetectContentType` equality says nothing about *how* the bytes reached the writer. It is a content and content-type correctness test, which is worth having — the empty / sub-512-byte / multi-MiB cases are exactly the right sniffing boundaries — but it does not test streaming.

The actual streaming guarantee is carried solely by `TestPublishDoesNotBufferFilesPerDestination` below, and that one does earn its name: I confirmed it fails on the buffering implementation with `"34114864" is not less than "2097152"`.

Worth noting for that test too: it substitutes `discardUploader`, so it measures `uploadData`/`openData` only and never touches `productionUploader` — it cannot bound production allocation (see the 1 MiB `readFromWriteTo` buffer in the review body). Its measured window is ~0.37-0.52 MiB against the 2 MiB budget across plain, `-race`, and CI-like runs, so ~4x headroom; but the largest component is `text/template` work in `tmpl.New`/`Apply` (~300 KB), not file data. A future change that adds one more templated field per destination could fail this test for a reason unrelated to buffering. Raising the budget to `2*fileSize` would keep the regression caught with a large margin (the bug allocates `16*fileSize`) while removing that trap.

Suggested fix: rename this to something like `TestProductionUploaderUploadsFileContentsAndDetectsContentType`, so the suite does not read as if streaming is covered in two places.

caarlos0 (commented):
Reviewed the full diff against `origin/main` (merge base), plus the gocloud.dev v0.46.0 writer source and the Go 1.27 `os`/`io` copy paths. The core change is correct and does what it says: `uploader.Upload` taking `io.Reader` is internal-only (`internal/pipe/blob`, three implementations, all in-package), content-type sniffing is preserved, the file is closed on every path, and the KMS path is correctly delimited. I confirmed the regression test genuinely fails without the fix (`34114864` vs the `2097152` budget) and that it is stable here (50x isolated, and whole-package `-count=5 -race -cover`), with roughly 4x headroom.

Four inline findings, none of which I consider blocking. Summary of what needs doing: add end-to-end coverage for the new KMS branch (the one part of the refactor nothing exercises), and drop the branch in `getData` that this PR made unreachable.

**Measured trade-off worth recording (no change requested).** `io.Copy(w, data)` resolves to `(*os.File).WriteTo` -> `os.genericWriteTo` -> `io.Copy` -> `blob.Writer.ReadFrom` -> `readFromWriteTo`, and `readFromWriteTo` unconditionally allocates a 1 MiB buffer (`gocloud.dev@v0.46.0/blob/blob.go:284`, `buf := make([]byte, 1024*1024)`), not the 32 KiB one might assume. Measured `TotalAlloc` per upload through `productionUploader` + `fileblob`: 100 B file 1.033 MiB (was 0.034), 64 KiB 1.032 MiB (was 0.104), 1 MiB 1.033 MiB (was 1.041), 64 MiB 1.033 MiB (was 64.047). So allocation is now O(1) in file size, which is the win, but the crossover is ~1 MiB and it goes *up* for the many small files a release uploads (checksums, metadata, sigs, SBOMs). Peak concurrency is `Parallelism` x `Parallelism` (outer group over destinations in `blob.go`, inner group over artifacts in `upload.go`), so worst case is `Parallelism^2 x 1 MiB` regardless of artifact size — 16 MiB at the defaults. Still a large net improvement over `Parallelism^2 x filesize`; just not what the description implies.

**Verification gaps I could not close.** All upload experiments used `memblob` and `fileblob`. Real driver behaviour for s3blob/gcsblob/azureblob — in particular whether a partial multipart upload is aborted or completed — was not tested against a live or emulated backend.

_Disclosure: this review was produced by an AI assistant. Every claim above was traced to source or reproduced locally, but please sanity-check it._

internal/pipe/blob/upload.go
caarlos0: Fixed in 54490ca14.

`Upload` now cancels the writer context before closing on the error path, so a failed copy aborts the write instead of committing what was already sent. This also removes the double `Close()` on the success path:

```go
wctx, cancel := stdctx.WithCancel(ctx)
defer cancel()
w, err := u.bucket.NewWriter(wctx, filepath, opts)
if err != nil {
	return err
}
if _, err := io.Copy(w, data); err != nil {
	cancel()
	_ = w.Close()
	return err
}
return w.Close()
```

Covered by the new `TestProductionUploaderDoesNotCommitPartialUploads`, which uploads through a reader that yields N bytes then a non-EOF error and asserts both the returned error and `bucket.Exists(...) == false`. It runs at 100 and 100,000 bytes, i.e. either side of the 512-byte sniff threshold, so both the still-buffered writer and the already-open writer are exercised.

Proved decisive: with the old `defer func() { _ = w.Close() }()` restored via `go test -overlay`, both subtests fail with `a failed read must not leave a truncated object behind`.

internal/pipe/blob/upload.go
caarlos0: Fixed in 54490ca14.

The KMS table test was repointed from `getData` to `uploadData`, running through a real `productionUploader` backed by `memblob`, and now asserts what actually lands in the bucket:

```go
require.Equal(t, []byte("ciphertext"), got)
require.NotEqual(t, plaintext, got, "the plaintext must never reach the bucket")
```

The failure case additionally asserts `bucket.Exists(...) == false`, so a rejected oversize file cannot leave anything behind. Renamed to `TestUploadDataAWSKMSEncryptsBeforeUploading`, since it now covers the whole path rather than just the size limit.

Coverage on this branch went from 0 to covered on every statement:

```
upload.go:225.2,226.16 2 1
upload.go:227.3,228.1 1 1
upload.go:229.2,229.49 1 1
```

Package coverage 76.1% -> 79.4%. Proved decisive: forcing `openData` to always stream (i.e. breaking the `conf.KMSKey == ""` guard) via `go test -overlay` fails both subtests — the success case sees 0 KMS requests instead of 1, and the oversize case stops erroring.

internal/pipe/blob/upload.go
caarlos0: Fixed in 54490ca14. Dropped the unreachable branch and documented the contract so the call graph no longer has to be re-derived:

```go
// getData reads the whole file and encrypts it with KMS.
// It is only used when a KMS key is set, as encryption needs the full
// plaintext; everything else is streamed by openData.
func getData(ctx *context.Context, conf config.Blob, path string) ([]byte, error) {
	data, err := os.ReadFile(path)
	if err != nil {
		return data, fmt.Errorf("failed to open file %s: %w", path, err)
	}
	if err := validateKMSPlaintextSize(conf.KMSKey, len(data)); err != nil {
```

No coverage lost: `getData` is still fully exercised, now through `uploadData` in `TestUploadDataAWSKMSEncryptsBeforeUploading` (see the sibling thread on line 228).

internal/pipe/blob/blob_test.go
caarlos0: Fixed in 54490ca14. Renamed to `TestProductionUploaderUploadsFileContentsAndDetectsContentType`, which is what it actually asserts. Streaming is now covered in exactly one place, `TestPublishDoesNotBufferFilesPerDestination`, plus the new partial-upload test on the `productionUploader` side.

Also took the budget suggestion — it is now `2*fileSize` (4 MiB):

```go
// Buffering every file for every destination allocates
// files*destinations*fileSize bytes; streaming keeps the whole run under
// two files worth, which still leaves a wide margin over that bug.
require.Less(t, after.TotalAlloc-before.TotalAlloc, uint64(2*fileSize))
```

That removes the trap where unrelated per-destination `text/template` allocation could trip it, without weakening the assertion: restoring the buffering implementation via `go test -overlay` still fails at `"34109656" is not less than "4194304"`, so roughly 8x margin over the bug and 8x headroom over the ~0.4 MiB the streaming path actually allocates. Stable over 30 runs, and the whole package passes `-count=5 -race -cover`.

Left `discardUploader` in place: the point of that test is `uploadData`/`openData`, and the new `productionUploader` tests cover the real writer path.

PR #7087, opened 2026-09-06
fix(winget,krew): manifest validation and executable paths

caarlos0:
Fixes a group of related bugs in Winget and Krew pipes.

- Closes #6947: renders WinGet package_identifier templates before fallback, validation, filenames, manifests, and publish paths.
- Closes #6946: keeps same-name WinGet entries in separate publish groups so repository and skip_upload settings do not bleed across entries.
- Closes #6945: validates WinGet installer selections before writing or registering manifests.
- Closes #6948: includes archive wrapper directories in generated Krew executable paths.

PR #7140, opened 2026-09-15
fix: nine bugs found reviewing v2.18.1..main

caarlos0:
Fix nine bugs found while reviewing everything that landed between `v2.18.1` and `main`. One commit per bug, each with a regression test.

## What

| Commit | Fix |
| --- | --- |
| `49a5d5a` | **docker**: a bare `linux/amd64` platform matches any `goamd64` variant again |
| `866ec5d` | **docker**: arm64 variants land in the directory buildx actually reports |
| `54e9bf4` | **aur**: the duplicate-architecture error says how to fix it |
| `93509fd` | **cask**: warn when the cask file gets renamed to match its token |
| `2ff0cf0` | **gomod**: an unparseable `go.mod` no longer fails the release |
| `8be6ad0` | **sbom,sign**: default names disambiguate every CPU variant |
| `e300e7e` | **gitea**: find releases whose tag contains a slash |
| `d4607ec` | **nix**: template the description only once |
| `95e55a8` | **winget**: artifact IDs are named after the winget again |

### `docker`: bare `linux/amd64` no longer matched non-baseline `goamd64` builds

Regression from #7095. `parsePlatform` defaulted a bare `linux/amd64` to `amd64: v1`, and `contextArtifacts` turned that into `artifact.ByGoamd64("v1")`, which only matches `v1` or empty. A project building `goamd64: [v3]` with the default `platforms` got an **empty build context** and failed at `COPY $TARGETPLATFORM/`, with only a "no binaries found for the given platform" warning hinting at why.

| `goamd64` | v2.18.1 | before this PR |
| --- | --- | --- |
| `v1` / unset | 2 artifacts | 2 artifacts |
| `v2` / `v3` / `v4` | 2 artifacts | **1 — amd64 dropped** |

The build context directory is now derived from the **requested platform**, normalized the way buildx normalizes `TARGETPLATFORM`, instead of from the artifact. A platform without an explicit CPU variant accepts any variant, and when several match, the least specific one wins so the result does not depend on the order artifacts were added in — which keeps the determinism #7095 was after.

### `docker`: arm64 variants went to the wrong directory

`parsePlatform` and `byGoarm64` learned about `linux/arm64/vX.Y`, but placement still returned a plain `linux/arm64` for every arm64 artifact, so `platforms: [linux/arm64/v9.0]` copied the binary to `linux/arm64/` while buildx sets `TARGETPLATFORM=linux/arm64/v9`. `COPY` then failed.

The expected values are not guesses — they were taken from `docker buildx build --platform=X` runs against a Dockerfile echoing `$TARGETPLATFORM`:

| requested | `TARGETPLATFORM` |
| --- | --- |
| `linux/amd64/v1` | `linux/amd64` |
| `linux/amd64/v3` | `linux/amd64/v3` |
| `linux/arm64/v8` , `v8.0` | `linux/arm64` |
| `linux/arm64/v8.2` | `linux/arm64/v8.2` |
| `linux/arm64/v9` , `v9.0` | `linux/arm64/v9` |

### `aur`: duplicate-architecture error had no remediation

`ErrMultipleArchivesSameArch` rejects configs that used to produce output — for instance `archives: [{formats: [tar.gz, zip]}]` with an `aurs` entry that has no `ids` — but did not say what to do about it. brew and cask already point at `ids` in the same situation.

### `cask`: renamed files are orphaned in the tap

The file name now follows the cask token, which is always normalized, so a `name` with spaces or uppercase letters writes `myapp.rb` where it used to write `MyApp.rb`. Publishing only creates and updates files, so the old one stays in the tap declaring the same token, which makes `brew audit` fail. GoReleaser cannot safely delete it, so it warns instead.

### `gomod`: strict parsing turned an advisory check into a release failure

The replace check switched from a line scan to `modfile.Parse`, which is strict and rejects any directive the bundled `golang.org/x/mod` does not know about (verified: `go.mod:5: unknown directive: futuredirective`). A newer Go release adding a directive would abort the release over a check that only looks for `replace`. It now warns and skips.

### `sbom`,`sign`: silent overwrites between CPU variants

The default document and signature names only accounted for `goarm`, `gomips` and `goamd64`. Building more than one `goarm64`, `go386`, `goppc64` or `goriscv64` variant made every binary of an os/arch pair resolve to the same name, so syft overwrote the first document and two artifacts were registered with the same path. Unlike archives, neither pipe checks whether the file already exists, so this was **silent**.

### `gitea`: releases with a slash in the tag were never found

The SDK escapes the tag into a single path segment and Gitea routes it as one, so `GetReleaseByTag` can never match a tag like `release/1.2.3`. The 404 was read as "no release yet", and the release creation that followed failed because the release already existed. Those tags now fall back to paging through the release list.

### `nix`: descriptions were templated twice

The whole rendered derivation goes through the template engine again, so a description that renders to text containing `{{` was evaluated a second time and failed to parse. It is now escaped and quoted through a `nixString` template function that emits a literal for the final pass — the same shape brew and cask use for `desc`.

### `winget`: artifact IDs became opaque

The IDs became positional `winget-<index>`. They show up in `dist/artifacts.json` and can be matched with `ids:`, so they should say which winget they came from. They are now built from the name, after it is templated, keeping the index so two entries sharing a name stay distinct.

## Verification

Every test was proven to fail on the unfixed code before being kept: each source change was reverted with the new test in place, the failure confirmed, then restored. For example `goamd64: v2/v3/v4` returned no artifacts, arm64 `v9.0` produced `linux/arm64` instead of `linux/arm64/v9`, the sbom/sign variants collided on identical names, gitea returned nil for a slash tag, and winget produced `winget-0`.

Full test suite run against a `v2.18.1` worktree in the same environment: **12 failing packages on both sides, identical sets, zero new failures**. The remaining failures are environment-only (docker daemon, local gpg 2.5.22, node SEA) and reproduce unchanged at the baseline. `go build ./...`, `go vet ./...` and `golangci-lint` are clean across all touched packages.

## Notes

- No documentation changes: the docker docs already describe `COPY $TARGETPLATFORM/`, which is exactly the behavior these fixes restore.
- `contextArtifacts`/`makeContext` now carry `(platform, artifact)` pairs, so `toPlatform` is gone and its tests became `TestContextDir`. The change is contained to `internal/pipe/docker/v2`.

## Resources

Follow-up to the review of `v2.18.1..main`, covering #7095, #7100, #7086, #7096, #7091, #7097, #7082 and #7087.


caarlos0:
Superseded — split into one PR per fix so each can be reviewed and merged independently.

| PR | Fix |
| --- | --- |
| #7148 | `fix(docker)`: match any goamd64 variant on bare platforms |
| #7149 | `fix(docker)`: put arm64 variants where buildx looks (stacked on #7148) |
| #7141 | `fix(aur)`: tell users how to fix a duplicate architecture |
| #7142 | `fix(cask)`: warn when the cask file gets renamed |
| #7143 | `fix(gomod)`: do not fail the release on an unparseable go.mod |
| #7144 | `fix(sbom,sign)`: disambiguate every CPU variant in default names |
| #7145 | `fix(gitea)`: find releases whose tag contains a slash |
| #7146 | `fix(nix)`: template the description only once |
| #7147 | `fix(winget)`: name the generated artifact IDs after the winget |

Only the two docker fixes are coupled (the arm64 fix builds on the `contextDir` helper from #7148), so they are stacked. The other seven touch disjoint packages and branch off `main` directly.

PR #7141, opened 2026-09-15
fix(aur): tell users how to fix a duplicate architecture

caarlos0:
Tell users how to fix an AUR entry that gets two archives for the same architecture.

`ErrMultipleArchivesSameArch` (added in #7100) rejects configs that previously produced output. It is reachable with a plain `archives: [{formats: [tar.gz, zip]}]` plus an `aurs` entry without `ids`: both linux/amd64 archives pass the `ByTypes(UploadableArchive, UploadableBinary)` filter.

The old PKGBUILD emitted two `source_x86_64=` lines. That is valid bash — the last one wins — but the archive pipe builds formats concurrently, so which format shipped was not deterministic. Rejecting the config is still the right call; this only makes the fix discoverable.

`ids` alone is not enough, though. The archive pipe stamps **every format of a single `archives` entry with the same `ExtraID`** (`archive.go:259`), so `artifact.ByIDs` cannot separate them — a user who follows a bare "use ids" hint adds `ids` and hits the identical error. The message therefore names both causes:

> one aur can handle only one archive of each architecture. Consider using ids in the aurs section; if a single archives entry produces multiple formats, split it into separate archives with distinct ids

`TestRunPipeRejectsDuplicateArchitectures` gains two rows — `single id, multiple formats` and `single id, multiple formats, filtered by ids` — the second of which records that `ids` does not disambiguate that case. The assertion also moves from the literal string to `require.ErrorIs` on the sentinel, matching brew and cask, so future wording changes no longer require a test change.

Follow-up to #7100. Part of a series of fixes found reviewing `v2.18.1..main`.


internal/pipe/aur/aur.go
```diff
@@ -34,7 +34,7 @@ const (
 
 var (
 	ErrNoArchivesFound          = errors.New("no linux archives found")
-	ErrMultipleArchivesSameArch = errors.New("one aur can handle only one archive of each architecture")
+	ErrMultipleArchivesSameArch = errors.New("one aur can handle only one archive of each architecture. Consider using ids in the aurs section")
```
caarlos0: **The new hint does not correct the example given in this pull request's own description.**

**Trigger:** `archives: [{formats: [tar.gz, zip]}]` plus an `aurs` entry.

- `internal/pipe/archive/archive.go:136-151` makes one artifact for each format of a single `archives` entry.
- `internal/pipe/archive/archive.go:259` gives the same `artifact.ExtraID` (`arch.ID`) to every one of those artifacts.
- Thus the two linux/amd64 archives have one ID.
- `internal/pipe/aur/aur.go:139-141` filters with `artifact.ByIDs`. That filter cannot divide two artifacts that have the same ID.
- `dataFor` (line 360-364) keys the duplicate check on architecture only, and ignores the format.

**Impact:** the user reads the new message, adds `ids: [default]`, and gets the identical error. For the case this pull request names as the cause, the advice is a dead end.

**Evidence:** driving the real archive pipe with `formats: [tar.gz, zip]` gives two artifacts `id=default format=zip` and `id=default format=tar.gz`. Both `no ids` and `ids: [default]` return the same error. Only two `archives` entries with different IDs, and then `aurs.ids`, clear it. `config.AUR` has no filter for format, thus the `aurs` section alone cannot express the fix.

**Smallest fix:** name both causes in the message. For example:

```go
ErrMultipleArchivesSameArch = errors.New("one aur can handle only one archive of each architecture. Use ids in the aurs section to select one; if a single archives entry produces multiple formats, split it into separate archives with distinct ids")
```

**Needed test:** a row in `TestRunPipeRejectsDuplicateArchitectures` with two artifacts that have the same `ExtraID` and different `ExtraFormat`. See the related note at `internal/pipe/aur/aur_test.go:684`.

**Note on the precedent:** `brew.go:35` and `cask.go:34` carry the analogous wording, but they have the same blind spot (`brew.go:488` keys on `pkg.OS+pkg.Arch`). Thus consistency with them does not prove the wording is correct.

internal/pipe/aur/aur_test.go
```diff
@@ -681,7 +681,7 @@ func TestRunPipeRejectsDuplicateArchitectures(t *testing.T) {
 			require.NoError(t, Pipe{}.Default(ctx))
 			err := runAll(ctx, client.NewMock())
 			if tt.wantErr {
-				require.EqualError(t, err, "one aur can handle only one archive of each architecture")
+				require.EqualError(t, err, "one aur can handle only one archive of each architecture. Consider using ids in the aurs section")
```
caarlos0: `require.EqualError` compares the message text. This assertion is the only reason this pull request must change the test file.

`internal/pipe/brew/brew_test.go:964` and `internal/pipe/cask/cask_test.go:1033` compare the sentinel value instead. The sentinel returns unwrapped through `dataFor` -> `buildPkgFile` -> `doRun` -> `runAll`, thus `errors.Is` operates correctly here.

**Smallest fix:**

```go
require.ErrorIs(t, err, ErrMultipleArchivesSameArch)
```

Low severity. Keep the text comparison if you intend to pin the user-visible text on purpose. In that case, no change is necessary.

caarlos0 (commented):
## Verdict: one blocking problem

*(Posted as a comment: GitHub does not permit a change request on your own pull request.)*

The one goal of this change is to make the fix discoverable. For the configuration that the description names as the cause (`archives: [{formats: [tar.gz, zip]}]` with no `ids`), the new advice does not work. See the comment on `internal/pipe/aur/aur.go:37`.

The change itself is correct and safe in all other respects:

- The rejection is still the correct behavior. The old PKGBUILD emitted two `source_x86_64=` lines, and the artifact order is not deterministic, thus the shipped format was random.
- Nothing else in the repository, the documentation, or the golden testdata asserts the old string.
- `golangci-lint` reports 0 issues. The message does not start with a capital letter and does not end with punctuation, thus ST1005 is satisfied.

### Missing coverage (no single line)

The scenario in the description — one `archives` entry, two formats, one shared artifact ID — has no test in the repository. `TestRunPipeRejectsDuplicateArchitectures` covers only artifacts with different IDs. A test for that scenario would have shown that the new advice is inert.

### Verification gaps

- The claim that the pre-#7100 PKGBUILD was valid bash is confirmed by a template trace (`tmpl.go:65-67`), not by a build of the old tree.
- `TestRunPipeMultipleConfigurations` fails locally because of a `safe.bareRepository=explicit` git setting. That failure is environmental. With `safe.bareRepository=all`, `go test ./internal/pipe/aur/...` passes.

---
*This review was produced by an automated reviewer (GitHub Copilot CLI).*

internal/pipe/aur/aur.go
caarlos0: Fixed in af2ccf9ef.

Confirmed: `artifact.ByIDs` cannot separate two artifacts that share one `ExtraID`, thus the old hint was inert for `formats: [tar.gz, zip]`. The message now names both causes:

> one aur can handle only one archive of each architecture. Consider using ids in the aurs section; if a single archives entry produces multiple formats, split it into separate archives with distinct ids

Two rows were added to `TestRunPipeRejectsDuplicateArchitectures`: `single id, multiple formats`, and `single id, multiple formats, filtered by ids`. The second row records that `ids` does not help in this case. All 5 subtests pass, and `golangci-lint` reports 0 issues.

internal/pipe/aur/aur_test.go
caarlos0: Fixed in af2ccf9ef.

The assertion is now `require.ErrorIs(t, err, ErrMultipleArchivesSameArch)`, which agrees with `brew_test.go` and `cask_test.go`. Thus a change of the wording no longer needs a change of the test.

PR #7142, opened 2026-09-15
fix(cask): warn when the cask file gets renamed

caarlos0:
Warn when a cask's file name is normalized, because the previously published file is left behind in the tap.

#7086 changed the file name from `brew.Name + ".rb"` to `caskNameFor(brew.Name) + ".rb"`. That is correct — the file name has to match the cask token, which was already normalized — but `doPublish` only creates and updates files, it never deletes.

So a user with an uppercase or spaced `casks[].name` gets a new `myapp.rb` while the old `MyApp.rb` stays in the tap, still declaring token `myapp`. That is a duplicate-token tap and a `brew audit` failure.

GoReleaser cannot safely delete the old file, so it now warns:

> cask file is "Casks/myapp.rb" because it has to match its token: if your tap still has "Casks/MyApp.rb" from an older release, delete it

The warning lives in `doPublish`, after the skip checks, so `--snapshot`, `--skip=publish` and `skip_upload` stay quiet: it only fires on runs that actually touch the tap, and it names the full path, directory included.

The test extends the existing normalization table with a `wantWarn` column, so the already-normalized case asserts the warning is *not* emitted, and `TestRunPipeNoUpload` now uses a non-normalized name to pin that a skipped publish does not warn.

Follow-up to #7086. Part of a series of fixes found reviewing `v2.18.1..main`.


internal/pipe/cask/cask.go
```diff
@@ -310,6 +310,16 @@ func doRun(ctx *context.Context, brew config.HomebrewCask, cl client.ReleaseURLT
 	}
 
 	filename := caskNameFor(brew.Name) + ".rb"
+	if previous := brew.Name + ".rb"; previous != filename {
```
caarlos0: This warns on runs that publish nothing. `cask.Pipe` is in the release `Pipeline` (internal/pipeline/pipeline.go:158) and `Pipe.Skip` (L50-52) only checks `skips.Homebrew` and an empty cask list, so `doRun` still runs under `--snapshot` (cmd/release.go:158 sets `skips.Publish`, which only short-circuits `publish.Pipe`) and under `--skip=publish`. `skip_upload` is checked in `doPublish` (L154), i.e. after this line has already logged.

Trigger: `casks: [{name: MyApp}]` plus `goreleaser release --snapshot`. Impact: GoReleaser tells you to delete a file from a tap it never touched.

Smallest fix: move the check into `doPublish`, after the two skip returns at L154-160. Everything it needs is already there — the `brew` in the artifact extra is the post-template copy stored by `doRun`, and `cask.Name` is the new filename:

```go
if previous := brew.Name + ".rb"; previous != cask.Name {
```

Test: a `skip_upload: "true"` row asserting the warning is absent. The table in `TestRunPipeUsesNormalizedTokenForFilenames` already calls `publishAll`, so it is one more `wantWarn: false` case.

internal/pipe/cask/cask.go
```diff
@@ -310,6 +310,16 @@ func doRun(ctx *context.Context, brew config.HomebrewCask, cl client.ReleaseURLT
...
+		// the file name has to match the cask token, which is always
+		// normalized. GoReleaser only creates and updates files, so a
+		// previously published file with the old name would stay behind and
+		// declare the same token.
+		log.Warnf(
+			"cask file renamed from %q to %q to match its token: if %q was published to the tap before, delete it",
+			previous, filename, previous,
+		)
```
caarlos0: Three things about the message itself.

1. It names only the basename. The published path is `buildCaskPath(brew.Directory, filename)` (L345), so `Casks/MyApp.rb` by default and `<directory>/MyApp.rb` when `directory` is set — a user with a custom `directory` is not told where to look. `buildCaskPath(brew.Directory, previous)` costs nothing here.

2. "renamed from" asserts an event that never happened for anyone whose first release is on the upcoming version; only the trailing "if ... was published to the tap before" hedges it. Stating the state reads truer, e.g. `cask file is %q because it must match the token: if your tap still has %q from an older release, delete it`.

3. The condition is a property of the config, not of the tap, so it can never clear: the warning returns on every release forever. The obvious escape is a footgun — L77 derives the default binary from the *raw* name (`brew.Binaries = []string{brew.Name}`), so normalizing `name: MyApp` to `myapp` to silence this silently turns `binary "MyApp"` into `binary "myapp"` for anyone who does not set `binaries`. Reading the tap is not an option (`client.Client` has no read or delete), so this is a trade-off to take deliberately: keep the permanent warn, or drop it to info/summary. Related: the noise multiplies with the snapshot issue on L313.

caarlos0 (commented):
The production change is correct. `git show 924f5eb06` confirms the pre-#7086 name was `brew.Name + ".rb"` at the same position with the same templating above it, and `doPublish` writes `buildCaskPath(brew.Directory, cask.Name)`, so the warned basename really is the file left behind. `git tag --contains 924f5eb06` lists only nightlies (latest stable is v2.18.1), so every affected user hits this on the upgrade to the next stable — the warning is worth having.

Two problems with *where* and *what* it warns; both inline. Posting as a comment because GitHub does not allow requesting changes on your own PR.

**Test coverage.** The new assertions do fail without the fix — verified with `go test -overlay` against `origin/main`'s `cask.go`: `spaced` and `uppercase` fail on the `Contains`, `already-normalized` passes. The global-logger swap is safe here: the only two `t.Parallel()` tests in the package (`TestDefault`, `TestRunPipeUniversalArchiveWithLinuxBinary`) are top-level and resume only after the sequential tests return, and `go test -race -count=2 ./internal/pipe/cask/` reports no races. What is missing is a case pinning the *absence* of the warning when nothing is published (see the comment on L313).

**Verification gaps.** `TestFullPipe/git_remote` fails locally, but it fails identically with `origin/main`'s `cask.go` overlaid, so it is environmental and unrelated to this diff. I did not run anything outside `./internal/pipe/cask`.

Reviewed by a bot (GitHub Copilot CLI).

internal/pipe/cask/cask.go
caarlos0: Fixed in fa3a9bb9b: the check moved into `doPublish`, right after `gpath` is built and after both skip returns, comparing `buildCaskPath(brew.Directory, brew.Name+".rb")` with `gpath`. `--snapshot`, `--skip=publish` and `skip_upload` no longer warn.

Coverage: `TestRunPipeNoUpload` now uses a non-normalized `name: Foo Bar` and asserts the warning is absent across all three skip paths. Verified decisive — with the block moved back into `doRun` (same wording, via `go test -overlay`), `TestRunPipeNoUpload/skip_upload_auto` fails.

internal/pipe/cask/cask.go
caarlos0: Fixed in fa3a9bb9b. The message is now built from `buildCaskPath`, so it carries the directory, and states the state instead of an event:

> cask file is "Casks/foo-bar.rb" because it has to match its token: if your tap still has "Casks/Foo Bar.rb" from an older release, delete it

On (3): keeping the warning. Reading the tap is not possible (`client.Client` has no read or delete), and normalizing `name` is indeed a footgun because of the raw-name binaries default, so the message no longer suggests it. Moving it to `doPublish` at least limits it to runs that actually touch the tap.

caarlos0:
unlikely to be needed, as these previous casks were likely invalid anyways.

Issue #6892, opened 2026-09-05
gomod: replacement blocks bypass proxy validation


caarlos0:
### Trigger

Enable `gomod.proxy` for a tagged, non-snapshot Go build whose root `go.mod` contains a normal replacement block:

```go
replace (
    example.invalid/dependency => example.invalid/fork v1.0.0
)
```

### Observed

Production module loading, replacement checking, and proxying were exercised with local recording Go executables. The block passed validation and reached `go get example.invalid/project@v1.2.3`. Rewriting the same replacement as a single line correctly returned `ErrReplaceWithProxy`. Parsing the module with `modfile` found the replacement in both forms. No external download or full CLI release was performed.

[The line-based regular expression](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/gomod/gomod_proxy.go#L36) and [its caller](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/gomod/gomod_proxy.go#L42-L51) only recognize replacements whose directive and mapping occupy one line.

### Smallest fix

Use the existing Go module parser to inspect replacement directives instead of matching text lines. Add equivalent single-line and block cases, including multiple replacements. Preserve the current distinction between snapshot warnings and release rejection. This reproduces at the module root and does not depend on directory selection.

---
Audit priority: **P2**.

Audit source revision: [628c889d7](https://github.com/goreleaser/goreleaser/commit/628c889d7f2a34066649833be419fdb9abe01174). AI disclosure: this report comes from a Copilot-assisted audit, with discovery by `caarlos0` agents and independent verification by `anvil` agents. Reproduction methods and limits are stated above; no fix is included in this issue.

PR #7143, opened 2026-09-15
fix(gomod): do not fail the release on an unknown go.mod directive

caarlos0:
Do not fail the release when `go.mod` contains a directive the bundled `golang.org/x/mod` does not know about, while keeping the `replace` check working.

#7096 switched the check from a line scan for `^replace .* => .*$` to `modfile.Parse`, which is strict: it rejects **any** directive the bundled `golang.org/x/mod` does not know about. `CheckGoModPipe` then aborts the whole release over a check that only cares about `replace`.

`tool`, `ignore` and `godebug` parse fine against the currently pinned x/mod, but a future Go release adding a directive would break releases for everyone on an older GoReleaser. That is a bad failure mode for an advisory check.

The check now parses with `modfile.ParseLax`, which ignores unknown directives. `ParseLax` deliberately drops `replace` from the parsed file, so the replaces are read off `file.Syntax.Stmt` instead: the syntax tree keeps them in both the single-line and the block form, and carries the directive exactly as written.

```
unknown+replace     err=<nil> replaces=["replace example.invalid/a => ../a"]
unknown+block       err=<nil> replaces=["replace example.invalid/a => ../a"]
versioned replace   err=<nil> replaces=["replace example.invalid/a v1.0.0 => example.invalid/b v1.2.3"]
unterminated block  err=go.mod:4: syntax error (unterminated block started at go.mod:3:1)
```

That means an unknown directive no longer aborts the release **and** no longer hides a `replace` sitting next to it, which would otherwise let a release ship a binary silently resolved against the real upstream instead of failing with `ErrReplaceWithProxy`.

Warn-and-skip now only happens when `ParseLax` itself fails, i.e. the file is not a valid `go.mod` at all. `gomod.Pipe` already aborts on that earlier via `go list -m`, with a better message.

This also drops `formatReplace`: the syntax tree already carries the directive as written, so there is a single code path for both forms.

Tests cover an unknown directive alone, an unknown directive alongside a `replace` (single-line and block, both asserting `ErrReplaceWithProxy`), and a genuinely invalid `go.mod`. The two `replace`-plus-unknown-directive tests fail without this change.

Follow-up to #7096. Part of a series of fixes found reviewing `v2.18.1..main`.


internal/pipe/gomod/gomod_proxy.go
```diff
@@ -50,7 +50,12 @@ func (CheckGoModPipe) Run(ctx *context.Context) error {
 		}
 		file, err := modfile.Parse(path, mod, nil)
 		if err != nil {
-			return fmt.Errorf("could not parse %q: %w", path, err)
+			// modfile.Parse rejects any directive it does not know about, and
+			// a newer Go release may well add one. That is not a good enough
+			// reason to abort the release, since this check is only looking
+			// for replace directives.
+			log.WithError(err).Warnf("could not parse %s, skipping the %s check", logext.Keyword(path), logext.Keyword("replace"))
+			continue
```
caarlos0: **A `replace` in a `go.mod` with an unknown directive is now silently ignored.**

**Trigger:** non-snapshot release, `gomod.proxy: true`, and a `go.mod` that contains *both* a `replace` and any directive the pinned x/mod does not know. That is exactly the case the comment above describes — a newer Go adds a directive — except the user also has a `replace`, which is not unusual for local development.

**Impact:** `ErrReplaceWithProxy` stops firing. The release proceeds into `ProxyPipe`, which writes a synthetic `go.mod` and runs `go get <module>@<tag>`; Go ignores `replace` in a non-main module, so either the replaced dependency silently resolves to the real upstream (a different binary than the user expects) or `go get` fails with a module-resolution error that never mentions `replace`. Before #7096 this was a line scan and an unknown directive did not disturb it, so this is a regression against that behavior, and it partially undoes the `replace (...)` block handling #7096 added for #6892.

I reproduced it: with this branch, `Run` returns `nil` for a non-snapshot `go.mod` holding `futuredirective bar v1` plus `replace a => ../b`.

**Smallest fix:** on `Parse` error, fall back to `ParseLax` and rebuild the replaces from `f.Syntax.Stmt`. `parseToFile` always populates `f.Syntax` with the full statement tree, and I verified the raw `replace` statements survive `ParseLax` in both forms:

```
ParseLax err: <nil>   File.Replace len: 0
Line:      [replace a => ../b]
LineBlock: [replace]
   -> [c => ../d]
   -> [e => f v1.0.0]
```

Then warn-and-`continue` only when `ParseLax` itself fails, which means genuine syntax corruption — and `gomod.Pipe` already aborts on that earlier via `go list -m`, with a better message. This keeps the semantics you want (unknown directives never abort a release) without dropping the check.

If you would rather keep warn-and-skip as a deliberate trade-off, that is defensible, but the comment and the warning should say that a `replace`, if present, will no longer be found — right now neither does. Compare `internal/pipe/release/preflight.go:57`, which tells the user what the degraded check costs.

**Test needed:** see the comment on `gomod_proxy_test.go:101-129`.

internal/pipe/gomod/gomod_proxy_test.go
```diff
@@ -98,6 +98,36 @@ replace (
...
+					ID:     "foo",
+					Goos:   []string{runtime.GOOS},
+					Goarch: []string{runtime.GOARCH},
+					Main:   ".",
+					Dir:    ".",
+				},
+			},
+		}, withTestModulePath)
+
+		require.NoError(t, CheckGoModPipe{}.Run(ctx))
```
caarlos0: This subtest covers the unknown-directive-alone state, but not the state that actually changed in a way a user can feel: an unknown directive **next to a `replace`**, on a non-snapshot release. That is the gap behind the finding on `gomod_proxy.go:53-58`.

Add a subtest writing `module foo\n\ngo 1.25\n\nfuturedirective bar v1\n\nreplace foo => ../bar\n` without `testctx.Snapshot`, asserting `require.ErrorIs(t, CheckGoModPipe{}.Run(ctx), ErrReplaceWithProxy)`. As written, this branch fails that test. Worth a second one using a `replace (...)` block, since block handling is what #7096 added for #6892.

Minor, while you are here: this subtest writes to `filepath.Join(dir, "go.mod")` whereas every sibling subtest writes the relative `"go.mod"`. Equivalent, since `testlib.Mktmp` chdirs, but inconsistent.

caarlos0 (commented):
Reviewed the full diff against `origin/main` (merge-base `7ba40dc`). The change does fix the regression it targets, and the new test is decisive — I confirmed it fails on the base version with `could not parse "go.mod": go.mod:5: unknown directive: futuredirective`, and that it is hermetic (no `go` subprocess, no network, passes under `-shuffle=on`).

I also confirmed both claims in the description against `golang.org/x/mod v0.41.0`: `Parse` returns `nil, ErrorList` on an unknown directive (`rule.go:323-326`, `rule.go:378`), and `ParseLax` really does drop `replace` (`rule.go:340-354`), so it is not a drop-in.

One finding: in the scenario this PR is written for, the `replace` check stops working entirely rather than degrading. Details inline.

**Verification gaps**

- No currently released Go toolchain emits a directive that x/mod v0.41.0 rejects — `tool`, `ignore` and `godebug` all parse fine. Both this PR and my finding are forward-looking and rest on the same premise: goreleaser's x/mod lagging the user's Go. Reproduced against a cached x/mod v0.20.0, where `ignore` fails to parse while a sibling `replace` is present.
- Pipe-level verification only; no end-to-end release run.

_Automated review — I am a bot._

internal/pipe/gomod/gomod_proxy.go
caarlos0: Fixed in 14c65e401.

Took the suggested route, with one simplification: rather than keeping strict `Parse` and falling back to `ParseLax` on error, the check now always uses `ParseLax` and reads the replaces off `file.Syntax.Stmt`. Strict parsing was never wanted here — the pipe only needs to know whether a `replace` exists and how to print it — so a single code path avoids the two formatting paths diverging. `formatReplace` is gone, since the syntax tree already carries the directive as written.

Warn-and-skip now only happens when `ParseLax` itself fails, i.e. the file is not a valid `go.mod` at all (`syntax error (unterminated block ...)`), and `gomod.Pipe` already aborts on that earlier via `go list -m` with a better message.

Verified against x/mod v0.41.0 — unknown directive plus a replace, single-line and block:

```
unknown+replace   err=<nil> replaces=["replace example.invalid/a => ../a"]
unknown+block     err=<nil> replaces=["replace example.invalid/a => ../a"]
versioned replace err=<nil> replaces=["replace example.invalid/a v1.0.0 => example.invalid/b v1.2.3"]
unterminated blk  err=go.mod:4: syntax error (unterminated block started at go.mod:3:1)
```

internal/pipe/gomod/gomod_proxy_test.go
caarlos0: Added in 14c65e401: `replace with unknown directive` and `replace block with unknown directive`, both non-snapshot, both asserting `require.ErrorIs(..., ErrReplaceWithProxy)`.

Confirmed decisive — running them against the previous commit via `go test -overlay` (so the worktree stayed clean):

```
--- FAIL: TestCheckGoMod/replace_with_unknown_directive
    Expected error with "cannot use the go.mod replace directive with go mod proxy enabled" in chain but got nil.
--- FAIL: TestCheckGoMod/replace_block_with_unknown_directive
    Expected error with "cannot use the go.mod replace directive with go mod proxy enabled" in chain but got nil.
```

Both pass on this branch; all 12 subtests green.

Also: renamed the original subtest to `unknown directive` (what it actually covers) and repurposed `unparseable go mod` to a genuinely invalid file (`module foo\n\nreplace (\n`) so the warn-and-skip branch is covered too. Nit fixed — it now writes the relative `"go.mod"` with `0o666` like its siblings.

caarlos0:
honestly IMHO this is a non-issue.

PR #7144, opened 2026-09-15
fix(sbom,sign): disambiguate every target variant in default names

caarlos0:
Include every target variant in the default SBOM document and binary signature names, via a new `{{ targetVariant . }}` template function.

#7089 taught `GroupByPlatform` about `Goarm64`, `Go386`, `Goppc64`, `Goriscv64` and `Abi`, but the default name templates touched in the sibling #7091 only gained `.Arm`, `.Mips` and `.Amd64`.

So with `goarm64: [v8.0, v9.0]` plus `sboms: [{artifacts: binary}]`, both binaries resolve to the same `.sbom.json` path: the second syft run overwrites the first, and two `artifact.SBOM` entries with identical `Name`/`Path` get registered and later uploaded twice. `defaultSignatureName` had the same gap — there the signature *files* are fine, since `signone` renders the path from `env["artifact"] = art.Path`, which is already per-target, but the registered artifact name is ambiguous, so only one of the two survives as a release asset.

Unlike `archive.create`, neither pipe has an "already exists" guard, so this failed **silently** rather than erroring.

## `{{ targetVariant . }}`

Rather than grow the inline suffix in each default, this adds a template function:

```
{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ targetVariant . }}.sbom.json
```

The suffix was already duplicated between these two pipes, and `archive`, `nfpm`, `flatpak`, `makeself` and `snapcraft` each carry their own older copy, so a seventh copy was the direction of travel.

More importantly it is available to users, which matters because of the scope limit below: until those pipes are fixed, `{{ targetVariant . }}` is the workaround for colliding archive names.

```yaml
archives:
  - name_template: >-
      {{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ targetVariant . }}
```

Being a function also makes it degrade gracefully: on a template with no artifact in scope it renders empty, where the equivalent template string hit `missingkey=error`.

It is `targetVariant` rather than `archVariant` because the ABI is part of the target, not of the architecture: the rust builder parses `x86_64-unknown-linux-musl` into `Abi: musl` with the arch unchanged, so that target and its `gnu` sibling share an architecture.

Two details worth calling out:

- `goarm64` accepts feature suffixes (`^(v8\.[0-9]|v9\.[0-5])(,lse|,crypto)*$`), so a naive `{{ .Arm64 }}` would put a comma straight into a release asset name. The comma is replaced with a dash rather than stripped, because stripping would make `v9.0` and `v9.0,lse` collide again.
- `Abi` counts as a platform dimension for `GroupByPlatform`, and zig/rust targets `x86_64-linux-gnu` and `x86_64-linux-musl` both resolve to `Goos: linux, Goarch: amd64` with the ABI only in `Extra`. `tmpl.WithArtifact` did not expose `.Abi` at all, so users could not work around it with a custom template either — with `missingkey=error` it failed with `map has no entry for key "Abi"`. `.Abi` is now a documented single-artifact field.

## Compatibility

Names are unchanged for default builds — the omitted values are exactly the Go builder defaults in `internal/builders/golang/build.go`, confirmed by the untouched golden files.

They **do** change for anyone who pins a non-default variant, even with a single target and no collision to fix:

| config | before | after |
| --- | --- | --- |
| `go386: [softfloat]` | `foo_1.0.0_linux_386` | `foo_1.0.0_linux_386softfloat` |
| `goarm64: [v9.0]` | `foo_1.0.0_linux_arm64` | `foo_1.0.0_linux_arm64v9.0` |
| rust `x86_64-unknown-linux-gnu` | `foo_1.0.0_linux_amd64` | `foo_1.0.0_linux_amd64_gnu` |
| go `goarm: ["7,softfloat"]` | `foo_1.0.0_linux_armv7` | `foo_1.0.0_linux_armv7_softfloat` |

The `.Abi` row has the widest blast radius, since every Rust `*-linux-gnu` build carries one — but it is also the only way to tell gnu and musl apart. This affects the SBOM document and binary signature names only; it needs a changelog note.

## Known scope limit

`internal/pipe/archive/archive.go`, `nfpm`, `flatpak`, `makeself` and `snapcraft` still carry the old `Arm`/`Mips`/`Amd64`-only suffix, so the same configuration still collides there — and `skip()`, used by `formats: [binary]`, has no "already exists" guard either, unlike `create()`. Left for a follow-up, since changing archive names is a much larger compatibility decision than changing SBOM and signature names. `{{ targetVariant . }}` is the workaround in the meantime.

## Tests

`tmpl.TestVariant` covers the function directly: every knob at its toolchain default and at a non-default value, the always-rendered `Goarm`/`Gomips`, ABI, comma-bearing feature suffixes, a template with no artifact in scope, and misuse.

The sbom and sign pipe tests cover only what they alone can: that each default name is wired to `{{ targetVariant . }}`, and that the whole rendered name stays distinct and free of commas. Against the old templates they fail with:

```
variants share a document name: [mybin_1.0.0_linux_arm64.sbom.json mybin_1.0.0_linux_arm64.sbom.json]
```

The extracted function was diffed against the template string it replaces over all 57600 combinations of the eight knobs, to confirm rendering is byte-identical.

Follow-up to #7089 and #7091. Part of a series of fixes found reviewing `v2.18.1..main`.


internal/pipe/sbom/sbom.go
```diff
@@ -29,6 +29,19 @@ import (
...
 
+// variantSuffix disambiguates binaries that differ only by CPU variant.
+// Without it two binaries for the same os/arch resolve to the same document
+// name, and the second syft run silently overwrites the first. Each variant
+// is omitted when it is the Go toolchain default, to keep names stable for
+// the common case.
+const variantSuffix = `{{ with .Arm }}v{{ . }}{{ end }}` +
+	`{{ with .Mips }}_{{ . }}{{ end }}` +
+	`{{ if not (eq .Amd64 "v1") }}{{ .Amd64 }}{{ end }}` +
+	`{{ if not (eq .Arm64 "v8.0") }}{{ .Arm64 }}{{ end }}` +
```
caarlos0: `validGoarm64` accepts optional feature suffixes: `^(v8\.[0-9]|v9\.[0-5])(,lse|,crypto)*$` (`internal/builders/golang/targets.go:424`), and `Goarm64` keeps the whole string on the artifact (`internal/builders/golang/build.go:87`). So `goarm64: ["v8.0,lse"]` renders `mybin_1.0.0_linux_arm64v8.0,lse.sbom.json` — a comma in a release asset name and in a URL, for the first time. Until now the comma stayed inside the `dist/` folder name.

Note `.Arm` deliberately avoids this: `fullGoarm()` keeps `7,softfloat` for the target directory, but the artifact's `Goarm` is just `7` (`internal/builders/golang/targets.go:83-86`), so the ABI never reaches the name. `.Arm64` has no equivalent stripping.

Smallest fix: emit only the part before the comma, e.g. `{{ with .Arm64 }}{{ if not (eq . "v8.0") }}{{ index (split . ",") 0 }}{{ end }}{{ end }}`. That pairs with the `.Abi` point on line 37, since `lse`/`crypto` is the same kind of sub-variant.

Test needed: add `{Goos: "linux", Goarch: "arm64", Goarm64: "v8.0,lse"}` vs `"v9.0,lse"` to the new table and assert the rendered names contain no comma.

internal/pipe/sign/sign_binary.go
```diff
@@ -13,7 +13,19 @@ import (
...
+const defaultSignatureName = `${artifact}_{{ .Os }}_{{ .Arch }}` + variantSuffix
+
+// variantSuffix disambiguates binaries that differ only by CPU variant.
+// Without it two binaries for the same os/arch resolve to the same signature
+// name and silently overwrite each other. Each variant is omitted when it is
+// the Go toolchain default, to keep names stable for the common case.
+const variantSuffix = `{{ with .Arm }}v{{ . }}{{ end }}` +
+	`{{ with .Mips }}_{{ . }}{{ end }}` +
+	`{{ if not (eq .Amd64 "v1") }}{{ .Amd64 }}{{ end }}` +
+	`{{ if not (eq .Arm64 "v8.0") }}{{ .Arm64 }}{{ end }}` +
```
caarlos0: Same comma leak as `internal/pipe/sbom/sbom.go:40`: `goarm64: ["v8.0,lse"]` produces the signature name `${artifact}_linux_arm64v8.0,lse`, which becomes a release asset name. Fix both constants together — see that comment for the reasoning and the suggested template.

internal/pipe/sbom/sbom.go
```diff
@@ -29,6 +29,19 @@ import (
 // Environment variables to pass through to exec
 var passthroughEnvVars = []string{"HOME", "USER", "USERPROFILE", "TMPDIR", "TMP", "TEMP", "PATH", "LOCALAPPDATA"}
 
+// variantSuffix disambiguates binaries that differ only by CPU variant.
+// Without it two binaries for the same os/arch resolve to the same document
+// name, and the second syft run silently overwrites the first. Each variant
+// is omitted when it is the Go toolchain default, to keep names stable for
+// the common case.
+const variantSuffix = `{{ with .Arm }}v{{ . }}{{ end }}` +
```
caarlos0: The premise of this PR is `GroupByPlatform`, and `GroupByPlatform` counts `Abi` as a distinguishing dimension (`internal/artifact/artifact.go:511`). `variantSuffix` does not, so the bug you are fixing is still reachable — and for a more common configuration than `goarm64: [v8.0, v9.0]`.

Zig (and rust/deno) targets `x86_64-linux-gnu` and `x86_64-linux-musl` (`internal/builders/zig/all_targets.txt:88,90`) both produce `Goos: linux, Goarch: amd64` with the ABI only in `Extra` (`internal/builders/zig/build.go:120-131`). Both render `mybin_1.0.0_linux_amd64.sbom.json`, so the second syft run overwrites the first — exactly the failure this PR set out to remove. The Go builder is safe here only because `checkGoarmConflict` (`internal/builders/golang/targets.go:175-188`) rejects same-GOARM different-ABI targets.

Worse, users cannot work around it: `tmpl.WithArtifact` (`internal/tmpl/tmpl.go:198-214`) does not expose `Abi`, and with `missingkey=error` (`internal/tmpl/tmpl.go:274`) a custom `{{ .Abi }}` document template fails with `map has no entry for key "Abi"`.

Smallest fix: add `"Abi": artifact.ExtraOr(*a, "Abi", "")` to `tmpl.WithArtifact`, then append `{{ with .Abi }}_{{ . }}{{ end }}` to the suffix.

Test needed: a zig-style pair — `{Goos: "linux", Goarch: "amd64", Extra: {"Abi": "gnu"}}` vs `"musl"` — in both new tables.

internal/pipe/sign/sign_binary.go
```diff
@@ -13,7 +13,19 @@ import (
 	"github.com/goreleaser/goreleaser/v2/pkg/context"
 )
 
-const defaultSignatureName = `${artifact}_{{ .Os }}_{{ .Arch }}{{ with .Arm }}v{{ . }}{{ end }}{{ with .Mips }}_{{ . }}{{ end }}{{ if not (eq .Amd64 "v1") }}{{ .Amd64 }}{{ end }}`
+const defaultSignatureName = `${artifact}_{{ .Os }}_{{ .Arch }}` + variantSuffix
+
+// variantSuffix disambiguates binaries that differ only by CPU variant.
+// Without it two binaries for the same os/arch resolve to the same signature
+// name and silently overwrite each other. Each variant is omitted when it is
```
caarlos0: "silently overwrite each other" is not what happens, and the wrong mechanism here will mislead whoever debugs this next.

`signone` sets `env["artifact"] = art.Path` before rendering the signature path (`internal/pipe/sign/sign.go:190`), and a binary's path is `dist/<id>_<target>/<name>` (`internal/builders/base/options.go:26-30`) where `<target>` already encodes every variant (`internal/builders/golang/targets.go:33-39`). So the two signature *files* landed in different directories and never overwrote each other.

What actually collided is the registered artifact: the name is re-rendered with `env["artifact"] = art.Name` (`sign.go:294`) and stored as `artifact.Signature.Name` (`sign.go:306-309`) while `Path` stays unique. Two signatures with distinct paths and the same name are then published under one asset name, so one is clobbered at upload and the checksums entry cannot say which variant it covers. Nothing is lost on disk; it is lost at publish time.

Smallest fix: reword to name ambiguity at upload rather than file overwrite. The `sbom.go:32-36` comment is accurate as written — only this one is wrong.

internal/pipe/sbom/sbom.go
```diff
@@ -68,7 +81,7 @@ func setConfigDefaults(cfg *config.SBOM) error {
 	if len(cfg.Documents) == 0 {
 		switch cfg.Artifacts {
 		case "binary":
-			cfg.Documents = []string{`{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ with .Arm }}v{{ . }}{{ end }}{{ with .Mips }}_{{ . }}{{ end }}{{ if not (eq .Amd64 "v1") }}{{ .Amd64 }}{{ end }}.sbom.json`}
+			cfg.Documents = []string{`{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}` + variantSuffix + `.sbom.json`}
```
caarlos0: This changes the documented default, but `www/content/customization/sbom.md:43` still spells out the old three-knob template verbatim under `# Default: When "binary":`. Users copy that line into their config, so it is now actively wrong. Please update it in this PR (and see the sibling note on `internal/pipe/sign/sign_binary.go:16`).

internal/pipe/sign/sign_binary.go
```diff
@@ -13,7 +13,19 @@ import (
 	"github.com/goreleaser/goreleaser/v2/pkg/context"
 )
 
-const defaultSignatureName = `${artifact}_{{ .Os }}_{{ .Arch }}{{ with .Arm }}v{{ . }}{{ end }}{{ with .Mips }}_{{ . }}{{ end }}{{ if not (eq .Amd64 "v1") }}{{ .Amd64 }}{{ end }}`
+const defaultSignatureName = `${artifact}_{{ .Os }}_{{ .Arch }}` + variantSuffix
```
caarlos0: `www/content/customization/sign/binary_sign.md:36` documents the old value of this constant verbatim (`# Default: '${artifact}_{{ .Os }}_{{ .Arch }}{{ with .Arm }}...'`) and is now wrong. Same for the SBOM default — see `internal/pipe/sbom/sbom.go:84`.

internal/pipe/sbom/sbom_test.go
```diff
@@ -1046,3 +1046,50 @@ func TestDependencies(t *testing.T) {
...
+		t.Run(name, func(t *testing.T) {
+			var got []string
+			for _, art := range variants {
+				art.Name = "mybin"
+				art.Extra = map[string]any{artifact.ExtraBinary: "mybin"}
+				name, err := tmpl.New(ctx).WithArtifact(&art).Apply(document)
+				require.NoError(t, err)
+				got = append(got, name)
+			}
+			require.Len(t, slices.Compact(slices.Clone(got)), len(got), "variants share a document name: %v", got)
```
caarlos0: `slices.Compact` removes only *consecutive* duplicates, so this assertion holds for today's two-element inputs but silently stops detecting collisions the moment a third variant is added: `["a", "b", "a"]` compacts to length 3 and passes. Given the point of the test is to keep catching this class of bug as new variants appear (`goriscv64` already has three valid values), that is a trap.

Smallest fix: sort first, e.g. `require.Len(t, slices.Compact(slices.Sorted(slices.Values(got))), len(got), ...)`.

Minor, same line: `name, err := tmpl...` shadows the map key `name` used as the subtest name, so the failure message loses the knob it came from.

internal/pipe/sign/sign_binary_test.go
```diff
@@ -351,3 +353,45 @@ func TestBinarySignUniversalBinaryReplaced(t *testing.T) {
...
+	} {
+		t.Run(name, func(t *testing.T) {
+			var got []string
+			for _, art := range variants {
+				art.Name = "mybin"
+				name, err := tmpl.New(ctx).WithArtifact(&art).Apply(defaultSignatureName)
+				require.NoError(t, err)
+				got = append(got, name)
+			}
+			require.Len(t, slices.Compact(slices.Clone(got)), len(got), "variants share a signature name: %v", got)
```
caarlos0: Same consecutive-duplicates limitation and same `name` shadowing as `internal/pipe/sbom/sbom_test.go:1092` — fix both together.

Also worth noting for the table above: the `goamd64` subtest passes against the pre-fix template too (`.Amd64` was already in the old suffix), so it is not a regression test. Only `goarm64`, `go386`, `goppc64` and `goriscv64` fail without the fix — I confirmed that by rendering the old template strings against these same inputs.

caarlos0 (commented):
Automated review (GitHub Copilot CLI bot).

The SBOM half of this is a real fix and I reproduced it: rendering the pre-fix template for `goarm64`, `go386`, `goppc64` and `goriscv64` pairs yields identical names, the document path is joined flat onto `ctx.Config.Dist` (`internal/pipe/sbom/sbom.go:264-266`), so the second syft run really did overwrite the first. `goamd64` already passed before, so four of the five new subtests are genuine regression tests and one is not. `go build ./...` is clean and both new tests pass; `internal/pipe/sign` has pre-existing gpg-environment failures in `TestBinarySign/default`, `/filter` and `TestSignArtifacts/*` that reproduce identically on the base commit and are unrelated to this diff.

Two blocking items (posted as a comment review because the API will not let this account request changes on its own PR): a comma can now reach a release asset name for the first time, and the `.Abi` case this PR's own premise covers is still broken with no user workaround. Both inline.

**Cross-cutting: the fix stops halfway through the pipeline.**

`internal/pipe/archive/archive.go:30`, `internal/pipe/nfpm/nfpm.go:38`, `internal/pipe/flatpak/flatpak.go:40`, `internal/pipe/makeself/makeself.go:31` and `internal/pipe/snapcraft/snapcraft.go:106` still carry the old Arm/Mips/Amd64-only suffix. The PR body says `archive.create` errors on a duplicate, but that guard is only in `create()` (`archive.go:187`); `skip()` (`archive.go:311`), used by `formats: [binary]`, has none. So with `goarch: [arm64]`, `goarm64: [v8.0, v9.0]`, `archives: [{formats: [binary]}]` you still get two `UploadableBinary` artifacts named `foo_1.0.0_linux_arm64`, silently — and after this PR the SBOM describing them is named `foo_1.0.0_linux_arm64v9.0.sbom.json`, so the SBOM no longer matches the artifact it documents. Either share one suffix across all seven defaults, or reject the configuration the way `checkGoarmConflict` (`internal/builders/golang/targets.go:175-188`) already does for GOARM. Test needed: an archive-pipe case with two `goarm64` values and `formats: [binary]` asserting two distinct artifact names.

**Compatibility: the "names are unchanged" claim in the description is not accurate.** It holds for default builds — the omitted values match `internal/builders/golang/build.go:31-36` exactly — but not for a user who pins a single non-default variant. `go386: [softfloat]` alone (one target, no collision to fix) renames the SBOM from `foo_1.0.0_linux_386.sbom.json` to `foo_1.0.0_linux_386softfloat.sbom.json`, while the binary asset stays `foo_1.0.0_linux_386` because `archive.go:30` is untouched. Same for `goarm64: [v9.0]`, `goppc64: [power10]`, `goriscv64: [rva22u64]`. That breaks install scripts and any tool that derives the SBOM URL from the binary URL. It needs a changelog note at minimum. The golden files staying untouched does not confirm the claim — no golden or schema file contains a rendered name, so they could not have caught this.

**Verification gaps I could not close:** I did not run the full suite (machine load), only `./internal/pipe/sbom/...`, `./internal/pipe/sign/...`, `./internal/tmpl/...` and `go build ./...`. I did not exercise a real syft or gpg run end to end; the duplicate-asset consequence for signatures is traced through `signone`, not observed against a live release.

internal/pipe/sbom/sbom.go
caarlos0: Fixed in 6bd8585 — but not the way suggested.

`index (split . ",") 0` would have reintroduced the very collision this PR exists to fix: `goarm64: [v9.0, "v9.0,lse"]` would both strip down to `v9.0` and share a name again. Stripping is only safe for `v8.0,lse` because the bare `v8.0` is the omitted default.

So the comma is replaced rather than dropped: `{{ if not (eq .Arm64 "v8.0") }}{{ replace .Arm64 "," "-" }}{{ end }}`. `v9.0,lse` → `v9.0-lse`, `v9.0,lse,crypto` → `v9.0-lse-crypto`, all distinct, no comma in the asset name.

Covered by a new `goarm64 features` case with three values (`v9.0`, `v9.0,lse`, `v9.0,lse,crypto`) plus a `require.NotContains(t, rendered, ",")` on every rendered name in both tables. The three-value case is also what makes the `slices.Compact` fix on the assertion load-bearing.

internal/pipe/sign/sign_binary.go
caarlos0: Fixed in 6bd8585, same `replace .Arm64 "," "-"` as `internal/pipe/sbom/sbom.go` — see that thread for why replacing beats stripping. The signature table gained the same `goarm64 features` case and the same no-comma assertion.

internal/pipe/sbom/sbom.go
caarlos0: Fixed in 6bd8585, as suggested.

`internal/tmpl/tmpl.go` now has `KeyAbi` alongside the other target keys, `WithArtifact` populates it from `artifact.ExtraOr(*a, KeyAbi, "")`, and `buildOptsToFields` zeroes it with the rest so builders that expose no ABI cannot break templates. Both suffixes end with `{{ with .Abi }}_{{ . }}{{ end }}` — underscore-separated like `.Mips`, since ABIs are words (`gnu`, `musl`, `softfloat`) rather than version tokens.

Both tables gained an `abi` case: two `linux/amd64` artifacts whose only difference is `Extra{"Abi": "gnu"}` vs `"musl"`, which is the zig/rust gnu-vs-musl pair. It fails without the fix (both render `linux_amd64`).

Two consequences worth stating plainly rather than burying:

- `.Abi` is now a real single-artifact template field, documented in `www/content/customization/general/templates.md`, so the `missingkey=error` dead end is gone and a custom template can use it too.
- This renames artifacts for anyone already building an ABI-bearing target — every Rust `*-linux-gnu` build gets `_gnu`, and Go `goarm: ["7,softfloat"]` gets `_softfloat`. That is the same trade already made for `.Arm64`/`.I386`/`.Ppc64`/`.Riscv64` here, and it is the only way to make gnu/musl distinct, but it is a bigger blast radius than the rest of this PR and I have called it out in the description.

internal/pipe/sign/sign_binary.go
caarlos0: Fixed in 6bd8585. You are right, and the trace is right: `signone` renders the path from `env["artifact"] = art.Path`, which is already `dist/<id>_<target>/<name>`, so the files never collided. The comment now says what actually happens — the registered artifact name is ambiguous, so only one of them survives as a release asset, and the loss is at upload rather than on disk. The `sbom.go` comment was accurate and is unchanged apart from the new Goarm64 note.

internal/pipe/sbom/sbom.go
caarlos0: Fixed in 6bd8585. `www/content/customization/sbom.md` now spells out the current default, including `replace .Arm64 "," "-"` and the `.Abi` clause.

internal/pipe/sign/sign_binary.go
caarlos0: Fixed in 6bd8585. `www/content/customization/sign/binary_sign.md` now matches the constant. `www/content/customization/general/templates.md` also gained an `.Abi` row, since this PR makes it available to artifact templates for the first time.

internal/pipe/sbom/sbom_test.go
caarlos0: Fixed in 6bd8585: `slices.Compact(slices.Sorted(slices.Values(got)))`.

This was not hypothetical — the new `goarm64 features` case added for the comma fix has three values, so the old assertion would have started missing collisions in this very PR.

The shadowing is gone too: the rendered name is now `rendered`, so the map key `name` survives into the failure message and says which knob collided.

internal/pipe/sign/sign_binary_test.go
caarlos0: Fixed in 6bd8585 alongside `internal/pipe/sbom/sbom_test.go` — sorted before compacting, and `rendered` instead of shadowing `name`.

Agreed on `goamd64`: it passes against the old template and is a guard, not a regression test. Keeping it, since the point of the table is that every knob stays covered as new ones are added, but the genuinely new coverage is `goarm64`, `go386`, `goppc64`, `goriscv64`, and now `goarm64 features` and `abi`.

PR #7145, opened 2026-09-15
test(gitea): cover release lookup for tags containing a slash

caarlos0:
Cover release lookup for tags containing a slash.

This PR originally claimed Gitea could not find a release whose tag contains a slash, and added a `ListReleases` fallback. **That premise was wrong**, so the fallback is gone and only a regression test remains.

Gitea routes on the *escaped* path and unescapes the path parameter, so the `%2F` the SDK writes for `release/1.2.3` survives routing as a single segment and is decoded back for the lookup. Verified against live servers with the exact URL `code.gitea.io/sdk/gitea v0.25.1` builds:

```
GET gitea.com/api/v1/repos/gitea/go-sdk/releases/tags/gitea%2Fv0.25.1
200 {"id":868261,"tag_name":"gitea/v0.25.1",...}

GET codeberg.org/api/v1/repos/uzu/strudel/releases/tags/@strudel%2Fcore@1.2.4
200 {"id":4558522,"tag_name":"@strudel/core@1.2.4",...}
```

Probing the mechanism on both instances:

| request | result | meaning |
|---|---|---|
| `releases/tags/foo/bar` | `404 page not found` (text/plain) | no route matched — single-segment route |
| `releases/tags/foo%2Fbar` | `404 {"message":"not found"}` (JSON) | route matched, handler ran |
| `releases/tags/v0%2E9%2E2` | `200` real release | params *are* unescaped |

In Gitea's source this is `routers/common/middleware.go` → `ChiRoutePathHandler()` ("make sure chi uses EscapedPath(RawPath) as RoutePath, then `%2f` could be handled correctly") plus `services/context/base_path.go` → `PathParam()` doing `url.PathUnescape(chi.URLParam(...))`.

The original test only passed because its fake server switched on `r.URL.Path`, which Go has already decoded — a naive `net/http` server, not Gitea.

So `dce409c02` (#7097) is correct as-is and nothing needs fixing. What is left is one test that routes on `EscapedPath()` like Gitea does, asserts a slash tag is found via `GetReleaseByTag`, and asserts the release list is never hit — so it fails if the fallback is ever reintroduced.

Also worth noting: the SDK already degrades on its own. `GetReleaseByTag` falls back to paging `ListReleases` whenever `checkServerVersionGreaterThanOrEqual(version1_13_0)` fails, so servers that genuinely cannot serve the tag route are covered upstream.

Follow-up to #7097.


caarlos0 (commented):
The premise is wrong: Gitea **does** find releases whose tag contains a slash via `GetReleaseByTag`. I verified this against live servers using the exact URL the pinned SDK builds.

**Live gitea.com (1.27.0+dev)** — the SDK's `url.PathEscape`'d tag:

```
GET /api/v1/repos/gitea/go-sdk/releases/tags/gitea%2Fv0.25.1
200 {"id":868261,"tag_name":"gitea/v0.25.1",...}
```

**Live codeberg.org (Forgejo, `gitea-1.22.0` compat — the version this PR's mock reports)**:

```
GET /api/v1/repos/uzu/strudel/releases/tags/@strudel%2Fcore@1.2.4
200 {"id":4558522,"tag_name":"@strudel/core@1.2.4",...}
```

And end-to-end through `code.gitea.io/sdk/gitea v0.25.1` itself, the exact call `getExistingRelease` makes:

```
tag="gitea/v0.25.1"        status=200 FOUND id=868261 tag_name="gitea/v0.25.1"
tag="gitea/v0.0.0-nope"    status=404 err=not found
```

The mechanism, probed on both instances: Gitea routes on the **escaped** path and then unescapes the param.

| request | result | meaning |
|---|---|---|
| `releases/tags/foo/bar` | `404 page not found` (text/plain) | no route matched — single-segment route |
| `releases/tags/foo%2Fbar` | `404 {"message":"not found"}` (JSON) | route matched, handler ran |
| `releases/tags/v0%2E9%2E2` | `200` real `v0.9.2` release | params **are** unescaped |

So `%2F` survives routing as one segment and is decoded back to `/` for the lookup. `dce409c02` (#7097) is correct as-is, and the 404-means-no-release fallthrough described in the PR body cannot happen.

Two independent checks reached the same conclusion from Gitea's source: `routers/common/middleware.go` → `ChiRoutePathHandler()` ("make sure chi uses EscapedPath(RawPath) as RoutePath, then `%2f` could be handled correctly") and `services/context/base_path.go` → `PathParam()` doing `url.PathUnescape(chi.URLParam(...))`, unchanged since ~v1.16.

My recommendation is to close this. If you have a real reproduction, the likely cause is a reverse proxy in front of Gitea normalizing or rejecting `%2F` (Apache's `AllowEncodedSlashes` defaults to `Off`) — that is a 404 symptom, not a tag shape, and it wants the design in the inline comment on `gitea.go` rather than this gate.

**Verification gaps:** I could not test a Gitea older than the two live instances above, nor a deployment behind a proxy that rewrites `%2F`.

*Reviewed by a bot. Posted as a comment rather than "request changes" because GitHub does not allow requesting changes on your own pull request.*

internal/client/gitea.go
```diff
@@ -253,6 +253,12 @@ func (c *giteaClient) createRelease(ctx *context.Context, title, body string) (*
 }
 
 func (c *giteaClient) getExistingRelease(ctx *context.Context, owner, repoName, tagName string) (*gitea.Release, error) {
+	// the SDK escapes the tag into a single path segment, and Gitea routes
+	// the tag as a single segment too, so a tag containing a slash can never
+	// be found by name and has to be looked up in the release list.
+	if strings.Contains(tagName, "/") {
```
caarlos0: This gate is based on a premise I could not reproduce, and the comment states it as fact. Verified against live gitea.com and codeberg.org, plus the pinned SDK: `GetReleaseByTag("gitea", "go-sdk", "gitea/v0.25.1")` returns `200` with `tag_name="gitea/v0.25.1"`. A slash tag *can* be found by name — Gitea routes on the escaped path (so `%2F` stays one segment) and unescapes the param before lookup. Details and the probe table are in the review body.

**Impact:** every slash tag now pays N sequential `ListReleases` round-trips instead of one request, to work around a 404 that does not occur. Monorepo prefix tags are exactly the repos with the most releases, so the cost lands where it hurts most.

**Smallest fix:** drop this gate and `findReleaseByTag`, i.e. revert both files.

If you are chasing a real report, the plausible cause is a reverse proxy decoding or rejecting `%2F` before Gitea ever sees it (Apache defaults to `AllowEncodedSlashes Off`). That is a *404 symptom*, not a tag shape, so the condition belongs on the response, not on the tag: keep `GetReleaseByTag` first for every tag, and only when it 404s **and** the tag contains a slash, page the list. Normal tags then cost zero extra requests, and a proxy-mangled slash tag still resolves.

**Test needed for that variant:** a fake server that 404s the `releases/tags/` route and then serves the release from `releases`, asserting both endpoints were hit — which is also the test the current `default:` branch cannot provide (see the comment at line 376).

internal/client/gitea_test.go
```diff
@@ -351,6 +352,73 @@ func TestGiteaGetExistingReleaseByTag(t *testing.T) {
...
+				fmt.Fprint(w, `[{"id":1,"tag_name":"other/1.0.0"}]`)
+			case "2":
+				fmt.Fprint(w, `[{"id":123,"tag_name":"release/1.2.3"}]`)
+			default:
+				fmt.Fprint(w, `[]`)
+			}
+		default:
+			// this is what gitea does with the escaped tag: no route matches
+			// `releases/tags/release/1.2.3`.
+			http.Error(w, "unexpected "+r.URL.Path, http.StatusNotFound)
```
caarlos0: This is the core problem with the test: the comment says "this is what gitea does with the escaped tag", but the handler switches on `r.URL.Path`, which Go has already decoded. That models a naive `net/http` server, not Gitea.

Gitea routes on `EscapedPath()`/`RawPath` (`routers/common/middleware.go` → `ChiRoutePathHandler`) and then unescapes the captured param (`services/context/base_path.go` → `PathParam` → `url.PathUnescape`). A faithful mock would switch on `r.URL.EscapedPath()`, match `/api/v1/repos/owner/repo/releases/tags/release%2F1.2.3`, unescape to `release/1.2.3`, and serve the release — which is exactly what the live servers do (see the review body).

Because the mock bakes in the misconception, the test proves only that the new code is self-consistent; it cannot detect that the fallback is unnecessary.

**Smallest fix:** switch the handler on `r.URL.EscapedPath()`. That makes this test fail, which is the correct outcome.

internal/client/gitea_test.go
```diff
@@ -351,6 +352,73 @@ func TestGiteaGetExistingReleaseByTag(t *testing.T) {
...
+
+	ctx := testctx.WrapWithCfg(t.Context(), config.Project{
+		GiteaURLs: config.GiteaURLs{API: srv.URL},
+	})
+	client, err := newGitea(ctx, "giteatoken")
+	require.NoError(t, err)
+
+	release, err := client.getExistingRelease(ctx, "owner", "repo", "release/1.2.3")
+	require.NoError(t, err)
+	require.Nil(t, release)
```
caarlos0: This test guards nothing — it passes without the production change. I reverted just the `strings.Contains` gate in a scratch copy of the worktree and ran both new tests:

```
--- FAIL: TestGiteaGetExistingReleaseBySlashTag
--- PASS: TestGiteaGetExistingReleaseBySlashTagNotFound
```

Without the fix, `getExistingRelease` takes the `GetReleaseByTag` path, the `default:` branch 404s, and line 266 maps that to `nil, nil`. With the fix it returns `nil, nil` from the empty-list branch. The assertions cannot tell the two apart.

**Smallest fix:** if any fallback survives review, assert the list endpoint was actually reached — record requests and assert on them, the way the sibling test does with `pages` at line 394. Otherwise drop this test.

internal/client/gitea.go
```diff
@@ -265,6 +271,27 @@ func (c *giteaClient) getExistingRelease(ctx *context.Context, owner, repoName,
...
+			return c.client.ListReleases(owner, repoName, gitea.ListReleasesOptions{
+				Page: page,
+			})
+		})
+		if err != nil {
+			return nil, err
+		}
+		if len(releases) == 0 {
+			return nil, nil
+		}
```
caarlos0: Only relevant if a fallback survives review (see the comment at line 256), but the pagination has two avoidable problems, and this repo already has the right pattern in `internal/client/github.go:924-949` (`findDraftRelease`).

1. **`PageSize` is unset** (line 277-279), so the server default applies — Gitea's `DefaultPagingNum` is 30. Looking up a *new* slash tag never matches, so it scans every page: ~50 sequential requests for a repo with 1500 releases, each through the `giteaDo` retry wrapper. `findDraftRelease` uses `PerPage: 50`.

2. **Termination depends on the server returning an empty page** (line 284). Any server that ignores `page` and keeps returning results loops forever. Gitea sends the pagination headers, so use them: stop when `resp.NextPage == 0` instead of discarding `resp` at line 276. That guarantees an end and saves the extra trailing request. `findDraftRelease` does exactly this.

**Test needed:** a fake server whose last page is full (no trailing empty page) and where the tag is absent — it must terminate and return `nil, nil` rather than spin.

internal/client/gitea.go
caarlos0: Agreed, and thanks for the live-server evidence — I reproduced it too. Reverted in 36f572bea.

The gate and `findReleaseByTag` are gone; `getExistingRelease` is back to `dce409c02` exactly (verified: `git diff <merge-base> -- internal/client/gitea.go` is empty, so no upstream code was mixed into the revert).

I did not adopt the 404-gated variant. There is no issue behind this PR — it came out of my own `v2.18.1..main` read, not a report — so a fallback for a proxy that mangles `%2F` would be exactly the speculative defense the repo conventions say not to add. If someone hits it for real we will have a reproduction to design against.

Worth noting for the record: the SDK already degrades gracefully on its own. `GetReleaseByTag` calls `fallbackGetReleaseByTag` (which pages `ListReleases`) whenever `checkServerVersionGreaterThanOrEqual(version1_13_0)` fails, so servers that genuinely cannot do the tag route are already covered upstream.

internal/client/gitea_test.go
caarlos0: Correct — switching on `r.URL.Path` modelled a naive `net/http` server and baked the misconception into the fixture. Fixed in 36f572bea.

The replacement test routes on `r.URL.EscapedPath()` the way Gitea does, so the SDK's `%2F` arrives as one segment:

```go
case "/api/v1/repos/owner/repo/releases/tags/release%2F1.2.3":
	fmt.Fprint(w, `{"id":123,"tag_name":"release/1.2.3"}`)
```

Rather than deleting the coverage I inverted it: the test now asserts that a slash tag is found through `GetReleaseByTag` and that the release list is never touched (`require.False(t, listed)`, matching the existing `TestGiteaGetExistingReleaseByTag` pattern). So it documents the real behaviour and fails if the fallback is ever reintroduced — verified by restoring the old `gitea.go` over the new test:

```
--- FAIL: TestGiteaGetExistingReleaseBySlashTag
    Error: Received unexpected error: unknown API error: 500
    Request: '/api/v1/repos/owner/repo/releases' with 'GET' method
```

internal/client/gitea_test.go
caarlos0: You are right that it could not fail — both paths returned `nil, nil` and the assertions could not tell them apart. Removed in 36f572bea.

I dropped it rather than strengthening it: with the fallback gone there is no slash-specific 404 path left to cover, and the generic not-found behaviour is already exercised by `GetExistingReleaseSuite`. The one remaining slash test now asserts on the request actually made, so it cannot go vacuous the same way.

internal/client/gitea.go
caarlos0: Moot now — `findReleaseByTag` was removed entirely in 36f572bea, so there is no unbounded loop and no unset `PageSize` left.

Noting the substance for the future in case a fallback is ever justified by a real report: `internal/client/github.go` `findDraftRelease` is the pattern to copy (`PerPage: 50`, terminate on `resp.NextPage == 0` rather than on an empty page).

PR #7146, opened 2026-09-15
fix(nix): template the description only once

caarlos0:
Escape and quote the Nix `description` so it survives the final template pass as a literal.

`nix.Description` is templated in `doRun`, and then the whole rendered derivation is run through `tmpl.Apply` again in `doBuildPkg`. A description that *renders to* text containing `{{` fails that second pass with a template parse error.

This is exactly the bug the same-release #7086 fixed for brew and cask via `rubyString | printf "{{ %q }}"`; nix was left with the old shape.

The template now mirrors that approach:

```
description = {{ . | nixString | printf "{{ %q }}" }};
```

`nixString` becomes a template function that escapes for Nix and adds the quotes, and `printf "{{ %q }}"` wraps the result in an action so the final pass emits it verbatim.

It also now escapes newlines, carriage returns and tabs, bringing `nixString` to parity with `rubyString`. That is not needed to make the second pass parse — `%q` already escapes control characters inside the Go literal — but it prevents silent data loss further down: `doBuildPkg` sanitizes the rendered derivation line by line, so an unescaped multi-line description loses its carriage returns and trailing spaces. `line1\nline2\twith  \r\ntrailing` used to come out as a literal multi-line string with the `  \r` gone.

Output is unchanged for ordinary descriptions — the existing golden files all still pass untouched. The only differing output is the newline/tab case above, which is the point.

Tests:

- `TestPreparePkgDescriptionIsTemplatedOnce` covers a description coming from `{{ .Env.DESCRIPTION }}` that renders to `{{example}}`, plus one that needs escaping *and* has to survive the second pass, matching brew's `TestFormulaDescriptionIsTemplatedOnce`.
- `TestPreparePkgEscapesDescription` gains a `control characters` case pinning the new newline/CR/tab escaping. It fails without it.

Follow-up to #7082, mirroring #7086. Part of a series of fixes found reviewing `v2.18.1..main`.


caarlos0 (commented):
**Verdict: correct, no blocking issues.** The two-pass round-trip holds and the new test is a genuine regression test. Two non-blocking findings are inline; the rest has no single site.

**Premise corrections in the PR body / commit message** (prose only, code is fine):

1. "`nix.Description` is templated in `Default`" — it is not. `Default` never touches `Description`; the templating is in `doRun`, `internal/pipe/nix/nix.go:155-163`, via `tp.ApplyAll(&nix.Name, &nix.SkipUpload, &nix.Homepage, &nix.Description, &nix.Path, &nix.MainProgram)`. The test's hand-rolled `tmpl.New(ctx).ApplyAll(&nix.Description)` is a faithful mirror of that call, so the test is valid — but because it bypasses `doRun`, it would not catch someone dropping `&nix.Description` from that `ApplyAll` list.

2. "It also now escapes newlines/tabs, which is required since a template action cannot contain a literal newline" — this reason is false. `%q` always escapes the newline inside the Go literal, so pass 2 parses fine without the extra escaping. Measured: with a bare `\n`/`\r`/`\t` replacer omitted, pass 2 still parses. The escaping is still worth keeping, just for a different reason: `doBuildPkg`'s sanitizing scanner (`nix.go:517-530`) is what mangles the value. Input `"a  \r\nb"` comes out as `"a\nb"` without the escaping (`bufio.Scanner` eats the `\r`, `TrimRight(l, " ")` eats the trailing spaces) versus `"a  \r\nb"` with it. So it prevents silent data loss, and it brings `nixString` to parity with `rubyString` (`internal/pipe/brew/brew.go:382`).

**Scope (not introduced here, just noting the fix is narrower than the root cause).** `pname`, `homepage` and `mainProgram` are templated in the same `ApplyAll` and rendered with the same `"{{ . }}"` shape, so they still hard-error on the second pass. Confirmed on this branch with `homepage` set to an env var whose value contains `{{`:

```
homepage    err = template: failed to apply "...": function "example" not defined
mainProgram err = template: failed to apply "...": function "example" not defined
name        err = template: failed to apply "...": function "example" not defined
description err = <nil>                    <-- fixed by this PR
```

Sites: `internal/pipe/nix/tmpl.nix:93`, `:131`, `:137`. Same one-line remedy each. Fine to leave for the series, but `homepage: "{{ .Env.HOMEPAGE }}"` is a realistic trigger.

**Verification performed**

- Reproduced both template passes plus the trailing-whitespace scanner against 15 adversarial descriptions; all round-trip to the correct Nix literal with no parse error: embedded `"`, `\`, `${`, `$${`, `{{example}}`, a bare `}}`, `{{ and }}`, `{{- x -}}`, newline + tab, unicode, backtick, empty, whitespace-only, trailing space, `%q`/`%v` verbs, and a quoted action inside quotes. The `}}`-terminates-the-action-early worry is unfounded: the lexer consumes a quoted string atomically.
- Confirmed the new test fails at the merge base (`function "example" not defined`, both subtests) and passes on head.
- Confirmed output is byte-identical to base for every existing description shape; the only differing hunks are the intended newline and tab ones. No golden file needed updating — all goldens use `description = "my test";`.
- `go test ./internal/pipe/nix/... -count=1` → `ok`.

**Verification gap:** no Nix toolchain here, so "the emitted value is well-formed Nix" was checked against the manual's double-quoted-string rules, not `nix-instantiate`. NUL and form-feed bytes in a description are still emitted raw into the `.nix` file; that is unchanged from base and I did not test whether Nix's lexer accepts them.

_Reviewed by a bot (GitHub Copilot CLI)._

internal/pipe/nix/nix.go
```diff
@@ -623,11 +626,14 @@ func split(s string) []string {
...
 func nixString(s string) string {
-	return strings.NewReplacer(
+	return `"` + strings.NewReplacer(
 		"\\", "\\\\",
 		"\"", "\\\"",
 		"${", "\\${",
-	).Replace(s)
+		"\n", "\\n",
+		"\r", "\\r",
+		"\t", "\\t",
```
caarlos0, line 635: **The one output-changing part of this diff has no test.**

Everything else in the change is byte-identical to base — I diffed generated `.nix` output at base vs head across nine description shapes and the only differing hunks were the newline and tab ones produced by these three lines. So these three lines are the entire user-visible behavior change, and nothing asserts them.

Trigger: any `.goreleaser.yaml` with a multi-line description, e.g.

```yaml
nix:
  - description: |
      First line.
      Second line.
```

Impact if it regresses: without these escapes the value silently loses data on the way out — `doBuildPkg`'s sanitizing scanner (`nix.go:517-530`) drops the `\r` and any trailing spaces, so `"a  \r\nb"` becomes `"a\nb"`. That is a silent corruption, not an error, so nothing else in the suite would catch it. Grep confirms no test sets a `Description` containing `\n`, `\r` or `\t`, and every golden uses `description = "my test";`.

Smallest fix: no code change — extend the existing `TestPreparePkgEscapesDescription` (`nix_test.go:762`) to a small table and add a case with `Description: "line1\nline2\twith  \r\ntrailing"`, asserting the output contains `description = "line1\nline2\twith  \r\ntrailing";`. That test already runs the full `preparePkg` → `doBuildPkg` path, so it exercises both passes and the scanner.

Related: see the note on `nix_test.go:808-809` about reclaiming the redundant table slot.

internal/pipe/nix/nix_test.go
```diff
@@ -799,6 +800,61 @@ func TestPreparePkgEscapesDescription(t *testing.T) {
 	require.Contains(t, content, `description = "Say \"hello\" from C:\\tools and \${system}";`)
 }
 
+// The whole rendered derivation goes through the template engine once more,
+// so a description that renders to text containing `{{` must survive that
+// pass as a literal instead of being evaluated again.
+func TestPreparePkgDescriptionIsTemplatedOnce(t *testing.T) {
+	for _, description := range []string{
+		`Render {{ "{{example}}" }} templates`,
+		`{{ .Env.DESCRIPTION }}`,
```
caarlos0: **These two cases are the same test.**

Because the test calls `tmpl.New(ctx).ApplyAll(&nix.Description)` before `preparePkg`, both entries reduce to the identical string `Render {{example}} templates` before the code under test ever runs. I confirmed they produce a byte-identical intermediate derivation and, at the merge base, fail with the identical error — so the second case adds no incremental coverage.

This is why it diverges from brew's `TestFormulaDescriptionIsTemplatedOnce`: there, `rubyString` does the templating itself (`brew.go:382-385`), so the two inputs genuinely take different paths. Here the templating is hoisted out into the test.

Smallest fix: keep one case and spend the second on a materially different input — a description that needs escaping *and* survives the second pass, e.g. `has "quotes" and {{action}}` → `description = "has \"quotes\" and {{action}}";`. That is the combination neither this test nor `TestPreparePkgEscapesDescription` currently covers.

Alternatively, driving the test through `doRun` instead of hand-rolling the `ApplyAll` would make the two cases meaningfully distinct again and would also pin the `&nix.Description` entry in that `ApplyAll` list.

internal/pipe/nix/nix.go
caarlos0: Addressed in 1cca621ec.

Turned `TestPreparePkgEscapesDescription` into a table and added a `control characters` case with `Description: "line1\nline2\twith  \r\ntrailing"`, asserting the output contains `description = "line1\nline2\twith  \r\ntrailing";`. It runs the full `preparePkg` → `doBuildPkg` path, so it covers both template passes and the sanitizing scanner.

Confirmed it is a real regression test: with the three escape lines reverted in a scratch worktree, only that subtest fails, and the failure shows exactly the silent corruption described — the value comes out as

```
description = "line1
line2	with
trailing";
```

i.e. real newlines, and the `  \r` dropped by the scanner. The pre-existing `quotes and interpolation` case still passes, so the new case is what pins the new behavior.

internal/pipe/nix/nix_test.go
caarlos0: Addressed in 1cca621ec.

Made it a table of `{description, expect}` and replaced the duplicate. It now covers two materially different things:

- `from env`: `{{ .Env.DESCRIPTION }}` rendering to `Render {{example}} templates` — the realistic trigger.
- `needs escaping too`: `Say "hi" from {{ "{{example}}" }}` → `description = "Say \"hi\" from {{example}}";` — escaping and the second pass combined, which neither this test nor `TestPreparePkgEscapesDescription` covered before.

Left the templating in the test rather than routing through `doRun`, since `doRun` needs a repository and a client and would turn this into a much heavier test for little gain. The `ApplyAll` call still mirrors nix.go:155-163 faithfully.

Issue #6946, opened 2026-09-05
winget: entries sharing a display name publish through the first repository


caarlos0:
### Reproduction

Configure two WinGet entries with `name: tool`, but distinct package identifiers, such as `Acme.Tool` and `Other.Tool`, and distinct publication repositories. Give each entry otherwise valid installer metadata. Also compare entries with different `skip_upload` policies.

### Actual and expected results

Both entries generate their manifests, but all six files form one publication group. The first entry's repository and path are used for every file, and its skip policy controls both entries. Distinct publication configurations should remain independent even when their display names match.

Verification used actual YAML loading and manifest generation, then production `publishAll` with a recording client. It established the selected repository, paths, and skip behavior without writing a remote repository.

[Manifest registration](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/winget/template.go#L37-L40) uses `winget.Name` as the artifact ID. [Publication groups by that ID and takes the first configuration](https://github.com/goreleaser/goreleaser/blob/628c889d7f2a34066649833be419fdb9abe01174/internal/pipe/winget/winget.go#L389-L441).

### Narrow correction

Carry a private per-entry grouping key, separate from the display name and package identifier. Regressions should assert independent destinations and skip policies for same-name entries.

---
Audit priority: **P2**.

Audit source revision: [628c889d7](https://github.com/goreleaser/goreleaser/commit/628c889d7f2a34066649833be419fdb9abe01174). AI disclosure: this report comes from a Copilot-assisted audit, with discovery by `caarlos0` agents and independent verification by `anvil` agents. Reproduction methods and limits are stated above; no fix is included in this issue.

PR #7147, opened 2026-09-15
fix(winget): do not allow multiple wingets with the same name

caarlos0:
Require winget names to be unique, instead of working around name collisions.

#7087 fixed a real bug — `publishAll` groups manifests with `GroupByID()`, so two wingets sharing a name were merged into one group and both published under `wingets[0].Path`, with `wingets[0]`'s repository and `skip_upload`. The fix, though, was to replace the artifact ID with a positional `winget-<index>`, which leaks an internal detail into `dist/artifacts.json` (documented as "the artifact ID from the configuration"). Trying to keep both — the configured name as the ID, and separate publish groups — only added a second, hidden grouping key.

Same name means same package, so there is nothing sane to do with two of them. This drops the workaround and validates the config instead, with `internal/ids` in `Default`, exactly like builds, archives, nfpms, dockers, snapcrafts, flatpaks, and the rest do for their `id`:

```
found 2 wingets with the ID 'tool', please fix your config
```

Since it runs on defaults, it also fails on `goreleaser check`, before anything is built. Two entries that both omit `name` collide on the project name, and are caught too.

With duplicates rejected, `ExtraID` goes back to `winget.Name` and `publishAll` back to `.GroupByID()`; the `WingetIndex` extra and the index plumbing through `doRun`/`createYAML`/`doAdditionalLocales` are gone.

### Tests

`TestDefaultDuplicateNames` covers explicit duplicates, duplicates coming from the default name, and the unique case.

The two `TestPublishSameNameWingets*` tests from #7087 are removed: they pinned behavior for a config that is now an error. The `repos` field on the test client went with them, as nothing read it anymore.

Follow-up to #7087.


internal/pipe/winget/winget.go
```diff
@@ -153,6 +153,10 @@ func (p Pipe) doRun(ctx *context.Context, winget config.Winget, cl client.Releas
 		return err
 	}
 
+	// the index keeps the ID unique even when two entries share a name,
+	// which would otherwise make them publish over each other.
+	artifactID := fmt.Sprintf("%s-%d", winget.Name, index)
```
caarlos0: **The `-<index>` suffix changes released behaviour for every user, including the single-entry case that has no collision.**

The PR description compares against `winget-<index>`, but that shape never shipped: `git tag --contains 52e4c58d0` lists only `v2.19.0-*-nightly`, and `git show v2.18.1:internal/pipe/winget/template.go` has `artifact.ExtraID: winget.Name`. So the baseline every released user actually sees is the bare name.

**Trigger:** one winget entry named `tool`, upgrade v2.18.1 -> v2.19.0, read `dist/artifacts.json`.
**Impact:** `"ID": "tool"` becomes `"ID": "tool-0"`. `www/content/customization/general/artifacts.md:85` documents the field as "The artifact ID from the configuration", and `tool-0` is not in the configuration. It also doesn't fully deliver the PR's own goal — `other-tool-2` is still positional, so deleting `winget[0]` renames it to `other-tool-1`.

**Why it isn't necessary:** `GroupByID` in `publishAll` (winget.go:401) is the *only* functional consumer of this ID. Winget types are absent from `ReleaseUploadableTypes()` (`artifact/artifact.go:136-154`), so no `release`/`blob`/`sign`/`checksums`/`publishers` pipe can select them with `ids:`; the only other `artifact.ID()` callers are `sbom.go:290` and `sign.go:191`, neither of which reaches winget artifacts. Uniqueness is an internal grouping concern that doesn't need to leak into user-visible metadata.

**Smallest fix:** keep `artifact.ExtraID: winget.Name` in `template.go:39` and add an internal `wingetIndexExtra: index`, then group `publishAll` on `(id, index)` instead of `.GroupByID()`. That follows the `wingetConfigExtra`/`wingetLocaleExtra` pattern the pipe already uses, still fixes #6946, and changes no user-visible value.

Prefer that over "suffix only on duplicate names": with a bare name, a template like `name: "{{ .Env.EMPTY }}"` yields `""`, and `GroupByID` silently `continue`s on an empty ID (`artifact/artifact.go:486-488`), dropping the entry with no message. Grouping on an internal key sidesteps that entirely. (Worth noting the current `-0` suffix accidentally masks that pre-existing bug.)

**Test:** assert a lone winget keeps the bare `tool`, and keep `TestPublishSameNameWingetsUseTheirOwnRepositories` green to prove same-named entries still publish to their own repositories.

Unrelated, for a separate issue: `aur.go:195` pairs a bare templated `Name` with the same `GroupByID` publish loop (`aur.go:403`), which looks like the same bug as #6946.

internal/pipe/winget/winget_test.go
```diff
@@ -1239,6 +1239,73 @@ func TestPublishSameNameWingetsUseTheirOwnRepositories(t *testing.T) {
...
+					Repository:        config.RepoRef{Owner: "other", Name: "winget"},
+				},
+				{
+					Name:              "other-tool",
+					Publisher:         "Acme",
+					PackageIdentifier: "Acme.OtherTool",
+					License:           "MIT",
+					ShortDescription:  "tool",
+					IDs:               []string{"tool"},
+					Repository:        config.RepoRef{Owner: "acme", Name: "winget"},
```
caarlos0: **Coverage gap: no fixture has `additional_locales`, so the test filters `artifact.WingetLocale` but never produces one.**

Line 1297 is the only assertion on a winget-generated artifact ID anywhere in the repo (`grep -n ExtraID internal/pipe/winget/winget_test.go` — every other hit sets an ID on an input archive/binary fixture). None of the three entries here sets `AdditionalLocales`, so `doAdditionalLocales` (winget.go:357) never runs and the `WingetLocale` filter matches nothing.

**Why it matters:** locale manifests carrying the *parent's* `artifactID` is exactly the invariant `publishAll`'s grouping depends on. If that ID ever diverged, locale manifests would form their own publish group and get published against `wingets[0]`'s config — the #6946 failure mode, reintroduced for locales only, with a green suite.

**Smallest fix:** add `AdditionalLocales: []config.WingetLocale{{Locale: "pt-BR"}}` to one entry. The existing exact-set assertion at 1301-1306 then covers it with no new assertion, because a divergent locale ID would add an unexpected key.

See also the finding at `internal/pipe/winget/winget.go:158` — if you switch to an internal group key, this fixture is what pins locale artifacts to their parent.

caarlos0 (commented):
Automated code review (GitHub Copilot CLI). `COMMENT` rather than `REQUEST_CHANGES` only because GitHub does not allow a non-comment review on your own PR — treat finding 1 as blocking.

The mechanics are correct and I verified them: `winget` is a by-value copy, nothing between `ApplyAll` and line 158 mutates `Name`, `errNoRepoName` returns before the ID exists and every other skip returns before the first `createYAML`, the ID never reaches a filename or a golden file (`template.go:22-23`), no stale `winget-[0-9]` assertion survives anywhere in the repo, and the new test does fail against the merge base with `{winget-0, winget-1, winget-2}`. `go test ./internal/pipe/winget/... -race -count=5` passes, including under `TZ=Etc/GMT+12` and `LC_ALL=tr_TR.UTF-8`. `<name>-<index>` also cannot self-collide: the index is the all-digit suffix after the final `-`, so distinct indices give distinct IDs.

My objection is to the shape of the ID, not to the code — details inline at `internal/pipe/winget/winget.go:158`.

**Verification gaps:** GoReleaser Pro is out of this repo, so I could not check whether anything there filters winget artifacts by `ids:`; that is the one thing that would change the impact assessment of finding 1. External `dist/artifacts.json` consumers are unknowable. Tests were run on macOS only, and only for the `winget` package plus `go vet`.

internal/pipe/winget/winget.go
caarlos0: Agreed — fixed in 75bfb5608.

`artifact.ExtraID` goes back to the templated `winget.Name`, so a single entry is `tool` again, matching v2.18.1 and the `artifacts.md` contract. Uniqueness moved to a new internal `WingetIndex` extra, and `publishAll` now groups on that instead of `.GroupByID()` (sorted by index, so publish order is deterministic too).

This also drops the dependency on the ID being non-empty, so the `GroupByID` silent-skip you flagged for `name: "{{ .Env.EMPTY }}"` can no longer swallow an entry.

I verified the #6946 fix is still covered independently of the ID format: simulating a grouping regression (collapsing every artifact into one group) fails both `TestPublishSameNameWingetsUseTheirOwnRepositories` (`expected: "other", actual: "acme"`) and `TestPublishSameNameWingetsKeepSkipUploadSeparate`. `MustExtra[int]` is safe across an `artifacts.json` round-trip because `tryCastExtra` falls back to a JSON decode, so a `float64` still lands as `int`.

Left the `aur` parallel (`aur.go:195` bare name + `GroupByID` at `aur.go:403`) out of this PR rather than widening it — worth a separate issue.

internal/pipe/winget/winget_test.go
caarlos0: Fixed in 75bfb5608. The third entry now sets `AdditionalLocales: []config.WingetLocale{{Locale: "pt-BR"}}`, so `doAdditionalLocales` actually runs and the `artifact.WingetLocale` filter matches something.

The assertion now pins the ID and the group key together rather than just collecting IDs:

```go
require.Equal(t, map[group]int{
    {id: "tool", index: 0}:       3,
    {id: "tool", index: 1}:       3,
    {id: "other-tool", index: 2}: 4,
}, got)
```

The count is what covers your point — index 2 has 4 artifacts instead of 3 precisely because the pt-BR locale manifest inherits its parent entry. I confirmed it bites: making `doAdditionalLocales` receive `index+100` fails with a split `{id: "other-tool", index: 102}: 1` group, so a locale manifest drifting out of its parent group can no longer pass silently.

PR #7148, opened 2026-09-15
fix(docker): a bare platform means the baseline CPU variant

caarlos0:
Treat a Docker v2 `platforms` entry without an explicit CPU variant as the **baseline** variant buildx resolves it to, for every arch rather than amd64 alone.

`parsePlatform` pins the variant for amd64 only. `arm` and `arm64` are left unpinned, so they match any variant, and `toPlatform` then derives the build context directory from whichever binary happened to match. Two platforms can therefore never agree with the `$TARGETPLATFORM` buildx sets, and `COPY $TARGETPLATFORM/myprogram` fails:

| `platforms` | `$TARGETPLATFORM` | directory used | |
| --- | --- | --- | --- |
| `linux/arm` | `linux/arm/v7` | `linux/arm/v6` or `v7`, build-dependent | ✗ |
| `linux/arm64/v9` | `linux/arm64/v9` | `linux/arm64` | ✗ |

### The fix

```go
switch result.arch {
case "amd64": result.amd64 = "v1"   // already the case
case "arm":   result.arm   = "7"
case "arm64": result.arm64 = "v8.0"
}
```

plus the missing arm64 case in `toPlatform`, which appends the variant's major for anything above the baseline v8. The baselines come from containerd's `normalizeArch`, which is what buildkit uses to compute `TARGETPLATFORM`. Minor variants such as `v8.2` and `v9.2` are handled in #7149, stacked on top of this one.

Because the filter pins the variant, at most one variant per platform can match, so `toPlatform` stays correct on its own. The production change is **17 insertions, 2 deletions** in `docker.go`; the rest of the diff is tests.

### Why not "match any variant"

A bare `linux/amd64` deliberately does **not** pick up a `goamd64: [v3]` build. Doing so produces an image labelled baseline amd64 that dies with SIGILL on a pre-Haswell CPU, and with `goamd64: [v1, v3]` it copies both binaries to the same path where the last write wins. That is what v2.18.1 did, and #7095 fixed it. Use `platforms: [linux/amd64/v3]` for a non-baseline build.

### Tests

Every added case fails on `main`:

```
--- FAIL: TestToPlatform/linux/arm64/v9
--- FAIL: TestParsePlatform/linux/arm
--- FAIL: TestParsePlatform/linux/arm64
--- FAIL: TestContextPlacement/bare_arm_is_v7
--- FAIL: TestContextPlacement/bare_arm64_does_not_take_v9
--- FAIL: TestContextPlacement/arm64/v9_keeps_the_variant
```

`TestContextPlacement` goes end to end through `makeContext` and asserts the path a Dockerfile's `COPY $TARGETPLATFORM/` actually reads. The whole pre-existing suite passes without a single change.

Relates to #7095. Part of a series of fixes found reviewing `v2.18.1..main`.


internal/pipe/docker/v2/docker.go
```diff
@@ -585,53 +617,35 @@ func tagSuffix(plat string) string {
...
 }
 
+// contextDir returns the platform as buildx normalizes it for
+// TARGETPLATFORM, which is the build context directory artifacts for this
+// platform are copied into. buildx drops the baseline variant, so
+// linux/amd64/v1 becomes linux/amd64.
+func (p platform) contextDir() string {
+	parts := []string{p.os, p.arch}
+	switch {
+	case p.arm != "":
```
caarlos0: **Regression.** A bare `linux/arm` platform now gets the directory `linux/arm`, but buildx sets `TARGETPLATFORM=linux/arm/v7`. `COPY $TARGETPLATFORM/` then finds nothing and the image build fails — the same failure this PR fixes for amd64.

**Trigger:** `platforms: [linux/arm]` with `goarm: ["7"]`. This worked before this commit, because the directory came from the artifact (`toPlatform` appended `"v"+a.Goarm`). `docker.go:429` passes the raw configured string to `--platform`, and nothing validates it.

**Ground truth** is containerd `platforms.normalizeArch`, which buildkit uses to compute `TARGETPLATFORM`:

```go
case "arm":
    switch variant {
    case "", "7":
        variant = "v7"
```

**Measured on this branch:**

```
contextDir("linux/arm") = "linux/arm"
placed "mybin" (goarm 7) -> "linux/arm/mybin"
```

**Smallest fix** — in `parsePlatform`, not here:

```go
if len(parts) >= 2 {
    result.arch = parts[1]
    if result.arch == "arm" {
        result.arm = "7"
    }
}
```

This corrects the directory and the filter together, and the asymmetry is correct: `linux/amd64` means "any variant", but `linux/arm` means exactly `arm/v7`. I verified it with `go test -overlay`: `contextDir("linux/arm")` becomes `linux/arm/v7`, `linux/arm/v6` is unchanged, the selection stops depending on insertion order, and the whole package still passes.

**Tests:** add `"linux/arm": "linux/arm/v7"` to `TestContextDir` (see my comment on `docker_test.go:641`), plus a `TestContextArtifacts` subtest with `Platforms: ["linux/arm"]` and goarm `6` and `7` artifacts that asserts exactly `["linux/arm/v7/mybin"]` for both insertion orders.

internal/pipe/docker/v2/docker.go
```diff
@@ -535,48 +520,95 @@ func copyArtifact(src, dst string) error {
...
+		result = append(result, art)
+	}
+	return result
+}
+
+// variantRank returns the major CPU variant of an artifact, e.g. 3 for
+// goamd64 v3. Artifacts without a variant rank first.
+func variantRank(a *artifact.Artifact) int {
+	variant := a.Goamd64
+	if a.Goarch == "arm64" {
```
caarlos0: `variantRank` reads `Goamd64` and `Goarm64` only — it never reads `Goarm`. So every 32-bit arm artifact ranks 0, and the comment on `leastSpecificPerName` above (lines 578-582) claims more than the code delivers: the choice is *not* "independent of the order artifacts were added in" when ranks tie.

**Measured on this branch** with `platforms: [linux/arm]`:

```
added [7 6 5] -> kept mybin-v7
added [5 6 7] -> kept mybin-v5
```

The arm case disappears once `parsePlatform` defaults a bare arm to v7 (see my comment on line 634), because `ByGoarm` then leaves one candidate. A tie can still happen when two build IDs produce the same binary `Name` for the same platform:

```
add order [id1 id2] -> kept id1/mybin
add order [id2 id1] -> kept id2/mybin
```

**Not a regression** — the old code copied both to the same target path and the last write won. But the new comment is now false, so either weaken it or make the order total.

**Smallest fix:**

```go
return cmp.Or(
    cmp.Compare(variantRank(a), variantRank(b)),
    cmp.Compare(a.ID(), b.ID()),
)
```

**Test:** a `TestContextArtifacts` subtest with two IDs, the same artifact name and the same platform, asserting one fixed winner for both insertion orders.

internal/pipe/docker/v2/docker_test.go
```diff
@@ -631,96 +637,33 @@ func TestIsDockerDaemonAvailableNoDaemon(t *testing.T) {
...
-			Goos:   "linux",
-			Goarch: "s390x",
-		},
-		"linux/riscv64": {
-			Goos:   "linux",
-			Goarch: "riscv64",
-		},
+func TestContextDir(t *testing.T) {
+	// values asserted against `docker buildx build --platform=X` with a
+	// Dockerfile echoing $TARGETPLATFORM.
```
caarlos0: This comment says the values were measured with real buildx runs, but the table has **no bare `linux/arm` input and no `arm64` variant input** — exactly the two inputs where `contextDir` disagrees with buildx.

- `linux/arm` → buildx `linux/arm/v7`, this branch `linux/arm`. That is the regression in `docker.go:634`. Adding `"linux/arm": "linux/arm/v7"` here fails today and catches it.
- `linux/arm64/v9` → buildx `linux/arm64/v9`, this branch `linux/arm64`. containerd `normalizeArch` drops only `8`/`v8`/`v8.0` and maps `9`/`9.0`/`v9.0` to `v9`. **This also happens on `main`, so it is not a regression of this PR** and I would keep the fix out of it. I raise it only because this table now owns the buildx-parity contract, so the gap should at least be visible.

Suggested additions: `"linux/arm": "linux/arm/v7"`, `"linux/arm64/v8": "linux/arm64"`, and `"linux/arm64/v9": "linux/arm64/v9"` once `contextDir` handles it.

caarlos0 (commented):
## Verdict: changes requested — one new regression, otherwise the fix is correct

*(Posted as a comment review: GitHub does not allow `REQUEST_CHANGES` on your own pull request.)*

I confirmed the premise and the fix. On the base commit, `contextArtifacts` with `platforms: [linux/amd64, linux/arm64]` and one `goamd64: v3` binary returns **0 artifacts**. The new `TestContextArtifactsBarePlatformMatchesAnyAmd64Variant` fails on base for `v2`, `v3` and `v4`, and passes on this branch. The regression test is real and it is the right test.

**Blocker:** a bare `linux/arm` platform regresses — see `docker.go:634`. It worked before this commit and it fails now, with the same `COPY $TARGETPLATFORM/` failure this PR removes for amd64. The fix is 3 lines in `parsePlatform` and it also makes the arm selection deterministic.

The rest of the change holds up:

- Replacing the `Goos == "all"` branch with `ByType(artifact.PyWheel)` is equivalent. Only `PyWheel` and `PySdist` use `Goos: "all"` (`internal/builders/{uv,poetry}/build.go`), and `PySdist` never reached `makeContext` on either side.
- Wheels now land in `contextDir()` instead of the raw platform string, so `platforms: [linux/amd64/v1]` moves from `linux/amd64/v1` to `linux/amd64`. That is an unclaimed correction.
- Removing `toPlatform` loses no validation that matters. It rejected valid buildx platforms such as `mips64le`, so it failed correct configurations. Unsupported platforms now fail in buildx instead, which is still an error the user sees.
- `leastSpecificPerName` and the `contextArtifact` struct are both necessary. Without them, `goamd64: [v1, v3]` copies two files to one path and the last write wins.
- `len(d.Platforms) == 0` cannot reach `contextArtifacts`; `buildImage` returns `pipe.Skip` first (`docker.go:215`).
- Performance is noise. At a realistic 40 artifacts the difference is about 1.5 µs, once per image, before a `docker buildx build` that takes seconds.
- No public surface changes. Docs need no change; `dockers_v2.md` already shows `COPY $TARGETPLATFORM/myprogram /usr/bin/`.

### Coverage gaps (non-blocking, except the `linux/arm` one)

`variantRank` has no direct test; `leastSpecificPerName` dedup across IDs has no test; `contextDir` covers no `arm64` variant; and `goamd64: [v1, v3]` on a bare `linux/amd64` — the exact scenario the description promises — has no test, although I measured that it does work and is order-independent.

### Verification gaps

No live buildx run was possible here: no Docker daemon and no `docker buildx` on this machine. The buildx claims above are verified against `containerd/platforms` `normalizeArch`, which is the code buildkit uses to compute `TARGETPLATFORM`, not against a real build. Every other value in `TestContextDir` agrees with that source. `go test ./internal/pipe/docker/v2/` passes on this branch, also with `-count=10` and `-race`. The full repository suite was not run.

*Reviewed by an automated reviewer (GitHub Copilot CLI).*

internal/pipe/docker/v2/docker.go
caarlos0: Fixed in 920c43255.

Took the `parsePlatform` route, so the context directory and the artifact filter are corrected together:

```go
if len(parts) >= 2 {
    result.arch = parts[1]
    if result.arch == "arm" {
        // buildx normalizes a bare linux/arm to linux/arm/v7, so unlike
        // amd64 and arm64, it does not mean "any variant".
        result.arm = "7"
    }
}
```

`contextDir("linux/arm")` is now `linux/arm/v7`, and `linux/arm/v5|v6|v7` are unchanged.

Covered by the new `TestContextArtifactsBareArmIsV7`, which asserts exactly `["linux/arm/v7/mybin"]` for both insertion orders of goarm 6 and 7, plus `"linux/arm": "linux/arm/v7"` in `TestContextDir` and `"linux/arm": {os: "linux", arch: "arm", arm: "7"}` in `TestParsePlatform`. All three fail on `ce473bc3e`:

```
--- FAIL: TestContextDir/linux/arm       expected "linux/arm/v7"  actual "linux/arm"
--- FAIL: TestParsePlatform/linux/arm    expected arm:"7"         actual arm:""
--- FAIL: TestContextArtifactsBareArmIsV7/6-7  expected ["linux/arm/v7/mybin"] actual ["linux/arm/mybin"]
--- FAIL: TestContextArtifactsBareArmIsV7/7-6  expected ["linux/arm/v7/mybin"] actual ["linux/arm/mybin"]
```

internal/pipe/docker/v2/docker.go
caarlos0: Fixed in 920c43255. Made the order total instead of weakening the comment:

```go
slices.SortStableFunc(sorted, func(a, b *artifact.Artifact) int {
    return cmp.Or(
        cmp.Compare(variantRank(a), variantRank(b)),
        cmp.Compare(a.ID(), b.ID()),
    )
})
```

The arm half is gone for the other reason you gave: `parsePlatform` now pins a bare `linux/arm` to v7, so `ByGoarm` leaves one candidate and arm artifacts never compete. `variantRank` says so explicitly now:

> Only amd64 and arm64 are ranked: a platform always pins the arm variant, so arm artifacts never compete.

The cross-ID tie is covered by the new `TestContextArtifactsSameNameDifferentIDsIsDeterministic`, which adds two IDs with the same binary name for one platform in both orders and asserts `id1` wins each time. It fails on `ce473bc3e`:

```
--- FAIL: .../id2-id1   expected: "id1/mybin"   actual: "id2/mybin"
```

internal/pipe/docker/v2/docker_test.go
caarlos0: Addressed in 920c43255. Added to the table:

```go
"linux/arm":      "linux/arm/v7",
"linux/arm64/v8": "linux/arm64",
```

`"linux/arm"` fails on `ce473bc3e` (`expected "linux/arm/v7", actual "linux/arm"`), so it catches the regression. `"linux/arm64/v8"` pins the variant that containerd drops.

Left `linux/arm64/v9` out on purpose, for the reason raised in the comment itself: `main` behaves identically, so it is not a regression of this PR, and fixing it means changing `contextDir` for arm64 — a separate concern with its own blast radius. I will open a follow-up issue for it rather than grow this change.

That keeps the table honest about what it asserts: every value in it now matches containerd `normalizeArch`, and the one input that does not is not claimed.

PR #7156, opened 2026-09-19
feat(cask): opt-in use_steps for postflight/preflight stanzas

joaofsantos:
## Summary

Homebrew deprecated the `preflight`/`postflight`/`uninstall_preflight`/`uninstall_postflight` block stanzas in favor of `preflight_steps`/`postflight_steps`/`uninstall_preflight_steps`/`uninstall_postflight_steps`. Every cask GoReleaser generates with a `hooks.*` block (e.g. the common quarantine-removal hook, see the [docs](https://goreleaser.com/customization/homebrew_casks/#signing-and-notarizing)) now prints a deprecation warning on `brew install`/`upgrade`/`untap`.

I hit this via `go-task`'s tap, which uses this exact hook pattern generated from its `.goreleaser.yml`.

The new `_steps` stanzas use a restricted, declarative DSL instead of arbitrary Ruby (no `if`, no `system_command`, guards via `on_macos do ... end` instead). Because of that, unconditionally switching the template would silently break existing hook content for every GoReleaser cask user — so this adds an opt-in `hooks.use_steps` boolean (default `false`, fully backwards compatible) that switches the four stanzas GoReleaser emits to the new `_steps` variants when set.

## A real gotcha, found and fixed while verifying this end-to-end

Homebrew's `_steps` DSL uses `{{ }}`-delimited tokens for its own runtime substitution (`{{staged_path}}`, `{{appdir}}`, etc.). GoReleaser also renders hook content through its own `{{ }}` Go templates — twice, actually: once per-field (`cask.go`), and again as part of a whole-document pass in `doBuildCask` (the same pass that already protects the `desc` line, for the same reason). A natural, unescaped `{{staged_path}}` failed to parse on both passes.

Rather than requiring users to hand-escape every Homebrew token, the fix:
- Skips GoReleaser's per-field hook templating entirely when `use_steps` is set, since steps content is meant to carry Homebrew's own tokens verbatim.
- Renders each `_steps` stanza as a single `%q`-protected literal (same trick as `desc`) via a new `hookStanza` template func, so the whole-document pass reproduces it untouched.

Net effect: with `use_steps`, hook content needs zero escaping and can use Homebrew's `{{ }}` tokens exactly as their docs show.

## Changes

- `pkg/config/config.go`: new `HomebrewCaskHooks.UseSteps` field (`hooks.use_steps` in YAML).
- `internal/pipe/cask/cask.go`: skips GoReleaser's hook templating pass when `UseSteps` is set.
- `internal/pipe/cask/template.go`: new `hookStanza` helper that renders a hook block and protects it from the whole-document template pass.
- `internal/pipe/cask/templates/cask.rb`: all four hook stanzas now go through `hookStanza`, conditionally emitting the `_steps` variants.
- `internal/pipe/cask/cask_test.go` + new golden fixture: `TestFullCaskStepsHooks` covers all four hook stanzas with `use_steps: true`.
- `www/static/schema.json`: regenerated (`go run . schema -o ./www/static/schema.json`).
- `www/content/customization/publish/homebrew_casks.md`: documents the new field, and updates the quarantine-removal example with a `_steps` version.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/pipe/cask/...` (88 passed — all pre-existing golden fixtures byte-for-byte unchanged, confirming the non-`use_steps` path is untouched)
- [x] `go test ./pkg/config/...`
- [x] `go vet ./internal/pipe/cask/... ./pkg/config/...`
- [x] `gofmt -l` clean on edited files
- [x] End-to-end: rendered go-task's real quarantine-removal hook (unescaped `{{staged_path}}`, exactly as a user would write it) through the actual `doBuildCask` pipeline, then loaded the resulting cask file via a real Homebrew install's `Cask::CaskLoader.load` — parses with zero deprecation warnings, correct `postflight_steps` → `on_macos` → `run` nesting

AI was used to help write this change (implementation, tests, and docs); I reviewed and verified it against a live Homebrew install per the test plan above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


caarlos0:
duplicated of https://github.com/goreleaser/goreleaser/pull/6873

